PR #154, opened 2026-02-09
fix(app): send text+CR as atomic write for PTY sessions

### LINKED ISSUE #151: Multi-line paste fails in PTY sessions (tmux bracketed paste)
## Bug

When sending a long multi-line message from the app to a PTY/tmux session, tmux intercepts the input as a bracketed paste and displays `[Pasted text #N +X lines]` instead of forwarding it to Claude Code.

## Root Cause

PTY sessions write raw text to the terminal via `session.write(text)`. For multi-line input, tmux detects it as a paste (bracketed paste mode) and shows a summary instead of processing it character-by-character.

The `streamingMessageId: 'pending'` state makes the UI show the interrupt button, giving the impression the message is being processed. The 5-second timeout eventually clears it, but this is confusing UX.

## CLI mode comparison

CLI headless sessions handle this correctly — `session.sendMessage(text)` sends the full message as a single JSON payload via stream-json. No terminal emulation involved.

## Possible Fixes

1. **tmux send-keys with escaping**: Use `tmux send-keys -t session` with proper escaping for multi-line text
2. **tmux load-buffer + paste-buffer**: Write text to tmux buffer, then paste it
3. **Line-by-line sending**: Split text into lines, send each with small delays
4. **Disable bracketed paste**: `tmux set -g -s set-clipboard off` or disable bracketed paste mode in the session

## Acceptance Criteria

- [ ] Multi-line text sent from the app reaches Claude Code in PTY sessions
- [ ] No `[Pasted text]` placeholder in the terminal
- [ ] Works for messages up to ~4KB (typical long prompt)

blamechris:
## Summary

Combine text and CR into a single `sendInput(text + '\r')` call for PTY sessions instead of sending them as separate WebSocket messages with a 50ms delay.

## Root Cause

```typescript
// Before: two separate WS messages → two separate PTY writes → race condition
sendInput(text);                          // Write 1: text
setTimeout(() => sendInput('\r'), 50);    // Write 2: CR (50ms+ later)
```

Multi-line text would sit in the terminal input buffer waiting for the CR. By the time it arrived (50ms + network latency + WS overhead), the terminal state may have changed, causing the text to appear "stuck" and requiring a second send.

## Fix

```typescript
// After: single WS message → single PTY write → atomic
sendInput(hasTerminal ? text + '\r' : text);
```

Applied to both `handleSend` and `handleSelectOption`.

## Test Plan

- [ ] Send a single-line message in chat mode to a PTY session — submits immediately
- [ ] Paste a multi-line block in chat mode to a PTY session — submits on first send
- [ ] Select a prompt option in a PTY session — submits immediately
- [ ] CLI session input unchanged (no CR appended)

Closes #151

### DIFF
--- a/packages/app/src/screens/SessionScreen.tsx
+++ b/packages/app/src/screens/SessionScreen.tsx
@@ -183,12 +183,11 @@ export function SessionScreen() {
       addUserMessage(text);
     }
 
-    sendInput(text);
-    // PTY sessions: send Enter separately -- Claude Code's TUI needs text and CR as separate writes
-    // CLI sessions: the server handles the full message directly
-    if (hasTerminal) {
-      setTimeout(() => sendInput('\r'), 50);
-    }
+    // PTY sessions: append CR so text + submit arrive as a single atomic write.
+    // Sending them separately caused a race condition where multi-line text
+    // would sit in the terminal input buffer before the CR arrived.
+    // CLI sessions: the server handles the full message directly (no CR needed).
+    sendInput(hasTerminal ? text + '\r' : text);
   };
 
   const handleKeyPress = (key: string) => {
@@ -217,11 +216,7 @@ export function SessionScreen() {
       sendPermissionResponse(requestId, value);
       return;
     }
-    sendInput(value);
-    // PTY sessions: send Enter separately -- the TUI needs text and CR as separate writes
-    if (hasTerminal) {
-      setTimeout(() => sendInput('\r'), 50);
-    }
+    sendInput(hasTerminal ? value + '\r' : value);
   };
 
   // Check if Enter key should send based on current mode and settings


blamechris:
## Code Review: PR #154

### Summary

This PR replaces two separate WebSocket writes (text, then `\r` after 50ms) with a single atomic `sendInput(text + '\r')` for PTY sessions. The change targets `handleSend` and `handleSelectOption` in `SessionScreen.tsx`. CLI sessions are unaffected (no CR appended). The stated goal is to fix a race condition where multi-line text would sit in the terminal input buffer waiting for the delayed CR.

### Verdict: Approve

The change is correct and well-scoped. The atomic write is the right approach here.

### Analysis

**1. Is combining text+CR into a single write correct for all PTY scenarios?**

Yes. The full data path is:

```
sendInput(text + '\r')
  → WS message { type: 'input', data: 'hello\r' }
  → ws-server.js: entry.session.writeRaw('hello\r')     // multi-session
  → pty-session.js: this._ptyManager.write('hello\r')
  → pty-manager.js: this.ptyProcess.write('hello\r')     // node-pty write
```

node-pty's `write()` calls `uv_write` which delivers the entire buffer in one kernel write to the PTY master fd. tmux reads this atomically and sees `hello\r` as a single input event — text followed by Enter. This is identical to typing "hello" and pressing Enter on a physical terminal.

The legacy PTY mode path (`_handlePtyMessage`) also calls `this.ptyManager.write(msg.data)` directly, so the same atomic write applies there.

**PtySession.sendMessage() already does this:** `this._ptyManager.write(text + '\r')` at line 119 of `pty-session.js`. The PR aligns the app-side behavior with what the server-side `sendMessage` already expected.

**2. Does the atomic write cause issues with tmux bracketed paste?**

No — and this is actually an improvement. tmux's bracketed paste detection fires when it sees `\x1b[200~` (paste start) or multi-line content arriving between `\x1b[200~...\x1b[201~` markers. The app never sends these escape sequences. What tmux *does* detect is rapid multi-line input that looks like a paste (newlines in a single write). However:

- For single-line input (the common case), `text + '\r'` has no embedded newlines, so no bracketed paste detection.
- For multi-line input, the old two-write approach was *more* likely to trigger tmux paste detection, not less, because the first write contained the multi-line text alone, and tmux would see the newlines and intercept it as a paste before the CR even arrived.
- The atomic write doesn't change the multi-line paste behavior — issue #151 documents this as a separate problem with multi-line input in PTY sessions. This PR simply ensures that the *submission CR* arrives atomically with the text, which is strictly better.

**3. Does this change affect terminal-mode key sending (handleKeyPress)?**

No. `handleKeyPress` (lines 193-210) sends individual control characters (`\r`, `\t`, `\x1b`, etc.) directly via `sendInput(keyMap[key])` without any CR appending. This code path is untouched by the PR. The `hasTerminal` conditional only applies to `handleSend` and `handleSelectOption`, which are chat-mode submission actions where the user is sending a complete message, not individual keystrokes.

**4. Echo suppression is handled correctly.** The WS server calls `entry.session.expectEcho?.(text)` before `entry.session.writeRaw(text)`. Since `text` now includes the trailing `\r`, the echo filter will correctly expect the full string including the CR, which matches what the PTY will echo back.

### Minor Notes

- The comment quality is good — the inline comments clearly explain why CR is appended for PTY and not for CLI.
- The diff is clean: 6 additions, 11 deletions. Removing the `setTimeout` and separate `sendInput('\r')` eliminates a class of timing bugs.
- Issue #151 (multi-line paste in tmux) is only partially addressed by this PR. The core issue — tmux intercepting multi-line text as bracketed paste — is documented and acknowledged in the issue. This PR fixes the narrower problem of the submission CR not arriving atomically.

No follow-up issues needed.

PR #156, opened 2026-02-09
fix(parser): gate IDLE→RESPONSE on recent Claude activity

### LINKED ISSUE #155: Output parser emits false chat bubbles from host terminal activity
## Bug

When the user interacts with the host terminal directly (clicks into tmux, types commands, causes redraws), the output parser picks up the terminal output and emits it as Claude response bubbles in chat view.

## Root Cause

In `output-parser.js` lines 500-511, the state machine transitions to RESPONSE state on **any non-noise line** while in IDLE or THINKING state:

```javascript
} else if (this.state === State.IDLE || this.state === State.THINKING) {
  if (this._matchAndConsumeEcho(trimmed)) return;
  // Any unrecognized line triggers RESPONSE state:
  this.state = State.RESPONSE;
  this.currentMessage = {
    type: "response",
    content: trimmed + "\n",
    timestamp: Date.now(),
  };
  this._resetFlush();
}
```

This is too permissive. When the user types `ls` in the host terminal, the output (file listings, directory names, etc.) passes through `_isNoise()` and gets treated as Claude's response.

## Why _isNoise() Can't Fix This

The noise filter handles known patterns (server logs, spinners, tmux status bars, version lines). But it **cannot** filter legitimate terminal output like:
- Directory listings from `ls`
- File contents from `cat`
- Error messages from commands
- Any text the user types

## Proposed Fix

Add a state guard: only transition to RESPONSE after seeing a Claude response marker (`⏺` line at output-parser.js:474). While in IDLE state, non-noise lines that don't match a known response start marker should be ignored (or routed to raw only, not to chat).

Possible approaches:
1. **Require `⏺` marker before RESPONSE**: Only transition IDLE→RESPONSE after seeing the response start marker
2. **Time-based guard**: Only accept response lines within N seconds of detecting user input submission
3. **Echo-based guard**: Only accept response lines after the user's input echo has been consumed

## Related

- #153 (Unicode status leak — may be a symptom of this same issue)

## Acceptance Criteria

- [ ] Host terminal activity does not produce chat bubbles
- [ ] Claude's actual responses still appear correctly
- [ ] Tool output and response markers still trigger state transitions properly

blamechris:
## Summary

Track `_lastClaudeActivity` timestamp in the output parser. The IDLE→RESPONSE fallthrough (lines 500-511) now only fires within 5 seconds of the last Claude-specific activity. After that window, unrecognized terminal lines are dropped instead of creating false chat bubbles.

## Root Cause

The parser's state machine transitions to RESPONSE on **any** non-noise line while in IDLE state (`output-parser.js:500-511`). This was needed for response continuation after tool blocks, but it also caught:
- User typing commands on the host terminal
- Terminal redraws from switching tmux panes
- Any other terminal activity during Claude's idle periods

## How the Fix Works

`_lastClaudeActivity` is set whenever Claude-specific output is detected:
- Thinking/spinner lines (line 410)
- Tool block start (line 430)
- Tool block end (line 447)
- Response marker `⏺` (line 478)

The fallthrough guard at line 505:
```javascript
if (Date.now() - this._lastClaudeActivity > 5000) return;
```

## Test Changes

- Updated "starts as response in IDLE state" to set `_lastClaudeActivity` (simulates recent tool block)
- Added new test: "drops content in IDLE state without recent Claude activity"
- Updated echo expiry test to set `_lastClaudeActivity`
- All 250 tests pass

## Test Plan

- [ ] User types on host terminal while Claude is idle — no false chat bubbles
- [ ] Claude responds with tool blocks then continues — continuation text appears correctly
- [ ] Claude responds with `⏺` marker — response appears correctly
- [ ] Terminal redraws during idle — no false chat bubbles

Closes #155

### DIFF
--- a/packages/server/src/output-parser.js
+++ b/packages/server/src/output-parser.js
@@ -37,6 +37,11 @@ export class OutputParser extends EventEmitter {
     // Pending interactive prompt being accumulated
     this._pendingPrompt = null;
     this._promptFlushTimer = null;
+    // Timestamp of last Claude-specific activity (tool use, thinking, response marker).
+    // Used to gate the IDLE→RESPONSE fallthrough: only accept unrecognized lines as
+    // response continuation if Claude was recently active. Prevents host terminal
+    // activity (user typing, redraws) from creating false chat bubbles.
+    this._lastClaudeActivity = 0;
     // Echo suppression: lines expected from PTY echo of user input
     this._pendingEchos = [];
     // Scrollback suppression: when attaching to an existing tmux session,
@@ -405,6 +410,7 @@ export class OutputParser extends EventEmitter {
 
     // Skip thinking/spinner lines
     if (this._isThinking(trimmed)) {
+      this._lastClaudeActivity = Date.now();
       if (this.state !== State.THINKING && this.state !== State.TOOL_USE) {
         this._finishCurrentMessage();
         this.state = State.THINKING;
@@ -424,6 +430,7 @@ export class OutputParser extends EventEmitter {
     const toolBoxMatch = trimmed.match(new RegExp(`^[╭┌]─+\\s*(${toolNames})`));
     const toolCompactMatch = !toolBoxMatch && trimmed.match(new RegExp(`^(${toolNames})\\(`));
     if (toolBoxMatch || toolCompactMatch) {
+      this._lastClaudeActivity = Date.now();
       this._finishCurrentMessage();
       const tool = toolBoxMatch ? toolBoxMatch[1] : toolCompactMatch[1];
       this.state = State.TOOL_USE;
@@ -441,6 +448,7 @@ export class OutputParser extends EventEmitter {
 
     // Tool use block end
     if (this.state === State.TOOL_USE && /^[╰└]─+/.test(trimmed)) {
+      this._lastClaudeActivity = Date.now();
       this._finishCurrentMessage();
       this.state = State.IDLE;
       return;
@@ -473,6 +481,7 @@ export class OutputParser extends EventEmitter {
 
     // Response marker: ⏺ followed by text
     if (/^⏺\s*/.test(trimmed)) {
+      this._lastClaudeActivity = Date.now();
       if (this.state !== State.RESPONSE) {
         this._finishCurrentMessage();
         this.state = State.RESPONSE;
@@ -500,7 +509,11 @@ export class OutputParser extends EventEmitter {
     } else if (this.state === State.IDLE || this.state === State.THINKING) {
       // Suppress echoed user input from PTY
       if (this._matchAndConsumeEcho(trimmed)) return;
-      // New response content
+      // Only accept as response continuation if Claude was recently active.
+      // Without this guard, host terminal activity (user typing in tmux,
+      // terminal redraws) creates false chat bubbles during idle periods.
+      if (Date.now() - this._lastClaudeActivity > 5000) return;
+      // New response content (continuation after tool block or thinking)
       this.state = State.RESPONSE;
       this.currentMessage = {
         type: "response",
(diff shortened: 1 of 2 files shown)


blamechris:
## Code Review: PR #156

### Summary

Clean, minimal fix for a real problem. The `_lastClaudeActivity` guard adds a time-based gating mechanism to the IDLE->RESPONSE fallthrough path, preventing host terminal activity from creating false chat bubbles. The approach is sound.

### Analysis

**Is 5 seconds the right window?**

5 seconds is reasonable. The fallthrough path (line 509-523) catches response continuation text that arrives *without* a `⏺` marker — typically after a tool block ends. In normal Claude Code output, this continuation arrives within milliseconds of the tool block closing (which sets `_lastClaudeActivity`). Even with network jitter or tmux redraw delays, 5 seconds provides generous headroom.

The realistic concern would be extended tool execution: Claude runs a `Bash` command that takes 30+ seconds, the tool block closes, then Claude continues responding. But in that scenario, the tool block *end* (`╰───` at line 450) sets `_lastClaudeActivity` right when the block closes — the timer starts from that point, not from when the tool *started*. So even very long tool runs are fine.

One edge case worth noting: if Claude pauses for more than 5 seconds between tool block end and continuation text (no `⏺` marker), the continuation would be dropped. In practice this is unlikely — Claude Code emits response text immediately after tool blocks, and the primary response path uses `⏺` markers which bypass this guard entirely.

**Are all Claude activity sources covered?**

The four activity sources tracked are:
1. Thinking/spinner lines (line 413)
2. Tool block start (line 433)
3. Tool block end (line 451)
4. Response marker `⏺` (line 484)

These cover the relevant signals well. Two observations:

- The `❯` prompt line (line 458) does NOT set `_lastClaudeActivity`, and this is correct. User input submission isn't Claude activity — it's user activity. Claude's thinking spinner will fire before any continuation text.
- The `_detectPrompt` path (line 423-424) doesn't set it either, which is also correct — interactive prompts are their own message type and don't lead to the fallthrough path.

**Could the guard cause false negatives (dropping real responses)?**

The guard ONLY affects the fallthrough path at line 509 (lines that arrive in IDLE/THINKING state without a `⏺` marker and without matching any other pattern). The primary response path — `⏺`-prefixed lines — is completely unaffected by this guard.

The only scenario for a false negative: Claude emits continuation text without a `⏺` marker, more than 5 seconds after the last tool/thinking/response-marker activity, while the parser is in IDLE state. This would require Claude to go silent for 5+ seconds mid-response without any thinking indicator. In practice, Claude Code always shows spinner/thinking indicators during processing pauses, which resets the timer.

Verdict: safe. The guard targets a narrow fallthrough path and all realistic continuation scenarios set the timestamp within the window.

**Test coverage**

Three tests were modified/added:

1. "starts as response when content appears in IDLE state after recent Claude activity" — renamed and updated to set `_lastClaudeActivity` to `Date.now()`. Validates the happy path.
2. "drops content in IDLE state without recent Claude activity" — new test. Validates the guard works when `_lastClaudeActivity` is 0 (default). Good.
3. Echo expiry test — sets `_lastClaudeActivity` so the expired echo passes through the guard to test echo logic. Necessary fix.

The tests cover the two key branches (within window vs. outside window). Could be strengthened with a boundary test (e.g., set `_lastClaudeActivity = Date.now() - 4999` to test just inside the window), but the current coverage is adequate for the simplicity of the change.

### Verdict

**Approve.** The change is well-scoped, well-documented, and the 5-second window is defensible. All 250 tests pass. CI is green. No issues found that warrant blocking.

blamechris:
## Review Comments Addressed

Copilot reviewed 2/2 changed files and generated **no inline comments**.

| (none)  | N/A    | N/A          |

- Fixed: 0
- False positives: 0
- Follow-up issues created: 0

All CI checks pass (App Type Check, Server Tests, notify).
PR is ready for merge.

PR #157, opened 2026-02-09
Fix startup hint appearing as first Claude message

### LINKED ISSUE #139: Startup hint/auto-suggestion appears as first Claude message
## Description

When connecting to the server, the CLI output's hint/auto-suggestion text appears as the first Claude message in chat view before the user sends anything.

## Root Cause

In `cli-session.js` lines 344-363, the `assistant` event handler emits a `message` event with type `response`. On initial startup, `ctx` is `null` (no active message context), so the filter condition `(!ctx || ...)` passes and the startup noise gets through as a chat message.

## Fix

The `assistant` event should only emit messages when there's an active message context (`ctx` exists) — i.e., when we're actually processing a user request. Startup hints should be filtered.

## Acceptance Criteria

- [ ] No Claude messages appear in chat before the user sends their first message
- [ ] Legitimate Claude responses still display correctly

blamechris:
## Summary
- Filter startup hint/suggestion messages from appearing as chat bubbles
- Only emit `assistant` events when processing an actual user request
- Fixes issue where "Try '...'" text appeared as first message before user input

## Root Cause
The `assistant` event handler in `cli-session.js` was emitting messages when `ctx` was `null` (no active message context). During startup, Claude Code shows hint text like "Try 'help me refactor this function'" which would trigger the `assistant` event before any user interaction, causing it to appear as a chat message.

## Fix
Changed the condition in the `assistant` event handler from `(!ctx || (!ctx.didStreamText && !ctx.hasStreamStarted))` to only emit when `ctx` exists: `(ctx && !ctx.didStreamText && !ctx.hasStreamStarted)`. This ensures startup hints are filtered out while preserving legitimate non-streamed response handling.

## Test plan
- [ ] Connect to server via mobile app
- [ ] Verify no Claude messages appear before sending first user message
- [ ] Send a message and verify response appears correctly
- [ ] Test both streamed and non-streamed response scenarios

Closes #139

### DIFF
--- a/packages/server/src/cli-session.js
+++ b/packages/server/src/cli-session.js
@@ -352,7 +352,7 @@ export class CliSession extends EventEmitter {
         const content = data.message?.content
         if (Array.isArray(content) && ctx) {
           for (const block of content) {
-            if (block.type === 'text' && !ctx.didStreamText && !ctx.hasStreamStarted) {
+            if (block.type === 'text' && ctx && !ctx.didStreamText && !ctx.hasStreamStarted) {
               this.emit('message', {
                 type: 'response',
                 content: block.text,


blamechris:
## Agent Review Suggestion Addressed

Implemented the defense-in-depth ctx check suggested in the review.

**Change:** Added explicit `ctx &&` to inner condition for self-contained logic

```diff
- if (block.type === 'text' && !ctx.didStreamText && !ctx.hasStreamStarted)
+ if (block.type === 'text' && ctx && !ctx.didStreamText && !ctx.hasStreamStarted)
```

**Commit:** `6cc5b00`

**Rationale:** While the outer guard (`if (Array.isArray(content) && ctx)`) already ensures ctx exists, adding the explicit check makes the inner block self-contained and resilient to future refactoring.

All changes committed and pushed.

PR #166, opened 2026-02-09
fix(app): make terminal view text selectable

### LINKED ISSUE #141: Terminal view text is not selectable
## Description

The terminal tab displays output correctly but text cannot be selected or copied.

## Root Cause

`TerminalView.tsx` line 44 renders terminal content in a `Text` component without the `selectable={true}` prop.

## Fix

Add `selectable={true}` to the `Text` component in TerminalView.

## Acceptance Criteria

- [ ] Terminal text can be long-pressed to show copy menu
- [ ] Text selection works on both iOS and Android

blamechris:
## Summary
- Add `selectable={true}` to terminal text component
- Enables long-press to copy and text selection on both iOS and Android

Closes #141

## Test plan
- [ ] Long-press terminal text → shows native copy menu
- [ ] Text selection works on iOS
- [ ] Text selection works on Android

### DIFF
--- a/packages/app/src/components/TerminalView.tsx
+++ b/packages/app/src/components/TerminalView.tsx
@@ -41,7 +41,7 @@ export function TerminalView({ content, scrollViewRef }: TerminalViewProps) {
       keyboardDismissMode="on-drag"
       onContentSizeChange={() => scrollViewRef.current?.scrollToEnd()}
     >
-      <Text selectable={true} style={styles.terminalText}>{processed || 'Connected. Terminal output will appear here...'}</Text>
+      <Text selectable style={styles.terminalText}>{processed || 'Connected. Terminal output will appear here...'}</Text>
     </ScrollView>
   );
 }


blamechris:
## Code Review: PR #166

### Summary
One-line change adding `selectable={true}` to the `<Text>` component in `TerminalView.tsx`, enabling long-press copy and text selection on both iOS and Android. Closes #141.

### Strengths
- **Minimal, focused change.** One prop, one line, one component. Hard to get wrong.
- **Consistent with ChatView.** ChatView already uses `selectable` on message text and tool content (lines 245, 298). This brings TerminalView to parity.
- **Correct component to make selectable.** The `<Text>` inside TerminalView is the right target — it renders the entire processed terminal buffer as a single text node, so selection spans the full output naturally.

### Issues Found

#### Suggestion: Auto-scroll conflicts with text selection (Medium)

ChatView has an `isSelectingRef` guard that suppresses `scrollToEnd()` during multi-select mode:

```tsx
// ChatView.tsx ~line 367
onContentSizeChange={() => {
  if (!isSelectingRef.current) scrollViewRef.current?.scrollToEnd();
}}
```

TerminalView has no such guard:

```tsx
// TerminalView.tsx ~line 43
onContentSizeChange={() => scrollViewRef.current?.scrollToEnd()}
```

When the terminal is actively receiving output (which is most of the time — streaming Claude responses, tool execution), `onContentSizeChange` fires continuously. If the user long-presses to start a text selection, the scroll will jump to the bottom, breaking the selection and dismissing the copy menu.

This is the exact same bug that was fixed in ChatView (commit `e7fd047` — "stop auto-scroll" during selection). The terminal view needs the same treatment.

**Suggested fix:** Accept an `isAutoScrollEnabled` prop (or a ref) from `SessionScreen` and gate the `scrollToEnd` call, just like ChatView does. Alternatively, track a local "user is selecting" state using the `onTextLayout` or `onSelectionChange` callbacks on the `Text` component, though this is more complex.

#### Nitpick: `selectable={true}` vs `selectable`

In JSX, `selectable={true}` and `selectable` are equivalent. The codebase already uses the bare form in ChatView:

```tsx
// ChatView.tsx
<Text selectable style={styles.toolContentExpanded}>
<Text selectable style={[styles.messageText, ...]}>
```

For consistency, use `selectable` without `={true}`.

### Architecture Notes

- **Platform gotchas with `selectable`:**
  - **iOS:** `selectable` on `<Text>` works well. Long-press shows the native callout (Copy/Look Up/Share). No known issues with monospaced fonts or long text blocks.
  - **Android:** `selectable` works but has a historical quirk — on some Android versions (API < 28), very long text in a selectable `<Text>` inside a `ScrollView` can cause janky scrolling because Android's `TextView` with `textIsSelectable` intercepts touch events that conflict with the scroll gesture. React Native 0.76+ (which Expo 54 uses) handles this better, but it is worth testing on a physical Android device during QA.
  - **Neither platform** has issues with `selectable` on a `<Text>` that receives frequent content updates (like streaming terminal output), apart from the auto-scroll conflict noted above — the selection will be cleared when the text content changes, which is expected behavior.

- **Performance:** `selectable` adds a small overhead — React Native wraps the text in a native selectable container. For a single `<Text>` node (even a large one), this is negligible. The real performance concern would be if terminal output grows unbounded (thousands of lines), but that is an existing concern unrelated to this PR.

- **No interaction with multi-select.** The terminal view does not participate in the message multi-select system (that is chat-only). The `selectable` prop provides native OS-level text selection, which is the right UX for raw terminal output.

### Verdict: **Approve with Suggestion**

The change is correct and well-targeted. The auto-scroll conflict is the only real concern — it will cause a poor UX when selecting text during active streaming. That said, the same issue exists today without `selectable` (you simply cannot select at all), so this PR strictly improves the situation. The auto-scroll guard could be addressed in a follow-up.

Approving as-is, with a recommendation to file a follow-up for the scroll guard.

PR #182, opened 2026-02-09
fix(app): replace setTimeout with layout-driven scroll in SessionPicker

### LINKED ISSUE #169: fix(app): replace setTimeout race in SessionPicker auto-scroll with layout-driven scroll
## Context

PR #168 adds auto-scroll to the SessionPicker so the active pill is visible after switching. It uses a `100ms setTimeout` in a `useEffect` to defer the `scrollTo` call until layout settles.

This works for existing sessions (already laid out), but races when a **new session is created and immediately activated** — `onLayout` for the new pill may not fire within 100ms on slower Android devices, resulting in no scroll.

## Proposed Fix

Replace the timer-based approach with a layout-driven scroll. When `onLayout` fires for a pill that matches the current `activeSessionId`, scroll to it immediately:

```tsx
const activeSessionIdRef = useRef(activeSessionId);
activeSessionIdRef.current = activeSessionId;

const handlePillLayout = useCallback((sessionId: string, e: LayoutChangeEvent) => {
  const { x, width } = e.nativeEvent.layout;
  pillLayouts.current.set(sessionId, { x, width });
  if (sessionId === activeSessionIdRef.current) {
    scrollViewRef.current?.scrollTo({ x: Math.max(0, x - 40), animated: true });
  }
}, []);
```

This eliminates the `setTimeout` entirely and handles all cases deterministically.

## Acceptance Criteria

- [ ] New session creation scrolls to the new pill reliably
- [ ] Existing session switching still auto-scrolls
- [ ] No `setTimeout` in the scroll logic

---

Spawned from review of #168.

blamechris:
## Summary

Fixes #169 by replacing the fragile `setTimeout`-based auto-scroll in SessionPicker with a layout-driven approach.

**Changes:**
- Added `pendingScrollRef` to track scroll requests that arrive before layouts are ready
- Modified `scrollToSession` to mark pending scrolls when layout data is unavailable
- Updated `handlePillLayout` to check for pending scrolls and execute them when the layout completes
- Added `onContentSizeChange` handler to trigger scroll when new sessions are added (content grows)
- Removed the 100ms `setTimeout` race condition

**How it works:**
1. When `activeSessionId` changes, the effect immediately calls `scrollToSession`
2. If layout data is available, scroll happens immediately
3. If layout data is not ready, the scroll is marked as pending via `pendingScrollRef`
4. When the pill's `onLayout` fires, it checks if that pill is active or has a pending scroll, and executes it
5. When content size changes (e.g., new session added), scroll to show the active session

This ensures scrolling happens as soon as the layout system provides the necessary measurements, without arbitrary delays.

## Test plan

- [ ] Create multiple sessions (5+) to fill the horizontal scroll area
- [ ] Switch between sessions - active pill should center in viewport
- [ ] Create a new session - new pill should be visible (scrolled into view)
- [ ] Verify no jumpy scrolling or race conditions
- [ ] Test on both iOS and Android

### DIFF
--- a/packages/app/src/components/InputBar.tsx
+++ b/packages/app/src/components/InputBar.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import { View, Text, TextInput, TouchableOpacity, StyleSheet, Platform } from 'react-native';
+import { ICON_ARROW_UP, ICON_SQUARE, ICON_RETURN, ICON_PARAGRAPH } from '../constants/icons';
 
 // -- Props --
 
@@ -67,7 +68,7 @@ export function InputBar({
           accessibilityLabel={enterToSend ? 'Enter key sends message. Tap to switch to newline mode.' : 'Enter key inserts newline. Tap to switch to send mode.'}
           onPress={onToggleEnterMode}
         >
-          <Text style={styles.enterModeText}>{enterToSend ? '\u21B5' : '\u00B6'}</Text>
+          <Text style={styles.enterModeText}>{enterToSend ? ICON_RETURN : ICON_PARAGRAPH}</Text>
         </TouchableOpacity>
         <TextInput
           style={[styles.input, !enterToSend && styles.inputMultiline]}
@@ -85,11 +86,11 @@ export function InputBar({
         />
         {isStreaming ? (
           <TouchableOpacity style={styles.interruptButton} onPress={onInterrupt}>
-            <Text style={styles.interruptButtonText}>{'\u25A0'}</Text>
+            <Text style={styles.interruptButtonText}>{ICON_SQUARE}</Text>
           </TouchableOpacity>
         ) : (
           <TouchableOpacity style={styles.sendButton} onPress={onSend}>
-            <Text style={styles.sendButtonText}>{'\u2191'}</Text>
+            <Text style={styles.sendButtonText}>{ICON_ARROW_UP}</Text>
           </TouchableOpacity>
         )}
       </View>
--- a/packages/app/src/components/SessionPicker.tsx
+++ b/packages/app/src/components/SessionPicker.tsx
@@ -9,6 +9,7 @@ import {
   LayoutChangeEvent,
 } from 'react-native';
 import { useConnectionStore, SessionInfo } from '../store/connection';
+import { ICON_SQUARE } from '../constants/icons';
 
 interface SessionPillProps {
   session: SessionInfo;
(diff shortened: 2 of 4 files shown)


blamechris:
## Code Review: PR #182

### Summary

This PR fixes a race condition in `SessionPicker` auto-scroll by replacing the `100ms setTimeout` approach with a layout-driven scroll mechanism. The core improvement is the introduction of `pendingScrollRef` to track scroll requests that arrive before layouts are ready, ensuring scrolling happens as soon as layout data becomes available.

Additionally, the PR extracts hardcoded Unicode characters into a shared constants file (`icons.ts`) for better maintainability.

### Strengths

- **Eliminates race condition**: The `pendingScrollRef` pattern cleanly handles the case where `activeSessionId` changes before pill layouts are ready
- **Deterministic scrolling**: No arbitrary delays — scroll happens as soon as the layout system provides measurements
- **Handles content growth**: The `onContentSizeChange` handler ensures new sessions scroll into view when added
- **Clean code reuse**: Extracting Unicode constants to `icons.ts` reduces duplication and improves readability
- **TypeScript compliance**: All changes maintain strict typing

### Issues Found

#### Critical (Must Fix)

None.

| File | Lines | Suggestion | Rationale |
|------|-------|------------|-----------|
| `SessionPicker.tsx` | 56-62 | Consider clearing `pendingScrollRef` even when early-returning | If the function returns early (no layout/no scrollViewRef/no viewport), `pendingScrollRef` remains set indefinitely. While harmless in practice (it gets cleared on successful scroll), explicitly clearing it on early return would make the state machine more obvious. |
| `SessionPicker.tsx` | 148-153 | `onContentSizeChange` handler may fire more often than needed | This fires whenever ScrollView content size changes (any pill resize, not just additions). For pure session additions, checking if `sessions.length` changed in a `useEffect` might be more precise. However, the current approach is safe and simple. |

#### Nitpicks (Optional)

- The comment on line 88 says "Auto-scroll active session pill into view when it changes" — consider adding "and layout is ready" to reflect the pending scroll behavior
- Line 66: Comment says "If this pill is the active session or has a pending scroll, scroll to it now" — the logic now handles both cases (`activeSessionId` match and `pendingScrollRef` match), which is great, but worth noting that the two conditions are mutually exclusive in practice (pending scroll is only set when layout is missing)

### Architecture Notes

This change maintains the existing React Native pattern of using `onLayout` callbacks and refs to coordinate UI updates. The `pendingScrollRef` approach is a clean adaptation of the "defer until ready" pattern commonly used in React Native for layout-dependent operations.

The icon constant extraction is a minor DRY improvement that aligns with the existing codebase pattern (other constants files in the project).

None created — the suggestions above are minor optimizations that don't warrant separate issues.

### Verdict

This PR correctly solves the race condition described in #169 and introduces a more robust auto-scroll mechanism. The code is clean, well-structured, and follows project conventions. No blocking issues identified.

PR #264, opened 2026-02-09
fix(app): don't clear saved connection on retry exhaustion

blamechris:
## Summary
- Stop clearing saved URL/token from SecureStore when health-check retries are exhausted
- Add "Retry" and "Forget Server" buttons to the connection failed alert
- Only the explicit "Forget Server" action clears the saved connection

## Context
With stable tunnel URLs (Named Tunnels), clearing the saved connection on timeout destroys the pairing the user set up. The server might just be temporarily down — the user should be able to retry without re-scanning the QR code.

## Test plan
- [ ] Kill server, wait for retries to exhaust, verify alert shows Retry/Forget/OK
- [ ] Tap "Retry" — connection attempt restarts
- [ ] Tap "Forget Server" — saved connection is cleared, returns to ConnectScreen
- [ ] Tap "OK" — dismisses alert, saved connection preserved for next app launch
- [ ] App type check passes: `cd packages/app && npx tsc --noEmit`

### DIFF
--- a/packages/app/src/store/connection.ts
+++ b/packages/app/src/store/connection.ts
@@ -508,7 +508,7 @@ export const useConnectionStore = create<ConnectionState>((set, get) => ({
             'Could not reach the Chroxy server. Make sure it\'s running.',
             [
               { text: 'Retry', onPress: () => get().connect(url, token, 0) },
-              { text: 'Forget Server', style: 'destructive', onPress: () => { clearConnection(); set({ savedConnection: null }); } },
+              { text: 'Forget Server', style: 'destructive', onPress: () => { void get().clearSavedConnection(); } },
               { text: 'OK', style: 'cancel' },
             ],
           );


blamechris:
## Code Review: PR #264

### Summary
This PR addresses a critical UX issue where retry exhaustion on initial connection attempts incorrectly cleared the saved connection credentials. With stable Named Tunnels, the server may be temporarily unreachable, and destroying the saved pairing forces the user to re-scan the QR code unnecessarily. The fix adds explicit user control via "Retry" and "Forget Server" buttons in the failure alert.

### Strengths
1. **Correct problem identification**: The original behavior was indeed wrong for Named Tunnels — clearing saved credentials on timeout is too aggressive when the server is just temporarily down.

2. **Clean UX flow**: Three-button alert (Retry/Forget Server/OK) gives the user appropriate control:
   - Retry: immediate reconnection attempt
   - Forget Server: explicit credential clearing with destructive styling
   - OK: dismisses alert, preserves credentials for next app launch

3. **Minimal code change**: Only modifies the retry exhaustion branch, doesn't affect auto-reconnect or other connection flows.

4. **Good alert copy**: Simplified message removes the "scan the QR code again" instruction, which is now only needed if the user explicitly chooses "Forget Server".

### Issues Found

#### Critical
None.

#### Suggestions
1. **Retry button resets retry count**: The Retry button calls `get().connect(url, token, 0)`, which resets the retry counter. This is reasonable but means the user can trigger another full 6-attempt cycle. Consider whether you want to:
   - Keep current behavior (fresh 6-attempt cycle on manual retry)
   - OR resume from the exhausted count (though this would make the button useless)
   
   Current behavior seems correct, but worth documenting in a code comment.

2. **Alert on app foreground**: The `AppState.addEventListener` at line 1299 will attempt reconnect if the socket is stale when the app resumes. If that reconnect exhausts retries while the app is in foreground, the user will see this alert. Good UX, just worth confirming in testing.

3. **Auto-reconnect on unexpected disconnect**: Lines 998-1005 and 1015-1022 handle auto-reconnect for unexpected disconnections. These flows do NOT have retry limits — they call `connect(url, token)` without a retry count, so they'll run through the full 6-attempt cycle on every auto-reconnect. This is probably correct (you want aggressive reconnection for established connections), but the divergent behavior between initial connection and auto-reconnect could be documented.

#### Nitpicks
1. **Alert button order**: iOS HIG typically puts the primary action on the right (OK), destructive in the middle, and alternative on the left. Current order is [Retry, Forget Server, OK]. Consider [OK, Retry, Forget Server] or [Forget Server, Retry, OK] for more conventional alignment. Not critical — React Native Alert button ordering varies by platform.

2. **Destructive style on Forget**: Good use of `style: 'destructive'` on the Forget Server button.

### Verdict
**Approve**

This is a well-targeted fix that solves a real UX problem. The code is clean, the behavior is correct, and the only suggestions are documentation/minor UX polish. The change is scoped appropriately and doesn't introduce new risks.

**Test coverage notes:**
- Manual testing checklist in PR description is comprehensive
- Covers all three alert button paths
- Includes app type check verification
- Consider adding a test case for app foregrounding during retry exhaustion (suggestion #2 above)

**Ready to merge** once test plan is executed.

blamechris:
## Review Comments Addressed

| Use centralized clearSavedConnection action instead of inline duplication | Fixed | `c48d931` |

- Fixed: 1
- False positives: 0
- Follow-up issues created: 0

PR #316, opened 2026-02-09
fix(app): disable prompt buttons after selection, refactor PermissionDetail

### LINKED ISSUE #312: Disable AskUserQuestion option buttons after selection
## Context

From PR #304 review. When the user taps an option on an AskUserQuestion prompt bubble, the response is sent to the server, but the prompt bubble remains fully interactive — all buttons are still tappable.

## Impact

- A second tap calls `sendUserQuestionResponse` again. The server silently drops it (`_waitingForAnswer` is already false), so no harm is done, but the UX is confusing.
- The user has no visual feedback that their selection was registered.

## Suggested fix

After an option is selected:
1. Visually indicate the selected option (highlight or checkmark)
2. Disable the remaining buttons (gray them out or remove them)
3. Consider replacing the options with a static "You selected: X" label

This mirrors how permission prompts should ideally behave after Allow/Deny.

### LINKED ISSUE #313: Refactor PermissionDetailOrFallback to avoid direct component function call
In `ChatView.tsx`, `PermissionDetailOrFallback` calls `PermissionDetail` as a plain function:

```tsx
const detail = PermissionDetail({ tool, toolInput });
if (detail) return detail;
```

This bypasses React reconciliation (no fiber node). It works because `PermissionDetail` is currently pure (no hooks), but it is fragile — adding any hook to the component later would silently break.

**Options:**
1. Rename to `renderPermissionDetail` (non-component helper convention) to make the intent explicit
2. Refactor to use proper JSX with a `hasDetail` boolean check

Either approach prevents future hook-related bugs.

Ref: PR #303

blamechris:
## Summary

- Prompt option buttons now disable after the user taps one, preventing double-sends
- The chosen option gets highlighted (solid orange bg with white text), non-chosen options fade out
- Adds `answered` field to `ChatMessage` and `markPromptAnswered` store action
- Renames `PermissionDetail` to `renderPermissionDetail` to clarify it's a plain helper function, not a React component (avoids reconciliation issues if hooks are ever added)

Closes #312, closes #313

## Test plan

- [ ] Tap a permission prompt option — chosen button highlights, others fade and all become disabled
- [ ] Tap an AskUserQuestion option — same disable/highlight behavior
- [ ] Reconnect to a session with answered prompts — buttons remain disabled
- [ ] `npx tsc --noEmit` passes

### DIFF
--- a/packages/app/src/components/ChatView.tsx
+++ b/packages/app/src/components/ChatView.tsx
@@ -408,7 +408,7 @@ function MessageBubble({ message, onSelectOption, isSelected, isSelecting, onLon
       {isPrompt && message.options && (
         <View style={styles.promptOptions}>
           {message.options.map((opt, i) => {
-            const isAnswered = !!message.answered;
+            const isAnswered = message.answered != null;
             const isChosen = message.answered === opt.value;
             return (
               <TouchableOpacity


blamechris:
## Code Review: PR #316

### Summary

This PR addresses two issues: (1) disabling prompt option buttons after the user taps one (preventing double-sends with visual feedback), and (2) renaming `PermissionDetail` to `renderPermissionDetail` to clarify it is a plain helper function, not a React component. Clean, focused changes across three files (ChatView, SessionScreen, connection store).

### Strengths

- **Correct state approach**: Using an `answered` field on `ChatMessage` rather than local component state means the answer survives re-renders and navigation between chat/terminal views within the same session.
- **Good visual feedback**: Chosen option highlighted with solid orange + white text, non-chosen options faded out — clear, unambiguous UX.
- **Proper Zustand pattern**: `markPromptAnswered` uses `updateActiveSession` with immutable map, consistent with the rest of the store.
- **Rename is well-motivated**: Calling `PermissionDetail({ tool, toolInput })` as a plain function was fragile. Renaming to `renderPermissionDetail` makes intent explicit and prevents future hook bugs (exactly as #313 described).

### Issues Found

#### Critical (Must Fix)

| `packages/app/src/components/ChatView.tsx` | 274 | **Stale `PermissionDetail` component still exists on main** — the diff shows the rename to `renderPermissionDetail` at line ~272, but also shows `PermissionDetailOrFallback` calling the renamed function. However, the original `PermissionDetail` function component declaration (line 274 on main) must be fully replaced, not just have the call site changed. Verify the diff actually replaces the function declaration `function PermissionDetail(` with `function renderPermissionDetail(` — looking at the diff more carefully, it does. No issue here. | N/A — false alarm after closer read. |

*No critical issues.*

| `packages/app/src/components/ChatView.tsx` | ~93 | The `ChatViewProps.onSelectOption` signature was `(value, requestId?, toolUseId?)` and is now `(value, messageId, requestId?, toolUseId?)`. This puts `messageId` as the second positional arg. Consider whether a named options object `{ value, messageId, requestId?, toolUseId? }` would be more maintainable long-term, since this signature has grown to 4 params. | Positional optionals after a required param can be error-prone if the signature grows further. Low priority — current usage is fine. |
| `packages/app/src/store/connection.ts` | ~1295 | `markPromptAnswered` maps over ALL messages to find the one matching `messageId`. For typical conversations this is fine, but prompt messages could also be looked up by index or stored in a Map for O(1) access. | Minor — message arrays are small in practice. Not actionable now. |

#### Nitpicks (Optional)

- The `answered` field stores the raw option `value` string (e.g., `"allow"`, `"deny"`). This is the right choice — storing the label would couple display text to state, and the value is what was actually sent to the server.
- The `promptOptionChosen` and `promptOptionDisabled` style names are clear and consistent with existing naming conventions. No notes.

| `answered` state does not survive full reconnect | See below | Server replays raw `permission_request` / `user_question` events — the app reconstructs `ChatMessage` objects without the `answered` field. After reconnect, previously answered prompts re-appear as active. This is acceptable for now (answered prompts are in the past and the server won't re-ask), but a future improvement could have the server include an `answered` flag in replayed permission events. |

### Architecture Notes

The change threads `messageId` through the callback chain cleanly: `ChatView` -> `MessageBubble` -> `SessionScreen.handleSelectOption` -> `markPromptAnswered`. The store action fires synchronously before the WebSocket send, which means the UI updates instantly even if the network is slow — correct behavior.

The `renderPermissionDetail` rename is the right call. It was the only "component" in the file that was called as a plain function rather than rendered as JSX. Making it a lowercase helper function removes any ambiguity about React reconciliation expectations.

### Verdict

Clean, well-scoped fix for both issues. No critical problems. The reconnect limitation around `answered` state is pre-existing and acceptable for v0.1.

PR #328, opened 2026-02-10
feat(server): record user_question events in session history

### LINKED ISSUE #309: Record user_question events in session history for replay
## Context

From PR #304 review. The `user_question` event is added to `PROXIED_EVENTS` in `session-manager.js`, so it gets emitted via `session_event`. However, `_recordHistory()` has no `case 'user_question'` handler, so the event falls through the switch statement without being persisted.

## Impact

If a client reconnects while a `user_question` prompt is pending, the prompt won't be replayed — the user sees no prompt and the session appears stuck waiting.

## Suggested fix

Add a `case 'user_question'` in `_recordHistory()` that stores the event (with `toolUseId` and `questions`) in the session history buffer. On replay, `_replayHistory` in `ws-server.js` will forward it to the reconnecting client.

Consider also tracking `_waitingForAnswer` state so the replay can determine if the question is still pending vs already answered.

blamechris:
## Summary

- Adds `case 'user_question'` to `_recordHistory()` in session-manager.js
- Stores `toolUseId` and `questions` so the prompt gets replayed on reconnect
- Without this, a pending AskUserQuestion prompt is lost if the client disconnects

Closes #309

## Test plan

- [ ] Server tests pass
- [ ] Disconnect while AskUserQuestion prompt is pending, reconnect — prompt is replayed
- [ ] App marks replayed prompts as answered (via #322 fix)

### DIFF
--- a/packages/server/src/session-manager.js
+++ b/packages/server/src/session-manager.js
@@ -467,6 +467,15 @@ export class SessionManager extends EventEmitter {
           timestamp: Date.now(),
         })
         break
+
+      case 'user_question':
+        this._pushHistory(history, {
+          type: 'user_question',
+          toolUseId: data.toolUseId,
+          questions: data.questions,
+          timestamp: Date.now(),
+        })
+        break
     }
   }
 


blamechris (commented):
## Agent Review — PR #328

**Scope:** 1 file changed, +9 lines — `packages/server/src/session-manager.js`

### What it does

Adds a `case 'user_question'` to `_recordHistory()` in `SessionManager`, so that `AskUserQuestion` prompts are persisted in the session's message history. This enables replay on reconnect: if a client disconnects while a question prompt is pending, the prompt will be re-sent when the client reconnects.

### Correctness

- **Data shape matches the emitter.** `cli-session.js` emits `{ toolUseId, questions }` (line 405-408). The new case stores exactly those two fields plus a timestamp. This is consistent with how `ws-server.js` broadcasts the event (line 879-884) and how the app parses it (`connection.ts` line 1014+).
- **Replay path works.** `_replayHistory()` sends history entries directly via `this._send(ws, entry)`. The stored entry has `type: 'user_question'`, which the app's WebSocket handler dispatches on correctly. After replay, `history_replay_end` marks all prompt-type messages as `answered: '(resolved)'`, which handles the already-resolved case.
- **Event is already in `PROXIED_EVENTS`.** Line 487 of `session-manager.js` already includes `'user_question'` in the proxied events array, so `_recordHistory` is called for this event — the only missing piece was the `case` branch, which this PR adds.
- **Pattern consistency.** The new case follows the exact same pattern as `tool_start` and `result` (fields from `data` + `Date.now()` timestamp), fitting cleanly into the existing switch structure.

### Style

- ES modules, no semicolons, single quotes — matches server code style.
- No attribution in commit message or PR body.
- Branch naming (`feat/user-question-history`) follows conventions.
- Commit message format (`type(scope): summary`) is correct.

### CI

- Server tests: pass
- App type check: pass

### Verdict

Clean, minimal, correct change. No issues found.

PR #329, opened 2026-02-10
feat(server): add prefix filter to discoverTmuxSessions

### LINKED ISSUE #308: feat(server): add chroxy- prefix filtering to session discovery
## Context

PR #302 introduced the `chroxy wrap` command, which creates tmux sessions with a `chroxy-` prefix convention. Currently, `session-discovery.js` discovers all tmux sessions running a target process, with no way to filter by managed vs. unmanaged sessions.

## Proposal

Add an optional filter to `discoverTmuxSessions()` (or a separate function) that can restrict discovery to `chroxy-` prefixed sessions. This would allow the app to distinguish between:

- **Managed sessions** — created by `chroxy wrap`, prefixed with `chroxy-`
- **Discovered sessions** — any tmux session running the target process

This could be a boolean parameter, a prefix string parameter, or separate list endpoints.

## Motivation

As users adopt `chroxy wrap` for session management, being able to filter the session list in the app becomes useful — especially if the user has other tmux sessions running the target process for unrelated purposes.

## Related

- PR #302 (introduced `chroxy-` prefix convention)

blamechris:
## Summary

- Adds optional `{ prefix }` parameter to `discoverTmuxSessions()`
- When provided, only returns sessions whose name starts with the prefix
- Useful for filtering chroxy-managed sessions (`chroxy-*`) from unrelated tmux sessions
- Backward compatible — no prefix means all sessions returned (existing behavior)

Closes #308

## Test plan

- [ ] Existing discovery tests pass (7/7)
- [ ] `discoverTmuxSessions()` returns all sessions (backward compat)
- [ ] `discoverTmuxSessions({ prefix: 'chroxy-' })` filters correctly

### DIFF
--- a/packages/server/src/session-discovery.js
+++ b/packages/server/src/session-discovery.js
@@ -6,9 +6,11 @@ import { execSync } from 'child_process'
  * Scans all tmux panes, checks their child processes for claude,
  * and returns structured session info for the app to display.
  *
+ * @param {{ prefix?: string }} [options] - Optional filter options
+ * @param {string} [options.prefix] - Only return sessions whose name starts with this prefix (e.g. 'chroxy-')
  * @returns {Array<{ sessionName: string, cwd: string, pid: number }>}
  */
-export function discoverTmuxSessions() {
+export function discoverTmuxSessions({ prefix } = {}) {
   try {
     // Check if tmux is available
     execSync('which tmux', { stdio: 'pipe' })
@@ -51,6 +53,9 @@ export function discoverTmuxSessions() {
       }
     }
 
+    if (prefix) {
+      return results.filter((s) => s.sessionName.startsWith(prefix))
+    }
     return results
   } catch (err) {
     console.error(`[discovery] Failed to discover tmux sessions: ${err.message}`)


blamechris (commented):
## Agent Review

**PR:** feat(server): add prefix filter to discoverTmuxSessions
**Scope:** 1 file changed, +6/-1

### Assessment: LGTM

This is a clean, minimal change that adds an optional `{ prefix }` parameter to `discoverTmuxSessions()` for filtering sessions by name prefix. The change is well-scoped and backward compatible.

### Checklist

- [x] **Code style**: Matches project conventions (ES modules, no semicolons, single quotes, arrow functions)
- [x] **Backward compatibility**: Default parameter `{ prefix } = {}` ensures existing callers (session-manager.js line 317) continue to work without changes
- [x] **JSDoc**: Updated with proper `@param` tags for the options object and prefix field
- [x] **Branch naming**: `feat/discovery-prefix-filter` follows convention
- [x] **Commit message**: `feat(server): add prefix filter to discoverTmuxSessions` follows `type(scope): summary` format
- [x] **No attribution violations**: Clean commit, no AI mentions
- [x] **CI passing**: All checks green (Server Tests, App Type Check)

### Minor observations (non-blocking)

1. **Filter placement**: The prefix filter is applied after the full scan (all tmux panes are still enumerated, child processes checked, etc.). For the current use case this is fine -- the optimization of filtering early (before `_findClaudeChild` calls) would add complexity without meaningful performance gain given the expected number of tmux sessions.

2. **No new tests**: The PR body references "Existing discovery tests pass (7/7)" which confirms backward compatibility. The existing tests cover SessionManager's auto-discovery behavior rather than `discoverTmuxSessions()` directly. A unit test for prefix filtering would strengthen confidence but is not blocking for a change this small.

3. **Truthy check on prefix**: Using `if (prefix)` means an empty string `''` would be treated as "no filter" which is the intuitive behavior. Good.

### Verdict

Small, focused, backward-compatible change. No issues found.

PR #337, opened 2026-02-10
feat: background session state sync for multi-session

### LINKED ISSUE #148: Background session state sync (inactive sessions miss real-time updates)
## Description

Currently, the server only broadcasts messages to clients viewing the active session (`_broadcastToSession` filters by `client.activeSessionId`). Messages for inactive sessions are stored in server history but the client cache becomes stale.

This is the architectural root cause behind #145. While #145 fixes the symptom (replay dedup too aggressive), a better long-term solution is to keep all sessions synced in real-time.

### Option A: Broadcast all session events to connected clients
- Remove the `activeSessionId` filter from `_broadcastToSession` for message events
- Tag each message with its `sessionId`
- Client routes messages to the correct `sessionStates[sessionId]`
- Pro: Simple, real-time
- Con: More bandwidth (fine for 2-5 sessions, could be an issue at scale)

### Option B: Delta sync on session switch
- On switch, server sends a delta (messages since last switch-away timestamp)
- Client merges delta into cached state
- Pro: Bandwidth efficient
- Con: More complex server-side tracking

### Option C: Server-authoritative state
- On switch, server sends full session state
- Client replaces cached state entirely
- Pro: Simplest correctness guarantee
- Con: Redundant data transfer

## Recommendation

Option A for now (2-5 sessions is the expected range). Revisit if session count grows.

## Acceptance Criteria

- [ ] Inactive session state stays up-to-date on the client
- [ ] No duplicate messages when switching sessions
- [ ] Minimal latency on session switch

blamechris:
## Summary
- Broadcast session-scoped WS events to all connected clients (not just active session viewers) so inactive sessions stay in sync
- Session switching is now instant with always-fresh cached state — no stale messages
- Raw PTY data and `status_update` remain filtered to active session only (bandwidth)

## Changes

**Server (`ws-server.js`):**
- `_broadcastToSession` now tags every message with `sessionId` and sends to all authenticated clients
- Custom filter callbacks restrict `raw`, `raw_background`, and `status_update` to active session

**App (`connection.ts`):**
- New `updateSession(sessionId, updater)` helper routes messages to the correct session by `msg.sessionId`
- All 12 session-scoped WS handlers updated to use `msg.sessionId || get().activeSessionId`
- `updateActiveSession` simplified to delegate to `updateSession`
- Flat state only syncs when the target is the active session

**Tests:** 4 new tests covering background delivery, sessionId tagging, stream lifecycle, and raw PTY filtering

## Test plan
- [ ] All 386 server tests pass (4 new)
- [ ] TypeScript type check passes
- [ ] Multi-session: start two sessions, work in session A, switch to B and back — messages should be up-to-date
- [ ] Raw terminal data should only appear for the active session
- [ ] Session switching should be instant (no replay delay)

Closes #148

### DIFF
--- a/packages/app/src/store/connection.ts
+++ b/packages/app/src/store/connection.ts
@@ -721,7 +721,8 @@ export const useConnectionStore = create<ConnectionState>((set, get) => ({
           _receivingHistoryReplay = false;
           _isSessionSwitchReplay = false;
           // Mark all replayed prompts as answered — any prompt in history
-          // has already been resolved by the server
+          // has already been resolved by the server.
+          // Note: replay is always for the active session (connect or switch).
           updateActiveSession((ss) => {
             const hasUnansweredPrompts = ss.messages.some(
               (m) => m.type === 'prompt' && !m.answered
@@ -964,7 +965,11 @@ export const useConnectionStore = create<ConnectionState>((set, get) => ({
           }
           break;
 
-        case 'status_update':
+        case 'status_update': {
+          // Server filters status_update to active session only, but defend
+          // against misrouted messages on the client side too.
+          const statusSid = msg.sessionId || get().activeSessionId;
+          if (statusSid && statusSid !== get().activeSessionId) break;
           set({
             claudeStatus: {
               cost: msg.cost,
@@ -976,6 +981,7 @@ export const useConnectionStore = create<ConnectionState>((set, get) => ({
             },
           });
           break;
+        }
 
         case 'raw':
           get().appendTerminalData(msg.data);
@@ -1054,7 +1060,8 @@ export const useConnectionStore = create<ConnectionState>((set, get) => ({
         }
 
         case 'server_status': {
-          // Non-error status update (e.g., tunnel recovery notifications)
+          // Non-error status update (e.g., tunnel recovery notifications).
+          // Global broadcast (no sessionId) — route to active session.
           const statusMessage: string =
             typeof msg.message === 'string' && msg.message.trim().length > 0
               ? stripAnsi(msg.message)
@@ -1078,6 +1085,7 @@ export const useConnectionStore = create<ConnectionState>((set, get) => ({
         }
 
         case 'server_error': {
+          // Global broadcast (no sessionId) — route to active session.
           // Validate and coerce untyped JSON fields
           const allowedCategories = new Set<ServerError['category']>([
             'tunnel', 'session', 'permission', 'general',


blamechris:
## Code Review: PR #337

### Summary
This PR enables background session state synchronization for multi-session scenarios. Previously, `_broadcastToSession` only sent messages to clients actively viewing that session. Now, most session-scoped events are broadcast to all authenticated clients with a `sessionId` tag, allowing the app to keep inactive session caches up to date. Raw PTY data and `status_update` remain filtered to active-session-only clients to save bandwidth. On the app side, all 12 session-scoped WS handlers now route messages using `msg.sessionId || get().activeSessionId` and the new `updateSession(sessionId, updater)` helper.

### Strengths
- Clean separation between "broadcast to all" (metadata, chat messages, streaming) and "active-only" (raw PTY, status_update) is well-reasoned and bandwidth-aware
- The `updateSession` helper is a natural refactoring that avoids code duplication while preserving the flat-state sync behavior for the active session
- The `result` event on line 864 already had `sessionId: data.sessionId` in its payload, and the new `_broadcastToSession` tagging via `{ ...message, sessionId }` consistently overwrites it with the canonical session ID from the event envelope -- this is safe because they come from the same source
- Server-side delta buffering keyed by `sessionId:messageId` composite already handles concurrent streams from multiple sessions correctly
- 4 new tests cover the critical scenarios: tagging, background delivery, stream lifecycle, and raw PTY filtering
- All 50 ws-server tests pass; TypeScript type check passes clean

### Issues Found

#### Critical (Must Fix)

None.

| `packages/app/src/store/connection.ts` | 967-978 | `status_update` handler sets flat `claudeStatus` unconditionally without checking `msg.sessionId` | This currently works because the server filters `status_update` to active-session-only clients. But if the server filter were ever relaxed (or another server mode added), stale context data from a background session would overwrite the active session's status bar. Consider either (a) guarding with `if (!msg.sessionId \|\| msg.sessionId === get().activeSessionId)` on the client, or (b) making `claudeStatus` per-session in `SessionState`. Defense in depth. |
| `packages/app/src/store/connection.ts` | 1069-1077 | `server_status` handler still uses `updateActiveSession` instead of `updateSession(msg.sessionId \|\| ...)` | This is inconsistent with the pattern used by the other 12 handlers in this PR. Since `server_status` is broadcast via `_broadcast` (not `_broadcastToSession`), it currently lacks a `sessionId` tag, so this is not a bug -- but it breaks the otherwise uniform pattern established by this PR. |
| `packages/app/src/store/connection.ts` | 1113-1122 | `server_error` handler also still uses `updateActiveSession` | Same as above. If server errors originate from a specific session (e.g., crash), they should ideally route to that session's message list. Currently uses active session, which could misattribute errors. |
| `packages/app/src/store/connection.ts` | 720-738 | `history_replay_end` uses `updateActiveSession` without checking `msg.sessionId` | During session-switch replay, this should route to the target session, not necessarily the active session. In practice these are currently the same (you switch then replay), but it's fragile if replay ordering ever changes. |

#### Nitpicks (Optional)
- Line 864 in `ws-server.js`: The `result` message includes both `sessionId: data.sessionId` (explicit) and the `sessionId` added by `_broadcastToSession` tagging. The explicit one is always overwritten. Consider removing `sessionId: data.sessionId` from the message object to avoid confusion about where the sessionId comes from.
- The comment dedup cleanup (removing the "Strengthen signature" comment on the dedup check around line 749) is fine but could be a separate commit for cleaner history.

| Make `claudeStatus` per-session in `SessionState` | Requires touching the SettingsBar component and session switch logic. Out of scope for this sync PR. |
| Align `server_status` / `server_error` / `history_replay_end` handlers with session-routed pattern | These handlers use server-wide broadcasts (not session-scoped), so they don't receive `sessionId` tags today. Aligning them requires server-side changes too. |

### Architecture Notes
This change correctly distinguishes between high-bandwidth, display-coupled events (raw PTY data, status bar updates) and state-building events (messages, stream lifecycle, model/permission changes). The former remain filtered to active-session clients, while the latter broadcast to all clients so background session caches stay warm. This means session switching becomes instant -- no replay needed.

The `_broadcastToSession` change from "filter to active session" to "broadcast to all + tag" is the right architectural choice. The app's `updateSession` helper cleanly handles the routing, and the flat-state sync only fires for the active session, so UI performance is unaffected.

The delta batching on both server (50ms flush) and client (100ms flush via `flushPendingDeltas`) correctly uses `sessionId` for routing, preventing cross-session contamination during concurrent streams.

### Verdict

The core changes are solid and well-tested. The suggestions above are defensive improvements, not blockers. The PR achieves its stated goal of instant session switching with always-fresh cached state.

Issue #82, opened 2026-02-08
feat(app): background agent/task monitoring UI


blamechris:
## Background

When Claude Code runs background agents (via the Task tool), the Claude Code CLI shows an agent count in its status bar (e.g., "1 local agent"). These background agents are a core part of how Claude Code parallelizes work — spawning sub-agents for code review, search, test execution, etc.

Currently, Chroxy has no visibility into these background processes. Users see tool bubbles for the parent invocation but can't tell how many agents are running, their progress, or when they complete.

## Problem

From the mobile app, there's no way to:
1. See how many background agents are currently running
2. Know what each agent is doing (description/status)
3. See when agents complete or fail
4. Distinguish a Task tool invocation (agent spawn) from a regular tool use

### What exists today

- `cli-session.js` handles `content_block_start` with `type: 'tool_use'` and emits `tool_start` events
- `ws-server.js` forwards `tool_start` messages to the app with `tool` name and `input`
- `connection.ts` handles `tool_start` but treats all tools identically (lines 651-670)
- The stream-json output from `claude -p` includes tool use events, but we don't distinguish Task tool from others

### What's missing

- No detection of Task tool specifically (agent spawning)
- No tracking of active agent count or individual agent status
- No WS message types for agent lifecycle (`agent_spawned`, `agent_completed`, etc.)
- No app-side state for active agents
- No UI for displaying running agents

### Phase 1: Detect and track agents (server)

In `cli-session.js`, when a `tool_use` block starts with `tool: 'Task'`:
- Extract agent description from the tool input
- Track it as an active agent with a generated ID
- Emit a new `agent_spawned` event
- When the corresponding `content_block_stop` arrives, emit `agent_completed`

New events to emit from CliSession:
```javascript
// Agent spawned
{ type: 'agent_spawned', agentId, description, timestamp }

// Agent completed  
{ type: 'agent_completed', agentId, timestamp }

// Agent count changed (convenience)
{ type: 'agent_count', active: 2, total: 5 }
```

### Phase 2: WS protocol messages (server → app)

New message types in `ws-server.js`:
```javascript
// Server → Client
{ type: 'agent_spawned', agentId, description }
{ type: 'agent_completed', agentId }
{ type: 'agent_count', active, total }
```

### Phase 3: App state and UI

**Store** (`connection.ts`):
```typescript
interface AgentInfo {
  id: string
  description: string
  startedAt: number
  status: 'running' | 'completed'
}

// Add to ConnectionState
[...]

PR #359, opened 2026-02-10
feat(server,app): add background agent/task monitoring UI (#82)

blamechris:
## Summary

- Detect `Task` tool usage in `cli-session.js` and track active background agents by `toolUseId`
- Emit `agent_spawned`/`agent_completed` lifecycle events through the WS protocol (transient, not recorded in history)
- Display purple agent count badge in collapsed SettingsBar and expandable agent list with descriptions
- Resilient: agents cleared on process crash, destroy, and reconnect; dedup guard prevents duplicates

## Changes

| File | What changed |
|------|-------------|
| `packages/server/src/cli-session.js` | `_activeAgents` Map, Task tool detection at `content_block_stop`, emit spawned/completed at result/crash/destroy |
| `packages/server/src/session-manager.js` | Wire `agent_spawned`/`agent_completed` (transient, no history) |
| `packages/server/src/ws-server.js` | Forward events in session + legacy modes |
| `packages/app/src/store/connection.ts` | `AgentInfo` type, `activeAgents[]` state, WS handlers with dedup, stale-clear on replay, `updateSession` empty-patch optimization |
| `packages/app/src/components/SettingsBar.tsx` | Agent badge (collapsed), agent list (expanded), purple accent styling, a11y labels |
| `packages/app/src/constants/colors.ts` | `accentPurpleSubtle` color |
| `packages/app/src/screens/SessionScreen.tsx` | `activeAgents` selector + prop wiring |
| `CLAUDE.md` | Protocol docs: `agent_spawned`, `agent_completed` semantics |

## Test Plan

- [x] Server tests pass (392/392)
- [x] App TypeScript type check clean
- [ ] Manual: run server, connect app, trigger Task tool usage — verify purple badge appears
- [ ] Manual: expand SettingsBar — verify agent descriptions with purple dots
- [ ] Manual: badge clears when turn completes (result fires)
- [ ] Manual: interrupt mid-agents — badge clears (_clearMessageState)
- [ ] Manual: reconnect — stale agents cleared (history_replay_start)

### DIFF
--- a/packages/app/src/store/connection.ts
+++ b/packages/app/src/store/connection.ts
@@ -1087,11 +1087,14 @@ export const useConnectionStore = create<ConnectionState>((set, get) => ({
         case 'agent_completed': {
           const completeTargetId = msg.sessionId || get().activeSessionId;
           if (completeTargetId && get().sessionStates[completeTargetId]) {
-            updateSession(completeTargetId, (ss) => ({
-              activeAgents: ss.activeAgents.filter(
+            updateSession(completeTargetId, (ss) => {
+              const filtered = ss.activeAgents.filter(
                 (a) => a.toolUseId !== msg.toolUseId
-              ),
-            }));
+              );
+              // Skip no-op update if agent wasn't tracked (e.g. duplicate event)
+              if (filtered.length === ss.activeAgents.length) return {};
+              return { activeAgents: filtered };
+            });
           }
           break;
         }
--- a/packages/server/src/cli-session.js
+++ b/packages/server/src/cli-session.js
@@ -211,14 +211,6 @@ export class CliSession extends EventEmitter {
 
       if (this._destroying) return
 
-      // Emit completions for any tracked agents before clearing state
-      if (this._activeAgents.size > 0) {
-        for (const agent of this._activeAgents.values()) {
-          this.emit('agent_completed', { toolUseId: agent.toolUseId })
-        }
-        this._activeAgents.clear()
-      }
-
       // Safety net: if we were mid-message, close the stream
       if (this._isBusy && this._currentMessageId) {
         if (this._currentCtx?.hasStreamStarted) {
@@ -487,14 +479,6 @@ export class CliSession extends EventEmitter {
           this.emit('stream_end', { messageId })
         }
 
-        // Emit completions for any tracked background agents
-        if (this._activeAgents.size > 0) {
-          for (const agent of this._activeAgents.values()) {
-            this.emit('agent_completed', { toolUseId: agent.toolUseId })
-          }
-          this._activeAgents.clear()
-        }
-
         this.emit('result', {
           sessionId: data.session_id,
           cost: data.total_cost_usd,
@@ -517,10 +501,15 @@ export class CliSession extends EventEmitter {
     this._waitingForAnswer = false
     this._currentMessageId = null
     this._currentCtx = null
-    // Safety net: callers (close handler, result handler, destroy) emit
-    // agent_completed and clear the Map before calling this method, so this
-    // is typically a no-op. Kept as a defensive guard against future callers.
-    this._activeAgents.clear()
+    // Emit completions for any tracked agents so the app clears badges.
+    // Centralised here so every callsite (result, crash, timeout, interrupt,
+    // destroy) is safe by default.
+    if (this._activeAgents.size > 0) {
+      for (const agent of this._activeAgents.values()) {
+        this.emit('agent_completed', { toolUseId: agent.toolUseId })
+      }
+      this._activeAgents.clear()
+    }
     if (this._resultTimeout) {
       clearTimeout(this._resultTimeout)
       this._resultTimeout = null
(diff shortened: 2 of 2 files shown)


blamechris:
## Code Review: PR #359

### Summary

This PR adds background agent/task monitoring to the Chroxy app. When Claude Code's `Task` tool is invoked (spawning a sub-agent), the server detects it via `content_block_stop` events, tracks active agents in a `Map`, and emits `agent_spawned`/`agent_completed` lifecycle events over the WebSocket protocol. The app displays a purple badge in the collapsed SettingsBar and an expandable agent list with descriptions. Agents are cleared on process crash, destroy, reconnect, and turn completion.

### Strengths

- **Clean event lifecycle**: Three distinct cleanup paths (result, close, destroy) each emit `agent_completed` before clearing state -- good defensive programming pattern.
- **Dedup guard**: Both server-side (`Map.set` with toolUseId key) and client-side (`.some()` check before append) prevent duplicate tracking.
- **Transient by design**: Correctly excluded from history recording in `session-manager.js`, and stale entries cleared on `history_replay_start`. This is the right call for ephemeral UI state.
- **Empty-patch optimization**: The `Object.keys(patch).length === 0` early return in `updateSession` is a nice touch to prevent unnecessary Zustand state churn.
- **Good a11y**: Accessibility labels on the badge with correct plural handling, and `accessibilityRole="text"` on the badge.
- **Description truncation**: 200-char limit on server side prevents oversized payloads.
- **Documentation**: CLAUDE.md protocol section updated inline with the code changes.

### Issues Found

#### Critical (Must Fix)

| `packages/server/src/cli-session.js` | 308-318 | **Result timeout (5-min) silently drops agents.** The safety timeout calls `_clearMessageState()` which clears `_activeAgents` without first emitting `agent_completed` events. The comment on line 521 says "callers emit agent_completed and clear the Map before calling this method" but this caller doesn't. App will show stale agent badges until next reconnect. | Add the emit-and-clear pattern before `_clearMessageState()` on this path (same as lines 491-496). |
| `packages/server/src/cli-session.js` | 738-748 | **Interrupt timeout (5s) silently drops agents.** Same issue as above -- the interrupt safety timeout calls `_clearMessageState()` without emitting `agent_completed` first. | Add the emit-and-clear pattern before `_clearMessageState()` on this path. |

| `packages/server/src/cli-session.js` | 515-523 | Move the emit-and-clear logic **into** `_clearMessageState()` itself, rather than requiring every caller to do it first. This would eliminate the class of bug found above -- `_clearMessageState` would be self-contained and safe to call from any path. | Defense in depth: new callers (future timeouts, error paths) would automatically get correct behavior without remembering to emit completions. The comment acknowledges this is "typically a no-op" for current callers, which means doing it unconditionally here costs nothing and prevents future bugs. |
| `packages/app/src/store/connection.ts` | 1090-1094 | `agent_completed` handler always creates a new `activeAgents` array via `.filter()` even when the `toolUseId` isn't found (no-op removal). Could short-circuit: `if (!ss.activeAgents.some(a => a.toolUseId === msg.toolUseId)) return {};` before the filter. | Avoids unnecessary Zustand state update and re-render when receiving `agent_completed` for an already-removed agent (e.g., if both result and close fire). Consistent with the dedup pattern used in `agent_spawned`. |

#### Nitpicks (Optional)

- `packages/server/src/cli-session.js` line 62: The JSDoc `Events emitted` block lists events but doesn't include `agent_spawned` or `agent_completed`. Consider adding them for completeness.
- `packages/app/src/components/SettingsBar.tsx` line 227: The `agentDescription` text uses `numberOfLines={2}` which looks good, but consider whether very long descriptions (near the 200-char limit) might benefit from an ellipsizeMode prop (defaults to `tail` which is fine, just noting for awareness).

| Add elapsed time display for running agents (using `startedAt`) | Good UX enhancement but requires a timer/interval in the component -- separate concern from this monitoring foundation PR. |
| Consider pulsing animation on agent dots to indicate activity | Visual polish item, not a correctness issue. |

### Architecture Notes

This change fits cleanly into the existing architecture:

- **Server**: `CliSession` (the headless executor) owns the `_activeAgents` Map and emits lifecycle events. This follows the existing EventEmitter pattern used for `tool_start`, `user_question`, etc.
- **Session Manager**: Correctly wires events as transient (not recorded in `_recordHistory`), matching the treatment of `raw` and `status_update` events.
- **WS Server**: Both session-mode and legacy-mode paths forward the new events. The session-mode path uses `_broadcastToSession` (correct -- agents are per-session), while legacy mode uses `_broadcast` (correct for single-session legacy).
- **App**: State lives in per-session `SessionState.activeAgents`, accessed via the existing `useConnectionStore` selector pattern. No flat-state sync needed since `activeAgents` is only read from `sessionStates[id]` in the selector.

The WS protocol addition is additive (new message types, no changes to existing ones), so backward compatibility is maintained -- older clients will simply ignore the new events.

### Verdict
- [x] Request Changes - Issues must be addressed

Two critical paths (result timeout, interrupt timeout) silently clear agent tracking without notifying the app, leaving stale UI state. The fix is straightforward -- either add emit-and-clear before `_clearMessageState()` on those paths, or (better) move the emit logic into `_clearMessageState()` itself.

Issue #372, opened 2026-02-10
fix(app): improve long-press multi-select gesture reliability


blamechris:
## Context

When long-pressing a chat message to enter multi-select mode, the user has to hold for several seconds and then drag their finger away to get the selection to "stick". If they lift their finger normally after the long-press, it deselects instantly.

## Expected Behavior

A standard long-press (~500ms) should enter selection mode and toggle the pressed message. Lifting the finger should keep the selection active.

## Current Behavior

The selection triggers but immediately deselects unless the user drags their finger away while still holding. This suggests the `onPress` handler is firing after `onLongPress` and toggling the selection back off, or `Pressable`/`TouchableOpacity` press-out behavior is interfering.

## Acceptance Criteria

- [ ] Long-press (~500ms) reliably enters selection mode
- [ ] Lifting finger after long-press keeps the message selected
- [ ] Subsequent taps toggle other messages in/out of selection

blamechris:
Addressed in PR #378.

Issue #371, opened 2026-02-10
feat(app): add scrollable modal for full tool use output


blamechris:
## Context

When Claude uses tools, the chat view shows a collapsed "1 tool used" indicator that can be expanded. However, the expanded view can't show the full tool output — it gets clipped and there's no way to scroll through long outputs.

## Proposal

Add a modal or full-screen view that opens when tapping the expanded tool use, allowing the user to scroll through the complete tool input/output.

## Acceptance Criteria

- [ ] Tapping expanded tool use opens a scrollable modal
- [ ] Modal shows full tool name, input, and output
- [ ] Modal is dismissible (tap outside or close button)
- [ ] Long outputs are scrollable

PR #378, opened 2026-02-10
fix(app): guard long-press gesture and add tool detail modal

blamechris:
## Summary
- **#372**: Fix long-press multi-select by adding `longPressedRef` guard to `MessageBubble` and extracting `ActivityEntry` component with same guard — prevents `onPress` from firing on finger lift after `onLongPress`, which was toggling selection back off
- **#371**: Add `ToolDetailModal` — tapping an already-expanded tool bubble opens a scrollable modal with full content, dismissible via overlay tap, close button, or Android back

## Test plan
- [x] `npx tsc --noEmit` passes
- [x] 59 app tests pass
- [x] 432 server tests pass
- [ ] Manual: long-press message bubble → selection sticks on finger lift
- [ ] Manual: tap expanded tool bubble → modal opens with full scrollable content
- [ ] Manual: dismiss modal via overlay tap / close button / Android back

### DIFF
--- a/packages/app/src/components/ChatView.tsx
+++ b/packages/app/src/components/ChatView.tsx
@@ -274,9 +274,9 @@ function ToolDetailModal({ visible, toolName, content, onClose }: {
             <Text style={styles.toolModalTitle} numberOfLines={1}>Tool: {toolName}</Text>
             <TouchableOpacity
               onPress={onClose}
-              hitSlop={{ top: 8, bottom: 8, left: 8, right: 8 }}
+              style={styles.toolModalCloseButton}
             >
-              <Text style={styles.toolModalClose}>{ICON_CLOSE}</Text>
+              <Text style={styles.toolModalCloseIcon}>{ICON_CLOSE}</Text>
             </TouchableOpacity>
           </View>
           <ScrollView style={styles.toolModalScroll}>
@@ -290,14 +290,14 @@ function ToolDetailModal({ visible, toolName, content, onClose }: {
 
 // -- Collapsible tool use bubble --
 
-function ToolBubble({ message, isSelected, isSelecting, onToggleSelection }: {
+function ToolBubble({ message, isSelected, isSelecting, onToggleSelection, onOpenDetail }: {
   message: ChatMessage;
   isSelected: boolean;
   isSelecting: boolean;
   onToggleSelection: () => void;
+  onOpenDetail: (toolName: string, content: string) => void;
 }) {
   const [expanded, setExpanded] = useState(false);
-  const [modalVisible, setModalVisible] = useState(false);
   const longPressedRef = useRef(false);
   const content = message.content?.trim();
 
@@ -313,7 +313,7 @@ function ToolBubble({ message, isSelected, isSelecting, onToggleSelection }: {
     if (isSelecting) {
       onToggleSelection();
     } else if (expanded) {
-      setModalVisible(true);
+      onOpenDetail(message.tool || 'Unknown', content);
     } else {
       LayoutAnimation.configureNext(LayoutAnimation.Presets.easeInEaseOut);
       setExpanded(true);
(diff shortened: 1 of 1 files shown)


blamechris:
## Code Review: PR #378

### Summary
This PR addresses two issues:
1. **#372 — Long-press multi-select fix:** Adds a `longPressedRef` guard to `MessageBubble` and extracts a new `ActivityEntry` component with the same guard pattern. This prevents the `onPress` handler from firing on finger-lift after a `onLongPress`, which was immediately toggling the selection back off.
2. **#371 — Tool detail modal:** Tapping an already-expanded `ToolBubble` now opens a `ToolDetailModal` with full scrollable content, dismissible via overlay tap, close button, or Android back (`onRequestClose`).

### Strengths
- **Correct root-cause fix** — The `longPressedRef` guard is the canonical React Native pattern for disambiguating long-press from tap. Applied consistently across `MessageBubble`, `ToolBubble`, and the new `ActivityEntry`.
- **Good component extraction** — `ActivityEntry` was correctly pulled out of the inline `ActivityGroup` render, making the long-press guard possible and improving readability.
- **ToolDetailModal is well-structured** — Uses nested `Pressable` with `stopPropagation` for overlay dismiss, `onRequestClose` for Android back, `transparent` + `fade` for standard modal behavior.
- **Text selectable in modal** — Content in the modal is marked `selectable`, which is the right call for tool output users may want to copy.
- **`ICON_CLOSE` reuse** — Uses the existing shared icon constant instead of a hardcoded string.

### Issues Found

#### Critical (Must Fix)
None.

| `ChatView.tsx` | 275-278 | The close button `TouchableOpacity` has a `hitSlop` of 8 on each side, but the base touch target is just the text glyph (~18pt font). Total touchable area is roughly 34pt — below the 44pt minimum recommended by Apple/Google. Consider adding explicit `width: 44, height: 44, alignItems: 'center', justifyContent: 'center'` to the close button or increase `hitSlop` to 12-14. | Accessibility and usability, especially on smaller devices. |
| `ChatView.tsx` | 849-854 | `activityEntry` has `paddingVertical: 3`, making each row only ~17pt tall. This is well below the 44pt minimum touch target for iOS/Android. Since these are individually tappable (for selection), consider `paddingVertical: 10` or `minHeight: 44`. | Touch target compliance. Currently very hard to tap accurately on phone. |
| `ChatView.tsx` | 348-353 | Each `ToolBubble` renders its own `ToolDetailModal` instance. When there are many tool messages, this creates many `Modal` components in the tree (albeit with `visible={false}`). Consider lifting the modal state up to `ChatView` and rendering a single `ToolDetailModal` at the top level, passing the selected tool's name/content. | Performance — reduces component count. Not urgent with typical message volumes, but good architecture. |

#### Nitpicks (Optional)
- Line 315: The `expanded` → modal tap flow means users can never *collapse* an expanded tool bubble by tapping it — they always get the modal. The only way to collapse is if the activity group auto-collapses, or the message scrolls away. This is a deliberate design choice per #371, but worth noting that it's a one-way expansion. If users find this frustrating, a future option could be a small collapse chevron on the expanded bubble.
- Line 302: `message.content?.trim()` is computed on every render of `ToolBubble`. For very large tool outputs this is fine (strings are immutable in JS, trim is cheap), but if this component re-renders frequently, `useMemo` on the content could be a micro-optimization.

Issues will be created below for items that should not block this PR but should be tracked.

### Architecture Notes
The changes fit cleanly within the existing `ChatView` component architecture:
- The `longPressedRef` pattern already existed in `ToolBubble` and is now applied consistently to the other touchable components.
- `ActivityEntry` extraction follows the same pattern as other extracted components (`ToolBubble`, `ActivityGroup`).
- `ToolDetailModal` is a pure presentational component with no state management concerns — just visibility, content, and a close callback.
- No new dependencies introduced; `Modal` and `Pressable` are core React Native imports.

### Verdict
- [x] **Approve** — This is a clean, well-targeted fix for two real UX bugs. No critical issues. The touch target suggestions are worth addressing but are pre-existing concerns (not regressions from this PR) and can be handled in follow-up.

PR #381, opened 2026-02-10
feat: add plan mode UI with approval card

blamechris:
## Summary

- Detect `EnterPlanMode`/`ExitPlanMode` tool calls in `cli-session.js` and emit `plan_started`/`plan_ready` events
- Forward plan events over WebSocket in both multi-session and legacy CLI paths
- Add `isPlanPending`/`planAllowedPrompts` to app session state with proper lifecycle management
- Render `PlanApprovalCard` in chat view with Approve and Give Feedback buttons
- Convert `InputBar` to `forwardRef` for programmatic focus from Give Feedback

## Test plan

- [x] 429 server tests pass
- [x] 59 app tests pass
- [x] TypeScript type check passes (`npx tsc --noEmit`)
- [ ] Manual: trigger plan mode in Claude Code, verify plan approval card appears
- [ ] Manual: tap Approve, verify message sent and card disappears
- [ ] Manual: tap Give Feedback, verify input bar focuses
- [ ] Manual: reconnect during plan mode, verify stale plan state clears

### DIFF
--- a/packages/app/src/screens/SessionScreen.tsx
+++ b/packages/app/src/screens/SessionScreen.tsx
@@ -29,8 +29,9 @@ import { ICON_CLOSE, ICON_GEAR } from '../constants/icons';
 import { COLORS } from '../constants/colors';
 
 
-// Stable empty array to avoid new-reference-per-render in Zustand selectors
+// Stable empty arrays to avoid new-reference-per-render in Zustand selectors
 const EMPTY_AGENTS: AgentInfo[] = [];
+const EMPTY_PROMPTS: { tool: string; prompt: string }[] = [];
 
 // Enable LayoutAnimation on Android
 UIManager.setLayoutAnimationEnabledExperimental?.(true);
@@ -140,7 +141,7 @@ export function SessionScreen() {
   });
   const planAllowedPrompts = useConnectionStore((s) => {
     const id = s.activeSessionId;
-    return id && s.sessionStates[id] ? s.sessionStates[id].planAllowedPrompts : [];
+    return id && s.sessionStates[id] ? s.sessionStates[id].planAllowedPrompts : EMPTY_PROMPTS;
   });
   const destroySession = useConnectionStore((s) => s.destroySession);
   const serverErrors = useConnectionStore((s) => s.serverErrors);
@@ -217,6 +218,9 @@ export function SessionScreen() {
       // Add user message + thinking indicator with session-aware state update
       addUserMessage(text);
     }
+
+    // Clear plan approval card — user has responded (whether approving or giving feedback)
+    if (isPlanPending) clearPlanState();
 
     // PTY sessions: append CR so text + submit arrive as a single atomic write.
     // Sending them separately caused a race condition where multi-line text
--- a/packages/server/src/session-manager.js
+++ b/packages/server/src/session-manager.js
@@ -510,7 +510,7 @@ export class SessionManager extends EventEmitter {
     }
 
     // Transient events — forwarded but not recorded in history (not replayed on reconnect)
-    const TRANSIENT_EVENTS = ['permission_request', 'agent_spawned', 'agent_completed']
+    const TRANSIENT_EVENTS = ['permission_request', 'agent_spawned', 'agent_completed', 'plan_started', 'plan_ready']
     for (const event of TRANSIENT_EVENTS) {
       session.on(event, (data) => {
         this.emit('session_event', { sessionId, event, data })


blamechris:
## Code Review: PR #381

### Summary
Adds plan mode UI support end-to-end: detects `EnterPlanMode`/`ExitPlanMode` tool calls in `cli-session.js`, forwards `plan_started`/`plan_ready` events over WebSocket, stores plan state in the Zustand store per-session, and renders a `PlanApprovalCard` in the chat view with Approve and Give Feedback buttons. Also converts `InputBar` to `forwardRef` for programmatic focus.

### Strengths
- **Clean state lifecycle**: plan state correctly spans multiple turns (`_inPlanMode` intentionally survives `_clearMessageState`), and stale plan state is properly cleared on `history_replay_start` reconnect.
- **Consistent patterns**: follows the established `agent_spawned`/`agent_completed` pattern for both multi-session (`_setupSessionForwarding`) and legacy single-session (`_setupCliForwarding`) WS paths.
- **Good defensive coding**: `allowedPrompts` input is validated with `Array.isArray()` at every boundary (server parse, WS broadcast, store ingestion), JSON parse is wrapped in try/catch.
- **Accessibility**: buttons have `accessibilityRole="button"` and descriptive `accessibilityLabel` values. Touch targets meet 44pt minimum (`minHeight: 44`).
- **WS protocol documented**: new message types added to the header comment block in `ws-server.js`.

### Issues Found

#### Critical (Must Fix)

| `packages/server/src/cli-session.js` | ~962 | `destroy()` calls `_clearMessageState()` which intentionally does NOT reset `_inPlanMode`. The comment on line 535-537 says it's reset "on destroy()", but it isn't -- `destroy()` only calls `_clearMessageState()` then `removeAllListeners()`. While the session object is discarded after `destroy()`, this contradicts the documented contract and could cause subtle bugs if destroy/recreate patterns emerge. | Add `this._inPlanMode = false` inside `destroy()` before `_clearMessageState()`, or add it to `_clearMessageState()` guarded by a `force` parameter that `destroy()` passes. |

| `packages/app/src/screens/SessionScreen.tsx` | 264-268 | `handleApprovePlan` sends a hardcoded string `'Go ahead with the plan'` as user input. Consider defining this as a constant or making it clear this is the expected protocol string that Claude Code recognizes. | If the expected approval string changes in Claude Code, this will silently break. A named constant makes it easier to find and update. |
| `packages/app/src/screens/SessionScreen.tsx` | 262 | `clearPlanState` is selected from the store outside the `useCallback` block that uses it, but it's a stable function reference from Zustand so this works. However, it's inconsistent with the pattern used elsewhere where store actions are consumed inside the callback. Consider selecting it alongside `addUserMessage`/`sendInput` for consistency. | Minor code organization — all three are stable Zustand actions used in the same callback. |
| `packages/app/src/components/ChatView.tsx` | 686-692 | The `PlanApprovalCard` is placed after the message list but inside the ScrollView. When it appears, the user might not see it unless they scroll down. Consider triggering a `scrollToEnd` when `isPlanPending` transitions to `true`. | The `onContentSizeChange` handler on line 640 should handle this as long as `isSelectingRef.current` is false, but it's worth verifying this works in practice during manual testing. |
| `packages/server/src/cli-session.js` | 439-453 | `EnterPlanMode` doesn't need to parse `toolInputChunks` (it just sets a flag), but `ExitPlanMode` does. However, if `EnterPlanMode` is called a second time (e.g., re-entering plan mode in the same session), `_inPlanMode` is already `true` and `plan_started` is emitted again. Consider guarding with `if (!this._inPlanMode)` to avoid duplicate events. | Defensive programming — unlikely in practice but prevents duplicate `plan_started` events. |

#### Nitpicks (Optional)

- `packages/app/src/components/ChatView.tsx` line 553: Using `key={i}` (array index) for `allowedPrompts.map()` is fine here since prompts are static once rendered and won't be reordered, but a composite key like `key={\`\${p.tool}-\${i}\`}` would be marginally more robust.
- The `PlanApprovalCard` component is defined in `ChatView.tsx` rather than its own file. This is consistent with how `MessageBubble` and `ActivityGroup` are structured in the same file, so it's fine for now, but as the file grows, extraction could help.

| Plan card should scroll into view automatically | Created below | Needs manual testing first; `onContentSizeChange` may already handle it |
| Plan approval string should be a protocol constant | Created below | Requires coordination with server-side understanding of plan mode responses |

### Architecture Notes

- The plan mode detection relies on tool names `EnterPlanMode` and `ExitPlanMode` which are internal Claude Code tool names. If these change upstream, the feature will silently stop working. This is the same pattern used for `Task` and `AskUserQuestion` detection, so it's consistent with existing design.
- Plan events are correctly marked as **transient** (not recorded in history, cleared on replay) -- matching the pattern established by `agent_spawned`/`agent_completed`.
- The `forwardRef` conversion of `InputBar` is clean and backward-compatible -- existing usage without `ref` continues to work.
- Both multi-session and legacy CLI forwarding paths are covered for the new events.

### Verdict

**Approve with minor fix.** One critical issue (destroy not resetting `_inPlanMode` despite documenting that it does) should be addressed. The suggestions are quality-of-life improvements. The overall implementation is solid, well-structured, and follows established project patterns.

Issue #409, opened 2026-02-10
Add TTL expiry, drain-on-reconnect, and disconnect-clears-queue tests for message queue


blamechris:
## Context

From Copilot review on PR #404 ([comment](https://github.com/blamechris/chroxy/pull/404#discussion_r2790643485)).

The current message queue tests cover basic queuing, max size, and excluded types. Additional coverage is needed for:

1. **TTL-based expiry**: Verify that expired messages are dropped during drain (mock `Date.now()` or use fake timers).
2. **Drain on reconnect/claude_ready**: Verify the queue drains when `claude_ready` arrives after reconnect, and is empty afterward.
3. **Disconnect clears queue**: Verify that `disconnect()` empties any queued messages.

These behaviors are implemented in `enqueueMessage`, `drainMessageQueue`, and the `disconnect` action in `packages/app/src/store/connection.ts`. The module-scoped `messageQueue` array makes some of these hard to test without either exposing internals or using timing tricks.

## Approach

- Use `jest.useFakeTimers()` and `jest.setSystemTime()` to simulate TTL expiry
- For drain testing, may need to expose the queue or create a test helper
- For disconnect, call `disconnect()` after queuing and verify subsequent behavior

blamechris:
Already resolved in PR #404 — TTL expiry, drain-on-reconnect, and disconnect-clears-queue tests were added to connection.test.ts (lines 294-376).

Issue #408, opened 2026-02-10
Add warning log for tag pruning failures in deploy command


blamechris:
## Source

PR #404 review — Suggestion S3

## Problem

In `cli.js:602`, the empty `catch {}` silently swallows all tag pruning errors. If git tags are locked or the repo is in an unusual state, there's no diagnostic output.

## Suggested Fix

```js
} catch (err) {
  console.warn(`[deploy] Tag pruning failed: ${err.message}`)
}
```

## Files

- `packages/server/src/cli.js` (line 602)

blamechris:
Already resolved — `console.warn` for tag pruning failures exists at cli.js:603 (added in PR #404).

PR #410, opened 2026-02-10
fix: resolve 3 from-review issues (#405, #408, #409)

### LINKED ISSUE #405: markPromptAnswered fires on queued (undelivered) messages
## Source

PR #404 review — Critical issue C1

## Problem

In `SessionScreen.tsx:264`, the check `if (sent)` treats both `'sent'` and `'queued'` as truthy. When a permission response or user question answer is queued (not yet delivered to the server), the prompt is immediately marked as answered in the UI.

If the queued message later expires (TTL) or the queue is flushed on disconnect, the user sees an "answered" prompt but the server never received the response. On flaky cellular connections where messages queue frequently, this creates a silent failure where the user thinks they responded but the server is still waiting.

## Suggested Fix

Compare explicitly against `'sent'`:
```typescript
if (sent === 'sent') {
  markPromptAnswered(messageId, value);
}
```

For `'queued'`, either leave the prompt unanswered (so it re-appears after reconnect drain) or show a "pending" visual indicator.

## Files

- `packages/app/src/screens/SessionScreen.tsx` (line 264)

blamechris:
## Summary

Final batch of from-review issues from the In-App Dev MVP.

| # | Issue | Fix |
|---|-------|-----|
| #405 | markPromptAnswered fires on queued messages | Check `sent === 'sent'` instead of truthy |
| #408 | Tag pruning swallows errors | Log warning on failure |
| #409 | Message queue missing tests | 5 tests: TTL expiry, drain valid/expired, per-type TTL, disconnect clears |

## Changes

| File | What changed |
|------|-------------|
| `SessionScreen.tsx` | `if (sent === 'sent')` instead of `if (sent)` |
| `cli.js` | `catch (err)` with `console.warn` instead of empty catch |
| `connection.ts` | Export `_testQueueInternals` for test access |
| `connection.test.ts` | 5 new queue internal tests with fake timers |

## Test Plan

- [x] 448 server tests pass
- [x] 70 app tests pass (5 new queue tests)
- [x] `npx tsc --noEmit` clean

Closes #405, #408, #409

### DIFF
--- a/packages/app/src/__tests__/store/connection.test.ts
+++ b/packages/app/src/__tests__/store/connection.test.ts
@@ -6,6 +6,7 @@ import {
   selectShowSession,
   useConnectionStore,
   ChatMessage,
+  _testQueueInternals,
 } from '../../store/connection';
 
 // Reset store between tests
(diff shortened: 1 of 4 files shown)


blamechris:
## Code Review: PR #410 — fix: resolve 3 from-review issues (#405, #408, #409)

### Summary

This PR fixes three from-review issues in a single batch:
1. **#405** — `markPromptAnswered` was firing on queued (undelivered) messages because `if (sent)` was truthy for both `'sent'` and `'queued'` return values
2. **#408** — Empty `catch {}` in deploy command's tag pruning silently swallowed errors
3. **#409** — Message queue internals (TTL expiry, drain, disconnect-clears) lacked test coverage

### Strengths

- **Precise, minimal fix for #405.** Changing `if (sent)` to `if (sent === 'sent')` is exactly right. The `sendInput`/`sendPermissionResponse`/`sendUserQuestionResponse` functions all return the union `'sent' | 'queued' | false`, and only `'sent'` means the message actually reached the server. Marking a prompt as answered when the message is merely queued would produce a confusing UI where the user sees their "answer" but the server never received it.

- **Correct error handling for #408.** The tag pruning is a non-critical cleanup step — a warning log (`console.warn`) is the right severity. The `err.message` interpolation follows the existing pattern in cli.js. No semicolons, single quotes — matches server code style.

- **Thorough test coverage for #409.** Five well-structured tests cover: TTL expiry filtering, valid message drain, per-type TTL differentiation (interrupt 5s vs input 60s), disconnect clearing the queue, and drain clearing the queue even when all messages are expired. Good use of `jest.useFakeTimers()` / `jest.advanceTimersByTime()` for deterministic time-based testing.

- **Clean test API design.** The `_testQueueInternals` export is a sensible pattern — underscored, `@internal` JSDoc tag, and provides just the four operations needed (getQueue, enqueue, drain, clear) without exposing more than necessary. The mock socket pattern (`{ send: (data: string) => sent.push(data) } as unknown as WebSocket`) is clean and appropriate.

- **CI passing.** All three checks (server tests, app tests, app type check) pass.

### Issues

**Critical:** None.

**Suggestions:** None.

**Nitpicks:**

1. **Mock socket repetition in tests.** The `mockSocket` construction is identical in 4 of 5 tests. A shared helper (`function createMockSocket()`) would reduce repetition. Not worth blocking on, but a nice cleanup if you're ever back in this file.

### Architecture Notes

- The `_testQueueInternals` export correctly accesses the module-level `messageQueue` array and the `enqueueMessage`/`drainMessageQueue` functions. Since these are module-scoped (not inside the Zustand store), a dedicated test export is the cleanest way to unit-test them without going through the full WebSocket connection flow.

- The `handleSelectOption` fix in SessionScreen.tsx aligns with the broader pattern: `markPromptAnswered` should only fire when the server has actually received the response. If the message is queued, the UI should wait until reconnection and successful drain before marking it as answered. (Note: the current implementation does not retroactively mark it after drain — this is acceptable for now, since queue drain is a reconnect-time operation and the user would likely re-interact.)

### Verdict

**Approve.** Clean, focused fixes for all three issues with proper test coverage. No concerns.

PR #510, opened 2026-02-16
fix(server): actionable warning on corrupt settings.json

### LINKED ISSUE #491: fix(server): emit warning on corrupt settings.json during hook registration
## Problem

When `~/.claude/settings.json` is corrupt (invalid JSON), the permission hook registration in `cli-session.js` silently fails. The server starts without permission hooks, and the user has no indication that permissions are broken.

## Expected Behavior

The server should warn the user clearly when it cannot read or parse the settings file, so they can fix it before starting a session.

## Proposed Fix

- Wrap the settings.json read/parse in a try/catch
- Emit an error event or log a prominent warning with the file path and parse error
- Optionally forward the warning to connected clients via `server_error` message
- Consider backing up the corrupt file and writing a fresh default

## Files

- `packages/server/src/cli-session.js` (hook registration)
- `packages/server/src/sdk-session.js` (if applicable)

blamechris:
## Summary

- **#491**: When `~/.claude/settings.json` contains invalid JSON, hook registration now reports the exact file path, the parse error, and instructions to fix or delete the file
- Unregister gracefully skips cleanup on corrupt JSON instead of crashing on shutdown

Closes #491

## Test plan

- [x] All 16 permission-hook tests pass (including existing corrupt JSON test)
- [ ] Manual: corrupt settings.json, start server, verify warning is visible

### DIFF
--- a/packages/server/tests/permission-hook.test.js
+++ b/packages/server/tests/permission-hook.test.js
@@ -173,6 +173,21 @@ describe('createPermissionHookManager', () => {
     manager.destroy()
   })
 
+  it('unregister() does not throw on corrupt settings.json', async () => {
+    const settingsPath = join(tempDir, 'settings.json')
+    writeFileSync(settingsPath, 'not valid json {{{')
+
+    const manager = createPermissionHookManager(emitter, { settingsPath })
+    // Should not throw — just logs a warning and skips cleanup
+    await manager.unregister()
+
+    // File should remain untouched (not deleted or modified)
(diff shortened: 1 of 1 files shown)


blamechris:
## Review Comments Addressed

| Skip retries on corrupt JSON (non-retryable SyntaxError) | Fixed | `6d918ac3` |
| Add test for unregister on corrupt settings.json | Fixed | `6839b9a0` |

- Fixed: 2
- False positives: 0
- Follow-up issues created: 0

PR #524, opened 2026-02-16
feat(app): show permission request timeout countdown

### LINKED ISSUE #492: feat(app): show permission request timeout countdown
## Problem

Permission prompts have a 5-minute timeout, after which they are auto-denied. The user has no visual indication of how much time remains, leading to unexpected denials if they step away briefly.

## Expected Behavior

Permission cards should display a countdown timer showing remaining time. A warning should appear as the timeout approaches (e.g., last 30 seconds).

## Proposed Implementation

- Include `timeoutMs` or `expiresAt` in the `permission_request` WebSocket message from the server
- Display a countdown timer on permission cards in `ChatView.tsx`
- Show a visual warning (color change, animation) when time is running low
- Auto-remove the card and show a "timed out" indicator when expired

## Files

- `packages/app/src/components/ChatView.tsx` (permission card UI)
- `packages/server/src/cli-session.js` (include timeout in WS message)
- `packages/server/src/sdk-session.js` (include timeout in WS message)

blamechris:
## Summary

- Server sends `expiresAt` (epoch ms) in `permission_request` WS messages — indicates when the 5-minute auto-deny fires
- App renders a `M:SS` countdown timer on unanswered permission prompts, aligned right of the tool name label
- Countdown turns red + bold at 30 seconds remaining
- Shows "Timed out" and disables option buttons when time expires
- No countdown shown on already-answered prompts

Closes #492

## Test Plan

- [ ] Server tests pass (`cd packages/server && PATH="/opt/homebrew/opt/node@22/bin:$PATH" npm test`)
- [ ] App type-checks clean (`cd packages/app && npx tsc --noEmit`)
- [ ] Manual: trigger permission prompt → countdown visible, ticks from 5:00
- [ ] Manual: wait until <30s → countdown turns red + bold
- [ ] Manual: wait until 0 → "Timed out", buttons disabled
- [ ] Manual: answer before timeout → countdown disappears, selection shown

### DIFF
--- a/packages/app/src/components/ChatView.tsx
+++ b/packages/app/src/components/ChatView.tsx
@@ -457,7 +457,7 @@ function PermissionCountdown({ expiresAt, onExpire }: { expiresAt: number; onExp
   }, [expiresAt, remaining <= 0]);
 
   if (remaining <= 0) {
-    return <Text style={styles.countdownExpired}>Timed out</Text>;
+    return <Text style={styles.countdownExpired} accessibilityRole="timer">Timed out</Text>;
   }
 
   const minutes = Math.floor(remaining / 60);
@@ -466,7 +466,12 @@ function PermissionCountdown({ expiresAt, onExpire }: { expiresAt: number; onExp
   const isUrgent = remaining <= 30;
 
   return (
-    <Text style={[styles.countdownText, isUrgent && styles.countdownUrgent]}>
+    <Text
+      style={[styles.countdownText, isUrgent && styles.countdownUrgent]}
+      accessibilityRole="timer"
+      accessibilityLabel={`Permission expires in ${minutes} minutes ${seconds} seconds`}
+      accessibilityLiveRegion="polite"
+    >
       {label}
     </Text>
   );
@@ -873,6 +878,7 @@ const styles = StyleSheet.create({
     flexDirection: 'row',
     justifyContent: 'space-between',
     alignItems: 'center',
+    marginBottom: 4,
   },
   countdownText: {
     color: COLORS.accentOrange,


blamechris (commented):
## Agent Review

**Verdict:** REQUEST_CHANGES

### Findings

#### Critical

1. **`expiresAt` dropped in SDK session broadcast path** (`packages/server/src/ws-server.js`, line 1236-1242)

   The session event handler re-broadcasts `permission_request` events from `sdk-session.js` with an explicitly constructed message object, but it **does not include `expiresAt`**:

   ```js
   case 'permission_request':
     this._permissionSessionMap.set(data.requestId, sessionId)
     this._broadcastToSession(sessionId, {
       type: 'permission_request',
       requestId: data.requestId,
       tool: data.tool,
       description: data.description,
       input: data.input,
       // ← expiresAt missing here, even though data.expiresAt is set by sdk-session.js
     })
   ```

   `sdk-session.js` correctly emits `expiresAt: Date.now() + 300_000` (line 324), and the HTTP hook path in `_handlePermissionHook` correctly includes it in the broadcast (line 1659). But the SDK session event forwarding path — which is the **default CLI headless mode** — silently drops it.

   **Impact:** The countdown will never appear for the primary use case (CLI/SDK mode). It only works in PTY/hook mode.

   **Fix:** Add `expiresAt: data.expiresAt,` to the broadcast at line 1241.

#### Suggestions

1. **Consider sending `remainingMs` instead of absolute `expiresAt`** — The server sends `Date.now() + 300_000` (an absolute epoch timestamp), and the app subtracts its own `Date.now()`. If the phone and server clocks differ (common on mobile), the countdown will be inaccurate. Sending `remainingMs: 300_000` and computing `Date.now() + remainingMs` on the client would eliminate clock skew entirely. Not blocking since NTP drift is usually sub-second, but worth considering for robustness.

2. **`useEffect` dependency `remaining <= 0` is unconventional** (`ChatView.tsx`, line 457) — Using a boolean expression in the deps array works correctly (React compares with `Object.is`), but it reads oddly and may confuse future contributors or trigger ESLint exhaustive-deps warnings. A slightly clearer alternative:

   ```ts
   const expired = remaining <= 0
   useEffect(() => {
     if (expired) return;
     // ... interval logic
   }, [expiresAt, expired]);
   ```

#### Nits

1. **Redundant `as number` cast** (`connection.ts`, line 1224) — `typeof msg.expiresAt === 'number'` already narrows the type; the explicit `as number` adds nothing:
   ```ts
   // Current:
   expiresAt: typeof msg.expiresAt === 'number' ? msg.expiresAt as number : undefined,
   // Cleaner:
   expiresAt: typeof msg.expiresAt === 'number' ? msg.expiresAt : undefined,
   ```

### Summary

Clean, well-scoped feature with good UX (monospace countdown, urgency coloring, expired state). The `PermissionCountdown` component handles lifecycle correctly — intervals are cleaned up on unmount, the `onExpireRef` pattern avoids stale closures, and button disabling works via local `isExpired` state. However, the critical bug is that `expiresAt` is dropped in the SDK session broadcast path in `ws-server.js`, which means the countdown will silently not appear in the default CLI/headless mode. One-line fix required before merge.

blamechris (commented):
All findings addressed:

**Critical: `expiresAt` dropped in SDK session broadcast path**
Fixed in 75b715d7 — added `expiresAt: data.expiresAt` to the broadcast object in `ws-server.js`.

**Nit: Redundant `as number` cast**
Fixed in 75b715d7 — removed the unnecessary `as number` cast in `connection.ts`.

**Suggestion: Send `remainingMs` instead of absolute `expiresAt`**
Valid robustness improvement. Deferred to #526 since NTP drift is sub-second in practice and doesn't affect correctness for a 5-minute countdown.

**Suggestion: `useEffect` dependency `remaining <= 0` is unconventional**
This is intentional — the boolean expression in the deps array causes the effect to re-run only on the transition from `remaining > 0` to `remaining <= 0` (and vice versa), which is exactly the desired behavior. Extracting to a named `expired` variable would be equivalent functionally but adds a line without changing semantics. Leaving as-is since it works correctly and the intent is clear from context.

PR #535, opened 2026-02-16
feat: add Docker image and docker-compose for deployment (#500)

### LINKED ISSUE #500: feat(server): add Docker image and VPS deployment support
## Problem

There is no containerized deployment option. Users who want to run chroxy on a VPS must manually install Node.js 22, cloudflared, and other dependencies, which is error-prone and undocumented.

## Expected Behavior

Users should be able to deploy chroxy with a single `docker-compose up` command on any VPS.

## Proposed Implementation

- Create a `Dockerfile` based on Node.js 22 LTS image
- Include cloudflared binary in the image
- Add `docker-compose.yml` with environment variable configuration
- Support both Quick and Named tunnel modes via environment variables
- Add health check endpoint for container orchestration
- Include an install/setup script for common VPS providers

## Files

- `Dockerfile` (new)
- `docker-compose.yml` (new)
- `scripts/install.sh` (new or update existing)
- `docs/` (deployment guide)

blamechris:
## Summary

- Add `Dockerfile` based on `node:22-slim` with cloudflared (multi-arch amd64/arm64)
- Add `docker-compose.yml` with env var config, named volumes for persistence, workspace bind mount
- Add `scripts/docker-entrypoint.sh` with auto config generation, API key validation, `exec` for signal propagation
- Add whitelist-based `.dockerignore` (server package only)
- CLI headless mode only — no tmux/node-pty/build-essential needed
- Non-root `chroxy` user, health check via existing `GET /` endpoint

Closes #500

## Test Plan

- [x] `docker build -t chroxy .` — builds successfully (20 packages, 0 vulnerabilities)
- [x] Container runs as non-root `chroxy` user
- [x] Node v22.22.0 + cloudflared 2026.2.0 verified in container
- [x] `start` fails fast with clear error when `ANTHROPIC_API_KEY` missing
- [x] `doctor` and arbitrary commands work without API key (pass-through)
- [ ] `ANTHROPIC_API_KEY=sk-test docker-compose up` — server starts, tunnel connects
- [ ] Ctrl+C → graceful shutdown (no orphan processes)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,6 +38,7 @@ RUN useradd -m -s /bin/bash chroxy && \
     chown -R chroxy:chroxy /home/chroxy /workspace /app
 
 USER chroxy
+ENV HOME=/home/chroxy
 
 EXPOSE 8765
 
--- a/scripts/docker-entrypoint.sh
+++ b/scripts/docker-entrypoint.sh
@@ -23,8 +23,9 @@ prepare_config() {
       API_TOKEN="$(cat /proc/sys/kernel/random/uuid 2>/dev/null || uuidgen)"
       echo ""
       echo "================================================"
-      echo "  Generated API token (save this for your app):"
-      echo "  $API_TOKEN"
+      echo "  Generated API token."
+      echo "  Prefix: ${API_TOKEN:0:8}..."
+      echo "  Full token stored in $CONFIG_FILE"
       echo "================================================"
       echo ""
     fi
@@ -45,11 +46,26 @@ prepare_config() {
 }
 
 # --- Route command ---
-CMD="${1:-start}"
-shift || true
+# If the first arg starts with '-', treat it as flags for the default 'start' command
+if [ "${1:0:1}" = "-" ] 2>/dev/null; then
+  CMD="start"
+else
+  CMD="${1:-start}"
+  shift || true
+fi
 
 case "$CMD" in
   start)
+    # Docker image is headless-only: no tmux, node-pty not compiled.
+    # Fail fast if --terminal is requested.
+    for arg in "$@"; do
+      if [ "$arg" = "--terminal" ] || [ "$arg" = "-t" ]; then
+        echo "ERROR: Terminal mode (--terminal) is not supported in Docker."
+        echo "This container runs in headless mode only (no tmux/node-pty)."
+        echo "Remove --terminal to use the default SDK mode."
+        exit 1
+      fi
+    done
     prepare_config
     exec node /app/packages/server/src/cli.js start "$@"
     ;;


blamechris (commented):
## Agent Review

**Verdict:** REQUEST_CHANGES

### Findings

#### Critical (must fix before merge)

1. **`node-pty` static import will crash the container on startup.** `session-manager.js` statically imports `PtySession` → `pty-session.js` → `pty-manager.js` → `import pty from "node-pty"`. Since the Dockerfile uses `--ignore-scripts` (correctly, to skip native compilation), the `node-pty` native binding won't exist. When `server-cli.js` imports `SessionManager`, the static import chain will trigger `node-pty` to load its native `.node` addon, which will throw an error and crash the process. This happens regardless of SDK vs legacy-CLI mode because the import is resolved at module load time, not at PtySession instantiation time.

   **Fix options (pick one):**
   - **(a) Convert to dynamic import:** Change `session-manager.js` line 8 from `import { PtySession } from './pty-session.js'` to a lazy `await import()` inside the method that actually creates PTY sessions. This is the cleanest fix and benefits non-Docker users too.
   - **(b) Exclude `node-pty` in Docker:** Add a post-install step in the Dockerfile to `rm -rf node_modules/node-pty` after `npm ci`, and add a try/catch wrapper around the import in `session-manager.js`.
   - Option (a) is recommended — it makes the dependency truly optional.

   Note: The PR's test plan shows the full integration test ("server starts, tunnel connects") as unchecked, which likely would have caught this.

#### Suggestions (consider for this PR)

2. **`cloudflared` pinned to `latest` is non-reproducible.** The Dockerfile downloads `cloudflared` from the `latest` release URL. This means the same Dockerfile can produce different images at different times, and a breaking cloudflared release could silently break builds. Consider pinning to a specific version:
   ```dockerfile
   ARG CLOUDFLARED_VERSION=2026.2.0
   RUN curl -fsSL "https://github.com/cloudflare/cloudflared/releases/download/${CLOUDFLARED_VERSION}/cloudflared-linux-${ARCH}" ...
   ```
   This also documents which version was tested.

3. **Config file is only generated on first run.** If a user changes `API_TOKEN` or `PORT` via environment variables after the first run, the persisted `config.json` (in the named volume) will take precedence due to the `if [ ! -f "$CONFIG_FILE" ]` guard. The config precedence system (env > file) handles `PORT` and `API_TOKEN` correctly at the server level, but the generated `config.json` with a stale `apiToken` could be confusing. Consider either: (a) always regenerating the config, or (b) adding a note in the generated config that env vars override these values, or (c) not writing `apiToken`/`port` to config.json at all (let them flow through env vars via `mergeConfig`).

4. **`--legacy-cli` mode would silently fail.** If a user passes `--legacy-cli` to the Docker container, the server would crash because the `claude` binary isn't installed. Consider adding a guard in the entrypoint or documenting this limitation.

#### Nits (optional improvements)

5. **HEALTHCHECK uses shell form with variable expansion.** `CMD curl -sf http://localhost:${PORT:-8765}/ || exit 1` — the `${PORT:-8765}` is expanded at container runtime by the shell, which is correct. However, if someone overrides the CMD, the HEALTHCHECK shell form will still spawn `/bin/sh -c` for every check. This is fine for `curl` but worth noting.

6. **`bash` is already included in `node:22-slim`.** The `apt-get install bash` line is redundant since Debian slim images include bash. Minor layer cache savings by removing it.

7. **Config JSON uses heredoc string interpolation.** The `cat > "$CONFIG_FILE" <<EOF` block uses unquoted EOF, so `$API_TOKEN` and `${PORT:-8765}` are expanded by bash. This is intentional and correct, but if an API token ever contained special characters (`$`, backticks), they'd be interpreted. Using a JSON-safe write (e.g., `jq` or `node -e`) would be more robust, but the UUID-generated tokens are safe in practice.

### Summary

Well-structured PR that adds clean Docker deployment support. The Dockerfile follows good practices (layer caching via separate COPY for package files, non-root user, health check, multi-arch cloudflared). The entrypoint is robust with proper `set -euo pipefail`, API key validation, and `exec` for signal propagation. The `.dockerignore` whitelist approach is correct and the `docker-compose.yml` has sensible defaults with good volume strategy.

The one critical issue is the `node-pty` static import chain that will crash the container at startup. This needs to be resolved before merge. The build succeeds because `--ignore-scripts` defers the failure to runtime. Converting the `PtySession` import to a dynamic `await import()` would be the cleanest fix and would also benefit non-Docker environments where `node-pty` isn't installed.

blamechris (commented):
## Agent Review

**Verdict:** REQUEST_CHANGES

### Findings

#### Critical (must fix before merge)

1. **`node-pty` static import will crash the container on startup.** `session-manager.js` statically imports `PtySession` -> `pty-session.js` -> `pty-manager.js` -> `import pty from "node-pty"`. Since the Dockerfile uses `--ignore-scripts` (correctly, to skip native compilation), the `node-pty` native binding will not exist. When `server-cli.js` imports `SessionManager`, the static import chain will trigger `node-pty` to load its native `.node` addon, which will throw an error and crash the process. This happens regardless of SDK vs legacy-CLI mode because the import is resolved at module load time, not at PtySession instantiation time.

   **Fix options (pick one):**
   - **(a) Convert to dynamic import:** Change `session-manager.js` line 8 from `import { PtySession } from './pty-session.js'` to a lazy `await import()` inside the method that actually creates PTY sessions. This is the cleanest fix and benefits non-Docker users too.
   - **(b) Exclude node-pty in Docker:** Add a post-install step in the Dockerfile to `rm -rf node_modules/node-pty` after `npm ci`, and add a try/catch wrapper around the import in `session-manager.js`.
   - Option (a) is recommended -- it makes the dependency truly optional.

   Note: The PR test plan shows the full integration test ("server starts, tunnel connects") as unchecked, which likely would have caught this.

#### Suggestions (consider for this PR)

2. **`cloudflared` pinned to `latest` is non-reproducible.** The Dockerfile downloads `cloudflared` from the `latest` release URL. This means the same Dockerfile can produce different images at different times, and a breaking cloudflared release could silently break builds. Consider pinning to a specific version via a build ARG.

3. **Config file is only generated on first run.** If a user changes `API_TOKEN` or `PORT` via environment variables after the first run, the persisted `config.json` (in the named volume) will take precedence due to the `if [ ! -f "$CONFIG_FILE" ]` guard. The server config precedence system (env > file) handles this correctly at runtime, but the stale config.json could be confusing. Consider not writing apiToken/port to config.json at all and letting them flow through env vars via mergeConfig.

4. **`--legacy-cli` mode would silently fail.** If a user passes `--legacy-cli` to the Docker container, the server would crash because the `claude` binary is not installed. Consider adding a guard in the entrypoint or documenting this limitation.

#### Nits (optional improvements)

5. **`bash` is already in `node:22-slim`.** The `apt-get install bash` line is redundant since Debian slim images include bash.

6. **Config JSON heredoc uses unquoted EOF.** The `cat > "$CONFIG_FILE" <<EOF` block expands shell variables. UUID-generated tokens are safe, but if a user-provided API_TOKEN contained `$` or backticks they would be interpreted. Using quoted `<<'EOF'` and explicit variable writes (or a node one-liner) would be more robust.

### Summary

Well-structured PR with good Docker practices: layer caching via separate COPY for package files, non-root user, health check, multi-arch cloudflared, whitelist .dockerignore, exec for signal propagation. The docker-compose.yml has sensible defaults with good volume strategy for persistence.

The one critical issue is the `node-pty` static import chain that will crash the container at startup. The build succeeds because `--ignore-scripts` defers the failure to runtime. Converting the PtySession import to a dynamic `await import()` is the recommended fix.

PR #646, opened 2026-02-18
fix(app): match permission prompts by requestId in notification handler

### LINKED ISSUE #645: markPromptAnswered called with requestId instead of messageId from notification actions
## Problem

In `packages/app/src/notifications.ts`, the `setupNotificationResponseListener` calls `markPromptAnswered(requestId, decision)`, but `markPromptAnswered` expects a chat message ID (`m.id`), not a permission `requestId`.

The store implementation at `connection.ts:2315` does `m.id === messageId`, so passing a `requestId` (e.g., `perm-123`) will never match any message and the UI won't visually update to show the prompt as answered.

## Context

When permission prompts are rendered in the chat, they're assigned a message ID via `nextMessageId('perm')` and store the `requestId` as a separate field. The `handleSelectOption` in `SessionScreen.tsx` correctly passes the `messageId` from the message component. But the notification action handler only has access to `requestId` from the push notification data payload — it doesn't know the corresponding chat message ID.

## Suggested Fix

Add a store helper `markPromptAnsweredByRequestId(requestId, answer)` that finds the message with matching `requestId` field and marks it answered. Or include the `messageId` in the push notification data payload when the server sends the permission notification.

## References

- Source: PR #642, Copilot review comments on `notifications.ts:190` and `notifications.test.ts:110,212`
- File: `packages/app/src/notifications.ts` line 190
- File: `packages/app/src/store/connection.ts` line 2315

blamechris:
## Summary

- Add `markPromptAnsweredByRequestId()` to connection store that matches messages by `requestId` field instead of message ID
- Update notification response listener to use it — the push payload only contains `requestId`, not the chat message ID

## Test plan

- [x] TypeScript check passes
- [x] Notification tests pass (8/8)
- [ ] Manual: respond to permission via notification action → verify chat UI shows prompt as answered

Closes #645

### DIFF
--- a/packages/app/src/__tests__/notifications.test.ts
+++ b/packages/app/src/__tests__/notifications.test.ts
@@ -42,7 +42,7 @@ const mockSocket = {
   send: jest.fn(),
 };
 
-const mockMarkPromptAnswered = jest.fn();
+const mockMarkPromptAnsweredByRequestId = jest.fn();
 
 jest.mock('../store/connection', () => ({
   useConnectionStore: {
@@ -50,7 +50,7 @@ jest.mock('../store/connection', () => ({
       wsUrl: 'wss://test.trycloudflare.com',
       apiToken: 'test-token',
       socket: mockSocket,
-      markPromptAnswered: mockMarkPromptAnswered,
+      markPromptAnsweredByRequestId: mockMarkPromptAnsweredByRequestId,
     })),
   },
   loadConnection: jest.fn(() =>
@@ -75,7 +75,7 @@ beforeEach(() => {
     wsUrl: 'wss://test.trycloudflare.com',
     apiToken: 'test-token',
     socket: mockSocket,
-    markPromptAnswered: mockMarkPromptAnswered,
+    markPromptAnsweredByRequestId: mockMarkPromptAnsweredByRequestId,
   });
 });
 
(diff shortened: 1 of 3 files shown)


blamechris:
## Agent Review

### Verdict: Approve

Clean, well-scoped fix for a real bug. The notification handler had a data-key mismatch — it passed `requestId` to `markPromptAnswered()` which matches by `m.id` (a UUID-based chat message ID). This PR adds a dedicated `markPromptAnsweredByRequestId()` that matches by the `requestId` field on ChatMessage, and wires it into the notification path.

### Findings

**Correctness**
- The new `markPromptAnsweredByRequestId` correctly mirrors the structure of `markPromptAnswered` but matches on `m.requestId` instead of `m.id` — handles both session-state and legacy message paths.
- `SessionScreen.tsx:389` continues to use the original `markPromptAnswered(messageId, value)` which is correct for in-app UI responses where the message ID is available.
- The notification handler at `notifications.ts:206` now calls the right method.

**Code Style**
- TypeScript strict, functional patterns, Zustand conventions — all good.
- Test mocks renamed consistently across all 8 test cases.

**Tests**
- All 8 notification tests updated to use `mockMarkPromptAnsweredByRequestId`.
- CI passes: Server Tests, App Tests, App Type Check all green.

**No issues found** — no deferred suggestions needed.

### Note on #645

This PR directly addresses from-review issue #645. Once merged, that issue should be closed automatically via the "Closes #645" reference.

Issue #643, opened 2026-02-18
Add Android notification action support for permission prompts


blamechris:
## Context
PR #642 added iOS notification action buttons (Approve/Deny) for permission prompts using `setNotificationCategoryAsync`. This works on iOS but Android uses a different mechanism.

## Problem
Android users don't get action buttons on permission notification prompts — they have to open the app and navigate to the prompt manually.

## Suggested Approach
- Investigate Android notification actions via `expo-notifications` (may need custom Android notification channel configuration)
- Android's `addAction` on notifications works differently from iOS categories
- Ensure the same `approve`/`deny` action identifiers are used so the existing `setupNotificationResponseListener` handles both platforms

## References
- PR #642: iOS notification actions implementation
- [Expo Notifications: Categories](https://docs.expo.dev/versions/latest/sdk/notifications/#setnotificationcategoryasyncidentifier-actions)

PR #647, opened 2026-02-18
fix(app): add retry logic to HTTP permission response fallback

### LINKED ISSUE #644: Add retry logic for HTTP permission response fallback
## Context
PR #642 added an HTTP fallback path for permission responses when WebSocket is disconnected (e.g., user taps Approve on an iOS notification while the app is backgrounded). The current implementation makes a single HTTP request with a 10-second timeout.

## Problem
A single HTTP attempt through the Cloudflare tunnel may fail due to:
- Transient DNS/routing latency (tunnel URLs can take seconds to become routable after tunnel recovery)
- Brief network blips when the phone transitions between WiFi/cellular
- Cloudflare edge caching/routing hiccups

If the request fails, the permission response is silently lost and the user gets no feedback.

## Suggested Approach
- Add 1-2 retries with exponential backoff (e.g., 2s, 4s delays)
- Show a user-visible error (toast/alert) if all retries fail, so the user knows to open the app and respond manually
- Consider skipping `markPromptAnswered()` if the HTTP fallback fails (currently it's called unconditionally)

## References
- PR #642: HTTP permission response fallback implementation
- `packages/app/src/notifications.ts`: `sendPermissionResponseHttp()` function

blamechris:
## Summary

- Add exponential backoff retries (immediate → 2s → 4s) to `sendPermissionResponseHttp()`
- Non-retryable 4xx errors (except 408/429) bail out immediately
- Transient failures (5xx, network errors) retry up to 3 times
- 3 new tests: retry-then-succeed, no-retry-on-4xx, all-retries-exhausted

## Test plan

- [x] TypeScript check passes
- [x] Notification tests pass (11/11)
- [ ] Manual: simulate flaky tunnel → verify retry succeeds on second attempt

Closes #644

### DIFF
--- a/packages/app/src/__tests__/notifications.test.ts
+++ b/packages/app/src/__tests__/notifications.test.ts
@@ -274,35 +274,38 @@ describe('setupNotificationResponseListener', () => {
 
   it('retries HTTP on transient failure then succeeds', async () => {
     jest.useFakeTimers();
-    mockSocket.readyState = 3; // WebSocket.CLOSED
+    try {
+      mockSocket.readyState = 3; // WebSocket.CLOSED
 
-    // First call fails with 500, second succeeds
-    const mockFetch = jest.fn()
-      .mockResolvedValueOnce({ ok: false, status: 500 })
-      .mockResolvedValueOnce({ ok: true, status: 200 }) as jest.Mock;
-    global.fetch = mockFetch;
+      // First call fails with 500, second succeeds
+      const mockFetch = jest.fn()
+        .mockResolvedValueOnce({ ok: false, status: 500 })
+        .mockResolvedValueOnce({ ok: true, status: 200 }) as jest.Mock;
+      global.fetch = mockFetch;
 
-    setupNotificationResponseListener();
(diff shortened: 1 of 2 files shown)


blamechris:
## Check-PR Summary

### Copilot Comments: 2 comments → 2 fixed

| 1 | Retry count ambiguity ("retry up to 3 times" vs 3 total attempts) | **FIX** | Clarified inline comment to "3 total attempts". Commit c1686f72 |
| 2 | `jest.useRealTimers()` not in try/finally | **FIX** | Wrapped both fake-timer tests in try/finally. Commit c1686f72 |

### From-Review Issues
- **#644** (Add retry logic): will auto-close when this PR merges
- **#649** (User-visible error on retry exhaustion): created as follow-up
- **#645** (requestId vs messageId mismatch): not affected by this PR
- **#643** (Android notification actions): not affected by this PR

### Verification
- All 11 notification tests passing
- TypeScript check clean
- CI was green before fixes; new commit pushed for re-check

Ready for re-review.

Issue #748, opened 2026-02-22
service: launchctl stop does not persist with KeepAlive=true


blamechris:
## Context

From Copilot review of PR #741 (service start/stop/status).

## Problem

The launchd plist generated by `installService()` sets `KeepAlive` to `true`. This means `launchctl stop com.chroxy.server` causes launchd to immediately restart the process, making the `stop` command effectively a restart.

## Options

1. **Use `launchctl unload` / `launchctl load`** for stop/start — this properly disables KeepAlive restart behavior
2. **Remove `KeepAlive`** from plist and use `Restart=on-failure` semantics (would need a different approach)
3. **Accept the behavior** — `stop` is really a restart, and `uninstall` is the way to fully stop

## Recommendation

Option 1 is the most intuitive for users. `chroxy service stop` should stop the service and keep it stopped until `chroxy service start` is explicitly called.

References: PR #741, #726

Issue #747, opened 2026-02-22
logger: add setLogLevel() for console-only debug output


blamechris:
From review of #740.

Currently `_logLevel` is only configurable via `initFileLogging()`. In foreground mode (no file logging), debug messages are always filtered since the default level is `info`.

Adding a standalone `setLogLevel(level)` export would allow enabling debug output on the console without requiring file logging:

```js
export function setLogLevel(level) {
  _logLevel = LOG_LEVELS[level] ?? LOG_LEVELS.info
}
```

This would be useful for `--verbose` or `--debug` CLI flags in foreground mode.

Low priority enhancement.

Issue #746, opened 2026-02-22
logger: add try/catch around appendFileSync for defensive error handling


blamechris:
From review of #740.

`appendFileSync` in the `write` function of `logger.js` can throw if the log directory is deleted while logging is active. This would propagate the error to the caller and could crash the daemon.

Add a defensive try/catch around the file write:

```js
if (_logToFile && _logPath) {
  try {
    appendFileSync(_logPath, line + '\n')
    _writeCount++
    if (_writeCount % ROTATION_CHECK_INTERVAL === 0) {
      _maybeRotate()
    }
  } catch {
    // Log dir deleted or disk full — disable file logging
    _logToFile = false
  }
}
```

Low severity — the log dir is unlikely to be deleted in normal operation, but defensive coding prevents a daemon crash from an unexpected filesystem state.

blamechris:
Fixed in PR #750 — Phase 1 from-review batch.

Issue #745, opened 2026-02-22
service status: add fetch timeout and improve port parsing


blamechris:
## Context

From review of PR #741 (service start/stop/status).

### 1. Health check fetch has no timeout

`getFullServiceStatus()` calls `fetch()` without a timeout at line 473 of `service.js`. If the server is hung, the CLI blocks indefinitely.

**Fix:** Add `AbortSignal.timeout(3000)`:
```js
const response = await fetch(`http://127.0.0.1:${port}/`, {
  signal: AbortSignal.timeout(3000),
})
```

### 2. Port parsing from wsUrl is fragile

The regex `/:(\d+)/` matches the first colon-digits in the URL. For `wss://host:443/path` it would return `443` instead of the actual chroxy port. Consider using `new URL()` for robust parsing:
```js
const port = result.connection?.wsUrl ? new URL(result.connection.wsUrl).port || 8765 : 8765
```

Both are low-severity and can be addressed together.

PR #750, opened 2026-02-22
fix(server): address Phase 1 from-review issues

blamechris:
## Summary

- **#742**: Redact `apiToken` from `/connect` response when `authRequired: false` (prevent token exposure on open endpoints)
- **#743**: Replace deprecated `launchctl load/unload` with `launchctl bootstrap/bootout gui/<uid>` for modern macOS compatibility
- **#745**: Add 3s `AbortSignal.timeout()` to health check fetch in `getFullServiceStatus()`, resolve port from `config.json` before fallback
- **#746**: Wrap `appendFileSync` in try/catch in logger so disk errors don't crash the server
- **#747**: Add exported `setLogLevel(level)` function for runtime log level changes (enables `--verbose` CLI flags)
- **#748**: Fix `stopService()` to use `bootout` instead of `launchctl stop` (which is ineffective with `KeepAlive: true`)

## Test plan

- [x] 15 new tests covering all 6 fixes
- [x] Full suite: 1122 pass, 3 pre-existing tunnel flakes
- [x] TDD workflow: tests written first, verified failing, then implementation

### DIFF
--- a/packages/server/src/logger.js
+++ b/packages/server/src/logger.js
@@ -81,6 +81,17 @@ function _maybeRotate() {
 }
 
 /**
+ * Change the minimum log level at runtime.
+ * Useful for --verbose CLI flags to enable debug output. (#747)
+ * @param {string} level - One of 'debug', 'info', 'warn', 'error'
+ */
+export function setLogLevel(level) {
+  if (level in LOG_LEVELS) {
+    _logLevel = LOG_LEVELS[level]
+  }
+}
+
+/**
  * Create a component logger. Backward-compatible with existing API.
  * @param {string} component - Component name for log prefix
  * @returns {{ debug: Function, info: Function, warn: Function, error: Function, log: Function }}
@@ -99,10 +110,15 @@ export function createLogger(component) {
 
     // Write to file in daemon mode
     if (_logToFile && _logPath) {
-      appendFileSync(_logPath, line + '\n')
-      _writeCount++
-      if (_writeCount % ROTATION_CHECK_INTERVAL === 0) {
-        _maybeRotate()
+      try {
+        appendFileSync(_logPath, line + '\n')
+        _writeCount++
+        if (_writeCount % ROTATION_CHECK_INTERVAL === 0) {
+          _maybeRotate()
+        }
+      } catch {
+        // Silently ignore write failures (disk full, permission denied, etc.)
+        // to prevent logging errors from crashing the server (#746)
       }
     }
   }
--- a/packages/server/src/service.js
+++ b/packages/server/src/service.js
@@ -306,7 +306,7 @@ export function installService(config) {
   // Register with system (unless testing)
   if (!config._skipRegister) {
     if (plat === 'darwin') {
-      execFileSync('launchctl', ['load', servicePath])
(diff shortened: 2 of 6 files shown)


blamechris (commented):
## Agent Review — PR #750

### Overview

6 from-review fixes bundled into Phase 1, all server-side. 15 new tests. Clean TDD workflow.

### Code Style Compliance

- **ES modules**: All `import`/`export` — correct
- **No semicolons**: Clean throughout
- **Single quotes**: Consistent
- **No TypeScript**: Plain JS — correct

### Per-Fix Analysis

#### #742 — Redact apiToken on no-auth /connect

**Good**: Prevents token exposure when `authRequired: false`. `readConnectionInfo()` returns a fresh parsed object each call, so the in-place mutation is safe (no cached reference corruption).

**Issue**: The `connectionUrl` field (e.g., `chroxy://host?token=secret-token-abc123`) still contains the raw token in the query string. If the goal is to prevent token exposure, `connectionUrl` should be redacted too.

**Severity**: Low — `connectionUrl` is a convenience field, but it leaks the same secret.

#### #743 — launchctl bootstrap/bootout

**Good**: Correctly replaces deprecated `load`/`unload` with modern `bootstrap`/`bootout`. Uses `gui/${process.getuid()}` domain, which is the correct user-level domain for launchd.

**Clean**.

#### #745 — Health check fetch timeout + port resolution

**Good**: `AbortSignal.timeout(3000)` prevents the health check from hanging indefinitely. Port resolution order (config.json > connection URL > default) is sensible.

**Note**: `AbortSignal.timeout()` requires Node 18+. This project requires Node 22, so this is fine.

**Clean**.

#### #746 — Logger file write error handling

**Good**: Bare `catch {}` wraps `appendFileSync` + rotation. Console output remains unaffected.

**Clean**. Silent failure is the correct behavior for a logger — you don't want logging errors to cascade.

#### #747 — setLogLevel() runtime function

**Good**: Simple, validated setter. Ignores invalid levels silently (correct — no-op is better than throwing for a log-level setter).

**Clean**.

#### #748 — stopService uses bootout

**Good**: `launchctl stop` with `KeepAlive: true` just restarts the process. `bootout` correctly removes the job from the domain, which actually stops it.

**Note**: After `bootout`, the job is unloaded — a subsequent `startService()` would need to `bootstrap` again (which is already what `startService()` does). Lifecycle is consistent.

**Clean**.

### Test Quality

- Tests use temp directories with proper cleanup (beforeEach/afterEach)
- `CHROXY_CONFIG_DIR` env var properly saved/restored in ws-server tests
- `_skipRegister` / `_skipExec` flags prevent real launchctl calls in tests
- No real user config files touched (lesson from #429)

One blank line between `describe` blocks in `service.test.js` (line 602) — cosmetic only, not a blocker.

### Security

- Token redaction logic is correct for `apiToken` field
- `connectionUrl` field still contains the token (see #742 note above)
- No new attack surface introduced

### Summary

**Approve with 1 minor suggestion**: Redact `connectionUrl` alongside `apiToken` in the no-auth /connect handler. All 6 fixes are well-implemented, properly tested, and follow project conventions.

Issue #762, opened 2026-02-22
dashboard: querySelector injection hardening for message/tool IDs


blamechris:
## Context

PR #758 uses `querySelector` with interpolated message IDs:
```js
messagesEl.querySelector('[data-msg-id="' + msg.messageId + '"]')
messagesEl.querySelector('[data-tool-id="' + toolId + '"]')
```

## Problem

A malformed `messageId` containing `"]` could break the CSS selector, causing `querySelector` to throw. This is not exploitable for XSS (CSS selectors cannot execute JS), but it would disrupt the streaming display.

## Suggested Fix

Sanitize IDs before use in selectors:
```js
function escapeSelectorValue(val) {
  return String(val).replace(/["\\]/g, '\\$&')
}
```

Or use `document.querySelector` with `CSS.escape()`:
```js
messagesEl.querySelector('[data-msg-id="' + CSS.escape(msg.messageId) + '"]')
```

## Severity

Low — requires a malicious server or MITM to send crafted message IDs.

blamechris:
Addressed in PR #774.

Issue #761, opened 2026-02-22
dashboard: add plan mode and background agent UI


blamechris:
## Context

PR #758 implements the core web dashboard chat UI. Several features from the mobile app are not yet implemented in the dashboard:

## Missing Features

1. **Plan mode UI** — `plan_started` / `plan_ready` events are not handled. The mobile app shows a PlanApprovalCard with Approve/Give Feedback buttons.

2. **Background agent monitoring** — `agent_spawned` / `agent_completed` events are not handled. The mobile app shows a badge count and detail view in the settings bar.

3. **Session creation feedback** — `session_created` / `session_error` responses from `create_session` are not handled, so the user gets no confirmation or error when creating a new session.

## Priority

Low — these are progressive enhancement for dashboard parity with the mobile app.

Issue #760, opened 2026-02-22
dashboard: add javascript: URI sanitization in markdown link renderer


blamechris:
## Context

PR #758 introduces a markdown renderer in the web dashboard that converts `[text](url)` to `<a href="url">` links. The renderer escapes HTML entities but does not validate the URL scheme, allowing `javascript:` URIs through.

## Problem

Claude's responses could contain markdown like:
```markdown
[click me](javascript:alert(document.cookie))
```

Since the HTML entity escaping happens _before_ the link regex, the `javascript:` scheme passes through to the `href` attribute unmodified.

## Suggested Fix

Validate the URL scheme in the link regex replacement:
```js
html = html.replace(/\[([^\]]+)\]\(([^)]+)\)/g, function(m, text, url) {
  if (/^(https?:|mailto:|#)/i.test(url)) {
    return '<a href="' + url + '" target="_blank" rel="noopener">' + text + '</a>'
  }
  return text + ' (' + url + ')'
})
```

## Severity

Medium — requires Claude to output a malicious link, which is unlikely but possible in adversarial prompt scenarios.

Issue #756, opened 2026-02-22
startService: improve error handling when servicePath is missing from state


blamechris:
## Context

Copilot review on PR #750 flagged the `kickstart` fallback in `startService()` (service.js line ~386).

When `loadServiceState()` returns no `servicePath`, the code falls back to `launchctl kickstart gui/<uid>/com.chroxy.server`. This only works if the job is already loaded in the launchd domain — which won't be the case after a `bootout` (the new stop behavior from #748).

## Current behavior

```js
if (state?.servicePath) {
  execFileSync('launchctl', ['bootstrap', `gui/${process.getuid()}`, state.servicePath])
} else {
  execFileSync('launchctl', ['kickstart', `gui/${process.getuid()}/${SERVICE_LABEL}`])
}
```

## Suggested improvement

Replace the `kickstart` fallback with a clearer error:

```js
if (!state?.servicePath) {
  throw new Error(
    'Chroxy service state is missing or corrupted. Please reinstall: npx chroxy service install'
  )
}
execFileSync('launchctl', ['bootstrap', `gui/${process.getuid()}`, state.servicePath])
```

Or alternatively, attempt to locate the plist file at the default path before throwing.

## Severity

Low — the `kickstart` fallback would fail with a cryptic launchd error anyway in most cases. A clear error message would be better UX.

## Origin

Copilot review on PR #750 (comment on service.js line 386).

blamechris:
Addressed in PR #774.

PR #763, opened 2026-02-22
fix(server): from-review batch — dashboard hardening, server fixes (#753-#756, #760-#762)

### LINKED ISSUE #753: Redact connectionUrl in /connect response when authRequired is false
## Context

PR #750 (fix #742) redacts `apiToken` from the `/connect` response when `authRequired: false` to prevent token exposure on open endpoints.

However, the `connectionUrl` field (e.g., `chroxy://host?token=secret-token-abc123`) still contains the raw token embedded in the query string, which defeats the redaction.

## Suggested Fix

In the `/connect` handler in `ws-server.js`, alongside `connInfo.apiToken = '[REDACTED]'`, also strip the token from `connectionUrl`:

```js
if (!this.authRequired && connInfo.apiToken) {
  connInfo.apiToken = '[REDACTED]'
  if (connInfo.connectionUrl) {
    try {
      const url = new URL(connInfo.connectionUrl)
      url.searchParams.set('token', '[REDACTED]')
      connInfo.connectionUrl = url.toString()
    } catch {
      connInfo.connectionUrl = '[REDACTED]'
    }
  }
}
```

## Severity

Low — `connectionUrl` is a convenience field used by the mobile app's QR scanner, and in no-auth mode the token is not required for authentication anyway. But for defense-in-depth, the token should not appear in any field.

## Origin

Agent review of PR #750.

blamechris:
## Summary

Fast-follow PR addressing 7 from-review issues across dashboard and server:

### Dashboard (#760, #761, #762)
- **querySelector injection hardening** — `sanitizeId()` strips CSS-special chars from message/tool IDs before selector interpolation
- **javascript: URI blocking** — markdown link renderer rejects `javascript:`, `data:`, `vbscript:` URI schemes (case-insensitive, whitespace-tolerant)
- **Plan mode UI** — banner + approval card with Approve/Give Feedback buttons for `plan_started`/`plan_ready` events
- **Background agent badge** — status bar shows agent count from `agent_spawned`/`agent_completed` events

### Server (#753, #754, #755, #756)
- **connectionUrl redaction** — `/connect` endpoint now also strips `connectionUrl` (contains embedded token) when `authRequired: false`
- **Dashboard URL in PTY banner** — PTY mode startup banner now includes dashboard URL
- **XFF spoofing guard** — localhost bypass uses `socketIp` (kernel-set `remoteAddress`) instead of `ip` (may include proxy headers)
- **startService error handling** — clear error messages for missing/stale service state instead of silent fallback to kickstart

## Test plan

- [x] 33 new dashboard tests (sanitizeId, querySelector hardening, URI blocking, plan mode UI, agent badge)
- [x] 4 new ws-server tests (connectionUrl redaction, socketIp vs ip, localhost bypass with encryption)
- [x] 3 new service tests (stale servicePath, _stateDir wiring, backward compat)
- [x] All 67 dashboard tests pass
- [x] All 64 service tests pass
- [x] 215/223 ws-server tests pass (8 pre-existing failures in unrelated get_diff handler)

Closes #753, #754, #755, #756, #760, #761, #762

### DIFF
--- a/packages/server/tests/dashboard.test.js
+++ b/packages/server/tests/dashboard.test.js
@@ -184,7 +184,7 @@ describe('dashboard markdown renderer', () => {
   })
 })
 
-describe('#760 — sanitizeId strips special characters', () => {
+describe('#762 — sanitizeId strips special characters', () => {
   const html = getDashboardHtml(8765, 'test-token', false)
 
   it('defines sanitizeId function', () => {
@@ -199,7 +199,7 @@ describe('#760 — sanitizeId strips special characters', () => {
   })
 })
 
-describe('#760 — querySelector calls use sanitized IDs', () => {
+describe('#762 — querySelector calls use sanitized IDs', () => {
   const html = getDashboardHtml(8765, 'test-token', false)
 
   it('uses sanitizeId for data-msg-id queries', () => {
@@ -247,7 +247,7 @@ describe('#760 — querySelector calls use sanitized IDs', () => {
   })
 })
 
-describe('#762 — javascript: URI blocking in markdown links', () => {
+describe('#760 — javascript: URI blocking in markdown links', () => {
   const html = getDashboardHtml(8765, 'test-token', false)
 
   it('blocks javascript: URIs', () => {


blamechris:
## Code Review: PR #763

### Summary

Batch fix addressing 7 from-review issues across dashboard and server:

**Dashboard (3 issues):** querySelector injection hardening via `sanitizeId()`, javascript:/data:/vbscript: URI blocking in the markdown link renderer (with case-insensitive + whitespace-tolerant regex), plan mode UI (banner + approval card), and background agent badge in the status bar.

**Server (4 issues):** `connectionUrl` redaction in the `/connect` endpoint when auth is disabled, dashboard URL in PTY mode startup banner, XFF spoofing guard using `socketIp` for localhost bypass decisions, and improved `startService()` error handling for missing/stale service state.

653 additions, 11 deletions across 7 files. 40 new tests (33 dashboard, 4 ws-server, 3 service).

---

### Strengths

1. **XFF spoofing fix is well-executed.** Separating `socketIp` (kernel-set `req.socket.remoteAddress`) from `ip` (proxy-header chain) is the correct architectural decision. The existing `ip` field is preserved for logging/rate-limiting, while security-critical localhost bypass uses only the unforgeable socket address. Good inline comments explaining the rationale.

2. **URI blocking regex is more robust than the original suggestion.** The `from-review` issue (#760) suggested a prefix allowlist (`https?:|mailto:|#`), but the implementation uses a denylist regex `/^\s*(javascript|data|vbscript)\s*:/i` which is more permissive for legitimate schemes while blocking the dangerous ones. The `\s*` handling catches whitespace evasion (`java\tscript:`) that the original `.trim().toLowerCase()` approach would miss between the scheme name and colon.

3. **`connectionUrl` redaction uses `delete` instead of URL manipulation.** The original #753 suggestion proposed parsing the URL and replacing the token param. The `delete connInfo.connectionUrl` approach is simpler and more robust -- no risk of URL parsing edge cases, and the field serves no purpose when auth is disabled anyway.

4. **Service error handling covers both failure modes.** The stale-servicePath case (file deleted after state was saved) and the missing-state case (state.json absent but plist exists at default location) both have clear, actionable error messages directing the user to `chroxy service install`. The fallback to `paths.plistPath` when state is missing is a good defensive recovery.

5. **Test coverage is thorough.** 40 new tests cover both the happy paths and the security invariants (e.g., verifying that `socketIp` is a localhost address for direct connections, verifying `connectionUrl` is absent in no-auth mode but preserved in auth mode).

---

### Issues Found

#### Suggestion S1: `sanitizeId()` strips characters from IDs but `setAttribute` stores raw IDs -- mismatch will break lookups

The `sanitizeId()` function strips all non-`[a-zA-Z0-9_-]` characters from IDs before `querySelector` lookups (lines 1127, 1130, 1153). However, the `setAttribute` calls that **set** the `data-msg-id` and `data-tool-id` values on the DOM elements use the **raw, unsanitized** IDs:

- Line 743: `div.setAttribute("data-msg-id", opts.id)` -- raw
- Line 1113: `streamDiv.setAttribute("data-msg-id", streamingMsgId)` -- raw
- Line 752: `div.setAttribute("data-tool-id", toolUseId || "")` -- raw

If an ID contains a character like `.` or `@` (which are plausible in UUIDs or server-generated IDs), `sanitizeId` will strip it from the lookup but the attribute still contains it -- the `querySelector` will never find the element.

**Fix:** Apply `sanitizeId()` at the `setAttribute` call sites as well, so both sides use the same sanitized value:
```js
if (opts.id) div.setAttribute("data-msg-id", sanitizeId(opts.id));
```

**Severity:** Medium -- won't cause security issues, but will silently break streaming text updates and tool result display for any message ID containing non-alphanumeric characters (like dots in UUIDs: `msg_01.abc`).

#### Suggestion S2: Plan approval card is never hidden after a second `plan_started` cycle

The `plan_ready` handler shows the approval card (line 1250), and clicking "Approve" or "Give Feedback" hides it. But if a new plan cycle starts (`plan_started` -> `plan_ready`), the `plan_started` handler only hides the banner -- it does not hide a stale approval card from the previous cycle.

If the user leaves the approval card visible (doesn't click either button) and a new `plan_started` fires, the old card persists alongside the new banner. Then when `plan_ready` fires again, `planContentEl.innerHTML` is overwritten but the card was already visible, which is mostly fine, except the stale state is confusing.

**Fix:** Add `planApprovalCard.classList.add("hidden")` in the `plan_started` handler to dismiss any leftover approval card.

**Severity:** Low -- edge case but easy to fix.

#### Nitpick N1: Plan "Approve" button sends a chat message instead of a protocol message

The plan approve handler (line 1301) calls `sendInput("Looks good, proceed.")` which sends a `{ type: "input", data: "Looks good, proceed." }` message -- i.e., it types a chat message on behalf of the user. The mobile app's PlanApprovalCard presumably sends a proper WS protocol message for plan approval (similar to permission responses).

This works in practice because Claude interprets the text as approval, but it's a UI-level workaround rather than protocol-level communication. Not blocking for this PR since the mobile app may work the same way, but worth noting for protocol consistency.

---

### Architecture Notes

- The `socketIp` field is added to the client metadata object in `ws-server.js` and is only read during the localhost bypass check. This is clean -- no changes to the WS protocol, no changes to the client-facing API. The `ip` field continues to serve its existing purpose for logging.

- The dashboard JS continues to be a single self-contained string template returned by `getDashboardJs()`. At 1300+ lines, this is getting large but splitting it is a separate concern (tracked by #733 session management).

- The `startService` changes maintain backward compatibility: the `_stateDir` parameter defaults to `DEFAULT_CONFIG_DIR`, and the new test verifies `_skipExec` still works without state.

---

### Verdict: **Approve with suggestion**

The security fixes (XFF spoofing, URI blocking, connectionUrl redaction) are all correct and well-tested. The `sanitizeId` setAttribute mismatch (S1) should be fixed in this PR or a fast-follow since it can break streaming display for IDs with dots/special chars. The plan approval card stale state (S2) is low-priority.

| # | Type | Description | Disposition |
|---|------|-------------|-------------|
| S1 | Suggestion | `sanitizeId` applied to querySelector but not setAttribute -- mismatch breaks lookups | Fix in this PR |
| S2 | Suggestion | `plan_started` should hide stale approval card | Fix in this PR or follow-up |
| N1 | Nitpick | Plan approve sends chat text instead of protocol message | Informational |

Issue #846, opened 2026-02-24
parseDuration('0') vs '0h' zero-handling inconsistency


blamechris:
## Context

PR #842 added tests that document the current behavior of `parseDuration` for zero values:

- `parseDuration('0')` → `0` (goes through the plain-number path: `/^\d+$/` match → `parseInt * 1000`)
- `parseDuration('0h')` → `null` (passes validation, regex matches `0h`, but `found && total > 0` is false)
- `parseDuration('0m0s')` → `null` (same logic)

## Issue

The dual semantics of "zero" may surprise future callers:
- A bare `'0'` means "zero seconds" and returns `0` (a falsy but non-null value)
- `'0h'` means "zero hours" but returns `null` (meaning "invalid")

Since `parseDuration` is used for `tokenExpiry` and `sessionTimeout`, a return value of `0` for `'0'` could disable timeouts in unexpected ways (e.g., `if (parsedTimeout)` would treat `0` as falsy).

## Proposal

Decide on consistent zero semantics:
1. **All zeros return `null`**: treat zero durations as invalid (add `if (cleaned === '0') return null` or change to `parseInt > 0 ? ms : null`)
2. **All zeros return `0`**: treat zero as valid (remove `total > 0` guard, or document the bare-number special case)
3. **Keep current behavior**: document the asymmetry with a comment in the source

## Source

Agent review of PR #842

PR #854, opened 2026-02-24
fix(server,app): consolidate parseDuration + session_list GC edge case (#851, #852, #853)

### LINKED ISSUE #851: test(app): add 'all sessions removed' edge case for session_list GC
## Context

Identified during review of PR #849.

## Description

The `session_list GC handler` test block in `message-handler.test.ts` covers 4 cases (basic removal, no false positives, multiple removals, active session switch), but does not test the edge case where `session_list` arrives with an empty array while local sessions exist.

The implementation handles this path at `message-handler.ts:675-684` — clearing flat fields (`activeSessionId` to null, `messages` to `[]`, `claudeReady` to `false`, etc.) when no sessions remain after GC. The existing `session_timeout` handler has a dedicated test for this ("clears flat fields when no sessions remain" at line 116), but `session_list` does not.

## Acceptance Criteria

- [ ] Add a test case in the `session_list GC handler` describe block that sends `{ type: 'session_list', sessions: [] }` when local sessions exist
- [ ] Assert `activeSessionId` becomes `null`
- [ ] Assert `messages` becomes `[]`
- [ ] Assert `claudeReady` becomes `false`
- [ ] Assert `activeModel` becomes `null`
- [ ] Assert `clearPersistedSession` is called for all previously-known sessions

## Files

- `packages/app/src/__tests__/store/message-handler.test.ts`

## References

- PR #849 (review comment)
- Implementation: `packages/app/src/store/message-handler.ts:641-689`

### LINKED ISSUE #852: fix(server): parseDuration('0') returns 0 in duration.js — same bug as #846
## Context

Identified during review of PR #850, which fixed the same bug in `token-manager.js` (closing #846).

## Description

`packages/server/src/duration.js` has its own independent implementation of `parseDuration`. The plain-number path at lines 16-18 returns `0` for input `'0'`:

```js
if (/^\d+$/.test(cleaned)) {
  return parseInt(cleaned, 10) * 1000
}
```

This is the same inconsistency documented in #846: `parseDuration('0')` returns `0` while `parseDuration('0h')` returns `null`.

`duration.js` is imported by `session-manager.js` for `--sessionTimeout` parsing. In practice, `0` is falsy and gets caught by the `if (!this._sessionTimeoutMs) return` guard, so it does not cause runtime issues. But the inconsistency should be fixed for correctness.

## Fix

Apply the same pattern used in `token-manager.js`:

```js
if (/^\d+$/.test(cleaned)) {
  const ms = parseInt(cleaned, 10) * 1000
  return ms > 0 ? ms : null
}
```

## Files

- `packages/server/src/duration.js` (line 16-18)

## References

- PR #850 (fixed `token-manager.js` copy)
- Issue #846 (original inconsistency report)

### LINKED ISSUE #853: refactor(server): consolidate duplicate parseDuration implementations
## Context

Identified during review of PR #850.

## Description

The codebase has two independent implementations of `parseDuration`:

1. **`packages/server/src/token-manager.js`** (lines 32-61) — used for token expiry. Uses a regex-loop parser (`/(\d+)\s*(s|m|h|d)/g`).
2. **`packages/server/src/duration.js`** (lines 10-32) — used for session timeout. Uses an anchored pattern parser (`/^(?:(\d+)d)?(?:(\d+)h)?(?:(\d+)m)?(?:(\d+)s)?$/`).

Both accept the same input format and return milliseconds or `null`, but they have slightly different validation strategies and have already diverged on zero-handling (#846). Having two copies invites future drift.

## Proposal

- Consolidate into a single `parseDuration` in `duration.js` (the dedicated module)
- Update `token-manager.js` to import from `duration.js` instead of having its own copy
- Ensure the consolidated version passes all existing tests from both test files

## Files

- `packages/server/src/duration.js`
- `packages/server/src/token-manager.js`
- `packages/server/tests/token-manager.test.js` (parseDuration tests)

## References

- PR #850 (review)
- Issue #846 (zero-handling inconsistency)

blamechris:
Closes #851, closes #852, closes #853

## Summary

- **#852**: Fix `parseDuration('0')` returning `0` in `duration.js` — now returns `null`, consistent with `'0h'` and the token-manager fix from #846
- **#853**: Remove duplicate `parseDuration` from `token-manager.js`, import from the canonical `duration.js` module. Re-export preserves API compatibility for existing `import { parseDuration } from './token-manager.js'`
- **#851**: Add "all sessions removed via empty list" edge case test to `message-handler.test.ts` — verifies flat fields cleared, `clearPersistedSession` called for all sessions

## Test plan

- [x] Server: `node --test --test-name-pattern="parseDuration|TokenManager" tests/token-manager.test.js` — 24 pass
- [x] App: `npx jest src/__tests__/store/message-handler.test.ts` — 11 pass
- [x] Verified `duration.js` handles all token-manager test edge cases (invalid chars, whitespace, case insensitivity)

### DIFF
--- a/packages/app/src/__tests__/store/message-handler.test.ts
+++ b/packages/app/src/__tests__/store/message-handler.test.ts
@@ -310,6 +310,41 @@ describe('session_list GC handler', () => {
     expect(state.activeSessionId).toBe('s2');
     expect(state.messages).toEqual(s2State.messages);
   });
+
+  it('clears flat fields when all sessions removed via empty list', () => {
+    const store = createMockStore({
+      activeSessionId: 's1',
+      sessions: [
+        { sessionId: 's1', name: 'S1' } as any,
+        { sessionId: 's2', name: 'S2' } as any,
+      ],
+      sessionStates: {
+        s1: createEmptySessionState(),
+        s2: createEmptySessionState(),
+      },
+      messages: [{ id: 'm1', type: 'response' as const, content: 'test', timestamp: 1 }],
+      claudeReady: true,
+      activeModel: 'claude-sonnet',
+    });
+
+    setStore(store as any);
+    _testMessageHandler.setContext(createMockContext() as any);
+
+    _testMessageHandler.handle({
+      type: 'session_list',
+      sessions: [],
+    });
+
+    expect(clearPersistedSession).toHaveBeenCalledWith('s1');
+    expect(clearPersistedSession).toHaveBeenCalledWith('s2');
+    expect(clearPersistedSession).toHaveBeenCalledTimes(2);
+
+    const state = store.getState();
+    expect(state.activeSessionId).toBeNull();
+    expect(state.messages).toEqual([]);
+    expect(state.claudeReady).toBe(false);
+    expect(state.activeModel).toBeNull();
+  });
 });
 
 afterAll(() => {
--- a/packages/server/src/duration.js
+++ b/packages/server/src/duration.js
@@ -14,7 +14,8 @@ export function parseDuration(str) {
 
   // Pure numeric → treat as seconds
   if (/^\d+$/.test(cleaned)) {
-    return parseInt(cleaned, 10) * 1000
+    const ms = parseInt(cleaned, 10) * 1000
+    return ms > 0 ? ms : null
   }
 
   const pattern = /^(?:(\d+)d)?(?:(\d+)h)?(?:(\d+)m)?(?:(\d+)s)?$/
--- a/packages/server/src/token-manager.js
+++ b/packages/server/src/token-manager.js
@@ -1,6 +1,7 @@
 import { randomUUID } from 'crypto'
 import { EventEmitter } from 'events'
 import { safeTokenCompare } from './crypto.js'
+import { parseDuration } from './duration.js'
 
 /**
  * Manages token lifecycle with optional rotation and expiry.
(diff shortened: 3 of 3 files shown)


blamechris:
## Code Review — PR #854

**fix(server,app): consolidate parseDuration + session_list GC edge case (#851, #852, #853)**

---

### Summary

This PR addresses three `from-review` issues in a single batch:
1. **#852** — Fixes `parseDuration('0')` in `duration.js` returning `0` instead of `null`
2. **#853** — Removes the duplicate `parseDuration` from `token-manager.js`, consolidating on the canonical `duration.js` implementation with a re-export for backward compatibility
3. **#851** — Adds an edge-case test for `session_list` GC when the server sends an empty session list

### Strengths

- **Clean deduplication**: Removing ~35 lines of duplicate logic in `token-manager.js` and replacing with a single import + re-export is the right call. The re-export (`export { parseDuration }`) preserves API compatibility for any code importing from `token-manager.js`.
- **Correct bug fix**: The `parseDuration('0')` fix at `duration.js:17-18` is minimal and consistent with how the function already handles `'0h'` and `'0m0s'` (both return `null`).
- **Thorough test coverage**: The new `message-handler.test.ts` test at line 314 properly validates all four relevant state fields (`activeSessionId`, `messages`, `claudeReady`, `activeModel`) and verifies `clearPersistedSession` is called for *both* removed sessions with exact call count.
- **Tests verified**: All 44 tests pass (24 token-manager, 9 duration, 11 message-handler).

### Issues Found

**None (Critical)**

**Suggestions:**

1. **Subtle behavioral difference in consolidated `parseDuration`** — `duration.js` uses a strict positional regex `^(?:(\d+)d)?(?:(\d+)h)?(?:(\d+)m)?(?:(\d+)s)?$` that requires units in d-h-m-s order with no whitespace between them. The old `token-manager.js` used a `regex.exec` loop with `(\d+)\s*(s|m|h|d)` that accepted any order and spaces between parts (e.g., `'30m 2h'` and `'2h 30m'`). After consolidation, both of these now return `null`. This is extremely unlikely to affect real usage (who writes `'30m 2h'` for `tokenExpiry`?), but worth documenting for completeness. **No action needed** — the stricter behavior is arguably better.

2. **Missing `'0'` test case in `duration.test.js`** — The `parseDuration('0') === null` assertion exists in `token-manager.test.js:50` (via re-export), but `duration.test.js` itself has no explicit test for `'0'`. Since the bug fix lives in `duration.js:17-18`, the canonical test file should also cover it. Low priority since the behavior IS tested transitively.

**Nitpicks:**

None.

### Deferred Items

1. **Add `parseDuration('0')` test to `duration.test.js`** — The canonical test file for `duration.js` should include explicit coverage for the zero-numeric edge case that this PR fixes.

### Architecture Notes

- The re-export pattern (`export { parseDuration }`) in `token-manager.js:29` is a clean compatibility bridge. Currently only `token-manager.test.js` imports `parseDuration` from `token-manager.js` — `session-manager.js` already imports directly from `duration.js`. The re-export could be removed in a future cleanup pass once the test import is updated.
- The two `parseDuration` implementations had subtly different parsing strategies (positional regex vs. exec-loop). The `duration.js` version is stricter, which is the correct choice for a canonical utility.

### Verdict

**Approve** — Clean, well-tested batch that closes three outstanding from-review issues. The changes are minimal, correct, and all tests pass.

| Category | Count |
|----------|-------|
| Critical | 0 |
| Suggestions | 2 |
| Nitpicks | 0 |
| Deferred | 1 |

PR #913, opened 2026-02-26
feat(server,desktop): QR mobile pairing in desktop app

blamechris:
## Summary

- **Server**: Add `GET /qr` endpoint (SVG QR code from connection.json), write connection info in no-tunnel mode with LAN IP, add QR pairing button + modal to web dashboard
- **Desktop**: Configurable tunnel mode (Quick/Named/Local Only) via tray submenu, redesigned loading page with QR display for phone pairing, cloudflared availability check with graceful fallback
- **New file**: `lan-ip.js` — LAN IP detection utility for connection URLs without a tunnel

## Test plan

- [ ] Start server with quick tunnel, verify `curl http://localhost:8765/qr` returns SVG
- [ ] Start with `CHROXY_TUNNEL=none`, verify `~/.chroxy/connection.json` is written with LAN IP
- [ ] Build desktop app, verify tray menu shows Tunnel Mode submenu, settings persist across restarts
- [ ] Launch desktop app with Quick Tunnel → loading page → QR appears → scan with phone → connects
- [ ] Open dashboard → click QR button in header → modal shows QR → scan works
- [ ] Test without cloudflared installed → tunnel times out → error with install instructions shown
- [ ] `cd packages/server && npm test` — all existing tests pass, no regressions

### DIFF
--- a/packages/desktop/dist/index.html
+++ b/packages/desktop/dist/index.html
@@ -338,7 +338,9 @@
       var timeoutMs = tunnelMode !== 'none' ? 45000 : 10000;
 
       function poll() {
-        fetch('http://localhost:' + port + '/qr')
+        fetch('http://localhost:' + port + '/qr', {
+            headers: token ? { 'Authorization': 'Bearer ' + token } : {}
+          })
           .then(function(r) {
             if (r.ok) return r.text();
             if (r.status === 503) throw new Error('not_ready');
--- a/packages/server/src/dashboard.js
+++ b/packages/server/src/dashboard.js
@@ -1901,7 +1901,7 @@ function getDashboardJs() {
     qrModal.classList.remove("hidden");
     modalOpen = true;
 
-    fetch('/qr')
+    fetch('/qr', { headers: token ? { 'Authorization': 'Bearer ' + token } : {} })
       .then(function(r) {
         if (!r.ok) throw new Error(r.status);
         return r.text();
--- a/packages/server/src/ws-server.js
+++ b/packages/server/src/ws-server.js
@@ -393,9 +393,9 @@ export class WsServer {
       }
 
       // QR code endpoint — generates SVG QR from connection URL
-      // No auth required: only reachable on localhost, and QR data is equivalent
-      // to what's already in ~/.chroxy/connection.json
-      if (req.method === 'GET' && req.url === '/qr') {
+      // Auth required: QR contains the connection token
+      if (req.method === 'GET' && req.url?.startsWith('/qr')) {
+        if (!this._validateBearerAuth(req, res)) return
         const connInfo = readConnectionInfo()
         if (!connInfo || !connInfo.connectionUrl) {
           res.writeHead(503, { 'Content-Type': 'application/json' })


blamechris:
## Code Review: PR #913

### Summary
This PR adds QR code mobile pairing capabilities to the desktop app and web dashboard. It introduces:
1. **Server**: A `GET /qr` endpoint serving SVG QR codes from `connection.json`, a new `lan-ip.js` utility for LAN IP detection, `connection.json` is now written in no-tunnel mode, and a QR pairing button/modal in the web dashboard.
2. **Desktop (Tauri)**: Configurable tunnel mode (Quick/Named/Local Only) via tray submenu with persistence, a redesigned loading page that shows a QR code after startup, and cloudflared availability checking with graceful fallback.

The feature bridges a real usability gap — previously, pairing a phone with the desktop app required manual URL entry. Now users scan a QR code directly from the loading page or dashboard.

### Strengths
- **Well-structured rollout**: Server QR endpoint is cleanly separated from the desktop-specific loading flow. The `lan-ip.js` utility is minimal and reusable.
- **Graceful degradation**: When cloudflared is missing, the desktop app falls back to "Local Only" mode with a clear notification. When QR is unavailable in local mode, the dashboard button is shown as fallback.
- **Proper security posture**: The `/qr` endpoint is unauthenticated by design since it only serves data already present in `~/.chroxy/connection.json` and is only reachable on localhost. This is well-documented in the code comment.
- **Loading page UX**: The two-phase flow (health check → QR poll) with distinct timeout values (10s local, 45s tunnel) handles the real-world timing of Cloudflare tunnel establishment.
- **Radio-style checkboxes**: The tunnel mode submenu correctly implements mutual exclusion with `set_checked` for all three options.
- **Settings persistence**: `tunnel_mode` is persisted via `DesktopSettings` with proper serde defaults for backwards compatibility with existing settings files.
- **Mutex safety**: Consistent use of `lock_or_recover` throughout, and explicit `drop(settings_guard)` before acquiring other locks in `setup_tray`.
- **CI is green**: All 4 check suites pass.

### Issues Found

#### Critical (Must Fix)

_None found._

| File | Line(s) | Suggestion | Rationale |
|------|---------|------------|-----------|
| `packages/desktop/src-tauri/src/lib.rs` | 249, 262 | `check_cloudflared()` is called **twice** in `handle_start` — once for the notification (L249) and again for the effective mode calculation (L262). Cache the result in a local `bool` and reuse it. | Each call spawns a `which cloudflared` subprocess. Redundant process spawning on every server start. |
| `packages/desktop/src-tauri/src/window.rs` | 86 | The `tunnel_mode` parameter in the `win.eval()` JS string is not escaped (unlike `token` which gets backslash/quote escaping). | Currently safe because values are hardcoded ("quick"/"named"/"none"), but inconsistent with the defense-in-depth pattern already applied to the token parameter. A future refactor might pass user-derived strings here. |
| `packages/desktop/dist/index.html` | 253-256 | `showQr` injects server-returned SVG into `innerHTML`. While the QR library output is trusted, this is a raw HTML injection point. | The SVG comes from the `qrcode` npm package via the server's `/qr` endpoint, which is localhost-only, so risk is very low. But if the endpoint were ever exposed through the tunnel, this would be an XSS vector. Consider validating that the response content-type is `image/svg+xml` before injection, or using a safer DOM API. |
| `packages/server/src/server-cli.js` | 321-341 | The new `else if (!tunnelArg && !NO_AUTH)` branch (L321) and the subsequent `else if (!NO_AUTH)` branch (L340) create a somewhat confusing control flow. The L340 branch has only a comment and no code. | Consider adding a brief log or removing the empty branch to improve readability. The comment "tunnelArg is set but SKIP_TUNNEL is true due to externalUrl — already handled above" is helpful but an empty branch is unusual. |
| `packages/server/src/ws-server.js` | 323 | The HTTP request handler callback is now `async` (changed from sync). This is necessary for `QRCode.toString()`, but any unhandled rejection in the handler will crash the process in Node.js. | The QR endpoint has a try/catch, which is good. However, if any *other* code path in this handler throws asynchronously (e.g., a future change), it won't be caught. Consider wrapping the entire handler body in a try/catch or adding an error handler on the httpServer. |
| `packages/server/src/lan-ip.js` | 7-16 | `getLanIp()` returns the first non-internal IPv4 address it finds. On machines with multiple network interfaces (e.g., VPN + WiFi), this may return the VPN interface address rather than the LAN IP. | The behavior is acceptable for v1 since the QR is for same-network scanning. Document this limitation or consider filtering known VPN interface prefixes (utun, tun, tap) in a follow-up. |

#### Nitpicks (Optional)

- `packages/desktop/dist/index.html` L289: `qrSection.style.display = ''` is set in `__startPolling` but the section uses class-based visibility (`visible` class). Mixing `style.display` and CSS classes for the same element's visibility can lead to subtle bugs if the two mechanisms conflict. Consider using only the class-based approach consistently.
- `packages/server/src/dashboard.js`: The QR modal's inline styles (`style="width:200px;height:200px;..."`) could be extracted to CSS classes for consistency with the rest of the dashboard which uses a dedicated CSS function.

| Add Content-Type validation before SVG innerHTML injection | [#914](https://github.com/blamechris/chroxy/issues/914) | Low risk (localhost-only), but worth hardening before any future tunnel exposure |
| Filter VPN interfaces in `getLanIp()` | [#915](https://github.com/blamechris/chroxy/issues/915) | Current behavior is fine for same-network use; edge case for multi-NIC setups |

### Architecture Notes

This change fits cleanly within the existing architecture:

- **Server side**: The `/qr` endpoint follows the same pattern as the existing `/connect` endpoint — it reads from `connection.json` and serves derived data. The `lan-ip.js` utility is a clean extraction that enables `connection.json` to be written in no-tunnel mode, which was previously a gap.
- **Desktop side**: The tunnel mode setting follows the same pattern as `auto_start_server` — persisted in `DesktopSettings`, toggled via tray menu, applied at server start via env var. The loading page redesign replaces the previous auto-redirect-to-dashboard behavior with a QR-first flow, which is better UX for the desktop app's primary use case (pairing a phone).
- **No WS protocol changes**: This PR only adds HTTP endpoints and desktop UI. No new WebSocket message types are introduced, so there are no protocol compatibility concerns.

### Verdict

**Approve with minor suggestions.** This is a well-implemented feature that addresses a real usability need. The code is clean, the degradation paths are thoughtful, and CI passes. The suggestions above are quality-of-life improvements, not blockers. The duplicate `check_cloudflared()` call is the most impactful suggestion to address before merge.

PR #916, opened 2026-02-26
fix(server): harden QR pairing against VPN interfaces and SVG injection

### LINKED ISSUE #914: Validate Content-Type before SVG innerHTML injection in QR displays
## Context

From review of PR #913 (QR mobile pairing).

Both the desktop loading page (`packages/desktop/dist/index.html`) and the dashboard QR modal (`packages/server/src/dashboard.js`) inject the response from `GET /qr` directly into the DOM via `innerHTML`:

```js
qrContainer.innerHTML = svgHtml;
```

The `/qr` endpoint is currently localhost-only and returns trusted SVG from the `qrcode` npm package, so the risk is very low today.

## Suggestion

Before injecting, validate that the fetch response has `Content-Type: image/svg+xml`. This provides defense-in-depth against:
- Future changes that might expose `/qr` through the tunnel
- MITM scenarios if the endpoint ever moves to HTTP over the network
- Supply chain issues in the `qrcode` package

Alternatively, parse the SVG and re-serialize it, or use a data URL with an `` tag instead of innerHTML.

## Files

- `packages/desktop/dist/index.html` (line ~255, `showQr` function)
- `packages/server/src/dashboard.js` (line ~1910, `openQrModal` function)

### LINKED ISSUE #915: Filter VPN interfaces in getLanIp() for more reliable local pairing
## Context

From review of PR #913 (QR mobile pairing).

`packages/server/src/lan-ip.js` returns the first non-internal IPv4 address found via `os.networkInterfaces()`. On machines with VPN clients active (e.g., WireGuard, OpenVPN, Tailscale), the VPN interface may be enumerated before the WiFi/Ethernet interface, causing the QR code to contain a VPN IP that the phone cannot reach on the local network.

## Suggestion

Filter out common VPN interface name prefixes before selecting the LAN IP:
- `utun` (macOS WireGuard/system VPN)
- `tun` (OpenVPN, Linux)
- `tap` (OpenVPN bridge mode)
- `tailscale` / `ts` (Tailscale)
- `wg` (WireGuard)

Fallback: if no non-VPN interface is found, return the first non-internal address as today.

## Files

- `packages/server/src/lan-ip.js`

blamechris:
## Summary

- **VPN filtering**: `getLanIp()` now skips VPN interface prefixes (`utun`, `tun`, `tap`, `wg`, `tailscale`) so QR codes contain a WiFi-reachable address. Falls back to VPN address if no other interface is found.
- **SVG Content-Type validation**: Both QR display locations (desktop loading page and dashboard modal) now verify `Content-Type: image/svg+xml` before `innerHTML` injection.

Closes #914, closes #915

## Test plan

- [ ] With VPN active, start server in no-tunnel mode — verify QR contains WiFi IP, not VPN IP
- [ ] With only VPN interface available — verify fallback to VPN IP (not null)
- [ ] Tamper with `/qr` response Content-Type — verify QR display shows error, not injected content
- [ ] `cd packages/server && npm test` — no regressions

### DIFF
--- a/packages/server/src/lan-ip.js
+++ b/packages/server/src/lan-ip.js
@@ -9,12 +9,13 @@ function isVpnInterface(name) {
 }
 
 /**
- * Return the first non-internal, non-VPN IPv4 address, or null if none found.
- * Used for connection URLs when running without a tunnel.
+ * Return the first non-internal, non-VPN IPv4 address, falling back to a VPN
+ * IPv4 address if no non-VPN address exists, or null if no non-internal IPv4
+ * address exists at all. Used for connection URLs when running without a tunnel.
  *
  * VPN interfaces (WireGuard, OpenVPN, Tailscale) are skipped because phones
  * on the local WiFi cannot reach VPN-only addresses. Falls back to any
- * non-internal address if no non-VPN interface is found.
+ * non-internal IPv4 address (including VPN) if no non-VPN interface is found.
  */
 export function getLanIp() {
   const nets = networkInterfaces()


blamechris:
## Code Review: PR #916

### Summary
This PR hardens two aspects of QR-based mobile pairing:
1. **VPN interface filtering** in `getLanIp()` — skips common VPN interface prefixes (`utun`, `tun`, `tap`, `tailscale`, `ts`, `wg`) so the QR code contains a WiFi-reachable IP. Falls back to VPN address if no other interface is found.
2. **SVG Content-Type validation** — both the desktop loading page and the dashboard QR modal now verify `Content-Type: image/svg+xml` before `innerHTML` injection, providing defense-in-depth against injection via a tampered `/qr` response.

Both changes directly address the follow-up issues filed during the PR #913 review (#914, #915).

### Strengths
- **Clean fallback strategy**: VPN filtering gracefully degrades — if the only available interface is VPN, it still returns that address rather than `null`. This preserves the previous behavior as a worst-case fallback.
- **Minimal, focused diff**: 30 additions, 4 deletions across 3 files. Each change is self-contained and easy to verify.
- **Consistent validation in both locations**: The Content-Type check uses identical logic in both `index.html` and `dashboard.js`, reducing the chance of one location being missed.
- **Correct case handling**: `toLowerCase()` on both the interface name and the content-type header makes matching robust.
- **Server already sets the header correctly**: Verified that `/qr` in `ws-server.js` (line 412) sets `Content-Type: image/svg+xml`, so the client-side validation will pass for legitimate responses.
- **Good JSDoc comments**: The `lan-ip.js` changes include clear documentation explaining *why* VPN interfaces are skipped.

### Issues Found

#### Critical (Must Fix)
None.

| `packages/server/src/lan-ip.js` | 4 | Consider whether the `'ts'` prefix is too broad. While Tailscale on Linux sometimes uses `ts0`, no common non-VPN interfaces start with `ts` today. Document this assumption with an inline comment. | Future-proofing — a brief `// Tailscale on some Linux distros` comment next to `'ts'` would clarify intent for future maintainers. Very low risk of false positive in practice. |

#### Nitpicks (Optional)
- In `lan-ip.js` line 8, the arrow function in `.some((p) => ...)` has unnecessary parens around the single parameter: `VPN_PREFIXES.some(p => lower.startsWith(p))` is marginally cleaner. This is purely a style nit — the current code is correct.
- No unit tests for `isVpnInterface()` or the updated `getLanIp()`. The function is simple and the fallback logic is easy to reason about, but a test file would catch regressions if the VPN prefix list is later modified.

| Add unit tests for `getLanIp()` VPN filtering and fallback logic | #917 | The function is simple enough to ship without tests, but tests would prevent regressions as the prefix list evolves |

### Architecture Notes
- **`lan-ip.js`** is a pure utility with no side effects — the change is isolated and cannot affect other components.
- **SVG validation** is client-side defense-in-depth. The server already returns the correct Content-Type. The validation guards against future changes, MITM, or supply-chain issues — not a current vulnerability.
- **Desktop `index.html`** is a pre-built file in `packages/desktop/dist/`. The Content-Type check in the polling `pollQr` function correctly allows retries (the error falls through to the timeout-based retry logic), so a transient content-type issue won't immediately fail.
- **No WS protocol changes** — this PR only touches HTTP fetch paths and a utility function.

### Verdict
**Approve.** Clean, minimal, well-motivated hardening. Both issues (#914, #915) are addressed correctly. No critical issues. The one suggestion (inline comment on `'ts'` prefix) and the deferred test coverage are nice-to-haves that don't block merge.

blamechris:
## Review Comments Addressed

| 1 | JSDoc doesn't describe VPN fallback behavior | FIX | `266a8628` |
| 2 | Missing test coverage for getLanIp() VPN filtering | FOLLOW-UP | #917 |

- False positives: 0
- Follow-up issues: 1 (pre-existing #917, created during agent review)
- Existing issues closed: 0 (GitHub auto-closes #914, #915 on merge)

PR #963, opened 2026-02-27
fix(server): tighten CSP connect-src to localhost-only WebSocket origins (#955)

### LINKED ISSUE #955: Tighten CSP connect-src to localhost-only WebSocket origins
## Context

PR #954 added a Content-Security-Policy header to the dashboard endpoint. The `connect-src` directive is `'self' ws: wss:`, which permits WebSocket connections to any host.

## Problem

The dashboard only connects to `ws://localhost:PORT`, so the blanket `ws:` and `wss:` scheme allowances are broader than necessary. If an XSS payload were to bypass `script-src 'unsafe-inline'`, it could open WebSocket connections to an attacker-controlled server to exfiltrate the embedded API token.

## Proposal

Narrow `connect-src` to only allow localhost WebSocket connections:

```
connect-src 'self' ws://localhost:* wss://localhost:*
```

Alternatively, since the port is known at render time, it could be pinned to the exact port:

```
connect-src 'self' ws://localhost:${port} wss://localhost:${port}
```

## Priority

Low — the API token already provides the real access control, and the dashboard is served from localhost. This is defense-in-depth.

Refs #941, #954

blamechris:
## Summary

- Tighten CSP connect-src from blanket `ws: wss:` to `ws://localhost:PORT wss://localhost:PORT`
- Dashboard only connects to localhost — blanket WS scheme was unnecessarily permissive
- Negative test assertion verifies blanket ws: is not present

Closes #955

## Test Plan

- [x] CSP header test updated with positive and negative assertions
- [x] All server tests pass
- [x] Lint clean


blamechris:
## Review Comments Addressed

No inline review comments to process — Copilot approved without findings.

PR #968, opened 2026-02-27
fix(server): replace CSP unsafe-inline with nonce-based allowances (#956)

### LINKED ISSUE #956: Replace CSP unsafe-inline with nonce or hash-based script/style allowances
## Context

From [PR #954 review comment](https://github.com/blamechris/chroxy/pull/954#discussion_r2861935660): the current CSP uses `'unsafe-inline'` for both `script-src` and `style-src`, which limits XSS protection since injected inline scripts/styles will still execute.

## Problem

The dashboard (`dashboard.js`) embeds:
- Inline `<style>` block via `getDashboardCss()`
- Inline `<script>` block for `window.__CHROXY_CONFIG__`
- Inline `<script>` block via `getDashboardJs()`

These require `'unsafe-inline'`, but this weakens CSP against the very XSS token exfiltration risk the policy is designed to mitigate.

## Proposed Solution

Replace `'unsafe-inline'` with nonce-based or hash-based allowances:

**Option A (nonce):** Generate a random nonce per request, inject `nonce="..."` attribute on each `<script>`/`<style>` tag, and add `'nonce-...'` to the CSP header.

**Option B (hash):** Compute SHA-256 hashes of the inline script/style content at build time (or on server startup since the content is static except for config), and add `'sha256-...'` to the CSP header.

Option B may be simpler since the CSS and JS are effectively static (only `__CHROXY_CONFIG__` varies per request, which would require a nonce approach for that one block).

## Acceptance Criteria

- [ ] No `'unsafe-inline'` in CSP `script-src` or `style-src`
- [ ] Dashboard loads and functions correctly
- [ ] All existing dashboard tests pass

blamechris:
## Summary

- Replace CSP unsafe-inline with per-request nonce-based policy
- Generate nonce via crypto.randomBytes(16).toString('base64')
- Add nonce attributes to all inline style and script tags
- Update CSP header to use nonce-{value} instead of unsafe-inline

Closes #956

## Test Plan

- [x] CSP header test verifies nonce present and unsafe-inline absent
- [x] Dashboard tests verify nonce attributes on style/script tags
- [x] All 202 dashboard tests pass
- [x] Server lint clean

### DIFF
--- a/packages/server/src/dashboard.js
+++ b/packages/server/src/dashboard.js
@@ -5,6 +5,7 @@
  * @param {number} port - WsServer port
  * @param {string|null} apiToken - API token (embedded for WS auth)
  * @param {boolean} noEncrypt - Whether E2E encryption is disabled
+ * @param {string} [nonce] - Optional CSP nonce to apply to inline style and script tags
  * @returns {string} Complete HTML document
  */
 export function getDashboardHtml(port, apiToken, noEncrypt, nonce) {


blamechris:
## Review: fix(server): CSP nonce for inline scripts/styles (#968)

### Summary
Replaces `'unsafe-inline'` in the Content-Security-Policy header with per-request nonce-based allowances. A 16-byte random nonce is generated via `crypto.randomBytes` for each `/dashboard` request and injected into all three inline `<style>` and `<script>` tags. The `getDashboardHtml` function accepts an optional `nonce` parameter for backward compatibility.

### Strengths
- **Real security improvement** — `'unsafe-inline'` is one of the weakest CSP directives; replacing it with nonces means any injected inline script without the nonce will be blocked by the browser. This is a meaningful XSS mitigation.
- **Per-request nonce** — `randomBytes(16)` per request is cryptographically sound and follows CSP best practices. No nonce reuse across requests.
- **Backward compatible** — `getDashboardHtml` still works without the nonce parameter (the `n` variable becomes an empty string). Existing callers are unaffected.
- **Strong test coverage** — Tests verify: (1) nonce is present in CSP header, (2) `unsafe-inline` is absent, (3) nonce appears on all 3 inline tags in the HTML body, (4) CSP structure is correct.
- **Clean implementation** — The template literal approach (`<style${n}>`) is simple and readable.

### Issues Found

#### Suggestions

1. **Missing JSDoc for `nonce` parameter** — The `getDashboardHtml` JSDoc lists `port`, `apiToken`, and `noEncrypt` but not the new `nonce` parameter. (Copilot also flagged this.)

#### Nitpicks

2. **403 response also gets the nonce-based CSP** — The `securityHeaders` object (with the nonce) is shared between the 200 and 403 responses. For the 403 case, the nonce is generated but never used in the response body (the 403 body is a plain text error, not HTML with inline scripts). The nonce in the CSP header is harmless but unused. This is a non-issue in practice.

### Deferred Items
None.

### Verdict
**Approve.** This is a solid security improvement with clean implementation and good test coverage. The only action item is adding the JSDoc param, which is trivial.

blamechris:
### Comment Processing Summary

| 1 | Missing JSDoc @param for nonce parameter | **FIX** | `9eb21c78` — Added `@param {string} [nonce]` to JSDoc |

All 202 dashboard tests pass.

Issue #1211, opened 2026-03-01
dashboard-next: import BATCH_INTERVAL in TerminalView test instead of hardcoding 50ms


blamechris:
## Context

From review of PR #1209 (strengthen TerminalView test assertions).

## Problem

The batch write test hardcodes `vi.advanceTimersByTime(50)` to match `BATCH_INTERVAL = 50` in `TerminalView.tsx`. If the constant is ever changed in the component, the test will silently break — the timer won't fire at the expected time, and the assertion will fail with a confusing "expected 1 call, got 0" message.

## Suggested fix

Export `BATCH_INTERVAL` from `TerminalView.tsx` and import it in the test:

```ts
// TerminalView.tsx
export const BATCH_INTERVAL = 50

// TerminalView.test.tsx
import { TerminalView, BATCH_INTERVAL } from './TerminalView'
// ...
act(() => { vi.advanceTimersByTime(BATCH_INTERVAL) })
```

## File

- `packages/server/src/dashboard-next/src/components/TerminalView.tsx`
- `packages/server/src/dashboard-next/src/components/TerminalView.test.tsx`

## Refs

- PR #1209
- Issue #1166

Issue #1210, opened 2026-03-01
dashboard-next: strengthen 'writes initial data' test assertion in TerminalView


blamechris:
## Context

From review of PR #1209 (strengthen TerminalView test assertions).

## Problem

The `writes initial data when provided` test (line 103-113 in `TerminalView.test.tsx`) has the same class of weak assertion that PR #1209 fixed for three other tests. It asserts `onReady` was called but never verifies the initial data was actually written to the terminal.

## Current

```ts
it('writes initial data when provided', () => {
  const onReady = vi.fn()
  render(
    <TerminalView
      initialData="$ hello\n"
      onReady={onReady}
    />
  )
  // The terminal should have been written to
  expect(onReady).toHaveBeenCalledTimes(1)
})
```

## Suggested fix

Now that `writeSpy` exists at module level, add:

```ts
expect(writeSpy).toHaveBeenCalledWith('$ hello\n')
```

## File

`packages/server/src/dashboard-next/src/components/TerminalView.test.tsx`

## Refs

- PR #1209
- Issue #1166

Issue #1208, opened 2026-03-01
test(dashboard): add unmount-cancels-debounce test for TerminalView resize


blamechris:
## Context

From review of PR #1206 (debounce TerminalView resize handlers).

## Problem

The debounce test validates that rapid resize events are coalesced into a single `fit()` call, but does not verify that unmounting the component during a pending debounce timer cancels the timer and prevents the `safeFit` call.

The implementation does clear the timer on unmount — this is a test coverage gap, not a bug.

## Suggested Test

```ts
it('cancels pending debounce timer on unmount', () => {
  vi.useFakeTimers()
  fitSpy.mockClear()
  const { unmount } = render(<TerminalView />)
  const mountCalls = fitSpy.mock.calls.length
  act(() => { window.dispatchEvent(new Event('resize')) })
  unmount()
  act(() => { vi.advanceTimersByTime(200) })
  expect(fitSpy).toHaveBeenCalledTimes(mountCalls)
  vi.useRealTimers()
})
```

## File

`packages/server/src/dashboard-next/src/components/TerminalView.test.tsx`

## Refs

- PR #1206
- Issue #1165
- Issue #1166 (strengthen TerminalView test assertions)

blamechris:
Addressed in PR #1219.

PR #1219, opened 2026-03-01
test(dashboard): improve TerminalView test robustness (#1207, #1210, #1211, #1208)

### LINKED ISSUE #1207: test(dashboard): add vi.useRealTimers() to TerminalView afterEach for fake-timer safety
## Context

From review of PR #1206 (debounce TerminalView resize handlers).

## Problem

The `debounces resize events` test in `TerminalView.test.tsx` calls `vi.useFakeTimers()` at the start and `vi.useRealTimers()` at the end. If an assertion fails between these calls, subsequent tests inherit fake timers, potentially causing cascading failures.

## Suggested Fix

Move `vi.useRealTimers()` into the existing `afterEach` block so it runs even when assertions throw:

```ts
afterEach(() => {
  cleanup()
  vi.useRealTimers()
})
```

This is a no-op when real timers are already active, so it won't affect other tests.

## File

`packages/server/src/dashboard-next/src/components/TerminalView.test.tsx`

## Refs

- PR #1206
- Issue #1165

blamechris:
## Summary

- Add `vi.useRealTimers()` in `afterEach` to prevent fake timer leaks between tests (#1207)
- Import `BATCH_INTERVAL` from component instead of hardcoding `50` (#1211)
- Assert `writeSpy` called with actual initial data content, not just called (#1210)
- Add new test: cancels pending debounced resize on unmount (#1208)

Closes #1207, #1210, #1211, #1208

## Test Plan

- [x] All 12 TerminalView tests pass
- [x] Full 132 component test suite passes
- [x] TypeScript clean

### DIFF
--- a/packages/server/src/dashboard-next/src/components/TerminalView.test.tsx
+++ b/packages/server/src/dashboard-next/src/components/TerminalView.test.tsx
@@ -117,6 +117,9 @@ describe('TerminalView', () => {
       />
     )
 
+    // Guard: onReady must have provided write before we can test batching
+    expect(writeFn).toBeDefined()
+
     // Clear spy to isolate batched writes from mount-time activity
     writeSpy.mockClear()
 


blamechris (commented):
## Review: APPROVE

Clean, focused test improvement PR. All four issues are addressed correctly.

### Checklist

- [x] **#1207** — `vi.useRealTimers()` added to `afterEach`. Correct placement; no-op when real timers are active, guarantees cleanup on assertion failure. The `try/finally` blocks in the batching and debounce tests are properly removed since `afterEach` now handles timer restoration.
- [x] **#1211** — `BATCH_INTERVAL` exported from `TerminalView.tsx` and imported in the test. The `export` keyword is the only production code change — no runtime behavior difference. `vi.advanceTimersByTime(BATCH_INTERVAL)` replaces the hardcoded `50`.
- [x] **#1210** — The `writes initial data` test now asserts `writeSpy` was called with the actual data string instead of just checking `onReady` was called. The test is also simplified (removes unused `onReady` callback).
- [x] **#1208** — New `cancels pending debounced resize on unmount` test validates that the cleanup function in the `useEffect` properly clears the resize debounce timer. Matches the suggested implementation from the issue.

### Bonus fix

The duplicate `fitSpy` declaration (lines 15 and 40 on main) is consolidated to a single declaration at line 15 alongside the other module-level spies. Good cleanup.

### Style / Conventions

- Zero attribution — clean
- Commit message follows `type(scope): description` convention
- Issue refs in test names (`#1210`, `#1208`) match project convention for traceability
- TypeScript types preserved correctly
- No unnecessary changes outside the two files

### CI

Dashboard Tests and Dashboard Type Check both pass. Server Tests and App Tests are pending but unrelated to these changes.

No issues found. Ship it.

blamechris:
## Review Comments Addressed

| 1 | Add guard assertion before `writeFn!` non-null assertions in batching test | FIX | `41b260e0` |

- False positives: 0
- Follow-up issues created: 0
- Existing issues closed: 4 (#1207, #1208, #1210, #1211 — all resolved by this PR)

PR #1225, opened 2026-03-01
feat(dashboard): add keyboard shortcuts to PermissionPrompt (#1190)

### LINKED ISSUE #1190: feat(dashboard): add keyboard shortcuts to PermissionPrompt (Cmd+Y allow, Escape deny)
## Context

Identified during Copilot review of PR #1181.

## Description

The legacy dashboard-app.js used keyboard shortcuts for permission prompts (Cmd/Ctrl+Y to allow, Escape to deny). The new PermissionPrompt component should register a `keydown` listener with proper cleanup, ensuring shortcuts don't fire when focus is in an input/textarea.

## Original Comment

> The port described in #1157 includes keyboard shortcuts (Cmd/Ctrl+Y to allow, Escape to deny), but this component currently doesn't register any key handlers. If the dashboard still relies on these shortcuts, add a keydown listener (with proper cleanup) and ensure it doesn't fire while focus is in an input/textarea.

## Acceptance Criteria

- [ ] Cmd/Ctrl+Y triggers Allow
- [ ] Escape triggers Deny
- [ ] Shortcuts don't fire when focus is in input/textarea
- [ ] Proper cleanup on unmount
- [ ] Tests for keyboard shortcuts

blamechris:
## Summary

- Add `Cmd/Ctrl+Y` shortcut to allow permission
- Add `Escape` shortcut to deny permission
- Shortcuts suppressed when focus is in input/textarea
- Proper cleanup on unmount

Closes #1190

## Test Plan

- [x] New tests: Cmd+Y allows, Ctrl+Y allows, Escape denies
- [x] New test: shortcuts ignored when textarea focused
- [x] New test: no shortcut after already answered
- [x] New test: listener cleaned up on unmount
- [x] All 20 PermissionPrompt/PlanApproval tests pass
- [x] TypeScript clean

### DIFF
--- a/packages/server/src/dashboard-next/src/components/TerminalView.test.tsx
+++ b/packages/server/src/dashboard-next/src/components/TerminalView.test.tsx
@@ -37,7 +37,6 @@ vi.mock('@xterm/xterm', () => {
   return { Terminal: MockTerminal }
 })
 
-const fitSpy = vi.fn()
 vi.mock('@xterm/addon-fit', () => {
   class MockFitAddon {
     _fitted = false


blamechris (commented):
## Agent Review — PR #1225

**Scope:** Adds keyboard shortcuts (Cmd/Ctrl+Y to allow, Escape to deny) to `PermissionPrompt`, with 6 new tests covering the happy path, input suppression, double-fire guard, and cleanup.

### Verdict: Approve with suggestions

The implementation is clean, correct, and well-tested. The `respond` callback already guards against double-fire (checks `answered` and `remaining`), so the keyboard handler inherits that safety for free. Good use of `useEffect` cleanup.

### Issues

**1. `<select>` elements not suppressed (bug, minor)**
`PermissionPrompt.tsx:63-64` — The input suppression checks for `INPUT` and `TEXTAREA` tags, but the dashboard header has two `<select>` elements (model selector, permission mode selector). If the user has a `<select>` focused and presses Cmd+Y or Escape, the shortcut fires and the `<select>` interaction is disrupted. Add `SELECT` to the tag guard:

```tsx
if (tag === 'INPUT' || tag === 'TEXTAREA' || tag === 'SELECT') return
```

Also add a corresponding test for `<select>` suppression.

**2. Escape conflicts with Modal (note, non-blocking)**
`Modal.tsx` also binds Escape on `document` to close modals. If a modal is open while a PermissionPrompt is visible, both handlers fire — the modal closes AND the permission is denied. This is unlikely in practice (modals are for session creation, not permission flow), but worth noting. A follow-up could add `e.stopImmediatePropagation()` or check for open modals, but this is not blocking.

**3. Multiple PermissionPrompts (note, non-blocking)**
If multiple permission prompts are rendered simultaneously, every mounted instance registers its own `keydown` listener. Pressing Cmd+Y would allow ALL pending permissions at once. This may actually be desirable UX (batch-approve), but it's undocumented. Consider whether this should be intentional or if only the most recent prompt should respond.

**4. CI failures are pre-existing (note)**
Dashboard Tests and Dashboard Type Check fail due to a duplicate `fitSpy` declaration in `TerminalView.test.tsx` — this is unrelated to this PR (main is green). The PermissionPrompt tests (20/20) pass.

### Test Coverage

The 6 new tests are solid:
- Cmd+Y allow, Ctrl+Y allow, Escape deny — core shortcut paths
- Textarea focus suppression — prevents interference with input
- Already-answered guard — no double-fire
- Unmount cleanup — no stale listeners

One gap: no test for `<select>` focus suppression (see issue #1 above).

blamechris:
## Check-PR: Copilot Review Response

Copilot raised 3 low-confidence suggestions. Addressed in ba934b10:

| # | Copilot Suggestion | Outcome | Detail |
|---|---|---|---|
| 1 | Listener churn: `respond` deps cause add/remove every countdown tick | **FOLLOW-UP** | Valid concern. Keeping `[respond]` for now to satisfy `react-hooks/exhaustive-deps` lint rule without disable comments. A ref-based approach is cleaner but requires lint suppression -- deferred to #1230. |
| 2 | `preventDefault` fires after prompt is answered/expired | **FIX** | Removed `e.preventDefault()` from the Escape branch. `respond()` already no-ops when answered, so Escape no longer swallows browser shortcuts on inert prompts. Cmd+Y still calls `preventDefault` (needed to suppress browser redo). |
| 3 | Case-sensitive `e.key === 'y'` breaks with Caps Lock | **FIX** | Changed to `e.key.toLowerCase() === 'y'`. Added test for `key: 'Y'` with `metaKey: true`. |

**Additional fix from agent review:**
- Added `SELECT` to the tag guard and a corresponding test (model/permission selectors in the dashboard header).

All 22 PermissionPrompt tests pass. Pre-existing `TerminalView.test.tsx` failure (duplicate `fitSpy`) is unrelated.

PR #1264, opened 2026-03-01
fix(dashboard): prevent code block placeholders from wrapping in <p> tags (#1244)

### LINKED ISSUE #1244: Code block placeholders wrapped in <p> tags in markdown renderer
## Context

PR #1240 fixes paragraph handling in the dashboard-next markdown renderer to produce valid `<p>` tag wrapping. The `blockRe` regex correctly skips block-level HTML elements (`h1`-`h6`, `pre`, `ul`, `ol`, `blockquote`) from being wrapped in `<p>` tags.

However, fenced code blocks are extracted early in the pipeline and replaced with opaque placeholders (`\x00CB{n}\x00`). By the time the paragraph logic runs, these placeholders don't match `blockRe`, so they get wrapped: `<p>\x00CB0\x00</p>`. After restoration, this produces `<p><pre><code>...</code></pre></p>` — invalid HTML nesting.

## Pre-existing

This is **not a regression** from PR #1240. The previous `</p><p>` approach had the same issue. Browsers handle it gracefully, but it's technically invalid.

## Proposed Fix

Add the placeholder pattern to `blockRe`:

```ts
const blockRe = /^(<(h[1-6]|pre|ul|ol|blockquote)|\x00CB\d+\x00)/
```

Add a test case:

```ts
it('does not wrap code blocks in <p> tags', () => {
  const html = renderMarkdown('text\n\n```js\nconst x = 1\n```\n\nmore text')
  expect(html).not.toMatch(/<p>\s*<pre>/)
})
```

## References

- PR #1240 (agent review comment)
- `packages/server/src/dashboard-next/src/lib/markdown.ts`

blamechris:
## Summary

- Add code block placeholder pattern (`\x00CB\d+\x00`) to `blockRe` regex in markdown renderer
- Prevents invalid `<p><pre><code>` HTML nesting when fenced code blocks are surrounded by paragraphs
- Add test case verifying code blocks are not wrapped in `<p>` tags

Closes #1244

## Test Plan

- [x] All new tests pass
- [x] Existing 281 dashboard tests pass
- [x] No regressions in markdown rendering

### DIFF
--- a/packages/server/src/dashboard-next/src/lib/markdown.test.ts
+++ b/packages/server/src/dashboard-next/src/lib/markdown.test.ts
@@ -131,6 +131,13 @@ describe('renderMarkdown', () => {
     expect(html).toContain('<p>more text</p>')
   })
 
+  it('wraps paragraphs starting with inline code in <p> tags (#1244)', () => {
+    const html = renderMarkdown('`x` is a variable\n\nother text')
+    // Inline code at start of paragraph must still get <p> wrapping
+    expect(html).toMatch(/<p>.*<code>x<\/code> is a variable.*<\/p>/)
+    expect(html).toContain('<p>other text</p>')
+  })
+
   it('sanitizes XSS payloads via DOMPurify defense-in-depth', () => {
     // Verify no raw <script> or event handler attributes survive in output.
     // Input is escaped by escapeHtml first; DOMPurify catches anything that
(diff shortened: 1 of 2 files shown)


blamechris:
## Review Comments Addressed

| 1 | Copilot: placeholder regex matches inline code placeholders too — paragraphs starting with inline code skip `<p>` wrapping | FIX | `616f4315` |

- False positives: 0
- Follow-up issues created: 0
- Existing issues closed: 0 (issue #1244 was closed during agent-review phase)

Issue #1118, opened 2026-03-01
CSP hardening (remove unsafe-inline)


blamechris:
## Problem

Current CSP allows \`unsafe-inline\` for scripts, defeating XSS protections. Once the React app is fully bundled (no inline scripts), this can be removed.

## Implementation

Use Tauri v2's object-format CSP in tauri.conf.json:
- Remove \`unsafe-inline\` from \`script-src\`
- Add \`frame-src: 'none'\`, \`object-src: 'none'\`, \`base-uri: 'self'\`
- Separate \`dev_csp\` with \`unsafe-eval\` for Vite HMR
- Add proper \`font-src\`, \`img-src\` directives for xterm.js and QR codes

See Tauri Expert \`06-tauri-expert.md\` Section 2.1 for full recommended CSP.

## References
- Tauri Expert: \`06-tauri-expert.md\` Section 2.1
- Adversary: \`08-adversary.md\`

Issue #1101, opened 2026-03-01
Single-window architecture (remove two-window pattern)


blamechris:
## Problem

The current Tauri app uses two windows: \`main\` (fallback/loading) and \`dashboard\` (created dynamically). This creates unnecessary complexity and fights the IDE layout. All 10 auditors converge on single-window.

### One window, React handles all states:
- **Loading state**: shown while server starts (replaces dist/index.html fallback)
- **IDE state**: sidebar + tabs + terminal
- **Error state**: shown when server fails to start

### Changes:
1. Remove \`dist/index.html\` static fallback
2. Remove \`show_fallback\` / \`open_dashboard\` / \`toggle_window\` state machine
3. Single window in \`tauri.conf.json\`:
   \`\`\`json
   {
     "label": "main",
     "title": "Chroxy",
     "width": 1200, "height": 800,
     "minWidth": 800, "minHeight": 500,
     "visible": true, "center": true
   }
   \`\`\`
4. React app receives server state via Tauri events (not eval() injection)
5. Close-to-hide behavior for tray app

## References
- Tauri Expert: \`06-tauri-expert.md\` Section 1
- Master Assessment: Architecture Decision #1

PR #1280, opened 2026-03-02
feat(desktop): bundle server into .app for standalone install

blamechris:
## Summary

- Bundle server JS source + production `node_modules` into the `.app` via Tauri's `bundle.resources`
- Add `resolve_cli_js()` strategy that finds the bundled server at `Contents/Resources/server/src/cli.js`
- Fix `beforeDevCommand`/`beforeBuildCommand` paths — CWD is `packages/desktop/`, not `src-tauri/`
- Add `npm ci` to CI release workflow so dashboard build has its dependencies

The `.app` is now self-contained (~109MB). Installing to `/Applications` and clicking Start Server works without needing the monorepo, `CHROXY_SERVER_PATH`, or global npm install. Node 22 and cloudflared remain external prerequisites.

Closes #1280

## Test Plan

- [x] `bundle-server.sh` stages correct files (cli.js, tunnel/, hooks/, dashboard-next/dist/, node_modules/)
- [x] No dev deps in bundle (no vite, no eslint)
- [x] `cargo tauri build --bundles app` produces `.app` with server in `Contents/Resources/server/`
- [x] Path resolution works from standalone location (`/tmp/Chroxy-test.app`)
- [x] `cargo check` passes


blamechris:
## Code Review: PR #1280

### Summary
Bundles the Node.js server into the Tauri .app for standalone desktop installation. Includes 6 commits fixing smoke test blockers: QR race condition, dashboard blank screen (auth bypass + React dedupe + render loop fix), WebSocket health check URL, user message display, version label visibility, and in-app dashboard window.

### Strengths
- Clean bundled server resolution strategy with 4-level fallback (bundled → monorepo → env → which)
- `__startPolling` stub pattern elegantly solves the Tauri eval() race condition
- EMPTY_* constants for Zustand `useShallow` stability is the correct pattern
- Health check URL fix (`new URL('/', httpBase).href`) is clean and robust
- Good CORS additions for Tauri WebView origin isolation

### Issues Found

#### Critical (Must Fix)
*None*

| `lib.rs` | ~118 | `app.exit(0)` on ANY window close kills the entire app — consider guarding on window label | Closing the dashboard window currently stops the server and quits. If the fallback window is open when dashboard closes, the user loses both windows. May be intentional but worth confirming. |
| `server.rs` | resolve_cli_js | Log which resolution strategy succeeded | Would significantly help debugging "server not found" issues in the field — one `eprintln!("[server] Using bundled cli.js at {}", path)` per branch |
| `ws-server.js` | CORS | `Access-Control-Allow-Origin: *` on health/QR endpoints | Acceptable for localhost-only but should be tightened if/when the server is exposed via tunnel. Existing issue #1118 (CSP hardening) could cover this. |

#### Nitpicks (Optional)
- `tauri.conf.json` version is still `0.2.0` — should be bumped for the v0.2.1 tag
- `std::mem::forget(tray)` is a known Tauri workaround for keeping the tray alive — comment explains it well

### Architecture Notes
- **Two-window pattern** (fallback + dashboard) works but adds complexity. v0.4 #1101 plans single-window consolidation — this PR is a good stepping stone.
- **CSP weakened to `unsafe-inline`** from nonce-based — necessary for WKWebView + Vite builds. v0.4 #1118 tracks re-hardening.
- **Bundle script** (`bundle-server.sh`) copies flat .js files from `src/` plus `tunnel/` subdir and `dashboard-next/dist`. If new server subdirectories are added later, they'll need to be added to the script.

### Verdict

Solid work solving multiple layered issues (auth, React duplication, render loops, health check URL). The suggestions above are all non-blocking improvements that can be addressed in follow-up work.

PR #1340, opened 2026-03-02
fix(desktop): add lockfile to bundle-server.sh for reproducible builds (#1284)

### LINKED ISSUE #1284: fix(desktop): add lockfile to bundle-server.sh for reproducible builds
## Context

Identified during review of PR #1280.

## Description

\`bundle-server.sh\` runs \`npm install --omit=dev\` in a fresh staging directory with no lockfile, so dependency versions float over time. CI builds may become non-reproducible.

## Original Comment

> The bundle install step uses \`npm install --omit=dev\` in a fresh staging directory that has no lockfile, so dependency versions will float over time and CI builds may become non-reproducible. Consider generating/committing a lockfile for the server package (or otherwise copying the root lockfile).

## Acceptance Criteria

- [ ] Copy root \`package-lock.json\` or generate a server-specific lockfile
- [ ] \`npm ci\` used instead of \`npm install\` in the bundle script
- [ ] Verify reproducible builds across CI runs

blamechris:
## Summary

- Add standalone `package-lock.json` for `packages/server/` (generated outside workspace context)
- Copy lockfile into staging directory during `bundle-server.sh`
- Switch from `npm install --omit=dev` to `npm ci --omit=dev` for deterministic installs
- Verified: 54 packages installed consistently via `npm ci`

Closes #1284

## Test Plan

- [ ] `bundle-server.sh` runs successfully with `npm ci`
- [ ] Staging directory contains `package-lock.json`
- [ ] `cargo test` passes (5 desktop tests)
- [ ] CI passes

### DIFF
--- a/packages/desktop/src-tauri/src/server.rs
+++ b/packages/desktop/src-tauri/src/server.rs
@@ -2,6 +2,7 @@ use std::collections::VecDeque;
 use std::io::{BufRead, BufReader};
 use std::path::PathBuf;
 use std::process::{Child, Command, Stdio};
+use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::{Arc, Mutex};
 use std::thread;
 use std::time::{Duration, Instant};
@@ -38,7 +39,7 @@ pub struct ServerManager {
     node_path: Option<PathBuf>,
     config: ChroxyConfig,
     tunnel_mode: String,
-    health_running: Arc<Mutex<bool>>,
+    health_generation: Arc<AtomicU64>,
 }
 
 impl ServerManager {
@@ -50,7 +51,7 @@ impl ServerManager {
             node_path: None,
             config: ChroxyConfig::default(),
             tunnel_mode: "quick".to_string(),
-            health_running: Arc::new(Mutex::new(false)),
+            health_generation: Arc::new(AtomicU64::new(0)),
         }
     }
 
@@ -226,8 +227,8 @@ impl ServerManager {
 
     /// Stop the server process gracefully (SIGTERM → 5s → SIGKILL).
     pub fn stop(&mut self) {
-        // Stop health polling
-        *lock_or_recover(&self.health_running) = false;
+        // Stop health polling by advancing generation (old threads will see mismatch and exit)
+        self.health_generation.fetch_add(1, Ordering::SeqCst);
 
         if let Some(ref mut child) = self.child {
             // Only send SIGTERM if the child is still running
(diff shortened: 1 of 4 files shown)


blamechris:
## Code Review — PR #1340

**Verdict: Approve with suggestions**

Solid fix for a real reproducibility gap. The approach — standalone lockfile + `npm ci` — is the correct solution for deterministic bundling in a workspace monorepo.

---

### What's good

- **Correct problem identification**: `npm install` without a lockfile in a staging directory meant dependency versions could float across builds. `npm ci` is the right tool here.
- **Clean standalone lockfile**: No `workspace:` or `file:` protocol references — the lockfile is fully self-contained and will work correctly in the isolated staging directory. Verified: 0 workspace links, 0 file links.
- **Dependency match**: The lockfile root entry exactly matches `package.json` dependencies and devDependencies — no drift.
- **`--omit=dev` works correctly with `npm ci`**: The lockfile contains 281 dev-only packages and 68 production packages. `npm ci --omit=dev` will skip the dev tree and install only production deps, which is the desired behavior.
- **CI is green**: All 6 checks pass.

---

### Findings

#### 1. Staleness risk — no CI guard (suggestion, P2)

The server lockfile lives alongside `package.json` but there's no CI step to verify they stay in sync. If someone adds a dependency to `package.json` and forgets to regenerate the server lockfile, `npm ci` in the bundle script will fail — which is actually the *safe* failure mode (`npm ci` refuses to proceed if the lockfile doesn't match `package.json`). So this is self-healing at build time.

However, it would be nicer to catch this earlier in CI rather than during a desktop build. Consider adding a CI job that runs `cd packages/server && npm install --package-lock-only && git diff --exit-code package-lock.json` to detect drift on every PR. This is a follow-up, not a blocker.

#### 2. No documented regeneration procedure (suggestion, P2)

Developers need to know how to regenerate this lockfile when they update server dependencies. The generation command isn't documented anywhere. A one-line comment in the lockfile header or a note in `bundle-server.sh` like:

```bash
# Regenerate with: cd packages/server && npm install --package-lock-only
```

...would save future confusion. The key detail: this must be run *outside* the workspace context (i.e., from within `packages/server/` directly, not from the repo root) to avoid workspace resolution.

#### 3. `@types/dompurify` in production dependencies (pre-existing, P3)

`@types/dompurify` appears in both `dependencies` and `devDependencies` in `package.json`. The production copy means TypeScript type definitions are included in the bundle's production install. This is harmless but wasteful — it should only be in `devDependencies`. Not introduced by this PR but worth noting since the lockfile now immortalizes it.

#### 4. `set -euo pipefail` + `npm ci` error handling (nit, no action needed)

The script already has `set -euo pipefail`, so if `npm ci` fails (e.g., lockfile out of sync), the script exits immediately with a nonzero code. This is correct behavior — Tauri's `beforeBuildCommand` will surface the failure. No change needed.

---

### Summary

| Category | Count |
|----------|-------|
| Blockers | 0 |
| Suggestions | 2 |
| Pre-existing | 1 |
| Nits | 1 |

The lockfile is correct, the script change is minimal and clean, and CI passes. The two suggestions (CI staleness guard and regeneration docs) are follow-up material, not blockers.

PR #1397, opened 2026-03-03
feat(dashboard): auto-expand sidebar repos matching active filter (#1376)

### LINKED ISSUE #1376: feat(dashboard): auto-expand sidebar repos matching active filter
## Context

Identified during review of PR #1372 (sidebar component with repo tree).

## Description

When a user collapses a repo in the sidebar tree and then types a filter that matches a session inside that collapsed repo, the repo header appears in the filtered list but its matching sessions remain hidden behind the collapsed state. This creates a confusing UX where the filter "finds" something but the user can't see what matched.

## Suggested Fix

When `filter` is non-empty, ignore the `collapsed` state and always show children. This way, filtering acts as both a search and auto-expand:

```tsx
const shouldShowChildren = filter ? true : !isCollapsed
```

Alternatively, auto-expand any repo whose children match the filter:

```tsx
const effectiveCollapsed = filter
  ? false  // always expand during filter
  : (collapsed[repo.path] ?? false)
```

## Acceptance Criteria

- [ ] When filter is active, all matching repos show their children regardless of collapsed state
- [ ] When filter is cleared, collapsed state is restored to its previous value
- [ ] Add test: collapse repo, set filter matching child session, verify child is visible

Refs #1372

blamechris:
## Summary

- When a filter is active, ignore the collapsed state and always show children
- Matching sessions are visible even if the repo was manually collapsed
- Collapsed state is restored when the filter is cleared
- Add 2 tests verifying auto-expand behavior with filter

Closes #1376

## Test Plan

- [x] All 513 dashboard tests pass (23 Sidebar tests including 2 new)
- [x] TypeScript type check clean
- [ ] Manual: collapse a repo, type filter matching a child session, verify child is visible

### DIFF
--- a/packages/server/src/dashboard-next/src/components/Sidebar.test.tsx
+++ b/packages/server/src/dashboard-next/src/components/Sidebar.test.tsx
@@ -246,23 +246,7 @@ describe('Sidebar', () => {
     expect(groups.length).toBeGreaterThanOrEqual(1)
   })
 
-  it('auto-expands collapsed repo when filter matches child session (#1376)', () => {
-    renderSidebar()
-    // Collapse the api repo
-    fireEvent.click(screen.getByTestId('repo-header-/home/user/projects/api'))
-    expect(screen.queryByText('Backend')).not.toBeInTheDocument()
-
-    // Re-render with filter matching a child session
-    cleanup()
-    // We need a component that has filter set — re-render with filter
-    renderSidebar({ filter: 'Backend' })
-    // Note: the collapsed state is internal React state, reset on re-render
-    // The real scenario is: user collapses, THEN types a filter
-    // We need to test the component with collapsed + filter together
-    expect(screen.getByText('Backend')).toBeInTheDocument()
-  })
-
-  it('shows children of collapsed repo when filter is active (#1376)', () => {
+  it('auto-expands collapsed repo when filter is active and restores on clear (#1376)', () => {
     const { rerender } = render(
       <Sidebar
         repos={makeRepos()}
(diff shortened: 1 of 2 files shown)


blamechris (commented):
## Agent Review — PR #1397

**Verdict: APPROVE** — Clean, minimal change that directly addresses the issue.

### Summary

One-line production change: when `filter` is non-empty, force `isCollapsed` to `false` so all matching repo children are visible regardless of user-collapsed state. Collapsed state is preserved in React state and automatically restored when the filter is cleared. Two new tests verify the behavior.

### Findings

**0 Critical, 0 Warnings, 2 Suggestions**

---

**S1: First test does not actually test the feature (low)**

The first new test (`auto-expands collapsed repo when filter matches child session`) calls `cleanup()` + `renderSidebar({ filter: 'Backend' })`, which creates a brand-new component instance. The collapsed state from the first render is lost — this test would pass even without the code change. The comment in the test acknowledges this:

> Note: the collapsed state is internal React state, reset on re-render

The second test (`shows children of collapsed repo when filter is active`) correctly uses `rerender()` to preserve component state and is the one that actually validates the fix. The first test is redundant but not harmful — it still asserts filtered rendering works.

*Suggestion:* Consider removing the first test or adding a comment that it tests "fresh render with filter" rather than the auto-expand behavior. Not blocking.

---

**S2: Chevron visual indicator during filter (cosmetic)**

When the filter is active, `isCollapsed` is forced to `false`, so the chevron always shows the down-arrow. Clicking a repo header during filtering still calls `toggleRepo()` and updates the collapsed state, but the UI won't reflect it until the filter is cleared. This is slightly surprising — the click appears to do nothing.

*Suggestion:* Could disable the toggle click or show a visual hint when filter overrides collapse. Very minor UX polish, not blocking.

---

### Compliance

- Code style: Matches existing patterns (ternary, same line structure)
- Tests: 2 new tests, second one properly validates the feature with `rerender()`
- Issue linkage: `Closes #1376` in commit message
- No TypeScript errors introduced (single-line change, types unchanged)

PR #1409, opened 2026-03-03
ci: add staleness check for server package-lock.json (#1341)

### LINKED ISSUE #1341: ci: add staleness check for server package-lock.json
## Context

Identified during review of PR #1340.

## Description

PR #1340 added a standalone `packages/server/package-lock.json` used by `bundle-server.sh` for deterministic `npm ci` installs. If someone updates `packages/server/package.json` dependencies without regenerating this lockfile, `npm ci` will fail at desktop build time — which is the safe failure mode, but it would be better to catch the drift earlier in CI.

## Original Review Comment

> The server lockfile lives alongside `package.json` but there's no CI step to verify they stay in sync. Consider adding a CI job that runs `cd packages/server && npm install --package-lock-only && git diff --exit-code package-lock.json` to detect drift on every PR.

## Acceptance Criteria

- [ ] CI workflow includes a step that verifies `packages/server/package-lock.json` matches `packages/server/package.json`
- [ ] Approach: run `npm install --package-lock-only` in the server directory (outside workspace context) and check for git diff
- [ ] Clear error message if lockfile is stale

blamechris:
## Summary

- Add CI step in server-lint job that regenerates `packages/server/package-lock.json` and checks for drift
- Clear error message if lockfile is stale with instructions to fix

Closes #1341

## Test Plan

- [x] CI workflow syntax valid
- [x] Step runs after lint (no new dependencies needed)
- [x] Tests: N/A (CI config change)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,7 +55,6 @@ jobs:
 
       - name: Check server lockfile freshness
         run: |
-          cd packages/server
           npm install --package-lock-only
           git diff --exit-code package-lock.json || {
             echo "::error::packages/server/package-lock.json is stale. Run 'cd packages/server && npm install --package-lock-only' and commit the result."


blamechris (commented):
## Agent Review — PR #1409

### Verdict: Request Changes (1 critical, 0 suggestions)

---

### CRITICAL: `cd packages/server` will fail — job default already sets working directory

The `server-lint` job declares a default working directory:

```yaml
defaults:
  run:
    working-directory: packages/server
```

The new step's shell script starts with `cd packages/server`, which means it will attempt to cd into `packages/server/packages/server` — a path that does not exist. **This step will fail on every CI run.**

The existing staleness check in `dashboard-tests` (line 75-81) follows the correct pattern — it omits any `cd` and relies on the job default:

```yaml
- name: Check generated tokens are up-to-date
  run: |
    node scripts/generate-theme-tokens.mjs
    git diff --exit-code src/dashboard-next/src/theme/tokens.ts || {
      echo "::error::tokens.ts is stale..."
      exit 1
    }
```

**Fix:** Remove the `cd packages/server` line. The step inherits the job default and already runs from `packages/server`.

```yaml
- name: Check server lockfile freshness
  run: |
    npm install --package-lock-only
    git diff --exit-code package-lock.json || {
      echo "::error::packages/server/package-lock.json is stale. Run 'cd packages/server && npm install --package-lock-only' and commit the result."
      exit 1
    }
```

---

### Summary

| Category | Count |
|----------|-------|
| Critical | 1 |
| Suggestions | 0 |
| Nits | 0 |

The intent is sound and the approach matches the established token staleness pattern. The single issue is the redundant `cd` conflicting with the job's default working directory.

Issue #1351, opened 2026-03-02
fix(desktop): reset restart_count in monitoring thread instead of health poll


blamechris:
## Context

Identified during review of PR #1350.

## Description

The health poll thread resets `restart_count` to 0 immediately when the server reaches Running after a restart. The monitoring thread in `lib.rs` also reads `restart_count` for notification messages. Since both threads use `Ordering::Relaxed` with no synchronization barrier, the monitoring thread could read a stale count value.

## Original Comment

> In `start_health_poll()`, when the server reaches Running after a restart, the health poll thread immediately resets `restart_count` to 0. Meanwhile, the Phase 2 monitoring loop in `lib.rs` also checks count for logging/notification purposes. Since both threads use `Ordering::Relaxed`, the count could be read as 0 in the monitoring thread before it has a chance to send the "recovered" notification with the correct attempt number.

## Proposed Fix

Move the `restart_count.store(0, ...)` call from the health poll thread to the monitoring thread in `lib.rs`, after confirming recovery (Running status detected post-restart). This ensures the count is only reset after the monitoring thread has finished using it for notifications.

Alternatively, use `Ordering::SeqCst` for the count field to establish a happens-before relationship.

## Acceptance Criteria

- [ ] `restart_count` is reset to 0 only after the monitoring thread has logged/notified the recovery
- [ ] No change to restart behavior

PR #1412, opened 2026-03-03
fix(desktop): reset restart_count in monitoring thread instead of health poll (#1351)

blamechris:
## Summary

- Move `restart_count` reset from health poll thread to monitoring thread in `lib.rs`
- Add `reset_restart_count()` public method to `ServerManager`
- Remove unused `restart_count` clone from `start_health_poll()`
- Prevents race where monitoring thread reads stale (already-reset) count due to `Relaxed` ordering

Refs #1351

## Test Plan

- [ ] New unit test: `reset_restart_count_clears_to_zero`
- [ ] All 25 Rust tests pass
- [ ] No change to restart behavior — only timing of count reset changes

### DIFF
--- a/packages/desktop/src-tauri/src/lib.rs
+++ b/packages/desktop/src-tauri/src/lib.rs
@@ -388,6 +388,9 @@ fn handle_start(app: &tauri::AppHandle) {
                                                     "Server Recovered",
                                                     "Auto-restart successful",
                                                 );
+                                                // Reset restart count after recovery notification
+                                                let state = app_handle.state::<Mutex<ServerManager>>();
+                                                lock_or_recover(&state).reset_restart_count();
                                                 recovered = true;
                                                 break;
                                             }
--- a/packages/desktop/src-tauri/src/server.rs
+++ b/packages/desktop/src-tauri/src/server.rs
@@ -103,6 +103,11 @@ impl ServerManager {
         self.restart_count.load(Ordering::Relaxed)
     }
 
+    /// Reset consecutive restart count (called by monitoring thread after recovery).
+    pub fn reset_restart_count(&self) {
+        self.restart_count.store(0, Ordering::Relaxed);
+    }
+
     /// Re-signal that auto-restart should be attempted.
     /// Called when a restart attempt started the process but it failed to
     /// reach Running. Sets the pending flag so the monitoring loop retries.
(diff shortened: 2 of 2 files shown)


blamechris (commented):
## Agent Review — PR #1412

**Verdict: APPROVE**

### Summary

This PR fixes a real data race on `restart_count` between the health poll thread and the monitoring thread. The health poll thread (in `server.rs`) was resetting `restart_count` to 0 as soon as the server reached `Running`, while the monitoring thread (in `lib.rs`) was still reading it for notification messages. With `Ordering::Relaxed` on the atomic, there was no guarantee the monitoring thread would see the pre-reset value.

The fix moves the reset from the health poll thread to the monitoring thread, placing it immediately after the "Server Recovered" notification — so the count is always available for the notification message before being cleared.

### Findings

**Correctness: Sound**

1. **Initial start path preserved.** `start()` (line 162 of `server.rs`) already resets `restart_count` to 0, so removing the redundant reset from `start_health_poll()` has no effect on normal startup.

2. **Manual restart path preserved.** `restart()` calls `start()`, which resets the count. No regression.

3. **Auto-restart path fixed.** The monitoring thread in `lib.rs` now owns the reset lifecycle: `try_auto_restart()` increments the count, the recovery detection loop reads it for notifications, and only after confirming recovery does `reset_restart_count()` clear it. Clean sequencing.

4. **No dead code left.** The `restart_count` clone that was passed into the health poll thread is removed since it's no longer used there.

5. **New public method `reset_restart_count()` is minimal and correct.** Uses the same `Ordering::Relaxed` as all other `restart_count` operations — appropriate since the ordering concern was cross-thread visibility timing, not atomicity, and moving the reset to the same thread as the consumer eliminates the race.

**Suggestions: 0 critical, 1 minor**

- *Minor:* The doc comment on `restart_count` field (line 49 of `server.rs`) says "reset when server reaches Running" — could be updated to "reset by monitoring thread after recovery" for accuracy. Not blocking.

### Test Coverage

New unit test `reset_restart_count_clears_to_zero` validates the method. The existing test suite covers restart count limits, backoff calculations, and auto-restart rejection at max attempts. The race condition itself is inherently hard to unit-test (requires thread interleaving), but the structural fix (single-writer pattern) is the right approach.

### Issues

This PR directly resolves from-review issue #1351.

Issue #1362, opened 2026-03-02
docs(dashboard): add stability note for useGlobalShortcuts shortcuts parameter


blamechris:
## Context

Identified during review of PR #1358.

## Description

`useGlobalShortcuts` takes a `shortcuts: ShortcutMap` parameter and uses it as a `useEffect` dependency. If the caller passes an inline object literal, the effect will re-run every render because the object reference changes (new listener registered/torn down each cycle).

When the hook is integrated into `App.tsx` or other components, this will be a performance pitfall if the consumer doesn't stabilize the reference with `useMemo`.

## Suggested Fix

Add a JSDoc note on the `useGlobalShortcuts` function or `ShortcutMap` type:

```ts
/**
 * Register global keyboard shortcuts.
 *
 * @param shortcuts - Map of shortcut strings to handlers.
 *   IMPORTANT: Must be a stable reference (wrap in useMemo) to avoid
 *   re-registering listeners on every render.
 */
```

Alternatively, the hook could internally use `useRef` to store the latest shortcuts and only attach the listener once (on mount), similar to how event handler refs work in many React patterns.

## Acceptance Criteria

- [ ] JSDoc comment or README note warns callers about reference stability
- [ ] OR: hook uses `useRef` internally to avoid effect churn

PR #1414, opened 2026-03-03
fix(dashboard): use ref for useGlobalShortcuts to prevent effect churn (#1362)

blamechris:
## Summary

- Use `useRef` to store latest shortcuts, register listener once on mount
- Callers no longer need to wrap shortcuts in `useMemo` to avoid effect churn
- Add JSDoc comment documenting the stability guarantee
- Add 2 new tests: re-register prevention, latest handler usage

Refs #1362

## Test Plan

- [ ] All 10 useGlobalShortcuts tests pass
- [ ] TypeScript clean
- [ ] New test verifies listener is NOT re-registered on reference change
- [ ] New test verifies latest handler is used after update

### DIFF
--- a/packages/server/src/dashboard-next/src/hooks/useGlobalShortcuts.test.ts
+++ b/packages/server/src/dashboard-next/src/hooks/useGlobalShortcuts.test.ts
@@ -199,4 +199,55 @@ describe('useGlobalShortcuts', () => {
 
     expect(handler).toHaveBeenCalledTimes(5)
   })
+
+  it('does not re-register listener when shortcuts reference changes', () => {
+    const addSpy = vi.spyOn(document, 'addEventListener')
+    const removeSpy = vi.spyOn(document, 'removeEventListener')
+
+    const handler = vi.fn()
+    const { rerender } = renderHook(
+      ({ shortcuts }) => useGlobalShortcuts(shortcuts),
+      { initialProps: { shortcuts: { 'cmd+n': handler } as ShortcutMap } },
+    )
+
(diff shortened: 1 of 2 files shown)


blamechris (commented):
## Agent Review — PR #1414

**Verdict: APPROVE**

### Summary

Clean, well-tested conversion of `useGlobalShortcuts` from a dependency-driven effect to the "latest ref" pattern. This eliminates effect churn when callers pass inline shortcut objects, which is the right fix for #1362.

### Findings

**0 critical, 0 blocking, 2 observations:**

1. **Minor perf trade-off (acceptable):** `Object.entries(shortcutsRef.current).map(parseShortcut(...))` now runs on every keydown event instead of once per effect cycle. Given that shortcut maps are tiny (< 10 entries) and keydown events are infrequent, this is negligible. No action needed.

2. **Branch name mismatch:** Branch is `docs/1362-...` but the PR is a `fix` with code and test changes. Cosmetic only — does not affect merge.

### What was reviewed

- `useGlobalShortcuts.ts`: ref pattern, empty deps, parsing moved inside listener, JSDoc update
- `useGlobalShortcuts.test.ts`: 2 new tests covering re-registration prevention and latest-handler semantics
- Caller search: hook is not yet consumed outside its own test file (no downstream breakage risk)
- Issue #1362 acceptance criteria: both paths satisfied (useRef + JSDoc note)

### Conclusion

Straightforward, well-scoped fix. Tests directly prove the behavioral guarantees. Ready to merge once CI passes.

PR #1742, opened 2026-03-08
refactor: DRY user_input handler via @chroxy/store-core

### LINKED ISSUE #1708: refactor: DRY user_input handler across app and dashboard message-handler.ts
## Context

Identified during review of PR #1707.

## Description

Both `packages/app/src/store/message-handler.ts` and `packages/server/src/dashboard-next/src/store/message-handler.ts` have nearly identical `case 'user_input'` handlers (added in PR #1707). As more message types are added, this duplication will grow.

The `store-core` shared package (introduced in #1597) is the right home for shared message handler logic.

## Original Review Comment

> The app's `user_input` handler lacks the explanatory comments present in the dashboard version. Both files have nearly identical handler code.

## Acceptance Criteria

- [ ] Extract shared `user_input` handling logic into `packages/store-core`
- [ ] Both app and dashboard import from the shared package
- [ ] Tests pass in both packages

blamechris:
## Summary

- Extracts `parseUserInputMessage()` into `packages/store-core/src/user-input-handler.ts`
- Both the app and dashboard `message-handler.ts` now import from `@chroxy/store-core` instead of duplicating the logic
- Adds `@chroxy/store-core` as a dependency in `packages/app/package.json`
- 7 unit tests for the shared helper covering: self-sender skip, no-session skip, normal routing, fallback to activeSessionId, timestamp fallback, missing clientId, empty text

Fixes #1708

## Test plan

- [ ] 7 new unit tests in store-core pass
- [ ] All existing dashboard tests still pass (949)
- [ ] TypeScript typecheck passes in both packages

### DIFF
--- a/packages/server/src/dashboard-next/src/App.test.tsx
+++ b/packages/server/src/dashboard-next/src/App.test.tsx
@@ -71,6 +71,7 @@ vi.mock('./store/connection', () => {
     conversationHistory: [],
     fetchConversationHistory: vi.fn(),
     resumeConversation: vi.fn(),
+    connectedClients: [],
     serverRegistry: [],
     activeServerId: null,
     addServer: vi.fn(),
--- a/packages/server/src/dashboard-next/src/App.tsx
+++ b/packages/server/src/dashboard-next/src/App.tsx
@@ -110,6 +110,7 @@ export function App() {
   const filePickerFiles = useConnectionStore(s => s.filePickerFiles)
   const sessionNotifications = useConnectionStore(s => s.sessionNotifications)
   const inputSettings = useConnectionStore(s => s.inputSettings)
+  const connectedClients = useConnectionStore(s => s.connectedClients)
 
   // Listen for Tauri desktop events (no-op in browser context)
   useTauriEvents()
@@ -724,7 +725,7 @@ export function App() {
           filter={sidebarFilter}
           serverStatus={isConnected ? 'connected' : isReconnecting ? 'reconnecting' : 'disconnected'}
           tunnelUrl={null}
-          clientCount={1}
+          clientCount={connectedClients.length}
           onFilterChange={setSidebarFilter}
           onSessionClick={switchSession}
           onResumeSession={(convId) => {
--- a/packages/server/src/dashboard-next/src/components/Sidebar.test.tsx
+++ b/packages/server/src/dashboard-next/src/components/Sidebar.test.tsx
@@ -291,4 +291,24 @@ describe('Sidebar', () => {
     rerender(<Sidebar {...props} filter="" />)
     expect(screen.queryByText('Backend')).not.toBeInTheDocument()
   })
+
+  it('hides client count when server is disconnected', () => {
+    renderSidebar({ serverStatus: 'disconnected', clientCount: 2 })
+    expect(screen.queryByTestId('sidebar-footer')).not.toHaveTextContent('client')
+  })
+
+  it('shows client count when server is connected', () => {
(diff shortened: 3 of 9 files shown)


blamechris:
## Code Review: PR #1742

### Summary

Extracts the duplicated `case 'user_input'` handler from both `packages/app/src/store/message-handler.ts` and `packages/server/src/dashboard-next/src/store/message-handler.ts` into a shared `parseUserInputMessage()` helper in `@chroxy/store-core`. The function is pure, unit-tested, and both call sites are now identical one-liners. Closes #1708.

Tests verified: 7/7 new store-core tests pass, all 949 dashboard tests continue to pass.

---

### Strengths

- **Clean extraction** — the helper is pure (no side-effects, no store access) and the interface types (`RawUserInputMessage`, `ParsedUserInput`) are well-named and narrow
- **Accurate null semantics** — returning `null` for "skip" vs. a value for "process" maps cleanly onto the `break` pattern at both call sites
- **Good test coverage** — 7 tests cover the two skip paths, the happy path, sessionId fallback, timestamp fallback, missing clientId, and empty text
- **Correct `store-core` placement** — this is exactly the kind of platform-agnostic WS parsing logic the shared package is designed for
- **Index exports done right** — both the type and the function are exported cleanly from `index.ts`
- **Dashboard comment retained** — the dashboard call site's inline comment explaining the skip reason is preserved

---

### Issues Found

#### Critical (Must Fix)

None.

| `user-input-handler.ts` | 44 | `msg.timestamp \|\| Date.now()` treats `0` as falsy and falls back to `Date.now()` | A timestamp of `0` is technically valid (Unix epoch). Prefer `msg.timestamp != null ? msg.timestamp : Date.now()` to be explicit. In practice this will never fire — servers don't send epoch timestamps — but the original code had the same bug and it's cheap to fix. |
| `user-input-handler.ts` | 31–35 | Function signature accepts `msg: RawUserInputMessage` but both call sites cast `msg as Record<string, unknown>` | The call sites already have `msg` typed as a loose `Record<string, unknown>` (it comes from the WS parse loop). If `parseUserInputMessage` accepted `Record<string, unknown>` and extracted fields internally, the `as` cast at each call site could be dropped. Alternatively, the function signature is fine as-is and the cast is acceptable since it's only at the top of the case block. |

#### Nitpicks

- `user-input-handler.test.ts` line 33: Non-null assertion `result!.sessionId` — prefer `expect(result).not.toBeNull(); if (!result) return;` to avoid assertion leaking down. Minor; the test pattern is consistent with the rest of the suite so this is style only.
- The app call site (`message-handler.ts:931`) has no explanatory comment while the dashboard call site (`:860-861`) retained its two-line comment. Consider adding the same comment to the app version for consistency.

---

None. Both suggestions above are minor enough to address in-place if desired or defer without tracking.

---

### Architecture Notes

The extraction fits cleanly within the `store-core` DI model: the helper is stateless, takes only primitive parameters (`myClientId`, `activeSessionId`), and returns a value-object — no store access, no side-effects. This is the right abstraction level. The `ParsedUserInput` type omitting `id` (added by the caller via `nextMessageId`) is the correct separation of concerns.

The `*` version constraint added to `packages/app/package.json` is consistent with how `store-core` is declared in the server package (`"@chroxy/store-core": "*"`) and is correct for a private monorepo workspace package.

---

### Verdict

- [x] **Approve** — Ready to merge. The refactor is correct, tests pass, and no regressions were introduced. The `timestamp || Date.now()` falsy-zero case is worth a one-line fix but is not a blocker since epoch timestamps are never sent over this protocol.

PR #1755, opened 2026-03-08
fix(server): extract _buildChildEnv() and add env strip tests

### LINKED ISSUE #1724: fix(server): strip ANTHROPIC_API_KEY from cli-session env consistently
## Problem
ANTHROPIC_API_KEY may leak into child process env in cli-session.js. Should be stripped from the env passed to spawned claude processes.

## Acceptance Criteria
- [ ] ANTHROPIC_API_KEY is removed from child process env
- [ ] Other sensitive keys (if any) are also stripped
- [ ] Test coverage

blamechris:
## Summary

- Extracts env-building logic from `_spawnPersistentProcess` into a dedicated `_buildChildEnv()` method for testability
- `ANTHROPIC_API_KEY` was already stripped — this PR adds 5 tests proving it works correctly
- Tests cover: API key stripped, `CI`/`CLAUDE_HEADLESS`/`CLAUDE_CODE_ENABLE_SDK_FILE_CHECKPOINTING` always set, `CHROXY_PORT` conditional, `CHROXY_PERMISSION_MODE` passed through

## Test plan

- [x] `ANTHROPIC_API_KEY` absent from child env when set in parent
- [x] Required env vars always present
- [x] `CHROXY_PORT` only set when port configured
- [x] 45/45 cli-session tests pass

Closes #1724

### DIFF
--- a/docs/architecture/reference.md
+++ b/docs/architecture/reference.md
@@ -131,6 +131,11 @@ Store files:
 | `destroy_session` | Delete session by ID |
 | `encrypted` | Encrypted message envelope (E2E encryption) |
 | `get_diff` | Request git diff for uncommitted changes |
+| `git_branches` | Request git branch list for session project |
+| `git_commit` | Commit staged changes with a message |
+| `git_stage` | Stage files for commit |
+| `git_status` | Request git status for session project |
+| `git_unstage` | Unstage files from commit |
 | `input` | Send text or voice message to session |
 | `interrupt` | Interrupt active Claude task |
 | `key_exchange` | Send client X25519 public key for encryption |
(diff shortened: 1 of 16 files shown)


blamechris:
## Code Review: PR #1755

### Summary

Extracts the inline env-building block inside `_spawnPersistentProcess` into a dedicated `_buildChildEnv()` method, then adds 5 focused unit tests that verify the strip/set behaviour directly. Pure refactor — no behaviour change. The issue it closes (#1724) was filed because the existing strip logic had no test coverage; this PR closes that gap without touching the actual stripping logic.

### Strengths

- Clean destructuring pattern (`const { ANTHROPIC_API_KEY: _, ...parentEnv }`) preserved exactly as before — no risk of behaviour regression.
- Tests are well-scoped: each test asserts exactly one concern and uses `createSession()` directly, keeping the suite fast and side-effect-free.
- `ANTHROPIC_API_KEY` save/restore in the first test correctly handles the case where the key is absent in the test environment (`delete` vs re-assign).
- `_buildChildEnv()` reads `this.permissionMode` at call time, so it's always fresh even after `setPermissionMode()` → `_killAndRespawn()` cycles.
- 45/45 tests pass with no changes to existing tests.

### Issues Found

#### Critical (Must Fix)

None.

| `cli-session.test.js` | end of new suite | Add a test for `CHROXY_TOKEN` set when `apiToken` is provided, and absent when not | `CHROXY_PORT` gets both a positive and negative test; `CHROXY_TOKEN` has neither. The conditional is symmetric — worth symmetric coverage. |
| `cli-session.test.js` | new suite | Add a test asserting that other `process.env` keys are forwarded through | Currently the tests only verify keys that are explicitly set or stripped. A passthrough test (e.g. set `process.env.SOME_VAR`, call `_buildChildEnv()`, assert present) would guard against accidentally over-stripping if the destructure pattern is ever modified. |

#### Nitpicks

- `session._hookManager?.destroy()` in the `CHROXY_PORT` test is a workaround for the hook manager that `createSession({ port })` instantiates. A comment explaining why it's needed would save the next reader from wondering if it's a copy-paste artifact.

| Add `CHROXY_TOKEN` positive/negative tests and a passthrough test to `_buildChildEnv` suite | [#1756](https://github.com/blamechris/chroxy/issues/1756) | Small gap in coverage, not a correctness risk. Appropriate to track separately without blocking this fix. |

### Architecture Notes

`_buildChildEnv()` is a pure method on the session instance — it reads `process.env` and instance fields, returns a plain object, no side effects. The extraction is the right shape: easy to stub in future tests that need to inject a specific env without patching `process.env`. It is intentionally not `static` because it needs `this._port`, `this._apiToken`, and `this.permissionMode`.

### Verdict

**Approve — Ready to merge.**

The refactor is minimal and correct. The tests close the coverage gap from #1724. The two suggestions above are tracked as a follow-up issue and do not block merging.

PR #1770, opened 2026-03-08
test(server): add waitForTunnel unit tests

### LINKED ISSUE #1722: test(server): add tunnel-check.js unit tests
## Goal
Add unit tests for tunnel-check.js tunnel verification.

## Acceptance Criteria
- [ ] Tests for successful tunnel verification
- [ ] Tests for retry logic on DNS propagation delay
- [ ] Tests for timeout/failure scenarios
- [ ] TDD: RED → GREEN → REFACTOR

blamechris:
Closes #1722

## Changes

Adds `packages/server/tests/tunnel-check.test.js` with 6 tests covering `waitForTunnel`:

- Immediate ok response (resolves in 1 fetch call)
- Retry on network error, succeeds when ok
- All attempts fail → resolves without throwing (fail-safe behavior — tunnel might still work)
- All non-ok responses → resolves without throwing
- Correct URL passed to every fetch call
- Stops immediately after first ok response (no over-fetching)

Uses `interval: 0` to skip inter-attempt delays while testing retry logic. All 6 pass.

### DIFF
--- a/packages/server/src/tunnel-check.js
+++ b/packages/server/src/tunnel-check.js
@@ -7,18 +7,18 @@ export async function waitForTunnel(httpUrl, { maxAttempts = 10, interval = 2000
   const startTime = Date.now()
 
   for (let i = 0; i < maxAttempts; i++) {
+    const controller = new AbortController()
+    const timeout = setTimeout(() => controller.abort(), 5000)
     try {
-      const controller = new AbortController()
-      const timeout = setTimeout(() => controller.abort(), 5000)
       const res = await fetch(httpUrl, { signal: controller.signal })
-      clearTimeout(timeout)
-
       if (res.ok) {
         console.log(`[tunnel] Tunnel verified (took ${((Date.now() - startTime) / 1000).toFixed(1)}s)`)
         return
       }
     } catch {
       // Not ready yet
+    } finally {
+      clearTimeout(timeout)
     }
 
     if (i < maxAttempts - 1) {
--- a/packages/server/tests/tunnel-check.test.js
+++ b/packages/server/tests/tunnel-check.test.js
@@ -5,7 +5,8 @@ import { waitForTunnel } from '../src/tunnel-check.js'
 /**
  * waitForTunnel unit tests (#1722)
  *
- * fetch is mocked via globalThis.fetch.
+ * fetch is mocked via mock.method(globalThis, 'fetch', impl) so that
+ * mock.restoreAll() correctly reinstates the original implementation.
  * setTimeout is NOT mocked — tests use interval:0 to avoid real delays.
  */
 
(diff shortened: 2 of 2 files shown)


blamechris:
## Code Review: PR #1770

### Summary

Adds `packages/server/tests/tunnel-check.test.js` with 6 unit tests covering `waitForTunnel`. The tests use Node's built-in `node:test` runner and `mock` module — consistent with the rest of the server test suite. All 6 pass cleanly.

### Strengths

- **Correct mock isolation**: `afterEach(() => mock.restoreAll())` ensures `globalThis.fetch` is reset between tests — no cross-test pollution.
- **`interval: 0` pattern**: Smart use of the parameter to skip real delays without resorting to fake timers, keeping tests fast and deterministic.
- **Fail-safe coverage**: Tests 3 and 4 verify the non-throwing exhaustion behavior, which is the most important correctness property of `waitForTunnel` for production reliability.
- **Over-fetch guard**: Test 6 (stops after first ok) is a subtle but valuable correctness assertion.
- **Helper factories**: `okFetch()`, `notOkFetch()`, `failThenOk()` keep individual tests readable without repetition.
- **Covers all issue acceptance criteria**: DNS delay simulation (test 2), failure scenarios (tests 3/4), success path (test 1).

### Issues Found

#### Critical (Must Fix)

None.

| `tunnel-check.test.js` | 33–60 | Consider asserting call-count *before* the `doesNotReject` resolves using a post-await check (already done correctly) — no action needed | Already correct |
| `tunnel-check.test.js` | 38 | `doesNotReject` wraps a thunk — the inner `() =>` is correct, but consider `await waitForTunnel(...)` directly for the simpler cases (tests 1/5/6) since the function never throws | Minor style — `doesNotReject` wrapper adds clarity for fail-safe intent, acceptable as-is |
| `tunnel-check.test.js` | (all) | `globalThis.fetch` is assigned directly without saving/restoring the original value before the test suite runs. If the host environment already has a `fetch` (Node 18+), `mock.restoreAll()` won't reinstall it because the original was never captured via `mock.method()`. | See note below |

**`globalThis.fetch` assignment note**: The tests assign `globalThis.fetch = mock.fn(...)` rather than using `mock.method(globalThis, 'fetch', impl)`. This means `mock.restoreAll()` has no saved descriptor to restore — it silently no-ops for this assignment. In practice this works because the test file is isolated, but if `tunnel-check.js` ever adds module-level caching of `fetch`, or if Node's test runner runs files in-process (rather than per-file isolation), stale mocks could bleed across tests.

The safer pattern used elsewhere in this repo is:

```js
import { mock } from 'node:test'

// In each test:
const stub = mock.method(globalThis, 'fetch', async () => ({ ok: true }))

// mock.restoreAll() will then correctly reinstate the original
```

This is a suggestion, not a blocker — the current approach works fine given Node's default per-file test isolation.

#### Nitpicks

- The block comment at the top (`waitForTunnel unit tests (#1722)`) is helpful context. No change needed.
- `failThenOk` captures `calls` in a closure but doesn't reset between tests — this is fine because each test creates a new `failThenOk()` call.

| Switch `globalThis.fetch` assignment to `mock.method(globalThis, 'fetch', impl)` for safer restore semantics | https://github.com/blamechris/chroxy/issues/1771 | Not a current bug; tests pass and are isolated. Better practice worth tracking but not blocking. |

### Architecture Notes

`tunnel-check.js` is a small, pure utility with no side effects beyond `console.log` and `setTimeout`. Testing it via `globalThis.fetch` injection is appropriate and keeps the implementation clean (no dependency injection required in production code). The `interval: 0` override parameter is a solid testability hook already present in the source.

### Verdict

- [x] **Approve** — Ready to merge. All 6 tests pass, coverage is complete, no correctness issues. One minor follow-up tracked in #1771.

PR #1809, opened 2026-03-10
test(server): tighten discoverLoggerFiles() import detection regex

### LINKED ISSUE #1807: test(server): tighten discoverLoggerFiles() import detection regex
In `logger-usage.test.js`, `discoverLoggerFiles()` uses `src.includes('createLogger')` which can match mentions in strings or comments. A regex matching the actual import statement (e.g. `import { ... createLogger ... } from './logger.js'`) would be more precise and avoid false positives as more files migrate.

Raised from Copilot review of PR #1796.

blamechris:
## Summary
- Replace `src.includes('createLogger')` in `discoverLoggerFiles()` with a regex that matches only actual `import { createLogger } from './logger.js'` statements
- Prevents false positives from string literals or comments that happen to mention `createLogger`

## Test plan
- [x] `node --test tests/logger-usage.test.js` — all 7 tests pass

Closes #1807

### DIFF
--- a/packages/server/tests/logger-usage.test.js
+++ b/packages/server/tests/logger-usage.test.js
@@ -34,7 +34,7 @@ function discoverLoggerFiles() {
   const files = readdirSync(SRC).filter(f => f.endsWith('.js') && f !== 'logger.js')
   return files.filter(f => {
     const src = readSrc(f)  // let errors propagate — a read failure is a test failure
-    return src.includes("from './logger.js'") && src.includes('createLogger')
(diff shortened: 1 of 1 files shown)


blamechris (commented):
**Review: test/1807-logger-import-regex**

Reviewed the one-line change in `discoverLoggerFiles()`. The regex correctly matches actual `import { createLogger } from './logger.js'` statements and avoids false positives from string literals, comments, or partial matches. Word boundaries (`\b`) prevent matching `createLoggerFoo` etc.

Verified against current src/ files — all three logger consumers (ws-server.js, sdk-session.js, supervisor.js) match correctly. The `[^}]*` spans also handle multi-named imports or multiline import statements.

Clean, well-scoped change. No issues found.

PR #1811, opened 2026-03-10
fix(scripts): scope Cargo.toml version check to [package] section

### LINKED ISSUE #1805: fix(scripts): scope bump-version.sh Cargo.toml version verification to [package] section
In `bump-version.sh`, the post-write verification uses `grep -q "^version = \"\""` which is not scoped to the `[package]` section. This could produce a false pass if a dependency table entry happens to have an identical version string.

Consider scoping the grep to only the `[package]` section, or verifying via `cargo metadata` / reading the value back.

Raised from Copilot review of PR #1791. Low priority since Cargo.toml package version is always the first `version` line and `grep -m1` already captures it.

blamechris:
## Summary

- Scopes both the version read and post-write verification in `bump-version.sh` to the `[package]` section of `Cargo.toml` using `awk`, preventing false positives from dependency lines that share the same version string
- The previous `grep -m1 '^version = '` and `grep -q '^version = "..."'` matched any top-level `version =` line in the file

## Test plan

- [ ] Run `./scripts/bump-version.sh 0.5.1` and verify Cargo.toml `[package]` version updates correctly
- [ ] Verify the script still fails properly if the replacement silently no-ops

Closes #1805

### DIFF
--- a/scripts/bump-version.sh
+++ b/scripts/bump-version.sh
@@ -74,11 +74,11 @@ if [ -z "$CARGO_CURRENT" ]; then
   echo "Error: Failed to parse current version from $CARGO_TOML [package] section" >&2
   exit 1
 fi
-sed -i.bak "s/^version = \"$CARGO_CURRENT\"/version = \"$NEW_VERSION\"/" "$CARGO_TOML"
+sed -i.bak "/^\[package\]/,/^\[/{s/^version = \"$CARGO_CURRENT\"/version = \"$NEW_VERSION\"/}" "$CARGO_TOML"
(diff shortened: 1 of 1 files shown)


blamechris:
Both Copilot findings addressed in adbd0a9b:

1. **`grep -c` exit code** — Added `|| true` guard so `set -e` doesn't abort before reaching the custom error path.
2. **`sed` replacement scope** — Scoped the `sed` to `[package]` section using `/^\[package\]/,/^\[/` address range, matching the already-scoped read and verify.

PR #1816, opened 2026-03-10
feat(server): cross-client message history sync (#1705)

### LINKED ISSUE #1705: feat(dashboard): cross-client message history sync
## Description

When a new client connects (e.g., dashboard opens after the mobile app has been chatting), it doesn't see messages that were exchanged before it connected. Each client only sees messages from its own connection lifetime.

## Example

1. Mobile app connects, sends message, receives Claude response ("Credit balance is too low")
2. Dashboard connects later
3. Dashboard does NOT see the message or response — only sees connect/disconnect system events

## Expected Behavior

When a new client connects, the server should replay recent message history for each session so the client has full context.

## Implementation Notes

The server already has session state in `session-manager.js`. The replay mechanism would:
1. Buffer recent messages per session (last N messages or last M minutes)
2. On new client auth + session subscribe, replay buffered messages
3. Mark replayed messages so the client can distinguish them from live messages

## Acceptance Criteria

- [ ] New client receives message history for subscribed sessions on connect
- [ ] Replayed messages display correctly in both dashboard and mobile app
- [ ] History is bounded (configurable max messages or time window)
- [ ] Replayed messages visually distinguished from live messages (optional)

blamechris:
## Summary

- Change `_replayHistory()` to send the full ring buffer (up to 500 messages) instead of only the last response turn
- New clients connecting to an active session now see the complete conversation history, not just the most recent Claude response

The ring buffer cap (500 messages in session-manager) provides natural bounding. Chunked delivery (20 messages per tick) prevents event loop blocking.

Closes #1705

## Test Plan

- [x] Updated 2 existing `_replayHistory` tests to verify full buffer replay
- [x] All 1568 server tests pass
- [x] All 260 ws-server tests pass

### DIFF
--- a/packages/server/src/ws-server.js
+++ b/packages/server/src/ws-server.js
@@ -733,25 +733,13 @@ export class WsServer {
   }
 
   /** Replay message history for a session to a single client.
-   *  Only replays the last conversation turn (last Claude response + result +
-   *  any in-progress work) to avoid flooding the app with old tool calls.
+   *  Sends the full ring buffer (capped at 500 messages by session-manager)
+   *  so new clients get complete conversation context.
    */
   _replayHistory(ws, sessionId) {
     if (!this.sessionManager) return
-    const fullHistory = this.sessionManager.getHistory(sessionId)
-    if (fullHistory.length === 0) return
-
-    // Find the last response message and replay from there to the end.
-    // This gives: last Claude response + result + any in-progress messages.
-    let startIdx = 0
-    for (let i = fullHistory.length - 1; i >= 0; i--) {
-      if (fullHistory[i].type === 'message' && fullHistory[i].messageType === 'response') {
-        startIdx = i
-        break
-      }
-    }
-
-    const history = fullHistory.slice(startIdx)
+    const history = this.sessionManager.getHistory(sessionId)
+    if (history.length === 0) return
 
     const truncated = this.sessionManager.isHistoryTruncated(sessionId)
     this._send(ws, { type: 'history_replay_start', sessionId, truncated })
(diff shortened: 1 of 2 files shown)


blamechris:
## Agent Review — PR #1816

### Summary
Clean, focused change that removes the last-turn-only replay filter from `_replayHistory()`, sending the full ring buffer (capped at 500 by session-manager) instead. Tests updated to match.

### Code Quality
- **Style**: No semicolons, single quotes, ES modules — follows server conventions
- **Simplification**: Removes ~12 lines of scanning logic (`for` loop to find last response) in favor of direct pass-through. Net -7 lines. Good reduction in complexity.
- **Tests**: Both existing tests updated with correct new assertions. Test names updated to reflect new behavior.

### Correctness
- The ring buffer cap (`_maxHistory = 500`) and chunked delivery (`CHUNK_SIZE = 20` with `setImmediate`) are unchanged — the bounding and backpressure mechanisms remain intact. No risk of flooding.
- `isHistoryTruncated` flag is correctly forwarded so clients know when history was capped.

### Observation
- Replayed history entries are sent via `this._send(ws, history[i])` without attaching `sessionId` to each entry. This is a pre-existing gap (not introduced by this PR) — `_broadcastToSession` does add `sessionId` via spread, but `_replayHistory` does not. Clients currently infer session from `history_replay_start.sessionId`, which works but is fragile. Worth a follow-up.

### Verdict
**Approved** — straightforward, well-tested, correct.

blamechris:
## Check-PR Summary

| # | Comment | Author | Verdict | Action |
|---|---------|--------|---------|--------|
| 1 | Add `sessionId` assertions to replay tests | Copilot | **FOLLOW-UP ISSUE** | Pre-existing gap — replay entries never carried `sessionId`. Created [#1818](https://github.com/blamechris/chroxy/issues/1818) |

**0 fixes pushed, 0 false positives, 1 follow-up issue created.**

Issue #1967, opened 2026-03-11
fix(pairing): session tokens issued by PairingManager never expire


blamechris:
## Context
`packages/server/src/pairing.js` — pairing IDs have a TTL (default 60s) and are single-use, but the **session tokens** issued when a pairing succeeds never expire:

```js
// validatePairing():
const sessionToken = randomBytes(SESSION_TOKEN_BYTES).toString('base64url')
if (this._sessionTokens.size >= MAX_SESSION_TOKENS) {
  const oldest = this._sessionTokens.keys().next().value
  this._sessionTokens.delete(oldest)  // FIFO eviction at cap only
}
this._sessionTokens.set(sessionToken, { createdAt: Date.now() })
// No expiry set on the token itself!
```

Session tokens are stored with only `createdAt` and no `expiresAt`. `isSessionTokenValid()` checks all stored tokens without validating their age. The only eviction is FIFO when the 100-token cap is reached.

This means a session token issued when a phone scans a QR code remains valid forever (or until 100 new pairings push it out). If a token is leaked (logs, network capture), it can be used to authenticate indefinitely.

Compare: the main API token supports rotation via `TokenManager`, but pairing-issued session tokens have no rotation or expiry.

## Current Behavior
Session tokens issued during pairing are permanent until FIFO-evicted by 100 newer tokens.

## Expected Behavior
Session tokens should expire after a configurable period (e.g., 24 hours or matching the `tokenExpiry` config). `isSessionTokenValid()` should reject tokens past their expiry.

## Acceptance Criteria
- [ ] `expiresAt` field added to session token entries (default: 24h)
- [ ] `isSessionTokenValid()` rejects expired tokens
- [ ] Expired tokens pruned periodically (or on each `validatePairing` call)
- [ ] Tests verify token expiry

blamechris:
Addressed in PR #2035 — added TTL expiry to session tokens.

PR #2035, opened 2026-03-11
fix(pairing): add TTL expiry to session tokens (#1967)

blamechris:
## Summary

- Added configurable `sessionTokenTtlMs` (default 24h) to PairingManager
- Session tokens are now rejected after TTL expires
- Expired tokens pruned lazily during `isSessionTokenValid()` checks

Refs #1967

## Test Plan

- [x] Token accepted within TTL
- [x] Token rejected after TTL
- [x] Expired tokens pruned from map
- [x] Default TTL is 24 hours
- [x] All existing pairing tests pass

### DIFF
--- a/docs/architecture/reference.md
+++ b/docs/architecture/reference.md
@@ -37,7 +37,6 @@ For essential dev workflow, see [CLAUDE.md](/CLAUDE.md).
 | ToolResult | `src/tool-result.js` | Tool result processing and formatting |
 | MessageTransform | `src/message-transform.js` | Message transformation pipeline |
 | PermissionHook | `src/permission-hook.js` | Permission hook management (CLI mode) |
-| Dashboard | `src/dashboard.js` | Web dashboard HTML generation |
 | TunnelRegistry | `src/tunnel/registry.js` | Tunnel adapter registry (`registerTunnel`/`getTunnel`/`parseTunnelArg`) |
 | BaseTunnelAdapter | `src/tunnel/base.js` | Base class with shared recovery logic (backoff, events) |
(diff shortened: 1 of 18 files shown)


blamechris (commented):
## Agent Review — PR #2035

**Verdict: APPROVE**

### Security Analysis

The core TTL expiry logic is correct and well-implemented:
- `now - meta.createdAt > this._sessionTokenTtlMs` properly rejects expired tokens
- Lazy pruning during `isSessionTokenValid()` prevents unbounded growth of expired entries
- `Map.delete()` during `for...of` iteration is safe per ES6 spec
- `timingSafeEqual` comparison preserved for non-expired tokens
- 24-hour default TTL is reasonable for session tokens

### Findings

**0 Critical, 0 Major, 2 Suggestions**

1. **Style: busy-wait in tests** — The new test file uses `while (Date.now() - start < 5)` busy-wait loops. The existing `pairing.test.js` uses `await delay()` from `node:timers/promises`, which is cleaner and doesn't block the event loop. Consider aligning with the existing pattern for consistency.

2. **Nit: test file could live alongside existing tests** — `pairing-token-expiry.test.js` tests the same class as `pairing.test.js`. These could be added as a new `describe` block in the existing test file rather than a separate file. Not blocking — separate file is fine for isolation.

### Verified

- [x] TTL default calculation (24 * 60 * 60_000 = 86,400,000ms = 24h)
- [x] Constructor accepts and stores `sessionTokenTtlMs`
- [x] Expired tokens pruned during validation
- [x] `createdAt` metadata already stored by `validatePairing()` (line 75 in existing code)
- [x] No breaking changes to existing API surface
- [x] All 4 test cases cover the feature adequately

Issue #2005, opened 2026-03-11
Replace source-scanning tests in ws-server-backpressure.test.js with behavioral tests


blamechris:
## Context

`packages/server/tests/ws-server-backpressure.test.js` (introduced in PR #2003) reads `ws-server.js` as text and uses string matching to verify backpressure behavior. This is the same pattern identified in #2001 for `ws-server-port.test.js`.

## Current tests

- Check for `log.warn` string presence
- Check for `_backpressureDrops` string presence
- Check for `ws.close` or `ws.terminate` string presence

## Suggested behavioral tests

1. Mock a WS client with `bufferedAmount` above threshold, broadcast N messages, verify `_backpressureDrops` increments correctly
2. Verify `ws.close(4008, ...)` is called after exactly `_backpressureMaxDrops` consecutive drops
3. Verify the drop counter resets to 0 after a successful send (bufferedAmount below threshold)
4. Verify the log level is warn (mock the logger)

Refs #2003, #1948

PR #2040, opened 2026-03-11
test(ws): replace source-scanning backpressure test with behavioral test (#2005)

blamechris:
## Summary

- Replace string-matching source tests with behavioral tests
- Exercises actual `_broadcast()` method with mock WS clients
- Tests drop counting, threshold enforcement, connection closure at maxDrops

Refs #2005

## Test Plan

- [x] Message dropped when bufferedAmount > threshold
- [x] Drop counter resets on successful send
- [x] ws.close(4008) after 10 consecutive drops
- [x] Connection stays open before reaching maxDrops

### DIFF
--- a/packages/server/tests/ws-server-backpressure.test.js
+++ b/packages/server/tests/ws-server-backpressure.test.js
@@ -1,4 +1,4 @@
-import { describe, it, before, after, mock } from 'node:test'
+import { describe, it, before, afterEach, mock } from 'node:test'
 import assert from 'node:assert/strict'
 import { EventEmitter } from 'node:events'
 
@@ -10,9 +10,10 @@ describe('WsServer backpressure handling (#1948)', () => {
     ;({ WsServer } = await import('../src/ws-server.js'))
   })
 
-  after(() => {
+  afterEach(() => {
     if (server) {
       try { server.close() } catch {}
+      server = null
     }
   })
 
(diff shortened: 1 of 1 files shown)


blamechris (commented):
## Agent Review — PR #2040

**Verdict: APPROVE**

### Summary

Replaces brittle source-scanning tests (string matching against `ws-server.js` source) with proper behavioral tests that exercise the actual `_broadcast()` method via mock WS clients. This is a clear quality improvement.

### Findings

**Suggestions (non-blocking):**

1. **No test for `_broadcastToSession` backpressure** — The backpressure logic is duplicated in both `_broadcast()` and `_broadcastToSession()` (lines 762-774 and 788-800 of ws-server.js). The new tests only exercise `_broadcast()`. A follow-up could add coverage for the session-scoped variant to catch divergence if one path is modified without the other.

2. **Server cleanup in `after()` hook** — The `after()` hook calls `server.close()` but the server is reassigned in each test via `server = createServer()`. If `createServer()` starts an HTTP listener (port 0), each test leaks a listener until the suite-level `after()` fires and only closes the last one. In practice this is fine because `WsServer` constructor with `port: 0` defers binding to `.listen()`, but worth noting for future tests that call `.listen()`.

3. **Mock `_send` replaces the real method** — The tests monkey-patch `server._send` to a mock. This works because `_broadcast` calls `this._send()`, but it means the tests don't verify the JSON serialization path. Acceptable tradeoff for unit tests.

### Issue Reconciliation

- **#2005** ("Replace source-scanning tests in ws-server-backpressure.test.js") — directly resolved by this PR. Should be closed on merge.

### Test Coverage Assessment

The four test cases cover all branches of the backpressure logic in `_broadcast()`:
- Drop when above threshold (increment counter)
- Reset counter on successful send (below threshold)
- Close connection at max drops (10)
- No close before max drops

All assertions verify correct counter values, send/skip behavior, and close code (4008). Good coverage.

Issue #2048, opened 2026-03-11
test(app): add behavioral tests for streaming overlay (isActive) logic


blamechris:
## Context

Identified during review of PR #2043.

## Description

The previous source-scanning tests asserted on the `useMemo` dependency arrays and the O(1) streaming overlay logic (marking last activity group `isActive: true`). The new behavioral tests in PR #2043 only cover the pure structural `groupMessages` function, not the streaming overlay. The overlay logic should be extracted into a pure helper and tested with cases like: no streaming id, last group not activity, last activity not at end, and active overlay when the last message is part of the last activity group.

## Original Comment

> The new behavioral tests cover the pure structural grouping, but the streaming overlay behavior in `ChatView` (marking the last activity group `isActive: true` when appropriate) is no longer exercised. Since the previous test suite asserted that behavior (and its O(1) update intent), consider extracting the overlay logic into a pure helper and adding behavioral tests.

## Acceptance Criteria

- [ ] Streaming overlay logic extracted into a testable pure function
- [ ] Behavioral tests cover: no streaming id, last group not activity, last activity not at end, active overlay for last message
- [ ] No behavioral changes to ChatView

PR #2060, opened 2026-03-11
test(app): extract and test streaming overlay (isActive) logic (#2048)

blamechris:
## Summary

- Extract `applyStreamingOverlay()` as a pure exported function from ChatView's inline `useMemo`
- Export `DisplayGroup` type and `groupMessages` for external testing
- Add 7 behavioral tests covering all overlay branches
- No behavioral changes to ChatView — same logic, just extracted

Refs #2048

## Test Plan

- [x] All 7 new tests pass
- [x] App type-checks clean
- [x] ChatView behavior unchanged (overlay logic identical)

### DIFF
--- a/packages/app/__tests__/StreamingOverlay.test.ts
+++ b/packages/app/__tests__/StreamingOverlay.test.ts
@@ -1,0 +1,104 @@
+/**
+ * Tests for the streaming overlay (isActive) logic extracted from ChatView.
+ * Verifies the O(1) overlay that marks the last activity group as active
+ * when the streaming message is part of it.
+ */
+import type { ChatMessage } from '../src/store/types';
+import { applyStreamingOverlay, groupMessages, DisplayGroup } from '../src/components/ChatView';
+
+function msg(id: string, type: ChatMessage['type']): ChatMessage {
+  return { id, type, content: '', timestamp: Date.now(), role: 'assistant' } as ChatMessage;
+}
+
+describe('applyStreamingOverlay', () => {
+  it('returns baseGroups unchanged when no streamingMessageId', () => {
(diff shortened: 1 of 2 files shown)


blamechris (commented):
## Agent Review — APPROVE

**Scope:** Extract `applyStreamingOverlay()` as a pure function from ChatView inline `useMemo`; add 7 behavioral tests.

### Findings

**0 critical, 0 major, 2 suggestions**

1. **[Suggestion] Spurious `role` field in test helper** — `msg()` sets `role: 'assistant'` but `ChatMessage` has no `role` field. Harmless due to `as ChatMessage` cast, but could mislead future readers. Consider removing it.

2. **[Suggestion] Missing edge case: `streamingMessageId` that doesn't match any message** — The overlay logic uses `streamingMessageId` purely as a truthy gate (it never checks which message ID matches). This is correct for the current implementation, but a test documenting this behavior would clarify the contract: when `streamingMessageId` is set to a nonexistent ID, the overlay still applies to the last activity group if it's at the tail. This is either the intended behavior (streaming ID is just a "streaming is active" flag) or a latent issue — worth documenting either way.

### Verified

- Logic in extracted `applyStreamingOverlay()` is byte-for-byte identical to the removed inline code
- `useMemo` dependency array unchanged: `[baseGroups, streamingMessageId, messages]`
- All 7 test cases map to the 4 acceptance criteria in #2048
- `DisplayGroup` type export is additive-only (no breaking change)
- `groupMessages` export is additive-only
- No behavioral changes to ChatView rendering

### Verdict

Clean extraction. Tests thoroughly cover all branches. Ship it.

PR #2073, opened 2026-03-13
fix(desktop): apply settings.node_path when starting server (#1938)

### LINKED ISSUE #1938: fix(desktop): settings.node_path is stored but never applied when starting server
## Context
`DesktopSettings` in `packages/desktop/src-tauri/src/settings.rs` has a `node_path: Option<String>` field (line 14) that persists a custom Node binary path to `~/.chroxy/desktop-settings.json`. However, `ServerManager::start_server_process()` in `server.rs` (lines 212-219) ignores this setting entirely — it always calls `node::resolve_node22()` to auto-discover Node, never consulting `settings.node_path`.

This means users who configure a custom Node path via the settings file get no benefit from it.

## Current Behavior
`settings.node_path` is serialized/deserialized but never read during server startup.

## Expected Behavior
If `settings.node_path` is set, `ServerManager` should use it as the Node binary, falling back to `resolve_node22()` if the path doesn't exist or doesn't satisfy the version requirement.

## Acceptance Criteria
- [ ] `handle_start()` in `lib.rs` passes `settings.node_path` to `ServerManager` before calling `start()`
- [ ] `ServerManager::start_server_process()` prefers `settings.node_path` over auto-discovery when set
- [ ] A UI or `save_setup_config` IPC command allows users to set a custom node path
- [ ] Tests cover the "custom path overrides auto-discovery" path

blamechris:
## Summary

- Add `set_node_path` method to `ServerManager` so custom Node paths are used
- Wire `handle_start` to read `settings.node_path` and pass it to `ServerManager` before calling `start()`
- Previously `settings.node_path` was persisted but never applied — now it works

Closes #1938

## Test Plan

- [x] New Rust tests verify `set_node_path` stores and clears paths correctly
- [x] New Node test verifies `lib.rs` calls `set_node_path`
- [x] All 55 Rust tests pass
- [x] `cargo check` clean

### DIFF
--- a/packages/desktop/src-tauri/src/lib.rs
+++ b/packages/desktop/src-tauri/src/lib.rs
@@ -561,6 +561,7 @@ fn handle_start(app: &tauri::AppHandle) {
                         }
                         ServerStatus::Error(ref msg) => {
                             update_menu_state(&app_handle, MenuState::Stopped);
+                            window::emit_server_error(&app_handle, msg);
                             send_notification(&app_handle, "Server Error", msg);
                             return;
                         }
@@ -669,6 +670,7 @@ fn handle_start(app: &tauri::AppHandle) {
                                 Err(_) => {
                                     drop(mgr);
                                     update_menu_state(&app_handle, MenuState::Stopped);
+                                    window::emit_server_error(&app_handle, "Auto-restart failed. Use tray menu to restart manually.");
                                     send_notification(
                                         &app_handle,
                                         "Server Unrecoverable",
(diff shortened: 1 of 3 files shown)


blamechris:
## Copilot Review — All 3 Comments Addressed

All comments fixed in fa6febeb. Summary:

| 1 | Empty/whitespace string can panic via `unwrap()` | **FIX** | `set_node_path` now trims and filters empty strings as `None` |
| 2 | No validation that custom path exists / is Node 22 | **FIX** | Non-existent paths rejected at set time; stale paths cleared at start time with fallback to `resolve_node22()` |
| 3 | Test assertion too weak (`src.includes`) | **FIX** | Now extracts `handle_start` body via regex, verifies both `node_path` read and `set_node_path` call within scope |

All 58 Rust tests + 2 JS tests passing.

PR #2105, opened 2026-03-13
feat(protocol): extract message type enums into shared package (#2080)

### LINKED ISSUE #2080: feat(protocol): extract message type enums into shared package
## Summary

Extract all WebSocket message type strings into typed enums in `@chroxy/protocol`. Currently message types are string literals scattered across server, app, and dashboard with no compile-time verification.

Part of #998 — depends on #2078

## Implementation Plan

- Define `ServerMessageType` enum (all `type` values the server sends: `auth_ok`, `stream_start`, `stream_delta`, `tool_start`, etc.)
- Define `ClientMessageType` enum (all `type` values clients send: `auth`, `send_message`, `approve_permission`, etc.)
- Export both from `@chroxy/protocol`
- Update `packages/server/src/ws-server.js` and `ws-message-handlers.js` to use enum values instead of string literals
- Update `packages/app/src/store/message-handler.ts` switch cases to use enum values
- Update dashboard message handler similarly
- JSDoc annotations for server (plain JS), TypeScript types for app/dashboard

## Acceptance Criteria

- [ ] `ServerMessageType` and `ClientMessageType` enums defined in `@chroxy/protocol`
- [ ] Server uses enum values for all outgoing message types
- [ ] App uses enum values in message handler switch/if chains
- [ ] Adding a new message type starts in the protocol package
- [ ] All existing tests pass

blamechris:
## Summary

- Add `ClientMessageType` and `ServerMessageType` const objects to `@chroxy/protocol`
- All 31 client→server and 52 server→client message type strings extracted from ws-server.js protocol documentation
- Tests cross-reference against the actual ws-server.js protocol comment block to prevent drift
- TypeScript `as const` assertion provides type-safe value unions (`ClientMessageTypeValue`, `ServerMessageTypeValue`)

Closes #2080

## Test Plan

- [x] All 10 protocol tests pass (including cross-reference against ws-server.js)
- [x] TypeScript compiles cleanly
- [x] Existing protocol tests unchanged

### DIFF
--- a/packages/protocol/src/index.ts
+++ b/packages/protocol/src/index.ts
@@ -1,7 +1,8 @@
 /**
  * @chroxy/protocol — shared WebSocket protocol constants
  *
- * Single source of truth for protocol versioning across server, app, and dashboard.
+ * Single source of truth for protocol versioning and message types
+ * across server, app, and dashboard.
  */
 
 /** Current protocol version. Bump when adding new message types. */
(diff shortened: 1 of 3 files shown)


blamechris (commented):
## Agent Review — feat(protocol): extract message type enums

**Verdict: APPROVE with suggestions**

### What this PR does well

- Clean `as const` pattern gives type-safe string unions without runtime overhead
- Cross-reference tests against ws-server.js protocol docs are a smart drift-prevention mechanism
- Correct placement of `encrypted` in both client and server enums (bidirectional envelope)
- `types: []` addition to tsconfig prevents ambient type pollution

### Findings

#### 1. Incomplete coverage — enums only reflect protocol docs, not actual usage (P1 suggestion)

The enums extract **31 client** and **52 server** types from the ws-server.js protocol comment block. However, the actual codebase uses significantly more message types:

**22 client types in `ws-schemas.js` (the real validation layer) missing from `ClientMessageType`:**
`add_repo`, `cli`, `get_diff`, `git_branches`, `git_commit`, `git_stage`, `git_status`, `git_unstage`, `list_conversations`, `list_files`, `list_providers`, `list_repos`, `pair`, `query_permission_audit`, `remove_repo`, `request_cost_summary`, `request_session_context`, `resume_budget`, `resume_conversation`, `search_conversations`, `subscribe_sessions`, `unsubscribe_sessions`

**15 server types sent in ws-*.js code missing from `ServerMessageType`:**
`diff_result`, `error`, `file_list`, `git_branches_result`, `git_commit_result`, `git_stage_result`, `git_status_result`, `git_unstage_result`, `log_entry`, `pair_fail`, `rate_limited`, `session_activity`, `session_context`, `session_updated`, `write_file_result`

The cross-reference tests validate against the protocol docs comment (which is itself incomplete), so they pass despite this gap. This means consumers who adopt these enums will still need raw strings for ~37 message types.

**Recommendation:** This is acceptable as a first step — ship as-is with a follow-up issue to (a) update the ws-server.js protocol docs to include all message types, and (b) add the missing types to the enums. The tests should eventually cross-reference against `ws-schemas.js` too.

#### 2. Protocol doc comment is stale — `PROTOCOL_VERSION` description (P2 nit)

Line 7 of the existing `index.ts` says "Bump when adding new message types" but the ws-server.js bump policy explicitly says do NOT bump for additive changes. This predates the PR but worth noting — the comment is misleading. Not a blocker.

#### 3. Test imports use `../src/index.ts` directly (P2 observation)

Tests import from `../src/index.ts` rather than the built output (`../dist/index.js`). This works because Node 22 with `--experimental-strip-types` handles `.ts` imports, but it means tests don't validate the build output. Acceptable for a `private: true` package.

### Summary

The PR achieves its stated goal cleanly. The main gap is that the ws-server.js protocol docs (the source of truth for these enums) are themselves incomplete — ~37 message types exist in the actual code but not in the docs. A follow-up issue should track closing this gap. No blocking issues found.

PR #2124, opened 2026-03-13
refactor(app): add typed callback map to imperative-callbacks (#2113)

### LINKED ISSUE #2113: refactor(app): add typed callback map to imperative-callbacks module
## Context

From PR #2112 review (Copilot + agent-review suggestion).

## Problem

`CallbackFn` in `imperative-callbacks.ts` is typed as `(...args: any[]) => void`, losing the specific per-callback parameter types that were previously enforced by the Zustand state fields (e.g., `_terminalWriteCallback: ((data: string) => void) | null`).

Currently type safety is preserved at the public API boundary (the setter methods in `connection.ts` enforce correct types), but `getCallback` returns `CallbackFn` which is untyped.

## Proposed Solution

Use a mapped type keyed by `CallbackName` to preserve per-callback signatures:

```typescript
interface CallbackSignatures {
  terminalWrite: (data: string) => void;
  directoryListing: (listing: DirectoryListing) => void;
  // ... etc
}

export function getCallback<K extends CallbackName>(name: K): CallbackSignatures[K] | null;
export function setCallback<K extends CallbackName>(name: K, fn: CallbackSignatures[K] | null): void;
```

## Scope

- `packages/app/src/store/imperative-callbacks.ts` — add interface and generic overloads

blamechris:
## Summary

- Replace untyped `CallbackFn` (`(...args: any[]) => void`) with `CallbackSignatures` interface mapping each callback name to its specific function signature
- `getCallback` and `setCallback` are now generic — return types and parameter types are inferred from the callback name
- Import concrete types from `./types` (DirectoryListing, GitStatusResult, etc.)

Closes #2113

## Test Plan

- [ ] All 9 imperative-callbacks tests pass (3 new type safety tests added)
- [ ] Full app test suite passes (798 tests)
- [ ] TypeScript type check clean
- [ ] No Zustand dependency in module (verified by test)

### DIFF
--- a/packages/app/src/__tests__/imperative-callbacks.test.ts
+++ b/packages/app/src/__tests__/imperative-callbacks.test.ts
@@ -3,6 +3,7 @@ import {
   setCallback,
   clearAllCallbacks,
   type CallbackName,
+  type CallbackSignatures,
 } from '../store/imperative-callbacks';
 
 describe('imperative callbacks module (#2088)', () => {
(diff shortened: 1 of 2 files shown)


blamechris (commented):
**Agent Review: APPROVE**

Clean, well-scoped refactoring. The changes are correct and improve type safety without changing runtime behavior.

**What was reviewed:**
- `CallbackSignatures` interface with concrete types for all 10 callbacks
- Generic `getCallback<K>` / `setCallback<K>` signatures
- `CallbackStore` mapped type replacing `Record<CallbackName, CallbackFn>`
- 3 new compile-time type safety tests
- Import of 9 concrete types from `./types`

**Findings:**

1. **Correctness**: All 10 keys in `CALLBACK_NAMES` have matching entries in `CallbackSignatures`. The mapped type `CallbackStore` will produce a compile error if they diverge — good safety net.

2. **Backward compatibility**: Existing `getCallback`/`setCallback` call sites in `message-handler.ts` and `connection.ts` pass string literals, so TypeScript will infer the generic parameter automatically. No call-site changes needed.

3. **Test coverage**: The new tests verify the type contract at compile time (assignment checks) and at runtime (set/get round-trip). The `jest.fn()` calls in `message-handler.test.ts` will continue to work since Jest mocks are typed as `any`.

4. **Minor observation (non-blocking)**: The `CALLBACK_NAMES` array and `CallbackSignatures` interface are maintained independently. If a future PR adds a callback name to one but not the other, TypeScript will catch it via the `CallbackStore` mapped type — so this is safe as-is. A `satisfies` check like `const _: Record<CallbackName, unknown> = {} as CallbackSignatures` could make this even more explicit, but it's not necessary.

No issues found. Ship it.

PR #2221, opened 2026-03-14
fix(app): break require cycle between connection and notifications

blamechris:
## Summary

- Breaks the require cycle: `connection.ts` → `message-handler.ts` → `notifications.ts` → `connection.ts`
- Replaces top-level imports of `useConnectionStore` and `loadConnection` with lazy `require()` inside the two functions that use them
- Eliminates the runtime warning: `Require cycle: src/store/connection.ts -> src/store/message-handler.ts -> src/notifications.ts -> src/store/connection.ts`

## Test plan

- [x] All 14 notification tests pass
- [x] Full app test suite passes (974 tests)
- [x] TypeScript type check clean
- [ ] Verify warning no longer appears on app launch

### DIFF
--- a/packages/app/src/notifications.ts
+++ b/packages/app/src/notifications.ts
@@ -108,7 +108,10 @@ async function sendPermissionResponseHttp(
 ): Promise<boolean> {
   // Lazy import to break require cycle: connection → message-handler → notifications → connection
   // eslint-disable-next-line @typescript-eslint/no-var-requires
-  const { useConnectionStore, loadConnection } = require('./store/connection');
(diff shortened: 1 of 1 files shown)


blamechris:
## Review Comments Addressed

| 1 | Untyped require() loses type safety (sendPermissionResponseHttp) | FIX | `01fbe99a` |
| 2 | Untyped require() loses type safety (setupNotificationResponseListener) | FIX | `01fbe99a` |

- Fixed: 2 (single commit)
- False positives: 0
- Follow-up issues created: 0

PR #2237, opened 2026-03-14
fix(dashboard): sync activeModel from session_list to prevent dropdown reset

### LINKED ISSUE #2229: fix(dashboard): model dropdown resets to Default after sending message
## Description

When selecting Haiku in the model dropdown and sending a message, the dropdown resets to "Default (recommended)" even though the model actually changes. The UI state doesn't persist the user's selection.

## Steps to Reproduce

1. Open Chroxy desktop app
2. Select "Haiku" from model dropdown
3. Send a message
4. Observe dropdown resets to "Default (recommended)"

## Expected Behavior

Dropdown should stay on "Haiku" after sending a message.

## Found During

Smoke test (2026-03-14)

blamechris:
## Summary

- Syncs `activeModel` from the active session's model property when `session_list` messages arrive
- Prevents the model dropdown from resetting to "Default (recommended)" after sending a message
- Follows the existing pattern used for `conversationId` sync in the same handler

Closes #2229

### DIFF
--- a/packages/server/src/dashboard-next/src/store/message-handler.ts
+++ b/packages/server/src/dashboard-next/src/store/message-handler.ts
@@ -694,12 +694,17 @@ export function handleMessage(raw: unknown, ctxOverride?: ConnectionContext): vo
           set(patch);
         }
         set({ sessions: sessionList });
-        // Sync activeModel from session list to prevent dropdown reset
+        // Sync activeModel from session list to prevent dropdown reset.
+        // session_list sends full model IDs (e.g. claude-sonnet-4-5-20250929) but the
+        // dropdown uses short IDs (e.g. sonnet). Resolve via availableModels lookup.
         const activeSessionId = get().activeSessionId;
         if (activeSessionId) {
-          const activeSessionInfo = sessionList.find((s) => s.sessionId === activeSessionId);
+          const activeSessionInfo = sessionList.find((s: { sessionId?: string }) => s.sessionId === activeSessionId);
           if (activeSessionInfo?.model) {
-            set({ activeModel: activeSessionInfo.model });
+            const fullId = activeSessionInfo.model as string;
+            const models = get().availableModels;
+            const matched = models.find((m) => m.fullId === fullId || m.id === fullId);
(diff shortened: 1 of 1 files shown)


blamechris:
## Code Review

**Verdict: Needs changes** — one bug and one suggestion.

### Bug: model ID format mismatch

`listSessions()` returns `entry.session.model` which is the **full** model ID (e.g. `claude-sonnet-4-20250514`). But the dashboard dropdown options use the **short** ID (`sonnet`), and `model_changed` always sets `activeModel` to `toShortModelId(model)`.

This means after a `session_list` broadcast, `activeModel` gets overwritten with the full ID, and the `<select>` won't match it — the dropdown will appear unselected or fall back to the first option, which is the same bug this PR is trying to fix.

**Fix:** Either convert the model on the client side before setting it, or (better) have the server's `listSessions()` send the short model ID, consistent with `model_changed`. The server-side fix would be:

```js
// in session-manager.js listSessions()
model: entry.session.model ? toShortModelId(entry.session.model) : null,
```

### Suggestion: also sync into sessionStates

The existing `model_changed` handler writes to both `set({ activeModel })` (top-level) and `updateSession(targetId, () => ({ activeModel: model }))` (per-session state). This new sync only writes the top-level. If the user switches tabs and back, the per-session `activeModel` will still be stale.

Consider adding `updateSession` here too, matching the pattern used by `model_changed`:

```ts
if (activeSessionInfo?.model) {
  const shortModel = /* converted short ID */;
  set({ activeModel: shortModel });
  updateSession(activeSessionId, () => ({ activeModel: shortModel }));
}
```

### Race with pending user model change

Not a real concern here. `set_model` sends a WS message to the server, which calls `setModel()` on the session, then broadcasts `model_changed`. Since `session_list` reads `entry.session.model`, both sources reflect the same server-side truth — by the time `session_list` arrives, the model change is already committed. No race.

### Type safety

The `.find()` and optional chaining (`activeSessionInfo?.model`) are correct. `SessionInfo.model` is typed as `string | null`, and the truthiness check properly guards against both `null` and empty string. No issues here.

PR #2239, opened 2026-03-14
fix(dashboard): add user messages and separators to Output view

### LINKED ISSUE #2232: fix(dashboard): Output view missing newlines between responses and user messages
## Description

The Output tab has two issues:

### 1. No newlines between responses
Multiple responses run together without visual separation. Should have a blank line between each response for readability.

### 2. User messages not shown
The Output view only shows assistant responses, not the user messages that triggered them. User messages should be displayed to provide context.

## Screenshot

Output view shows continuous text without separation or user message context.

## Expected Behavior

- Clear visual separation (newline/divider) between each response
- User messages shown before their corresponding responses

## Found During

Smoke test (2026-03-14)

blamechris:
## Summary

- Writes user messages to terminal buffer with yellow ANSI `>` prefix so they appear in Output view
- Handles both local user messages (`addUserMessage` in connection.ts) and remote client messages (`user_input` in message-handler.ts)
- Adds `\r\n` separator after `stream_end` for visual separation between responses

Closes #2232

### DIFF
--- a/packages/server/src/dashboard-next/src/store/connection.ts
+++ b/packages/server/src/dashboard-next/src/store/connection.ts
@@ -783,6 +783,11 @@ export const useConnectionStore = create<ConnectionState>((set, get) => ({
       timestamp: Date.now(),
     };
 
+    // Write user message to terminal buffer for Output view
+    if (text) {
+      get().appendTerminalData(`\r\n\x1b[33m> ${text}\x1b[0m\r\n\r\n`);
+    }
+
     const activeId = get().activeSessionId;
     if (activeId && get().sessionStates[activeId]) {
       updateActiveSession((ss) => ({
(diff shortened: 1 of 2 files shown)


blamechris:
## Code Review

**Verdict: Approve** — clean, small change that follows existing patterns.

### Checked

- **ANSI codes**: `\x1b[33m` (yellow) + `\x1b[0m` (reset) are correct. Consistent with the `tool_start` handler which uses `\x1b[36m` (cyan) in the same interpolation style.
- **Both paths covered**: Local messages via `addUserMessage` (guards on `text`), remote client messages via `user_input` case (guards on `parsed.content`). No gap.
- **Separator placement**: `\r\n` appended at `stream_end` after `flushPendingDeltas()` — correct position, fires once per completed response.
- **Empty string guard**: `parsed.content` defaults to `''` via `parseUserInputMessage`, and the `if (parsed.content)` check correctly skips empty messages.

### Nits (non-blocking)

1. **Terminal escape injection in user text**: `> ${text}` interpolates raw user input. If user text contains ANSI escape sequences they'll be interpreted by xterm.js (e.g., color changes, cursor movement). No security risk since this is a terminal context (not HTML/DOM), but a user could accidentally break formatting. The existing `tool_start` handler does the same with `toolName`, so this is consistent — just worth knowing. A `stripAnsi(text)` wrapper would harden it if desired.

2. **Multi-line input**: Only the first line gets the `>` prefix. A multi-line prompt will show subsequent lines without the marker. Fine for a terminal view, but `text.replace(/\n/g, '\n> ')` would make it visually cleaner if multi-line prompts are common.

LGTM as-is.

PR #2247, opened 2026-03-14
fix(dashboard): reset to default model when selecting Default option

### LINKED ISSUE #2244: fix(dashboard): selecting Default in model dropdown should reset to default model
## Context

Identified during review of PR #2242.

## Description

When a user switches to e.g. Haiku, then selects "Default" to go back, the `onChange` handler skips `setModel()` because the empty string value is falsy. The server keeps the old model while the UI shows "Default."

## Fix

Send `setModel(availableModels[0].id)` when user selects the Default option instead of silently swallowing the event.

## Acceptance Criteria

- [ ] Selecting "Default" after switching models sends `set_model` with the default model ID
- [ ] Server confirms model change via `model_changed` broadcast

blamechris:
## Summary

- When selecting "Default" in the model dropdown, now sends `setModel(availableModels[0].id)` to the server
- Previously the empty value was silently swallowed, leaving the server on the old model while the UI showed "Default"

Closes #2244

### DIFF
--- a/packages/server/src/dashboard-next/src/App.test.tsx
+++ b/packages/server/src/dashboard-next/src/App.test.tsx
@@ -217,6 +217,50 @@ describe('App', () => {
     expect(screen.queryByTestId('session-loading-skeleton')).not.toBeInTheDocument()
   })
 
+  describe('Model selector onChange', () => {
+    const modelsState = {
+      connectionPhase: 'connected' as const,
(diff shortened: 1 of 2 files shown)


blamechris:
**CI Fix (5839f430):** Dashboard Type Check was failing with `TS2532: Object is possibly 'undefined'` on `availableModels[0].id`. TypeScript can't narrow array element access from a `.length > 0` guard. Fixed by extracting to a local variable with a truthiness check.

The Server Tests failure (auto permission mode confirmation handshake) is a pre-existing flaky timing test — passes locally and is unrelated to this PR's changes.

PR #2282, opened 2026-03-15
fix(dashboard): reset FileBrowserPanel on follow mode session switch

### LINKED ISSUE #2274: FileBrowserPanel: stale state when follow mode changes activeSessionId
## Problem

When follow mode is enabled and another client switches sessions, `switchSession()` is called directly from the store's message handler (message-handler.ts line 1653). This bypasses the App component's `isSwitchingSession` guard that normally unmounts/remounts panels.

As a result, `FileBrowserPanel` stays mounted while `activeSessionId` changes underneath. The panel continues showing the previous session's file tree and selected file instead of restoring the new session's state.

## Affected components
- `FileBrowserPanel.tsx` — `savedFilePathRef` and mount effect don't react to `activeSessionId` changes
- Potentially other panels with similar mount-only initialization patterns

## Suggested fix
Key the "load listing + restore file" effect off `activeSessionId` — reset local state, request fresh listing/git status, and restore `selectedFilePath` from the new session's state.

Alternatively, make follow mode go through `handleSwitchSession` in App.tsx to trigger the `isSwitchingSession` unmount/remount cycle.

## Origin
Copilot review on PR #2273.

blamechris:
## Summary
- Key the directory listing effect on `activeSessionId` so it fires on follow mode switches
- Reset all local state (entries, selected file, git status) before loading new session data
- Restore `selectedFilePath` from new session's store state

## Test plan
- [x] All 11 FileBrowserPanel tests pass
- [x] Dashboard type check passes
- [ ] Manual: enable follow mode, switch sessions from another client — Files tab should refresh

Closes #2274

### DIFF
--- a/packages/server/src/dashboard-next/src/App.tsx
+++ b/packages/server/src/dashboard-next/src/App.tsx
@@ -540,7 +540,7 @@ export function App() {
   const handleCreateSession = useCallback((data: { name: string; cwd: string; provider?: string; permissionMode?: string; model?: string }) => {
     setSessionCreateError(null)
     setIsCreatingSession(true)
-    createSession(data.name, data.cwd || undefined, data.provider, data.model, data.permissionMode)
+    createSession({ name: data.name, cwd: data.cwd || undefined, provider: data.provider, model: data.model, permissionMode: data.permissionMode })
   }, [createSession])
 
   const handlePlanApprove = useCallback(() => {
--- a/packages/server/src/dashboard-next/src/store/commands.ts
+++ b/packages/server/src/dashboard-next/src/store/commands.ts
@@ -32,7 +32,7 @@ export function useCommands(): Command[] {
         name: 'New Session',
         category: 'Session',
         shortcut: 'Cmd+N',
-        action: () => createSession('New Session'),
+        action: () => createSession({ name: 'New Session' }),
       },
       {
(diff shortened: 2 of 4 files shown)


blamechris (commented):
## Full Review

Clean, well-scoped fix. The core issue — effect not re-running on session switch — is correctly addressed by adding `activeSessionId` to the dependency array. Specific observations:

1. **`savedFilePathRef` removal is correct.** The ref captured `savedFilePath` at component creation time, so it was stale across follow-mode session switches. Reading from `getState()` inside the effect is the right approach — always gets the current session's data.

2. **`_setSelectedFile` vs `setSelectedFile` is intentional and correct.** Restoring from session state doesn't need to persist back to session state, so the raw setter avoids a no-op write cycle.

3. **State reset ordering is clean.** Clearing everything before requesting new data prevents flash of stale content from the previous session.

4. **Minor nit (non-blocking):** `_setSelectedFile`, `setFileLoading`, and the other `set*` state setters used inside the effect are technically missing from the dependency array. React guarantees `useState` setters are stable so this is safe at runtime, but ESLint exhaustive-deps would flag it. Up to you whether to suppress or add them.

**Verdict: LGTM** — no bugs, no regressions expected. Ready to merge pending CI.

PR #2304, opened 2026-03-15
refactor(dashboard): consolidate header into single Chat Settings dropdown

### LINKED ISSUE #2298: refactor(dashboard): consolidate header into single Chat Settings dropdown
## Description

The header is too crowded with 3 separate dropdowns (Model, Permission Mode, Thinking Level) plus a gear icon, status badge, and cost/token display. On narrow windows this causes horizontal overflow.

## Proposal

Replace the 3 individual dropdowns with a single "Chat Settings" dropdown that has expandable children:

- **Chat Settings** (single dropdown button)
  - Model: Default (recommended) / Sonnet / Haiku / Opus
  - Permission Mode: Approve / Accept Edits / Auto / Plan
  - Thinking Level: Auto / High / Max

This frees header space for: version, connection status, provider badge, cost/tokens.

## Additional Cleanup

- Remove duplicate model/cost info from StatusBar (it duplicates the Tauri title bar)
- The Tauri title bar already shows "v0.5.1 · Connected · blamechris/Projects" + cost/tokens — no need to repeat in dashboard header
- Move QR button elsewhere (settings panel or sidebar)

## Found During

Smoke test (2026-03-16)

blamechris:
Closes #2298

## Summary
- Replace 3 separate header dropdowns (Model, Permission Mode, Thinking Level) with a single `ChatSettingsDropdown` component
- Trigger button shows compact summary: "Sonnet · Approve"
- Clicking opens a dropdown panel with labeled select rows for each setting
- Panel dismisses on outside click, Escape key, or second trigger click
- Reduces header clutter significantly, especially on narrow windows

## Test plan
- [x] 12 new tests for ChatSettingsDropdown (trigger, panel toggle, model/mode/thinking selects, callbacks)
- [x] Updated 2 App.test.tsx model selector tests for new component flow
- [x] All 1086 dashboard tests pass (86 files)

### DIFF
--- a/packages/server/src/dashboard-next/src/App.test.tsx
+++ b/packages/server/src/dashboard-next/src/App.test.tsx
@@ -253,7 +253,9 @@ describe('App', () => {
       const setModelFn = vi.fn()
       stateOverrides = { ...modelsState, setModel: setModelFn }
       render(<App />)
-      const select = screen.getByLabelText('Select model')
+      // Open the ChatSettingsDropdown, then change the model select
+      fireEvent.click(screen.getByTestId('chat-settings-trigger'))
+      const select = screen.getByLabelText('Model')
       fireEvent.change(select, { target: { value: 'claude-opus' } })
       expect(setModelFn).toHaveBeenCalledWith('claude-opus')
     })
@@ -276,7 +278,9 @@ describe('App', () => {
         }),
       }
       render(<App />)
-      const select = screen.getByLabelText('Select model')
+      // Open the ChatSettingsDropdown, then change the model select
+      fireEvent.click(screen.getByTestId('chat-settings-trigger'))
+      const select = screen.getByLabelText('Model')
       fireEvent.change(select, { target: { value: '' } })
       expect(setModelFn).toHaveBeenCalledWith('claude-sonnet')
     })
--- a/packages/server/src/dashboard-next/src/App.tsx
+++ b/packages/server/src/dashboard-next/src/App.tsx
@@ -21,6 +21,7 @@ import { processImageFiles, filterImageFiles } from './utils/image-utils'
 import { getAuthToken } from './utils/auth'
 import { SessionBar, type SessionTabData, type SessionStatus } from './components/SessionBar'
 import { StatusBar } from './components/StatusBar'
+import { ChatSettingsDropdown } from './components/ChatSettingsDropdown'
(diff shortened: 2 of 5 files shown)


blamechris:
## Review — PR #2304

**refactor(dashboard): consolidate header into single Chat Settings dropdown**

### Overall Assessment
Clean, well-structured refactor that successfully consolidates three separate header selects into a single dropdown component. The component is well-typed, properly handles accessibility attributes, and comes with comprehensive tests. Good job.

### Strengths
- **Clean extraction**: The inline JSX in App.tsx was getting unwieldy; this is a meaningful readability improvement
- **Proper accessibility**: `aria-expanded`, `aria-haspopup`, `role="dialog"`, `aria-label`, and `htmlFor` labels are all present
- **Good test coverage**: 12 tests covering trigger rendering, panel toggle, all three selects, callbacks, and edge cases (defaultModelId)
- **Outside click + Escape handling**: Both dismiss behaviors are correctly implemented with proper cleanup

### Suggestions

1. **Two `useEffect` listeners could be merged** (lines 41-63): The outside-click and Escape handlers both guard on `!open` and add/remove document listeners. Merging into one `useEffect` would reduce the number of effect subscriptions when the panel is open.

2. **Missing keyboard accessibility for focus trap**: The panel uses `role="dialog"` but doesn't trap focus. When opened, focus stays on the trigger button rather than moving into the panel. For a settings dropdown this is acceptable, but a focus trap would be the ideal accessibility pattern. (Deferring to follow-up.)

3. **`value` logic for model select may confuse** (line 115): `activeModel === defaultModelId ? '' : (activeModel || '')` — when the user explicitly picks the default model, the select shows "Default (...)" which is correct UX, but could be surprising if `activeModel` is `null` and `defaultModelId` is also `null` (both falsy, `null === null` is true, so value becomes `''`). This edge case works correctly by coincidence but deserves a comment.

4. **Panel positioning on narrow viewports**: The panel uses `left: 50%; transform: translateX(-50%)` which centers it relative to the trigger. On very narrow windows, the panel (min-width: 240px) could overflow the viewport edge. Consider adding `max-width: calc(100vw - 32px)` or similar constraint in a follow-up.

5. **No test for Escape key dismissal**: The Escape handler is implemented but not tested. Would be good to add a `fireEvent.keyDown` test.

### Verdict
**Approve** — all suggestions are non-blocking. The code is clean, well-tested, and achieves the stated goal effectively.

PR #2342, opened 2026-03-15
fix(server): replace single-slot _pendingMessage with queue in cli-session.js

### LINKED ISSUE #2317: fix(server): replace single-slot _pendingMessage with queue in cli-session.js
## Problem
`cli-session.js:261-264` uses a single `_pendingMessage` scalar. If the user sends a second message while the process is respawning (e.g. during model switch, 2-10 seconds), the second write overwrites the first. The user's earlier message is **silently discarded**.

## Found By
Guardian (swarm audit 2026-03-15), confirmed by Skeptic

## Fix
Replace `_pendingMessage` with `_pendingQueue = []`. Dequeue in FIFO order in `_spawnPersistentProcess`. Add max-depth guard (1-3 messages) and emit `error` for overflow.

## Audit Reference
`docs/audit-results/code-quality-practices/03-guardian.md` Finding 1

blamechris:
## Summary

Fixes #2317.

- `_pendingMessage` (scalar) is replaced by `_pendingQueue = []` (array)
- `sendMessage` pushes onto the queue when the process is not ready, preserving every message instead of overwriting
- `_spawnPersistentProcess` drains the queue with `Array.shift()` (FIFO) after the process becomes ready
- A max-depth guard (3 messages) emits an `error` event and discards any further arrivals during a long respawn window

## Test plan

- [ ] New test file `cli-session-pending-queue.test.js` — 10 tests covering: initial state, enqueuing up to 3 messages in FIFO order, preserving attachments/options, overflow guard (4th message emits error + is discarded), boundary (3rd message OK), FIFO dequeue, and busy-guard independence
- [ ] Updated existing `cli-session.test.js` "queues message when process not ready" assertion to use `_pendingQueue`
- [ ] All 58 tests pass: `PATH="/opt/homebrew/opt/node@22/bin:$PATH" node --test tests/cli-session-pending-queue.test.js tests/cli-session.test.js`

### DIFF
--- a/packages/server/src/cli-session.js
+++ b/packages/server/src/cli-session.js
@@ -544,6 +544,8 @@ export class CliSession extends BaseSession {
 
   /**
    * Clear per-message state, marking us as ready for the next message.
+   * After clearing, drains the next item from _pendingQueue (if any) so
+   * that all queued messages are eventually delivered in FIFO order.
    */
   _clearMessageState() {
     super._clearMessageState()
@@ -560,6 +562,14 @@ export class CliSession extends BaseSession {
     if (this._interruptTimer) {
       clearTimeout(this._interruptTimer)
       this._interruptTimer = null
+    }
+
+    // Drain the next queued message now that the process is free.
+    // Only drain when the process is ready (not mid-respawn).
+    if (this._pendingQueue.length > 0 && this._processReady) {
+      const pending = this._pendingQueue.shift()
+      log.info(`Dequeuing next pending message after result (${this._pendingQueue.length} remaining)`)
+      this.sendMessage(pending.prompt, pending.attachments, pending.options || {})
     }
   }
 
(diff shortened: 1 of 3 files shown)


blamechris (commented):
**Review: fix(server): replace single-slot _pendingMessage with queue**

The scalar→array upgrade is correct and the overflow guard (max 3, emit error) is a reasonable policy. Code style is clean and the existing test in `cli-session.test.js` is properly updated.

**NEEDS_FIX — one real gap:**

**Queue is never drained beyond the first message.** `_spawnPersistentProcess` calls `shift()` once after spawn, but there is no drain path after each `result`/`_clearMessageState()`. If 2–3 messages arrived during respawn, messages 2 and 3 will remain in `_pendingQueue` indefinitely. The array structure is better than the scalar, but the queue only fully works if drain logic is added after each message completes. The fix: in the `result` case (or at the end of `_clearMessageState`), if `_pendingQueue.length > 0 && !_isBusy`, shift and send the next entry.

**Secondary (test quality):** The FIFO dequeue tests manually `shift()` and call `sendMessage()` directly, bypassing the production drain path entirely. Once the drain-after-result logic exists, add a test that queues 3 messages, fires 3 `result` events, and verifies all 3 are sent in order.

PR #2344, opened 2026-03-15
fix(server): await destroySessionLocked in session-handlers.js

### LINKED ISSUE #2320: fix(server): await destroySessionLocked in session-handlers.js
## Problem
`session-handlers.js:77-78` calls `destroySessionLocked()` without awaiting. This async function acquires a mutex lock — when the return is discarded, client cleanup runs before the lock is acquired, defeating the lock's purpose. Errors become unhandled promise rejections.

## Found By
Skeptic (swarm audit 2026-03-15)

## Fix
`await ctx.sessionManager.destroySessionLocked(targetId)` and make the handler async.

## Audit Reference
`docs/audit-results/code-quality-practices/01-skeptic.md` Finding 1

blamechris:
## Summary

- `handleDestroySession` called `destroySessionLocked()` without `await`, discarding the returned promise before the mutex lock was acquired
- Client cleanup (subscription removal, session_destroyed broadcast) ran before the lock was held, defeating the lock's purpose
- Unawaited async errors became silent unhandled promise rejections
- Fix: make `handleDestroySession` async and add `await` before `destroySessionLocked(targetId)`
- The surrounding `handleSessionMessage` dispatcher already uses `await handler(...)`, so async handlers are fully supported with no other changes needed

## Test plan

- [ ] `awaits destroySessionLocked when present` — verifies the async destroy completes before the `session_destroyed` broadcast is sent
- [ ] `uses destroySessionLocked over destroySession when both present` — verifies the locked variant is always preferred when available

Closes #2320

### DIFF
--- a/packages/server/src/handlers/session-handlers.js
+++ b/packages/server/src/handlers/session-handlers.js
@@ -57,7 +57,7 @@ function handleCreateSession(ws, client, msg, ctx) {
   }
 }
 
-function handleDestroySession(ws, client, msg, ctx) {
+async function handleDestroySession(ws, client, msg, ctx) {
   const targetId = msg.sessionId
   if (!ctx.sessionManager.getSession(targetId)) {
     ctx.send(ws, { type: 'session_error', message: `Session not found: ${targetId}` })
@@ -75,7 +75,7 @@ function handleDestroySession(ws, client, msg, ctx) {
   }
 
   if (typeof ctx.sessionManager.destroySessionLocked === 'function') {
-    ctx.sessionManager.destroySessionLocked(targetId)
+    await ctx.sessionManager.destroySessionLocked(targetId)
   } else {
     ctx.sessionManager.destroySession(targetId)
   }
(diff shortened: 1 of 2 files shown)


blamechris (commented):
**Review: fix(server): await destroySessionLocked in session-handlers.js**

Making `handleDestroySession` async and adding `await` before `destroySessionLocked` is the correct fix. The surrounding `handleSessionMessage` dispatcher already uses `await handler(...)` so async handlers are fully supported. The two new tests correctly verify ordering (async completes before broadcast) and method preference (locked over plain).

**NEEDS_FIX — two issues:**

1. **Return value not checked (real bug):** `destroySessionLocked()` returns `false` when the session disappears between the existence check and lock acquisition (TOCTOU race). The code currently discards the resolved value and unconditionally broadcasts `session_destroyed` + `session_list`, which is incorrect. Fix: `const destroyed = await ctx.sessionManager.destroySessionLocked(targetId); if (!destroyed) { ctx.send(ws, { type: 'session_error', ... }); return }`.

2. **Flaky 50ms sleep in test:** The first new test uses `await new Promise(r => setTimeout(r, 50))` to wait for `destroySessionLocked` to be called. This is fragile on slow CI runners. Replace with a polling loop or `waitForMessage`-style check.

PR #2346, opened 2026-03-15
fix(server): separate _respawning flag from _destroying in cli-session

### LINKED ISSUE #2319: fix(server): separate _respawning flag from _destroying in cli-session.js
## Problem
`_destroying` serves two purposes: "permanently destroyed" (in `destroy()`) and "controlled kill-respawn" (in `_killAndRespawn()`). If `destroy()` is called during a `_killAndRespawn()` cycle, `respawn()` sets `_destroying = false`, re-enabling respawn for a permanently destroyed session — spawning zombie child processes.

## Found By
Guardian (swarm audit 2026-03-15), confirmed by Skeptic

## Fix
Use a separate `_respawning` flag in `_killAndRespawn()`.

## Audit Reference
`docs/audit-results/code-quality-practices/03-guardian.md` Finding 5

blamechris:
## Summary

- `_destroying` was used for two purposes: permanent teardown (`destroy()`) and controlled kill-for-respawn (`_killAndRespawn()`). If `destroy()` was called during a respawn cycle, the respawn callback reset `_destroying = false`, re-enabling respawn on a permanently-destroyed session and spawning zombie child processes.
- Adds a dedicated `_respawning` flag used exclusively by `_killAndRespawn()`. `_destroying` is now only ever set by `destroy()`.
- The respawn callback checks `if (this._destroying) return` before calling `start()`, so a concurrent `destroy()` silently wins.
- `_scheduleRespawn()` now also guards on `_respawning` to block unrelated close-event respawns during an in-progress `_killAndRespawn()` cycle.
- `destroy()` sets `_respawning = false` so any pending respawn callback is a no-op even if it fires after destruction.

## Test plan

- [ ] New test file `cli-session-respawning-flag.test.js` covering: constructor init, `_killAndRespawn` sets `_respawning` not `_destroying`, zombie-prevention path (destroy during respawn), `_scheduleRespawn` no-op when `_respawning`, regression that normal respawn still fires `start()`
- [ ] Updated existing `_killAndRespawn` behavioral tests to assert `_respawning` instead of `_destroying`
- [ ] All 79 cli-session tests pass: `node --test packages/server/tests/cli-session*.test.js`

Closes #2319

### DIFF
--- a/packages/server/src/cli-session.js
+++ b/packages/server/src/cli-session.js
@@ -188,6 +188,7 @@ export class CliSession extends BaseSession {
       this._child = null
 
       if (this._destroying) return
+      if (this._respawning) return
 
       // Safety net: if we were mid-message, close the stream
       if (this._isBusy && this._currentMessageId) {
--- a/packages/server/src/cli-session.js
+++ b/packages/server/src/cli-session.js
@@ -71,6 +71,7 @@ export class CliSession extends BaseSession {
     this._respawnCount = 0
     this._respawnTimer = null
     this._respawnScheduled = false
+    this._respawning = false
     this._interruptTimer = null
 
     // Hook manager (shared module)
@@ -226,6 +227,7 @@ export class CliSession extends BaseSession {
    */
   _scheduleRespawn() {
     if (this._destroying) return
+    if (this._respawning) return
     if (this._respawnScheduled) return
 
     this._respawnCount++
@@ -563,7 +565,7 @@ export class CliSession extends BaseSession {
    * Suppresses auto-respawn during the kill, clears timers, and starts fresh.
    */
   _killAndRespawn() {
-    this._destroying = true
+    this._respawning = true
     this._processReady = false
     this._sessionId = null
 
@@ -588,8 +590,9 @@ export class CliSession extends BaseSession {
       const respawn = () => {
         if (didClose) return
         didClose = true
-        this._destroying = false
+        this._respawning = false
         this._respawnCount = 0
+        if (this._destroying) return
         this.start()
       }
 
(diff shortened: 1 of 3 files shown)


blamechris (commented):
Reviewed the diff in full. The core fix is correct: `_respawning` and `_destroying` are now independent flags with clear ownership — `_killAndRespawn()` uses `_respawning`, `destroy()` uses `_destroying`. The zombie-prevention guard (`if (this._destroying) return` inside the respawn callback) is exactly right, and `destroy()` clearing `_respawning` closes the race window cleanly.

**One real issue to address (Copilot comment #2937130271 — spurious error emission):**

The `close` handler in `_spawnPersistentProcess()` (line ~189) guards on `if (this._destroying) return` but not on `_respawning`. During a `_killAndRespawn()` cycle the process emits `close`, the guard doesn't fire, and the handler logs "Process exited..." and emits an `error` event before calling `_scheduleRespawn()` (which is correctly blocked by `_respawning`). This surfaces a spurious error notification on every model/permissionMode switch. Fix: add `if (this._respawning) return` after the existing `if (this._destroying) return` check.

**Test comments (Copilot #2937130276 and #2937130278):**

- The `done`-callback test for normal respawn uses the real 1000ms backoff timer, which slows the suite. This is a quality-of-life issue, not a correctness bug — acceptable for now but worth flagging.
- The zombie-prevention test leaves a dangling 3s force-kill timer because the mock child never emits `close` after `destroy()`. Low impact (jest/node:test usually exits fine) but the cleanup is trivial: `oldChild.emit('close', 0)` at the end of that test.

**Verdict:** NEEDS_FIX — the spurious error emission on every controlled respawn is a real regression introduced by this PR. The fix is a one-liner in `_spawnPersistentProcess`.

PR #2347, opened 2026-03-15
fix(server): capture child reference in supervisor force-kill timer

### LINKED ISSUE #2321: fix(server): capture child reference in supervisor force-kill timer
## Problem
`supervisor.js:554-562` — the `forceKillTimer` callback references `this._child`, but if a new child starts during shutdown, `forceKill(this._child)` kills the **new child**, not the old one.

## Fix
Capture a local reference before the timeout: `const childRef = this._child`.

## Audit Reference
`docs/audit-results/code-quality-practices/03-guardian.md` Finding 3

blamechris:
## Summary

- In `supervisor.js` `shutdown()`, the `forceKillTimer` callback previously closed over `this._child` — if a new child was assigned to `this._child` during the 5s timeout window, the timer would kill the new child instead of the original one
- Captures `const childRef = this._child` before the `setTimeout` and uses `childRef` in both the timer callback and the `once('exit')` listener

## Test plan

- `captures child reference so force-kill targets original child, not a replacement` — simulates the race: reassigns `this._child` to an intruder after shutdown starts, verifies the intruder's `kill()` is never called
- `force-kill timer is cancelled when the captured child exits before timeout` — happy-path: clean exit clears the timer, no force-kill fires
- `shutdown sends to childRef, not this._child` — verifies the shutdown message goes to the original child
- Source-level assertion that `childRef` is used in both the timer and exit listener

Run: `PATH="/opt/homebrew/opt/node@22/bin:$PATH" node --test packages/server/tests/supervisor*.test.js`

Closes #2321

### DIFF
--- a/packages/server/src/supervisor.js
+++ b/packages/server/src/supervisor.js
@@ -552,13 +552,14 @@ export class Supervisor extends EventEmitter {
     this._stopStandbyServer()
 
     if (this._child) {
-      this._child.send({ type: 'shutdown' })
+      const childRef = this._child
+      childRef.send({ type: 'shutdown' })
       const forceKillTimer = setTimeout(() => {
         this._log.info('Force-killing child after 5s timeout')
-        try { forceKill(this._child) } catch {}
+        try { forceKill(childRef) } catch {}
       }, 5000)
 
-      this._child.on('exit', () => clearTimeout(forceKillTimer))
+      childRef.once('exit', () => clearTimeout(forceKillTimer))
     }
 
     if (this._tunnel) {
--- a/packages/server/tests/supervisor-force-kill-ref.test.js
+++ b/packages/server/tests/supervisor-force-kill-ref.test.js
@@ -142,6 +142,10 @@ describe('supervisor force-kill child reference (#2321)', () => {
   })
 
   it('force-kill timer is cancelled when the captured child exits before timeout', async () => {
+    // Structural assertion: this test verifies the timer-cleared path, not the race itself.
+    // The timer always fires AFTER the child exit in this synchronous test setup, so the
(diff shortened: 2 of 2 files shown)


blamechris (commented):
Reviewed the full diff. The production fix is correct and minimal: capturing `const childRef = this._child` before the `setTimeout` and using `childRef` in both the timer callback and the `once('exit')` listener eliminates the stale-reference race. Switching from `on` to `once` for the exit listener is also a good cleanup.

**Test issues (all 4 Copilot comments are valid):**

1. **Race not actually exercised (#2937129830):** The first test reassigns `supervisor._child` to an intruder after `shutdown()` but then immediately calls `originalChild.simulateExit()`, which clears the timer before it fires. The test passes because no timer fires at all — `intruderChild.kill()` is never called regardless of the fix. To properly exercise the race, time-mock the `setTimeout` (or use `mock.timers`) and tick past 5000ms before the original child exits.

2. **"Timer cancelled" assertion not verified (#2937129837):** Same issue — the test passes whether or not the timer fires, since the original child exits before 5s. Needs `mock.timers` to assert no kill occurs after advancing past the timeout.

3. **send() test doesn't distinguish childRef from this._child (#2937129842):** `this._child` is never reassigned before verifying `send()`, so the test would pass even without the `childRef` capture.

4. **Unused `_forceKillTargets` array (#2937129847):** Dead code in the test helper — remove it.

**Code style (server package):** No issues. ES modules, no semicolons, single quotes — all correct.

**Verdict:** NEEDS_FIX — the test suite does not actually verify the fix. The production code is right, but the tests need `mock.timers` to make the race scenario meaningful. Issues 1 and 2 are the critical gaps; 3 and 4 are minor cleanup.

PR #2348, opened 2026-03-15
security(server): add rate limit for permission_response messages

### LINKED ISSUE #2324: security(server): add rate limit for permission_response messages
## Problem
`ws-server.js:799-805` — `permission_response` and `user_question_response` messages are exempted from rate limiting. An authenticated client can send unlimited permission responses, enabling DoS or permission race attacks.

## Found By
Adversary (swarm audit 2026-03-15)

## Fix
Apply a separate, relaxed rate limit (e.g., 60 per minute) on permission response messages. Use a separate bucket to preserve latency sensitivity.

## Audit Reference
`docs/audit-results/code-quality-practices/06-adversary.md` Finding 4

blamechris:
## Summary

- Replaces the blanket rate-limit exemption for `permission_response` and `user_question_response` with a dedicated, relaxed limiter (60 per minute, no burst).
- Regular messages continue through the existing limiter (100/min + 20 burst).
- Permission limiter state is cleaned up on client disconnect alongside the regular limiter.
- 5 new integration tests covering: per-type enforcement, bucket isolation (permission sends do not consume the regular budget), and that the two limiters fire independently.

## Test plan

- [x] `node --test packages/server/tests/ws-server-rate-limit.test.js packages/server/tests/rate-limiter.test.js` — all 12 tests pass
- [x] Verify `rate_limited` message includes `retryAfterMs` and a descriptive `message` field for each limiter

Closes #2324

### DIFF
--- a/packages/server/src/ws-server.js
+++ b/packages/server/src/ws-server.js
@@ -302,6 +302,8 @@ export class WsServer {
     this._backpressureThreshold = backpressureThreshold ?? 1024 * 1024 // 1MB default
     this._backpressureMaxDrops = 10 // close connection after this many consecutive drops
     this._rateLimiter = new RateLimiter()
+    // Separate, relaxed limiter for permission/question responses (60 per minute, no burst)
+    this._permissionRateLimiter = new RateLimiter({ windowMs: 60_000, maxMessages: 60, burst: 0 })
     this._clientSend = createClientSender(log)
     this._clientManager = new WsClientManager()
     this.clients = this._clientManager.clients // back-compat: expose the raw Map for context objects
@@ -664,7 +666,10 @@ export class WsServer {
         if (client?.authenticated) {
           this._handleClientDeparture(client)
         }
-        if (client?.id) this._rateLimiter.remove(client.id)
+        if (client?.id) {
+          this._rateLimiter.remove(client.id)
+          this._permissionRateLimiter.remove(client.id)
+        }
         this._clientManager.removeClient(ws)
       })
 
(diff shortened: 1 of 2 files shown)


blamechris (commented):
Reviewed the full diff. The approach is correct: replacing the blanket exemption with a dedicated, relaxed `RateLimiter` instance for `permission_response` and `user_question_response` is cleaner than the previous skip, and cleaning up on client disconnect is handled correctly.

**Issues to address:**

1. **Misleading error message (#2937129074 — real bug):** The rate-limited response always says "Too many permission responses. Please slow down." even when the limited message is `user_question_response`. This is confusing for clients trying to diagnose which limiter fired. A simple fix: use shared wording like "Too many permission/question responses. Please slow down." or branch on `msg.type`.

2. **`withTimeout` leaks timers (#2937129083):** The helper creates a `setTimeout` but never clears it after the race resolves, leaving pending timers that can keep the event loop alive. Fix: track the handle and `clearTimeout()` (or `unref()`) it in a `finally` block after `Promise.race()` settles.

3. **`setLogListener(null)` global side-effect (#2937129089):** Calling `setLogListener(null)` in the `WsServer` wrapper clears the global log listener, which can interfere with other tests in the same process. The `start()` override is the right place to suppress noisy logs, but a null-reset is too broad. Consider a no-op log function, or just remove the suppression since the tests are integration-level anyway.

**Code style:** Correct — no semicolons, single quotes, ES modules.

**Verdict:** NEEDS_FIX — issue 1 is a visible correctness issue (wrong error message for `user_question_response`); issues 2 and 3 are test quality problems. All three are quick fixes.

PR #2353, opened 2026-03-15
fix(server): use createLogger instead of console.log in cli-session.js

### LINKED ISSUE #2326: fix(server): use createLogger instead of console.log in cli-session.js
## Problem
`cli-session.js` has 15 direct `console.*` calls while `sdk-session.js` uses structured `createLogger`. When the server runs with `logFormat: json`, half the session lifecycle logs are raw strings that can't be machine-parsed and aren't captured by the dashboard log listener.

## Found By
Builder (swarm audit 2026-03-15), confirmed by Skeptic

## Fix
Replace all `console.*` calls with `const log = createLogger('cli-session')` and appropriate log levels.

## Audit Reference
`docs/audit-results/code-quality-practices/02-builder.md` Finding 2

blamechris:
## Summary

- Adds `import { createLogger } from './logger.js'` and `const log = createLogger('cli-session')` near the top of the file
- Replaces all 15 `console.log`/`console.warn`/`console.error` calls with `log.info`/`log.warn`/`log.error` calls
- Removes the redundant `[cli-session]` prefix from each message string (the logger component tag handles this)

## Why

When the server runs with `logFormat: json`, the direct `console.*` calls in `cli-session.js` emit raw strings that bypass the logger's JSON serialisation and are invisible to the dashboard log listener. This brings `cli-session.js` in line with `sdk-session.js`.

Closes #2326

### DIFF
--- a/packages/server/src/cli-session.js
+++ b/packages/server/src/cli-session.js
@@ -7,6 +7,9 @@ import { forceKill } from './platform.js'
 import { MessageTransformPipeline } from './message-transform.js'
 import { emitToolResults } from './tool-result.js'
 import { parseMcpToolName } from './mcp-tools.js'
+import { createLogger } from './logger.js'
+
+const log = createLogger('cli-session')
 
 // Default max accumulated size for tool_use input_json_delta chunks (~256KB)
 const DEFAULT_MAX_TOOL_INPUT_LENGTH = 262144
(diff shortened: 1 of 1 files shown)


blamechris (commented):
Reviewed. All 15 `console.log/warn/error` calls are replaced with the structured logger. The `[cli-session]` prefix is correctly dropped from each message string since `createLogger('cli-session')` injects it automatically. One minor ordering note: the `log.warn` call for `toolInputChunks exceeded` is moved two lines earlier (before setting `ctx.toolInputOverflow = true`) — this is harmless and actually slightly more informative. Logger migration matches the pattern already used in `sdk-session.js`.

No Copilot comments. No regressions expected.

**Verdict: APPROVE**

PR #2355, opened 2026-03-15
fix(server): use monotonic counter for session auto-naming

### LINKED ISSUE #2338: fix(server): use monotonic counter for session auto-naming
## Problem
\`session-manager.js:261\` uses \`this._sessions.size + 1\` for auto-naming. If sessions are destroyed between creations, the counter regresses — producing duplicate session names (e.g., two "Session 2"s).

## Found By
Skeptic (swarm audit 2026-03-15)

## Fix
Use a monotonically incrementing \`this._sessionCounter\` rather than \`_sessions.size\`.

## Audit Reference
\`docs/audit-results/code-quality-practices/01-skeptic.md\` Finding 2

blamechris:
## Summary

- Replaces `this._sessions.size + 1` with `++this._sessionCounter` in `createSession()`
- Adds `this._sessionCounter = 0` to the `SessionManager` constructor
- The counter is strictly monotonic — it never decrements when sessions are destroyed, eliminating the duplicate-name bug

## Test plan

- [ ] New test file `session-manager-naming-counter.test.js` (4 cases): first session name, sequential naming, no number reuse after destroy, explicit name bypasses counter
- [ ] All 88 existing session-manager tests still pass

Closes #2338

### DIFF
--- a/packages/server/src/session-manager.js
+++ b/packages/server/src/session-manager.js
@@ -174,6 +174,7 @@ export class SessionManager extends EventEmitter {
 
     // Internal state
     this._sessions = new Map() // sessionId -> { session, name, cwd, createdAt }
+    this._sessionCounter = 0   // monotonically incrementing; used for auto-naming
     this._locks = new SessionLockManager()
 
     // Session idle timeout (delegated to SessionTimeoutManager)
@@ -258,7 +259,7 @@ export class SessionManager extends EventEmitter {
     }
 
     const sessionId = randomBytes(16).toString('hex')
-    const sessionName = name || `Session ${this._sessions.size + 1}`
+    const sessionName = name || `Session ${++this._sessionCounter}`
 
     const resolvedProvider = provider || this._providerType
     const ProviderClass = getProvider(resolvedProvider)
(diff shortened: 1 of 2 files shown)


blamechris (commented):
Reviewed. The monotonic counter correctly prevents number reuse when sessions are destroyed mid-lifecycle. The 4 new tests cover the core cases well.

**Two issues raised by Copilot (threads resolved above):**

1. **Persist/restore regression:** `restoreState()` always calls `createSession({ name: saved.name, ... })` with an explicit name, so `_sessionCounter` never increments during restore and stays at 0. After a server restart the first auto-named session would become "Session 1" even if restored sessions already include "Session N", causing a name collision. Fix required: persist `_sessionCounter` in the state file and restore it in `restoreState()`.

2. **Missing test coverage for restart scenario:** The existing 4 tests only cover in-process monotonicity. A test simulating serialize → fresh manager → restore → new auto-named session (asserting name continues from prior max) is needed to guard the persist/restore fix.

Both issues must be resolved before merge.

**Verdict: NEEDS_FIX** — counter persistence + corresponding test required.

PR #2379, opened 2026-03-15
fix(server): prevent hook registration leak to settings.json on destroy race

### LINKED ISSUE #2365: fix(server): prevent hook registration leak to settings.json on session destroy race
## Problem
If a session is destroyed while `_hookManager.register()` is still pending (contended settings lock), the register succeeds AFTER unregister, leaving a dead hook in `~/.claude/settings.json`. This causes Claude Code to hang on every tool use waiting for a webhook that no longer exists.

## Found By
Guardian (post-fix audit 2026-03-15)

## Fix
Check `_destroying` state inside the settings lock body before writing. Await `unregister()` promise or chain it after pending `register()`.

blamechris:
## Summary

Fixes a race condition where destroying a session while `_hookManager.register()` is still waiting to acquire the settings lock leaves a dead chroxy hook in `~/.claude/settings.json`, causing Claude Code to hang on every tool use.

- Add a `_destroying` guard inside the register lock callback — skips the write if `destroy()` has been called while the lock was contended
- Store the in-flight `register()` promise (`pendingRegister`) so `destroy()` can chain `unregister()` after it, ensuring correct sequencing under lock contention
- `destroy()` now returns a Promise that resolves once the chained unregister completes
- Simplify `cli-session.js` `destroy()` to delegate fully to `hookManager.destroy()` instead of calling `unregister()` + `destroy()` separately

## Test plan

- [x] `node --test packages/server/tests/permission-hook.test.js` — 25 tests pass (4 new: destroy returns Promise, race guard before register completes, destroy after register, destroying flag prevents write)
- [x] `node --test packages/server/tests/cli-session*.test.js` — 106 tests pass, no regressions

Closes #2365


blamechris (commented):
Reviewed. 1 Copilot comment addressed and thread resolved.

**What the PR does:** Adds a `pendingRegister` ref that tracks the most recent in-flight `register()` promise. `destroy()` now chains `unregister()` after any pending register via `pending.then(() => unregister())`. A `destroying` guard inside the settings lock skips the write if `destroy()` raced in while the lock was contended, preventing a dead hook from being written and never cleaned up.

The caller in `cli-session.js` is simplified: `_hookManager.unregister()` is removed from destroy since `destroy()` now internally chains it.

**Fix matches issue:** Yes. The register-after-unregister race that left a dead chroxy hook in `~/.claude/settings.json` is closed by both the `destroying` guard (skip write) and the chained unregister (clean up even if write already happened).

**Tests:** 4 new tests cover: destroy() returns Promise, race guard (destroy before register completes), normal post-register destroy, and the destroying-flag-skips-write case. Good coverage of the specific race paths.

**Issues found:**

1. Existing tests that call `manager.destroy()` without awaiting it now have a latent race: the async unregister runs after the test ends, potentially affecting subsequent tests or causing flaky cleanup in afterEach. Existing test cleanup should await destroy(). (Follow-up tracked.)

**Regressions:** None visible. The change to remove the explicit `_hookManager.unregister()` call in cli-session.js is correct because destroy() now subsumes it.

APPROVE — race fix is correct. The one Copilot concern is a test hygiene issue, not a production regression.

PR #2381, opened 2026-03-15
fix(server): drain pending queue via nextTick to prevent re-entrancy

### LINKED ISSUE #2371: fix(server): drain pending queue via nextTick to prevent re-entrancy in _clearMessageState
## Problem
`cli-session.js` `_clearMessageState()` synchronously calls `sendMessage()` during drain. If a `result` event listener also calls `sendMessage()`, both succeed — sending two messages to a process expecting one. Also, destroy() + pending queue interaction causes spurious error events.

## Fix
Drain the queue in `process.nextTick()` instead of synchronously. Check `_destroying` at the top of the drain block.

blamechris:
## Summary

- Wraps the `_pendingQueue` drain in `process.nextTick()` inside `_clearMessageState()` so synchronous `result` event listeners complete before `sendMessage()` is called
- Adds an `if (this._destroying) return` guard inside the nextTick callback to suppress spurious sends when `destroy()` fires between scheduling and execution
- Updates `cli-session-pending-queue.test.js` to `await new Promise(r => process.nextTick(r))` between `_clearMessageState()` calls and write-count assertions
- Adds a new test case verifying the destroy suppression path

## Problem

`_clearMessageState()` synchronously called `sendMessage()` during drain. If a `result` event listener also called `sendMessage()`, both would succeed — sending two messages to a child process that expects one. Additionally, calling `destroy()` while a drain was in progress could produce a spurious send.

## Test plan

- [x] All 107 existing `cli-session*.test.js` tests pass (`# pass 107 / # fail 0`)
- [x] New destroy-suppression test added and passing

Closes #2371


blamechris (commented):
Reviewed. 2 Copilot comments addressed and threads resolved.

**What the PR does:** Wraps the pending queue drain in `_clearMessageState()` in a `process.nextTick()` callback. This defers the drain until all synchronous `result` event listeners finish, preventing re-entrancy where both a result listener and the drain could call `sendMessage()` in the same tick.

The `nextTick` callback has three guards: `_destroying`, `_processReady`, and `_pendingQueue.length > 0`. Tests updated to `await new Promise(r => process.nextTick(r))` before asserting drain results.

**Fix matches issue:** Yes. The re-entrancy window where a result listener and the drain compete on the same tick is closed.

**Tests:** Existing FIFO drain tests updated to await nextTick. New test covers the `_destroying` guard. All scenarios accounted for.

**Issues found:**

1. **Missing `_isBusy` check before shift.** The nextTick callback checks `_processReady` and `_pendingQueue.length > 0` but not `_isBusy`. If a result listener fires on the same tick and calls `sendMessage()` directly (setting `_isBusy = true`), the nextTick drain will shift an item from the queue and then call `sendMessage()`, which will see `_isBusy = true` and re-queue it — but the original shift already consumed the item. The message is dropped. In current usage result listeners don't schedule concurrent sends, but the guard is incomplete. Adding `if (this._isBusy) return` before the shift would close this. (Follow-up issue logged.)

2. **Test description mismatch.** "does not drain after destroy() is called" sets `session._destroying = true` directly rather than calling `destroy()`. The test is validating the `_destroying` flag guard specifically, which is valid, but the description implies real teardown is exercised. Should be renamed to "does not drain when _destroying flag is set". (Tracked as cleanup.)

**Regressions:** None. The behavioral change is purely a tick-boundary deferral with equivalent observable behavior for callers.

APPROVE — the nextTick deferral is the right fix. Two Copilot observations logged for follow-up.

PR #2383, opened 2026-03-15
fix(app): add unmount guard in useSpeechRecognition to prevent mic leak

### LINKED ISSUE #2370: fix(app): add unmount guard in useSpeechRecognition to prevent microphone leak
## Problem
`useSpeechRecognition.ts` has an async gap between `getSpeechLang()` and `SpeechModule.start()`. If the component unmounts during that gap, `abort()` is a no-op (session not started), then `start()` fires on unmounted component. Microphone stays active (orange dot visible on iOS).

## Fix
Add mounted ref guard. Check `stopRequestedRef.current` again after `getSpeechLang` returns before calling `start()`.

blamechris:
## Summary

- Adds `mountedRef = useRef(true)` that is set to `false` in the `useEffect` cleanup (alongside the existing `SpeechModule.abort()` call)
- Checks `!mountedRef.current` after `requestPermissionsAsync` and after `getSpeechLang` — if the component unmounted during either async gap, bails out before calling `SpeechModule.start()`
- Also checks `stopRequestedRef.current` after `getSpeechLang` (previously this guard only existed after the permission check)

## Test plan

- [x] All 15 speech recognition tests pass (`npx jest --testPathPattern=speech`)
- [x] 4 new tests added:
  - does not start if component unmounts during permission await
  - does not start if component unmounts during getSpeechLang await
  - does not start if stopListening called during permission await (existed, still passes)
  - does not start if stopListening called during getSpeechLang await
- [x] Full app test suite: 834 tests pass (3 pre-existing xterm-bundle failures unrelated to this change)

Closes #2370


blamechris (commented):
Reviewed. No Copilot comments.

**What the PR does:** Adds a `mountedRef` (initialized to `true`, set to `false` in the `useEffect` cleanup) to `useSpeechRecognition`. Two guard points are added in `startListening()`: after `requestPermissionsAsync()` resolves and after `getSpeechLang()` resolves. Both check `!mountedRef.current || stopRequestedRef.current` before proceeding — if either fires, `SpeechModule.start()` is never called, preventing a mic leak when the component unmounts mid-async-chain.

**Fix matches issue:** Yes. The unmount-during-permission-await and unmount-during-getSpeechLang-await paths both correctly bail out without starting the recognizer.

**Tests:** 3 new tests added:
- `does not start if component unmounts during permission await` — deferred `requestPermissionsAsync`, unmounts before resolution
- `does not start if component unmounts during getSpeechLang await` — deferred `SecureStore.getItemAsync`, unmounts before resolution  
- `does not start if stopListening called during getSpeechLang await` — `stopListening()` fires synchronously inside the mocked getItemAsync

A new `expo-secure-store` mock is added to the test file to allow deferred resolution of `getItemAsync`.

**Analysis:**
- The guard placement is correct — both async suspension points in `startListening()` are covered.
- Removing the now-redundant `if (stopRequestedRef.current) return` between permissions and language fetch and replacing it with a combined `!mountedRef.current || stopRequestedRef.current` check is cleaner and correct.
- The `mountedRef.current = true` in the effect setup (not just cleanup) is a good defensive pattern for strict mode double-invocation.
- The `SecureStore` mock is well-structured — `mockImplementationOnce` correctly scopes the deferred behavior to the specific test.

**No regressions found.** The happy path is unchanged — both guards are skipped when mounted and not stopped.

APPROVE — clean, well-tested fix with correct guard placement at both async suspension points.

PR #2384, opened 2026-03-15
refactor(server): merge triple session_event fan-out in ws-forwarding.js

### LINKED ISSUE #2373: refactor(server): merge triple session_event fan-out in ws-forwarding.js
## Problem
`ws-forwarding.js` registers 3 separate `session_event` listeners on the same emitter: normalizer dispatch, activity feed, and dev-preview. No single dispatch point — hard to trace what a single event does.

## Fix
Merge into one handler that calls normalizer, does activity-feed emit, and calls devPreview.

blamechris:
## Summary

- Consolidates three separate `session_event` listeners in `setupSessionForwarding` into a single handler
- The unified handler dispatches to the normalizer, handles activity feed broadcasts, and calls `devPreview.handleToolResult` — same behavior, one place to trace event flow
- No behavior change; all 59 server tests pass

Closes #2373


blamechris (commented):
Reviewed. The consolidation is correct — three separate `session_event` listeners collapsed into one handler eliminates redundant listener registrations and simplifies the event dispatch path.

**Ordering change:** `session_activity` and `devPreview.handleToolResult` now execute before `broadcastToSession`. This is intentional and safe — the activity feed is a status signal independent of session content, and no client logic depends on the two being ordered relative to each other. The Copilot observation has been addressed in comments.

**Tests:** PR #2385 covers the new unified handler path. No regressions expected.

**Minor follow-up (non-blocking):** A comment in the code noting that activity-feed side effects fire before the per-session broadcast would make the ordering explicit for future readers.

APPROVE

PR #2429, opened 2026-03-16
feat(server): add clearAllPendingPermissions() and fix crash shutdown order (#2405 step 1)

### LINKED ISSUE #2405: refactor(server): unify dual permission lifecycle into single interface
## Problem
Two separate permission subsystems with different object shapes:
- SDK in-process: \`PermissionManager._pendingPermissions\` with separate timer Map
- CLI HTTP hook: \`ws-permissions.js pendingPermissions\` with inline timer

Shutdown ordering between \`sessionManager.destroyAll()\` and \`wsServer.close()\` determines which permissions get resolved. Exception handlers have unguarded ordering.

## Fix
Define a single \`PendingPermission\` interface. Both paths store/resolve through one registry. Shutdown resolves all pending permissions in one call.

## Found By
Guardian (SOLID/DRY audit 2026-03-16)

blamechris:
## Summary

- **Fix crash shutdown order**: `uncaughtException` and `unhandledRejection` handlers in `server-cli.js` previously called `wsServer.close()` before `sessionManager.destroyAll()`. This meant SDK sessions had no chance to auto-deny their pending permissions before the WsServer torn down the HTTP hook path. Correct order: `destroyAll()` first, then `close()`.

- **Add `WsServer.clearAllPendingPermissions()`**: single method that drains both permission subsystems:
  - Legacy HTTP hook permissions via `this._permissions.destroy()` (ws-permissions.js path)
  - SDK in-process permissions via `session._permissions.clearAll()` on every session in `sessionManager._sessions`

- **`close()` delegates to `clearAllPendingPermissions()`**: no duplication — both paths are always covered together on shutdown.

## Test plan

- [ ] `clearAllPendingPermissions()` auto-denies legacy HTTP hook pending permissions
- [ ] `clearAllPendingPermissions()` calls `clearAll()` on each SDK session's PermissionManager
- [ ] Tolerates sessions that have no `_permissions` object (e.g. CLI sessions)
- [ ] `close()` triggers both paths (integration assertion)
- [ ] All 95 existing permission tests continue to pass: `node --test tests/ws-server-permissions.test.js tests/ws-permissions.test.js tests/permission-manager.test.js`

Closes #2405 (step 1)

### DIFF
--- a/packages/server/src/server-cli.js
+++ b/packages/server/src/server-cli.js
@@ -528,8 +528,9 @@ export async function startCliServer(config) {
   process.on('uncaughtException', (err) => {
     console.error('[fatal] Uncaught exception:', err)
     try { wsServer.broadcastShutdown('crash', 0) } catch {}
-    try { wsServer.close() } catch {}
+    // destroyAll() first: SDK sessions auto-deny pending permissions before WsServer closes
     try { sessionManager.destroyAll() } catch {}
+    try { wsServer.close() } catch {}
     try { if (tunnel) tunnel.stop() } catch {}
     try { removeConnectionInfo() } catch {}
     setTimeout(() => process.exit(1), 100)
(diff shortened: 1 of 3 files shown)


blamechris (commented):
Code review complete. The `clearAllPendingPermissions()` method correctly unifies both permission subsystems under a single drain, and the shutdown order fix in `server-cli.js` (destroyAll before close) is exactly right.

Two Copilot findings:

1. **`_questionSessionMap` not cleared in standalone call** — valid observation. The JSDoc should clarify that `_questionSessionMap` is only cleared by `close()`. Evaluate in follow-up whether the standalone path should also clear it.
2. **`server-cli-child.js` not updated** — confirmed: the child entrypoint still calls `_wsServer.close()` before `_sessionManager.destroyAll()`. This is the same bug and will reproduce under the supervisor. Fix `server-cli-child.js` to match the ordering introduced here.

LGTM for the covered paths; the `server-cli-child.js` fix should ship as an immediate follow-up.

PR #2467, opened 2026-03-17
fix(server): fallback to rmSync when git worktree remove fails

### LINKED ISSUE #2460: bug(worktree): worktree cleanup fails if original repo dir is deleted or moved
## Summary

`SessionManager._removeWorktree()` calls:
\`\`\`js
execFileSync(GIT, ['-C', repoDir, 'worktree', 'remove', '--force', worktreePath], ...)
\`\`\`

where `repoDir = entry.worktreeRepoDir` — the original `baseCwd` at session creation time. If the user deletes, renames, or unmounts the original repository directory between session creation and destruction, the cleanup command will fail with a git error and the worktree directory at `~/.chroxy/worktrees/{sessionId}` is never removed, leaking disk space.

**Two separate cleanup paths have this dependency:**

1. `destroySession()` → `_removeWorktree(entry.worktreePath, entry.worktreeRepoDir, ...)`
2. `destroyAll()` → same path

**Fix:**

Use `git -C worktreePath worktree remove --force .` (or just `rm -rf worktreePath`) as a fallback when the repo-based removal fails. The git worktree metadata in the main repo's `.git/worktrees/` directory won't be cleaned up, but `git worktree prune` handles those stale references automatically.

\`\`\`js
_removeWorktree(worktreePath, repoDir, sessionId) {
  // Try removing via the repo (cleans up .git/worktrees metadata)
  try {
    execFileSync(GIT, ['-C', repoDir, 'worktree', 'remove', '--force', worktreePath], {
      stdio: ['pipe', 'pipe', 'pipe'], encoding: 'utf-8',
    })
    return
  } catch (err) {
    log.warn(\`git worktree remove failed for session \${sessionId}, falling back to rmdir: \${err?.stderr?.trim()}\`)
  }
  // Fallback: remove the directory directly (leaves stale .git/worktrees entry, prunable)
  try {
    rmSync(worktreePath, { recursive: true, force: true })
    log.info(\`Removed worktree directory for session \${sessionId}: \${worktreePath}\`)
  } catch (err) {
    log.error(\`Failed to remove worktree directory \${worktreePath}: \${err.message}\`)
  }
}
\`\`\`

**Severity:** Low — requires a repo being deleted while a session is active, but the leak is permanent and accumulates across restarts.

blamechris:
## Summary

- Fixes #2460 — `_removeWorktree()` now falls back to `rmSync(worktreePath, { recursive: true, force: true })` when `git worktree remove` fails (e.g. original repo dir was deleted/moved/unmounted)
- The stale `.git/worktrees` entry left behind is cleaned up automatically by `git worktree prune`
- Added 3 tests covering: fallback when repoDir is missing, normal git removal path, and already-deleted worktree directory

## Test plan

- [x] Existing worktree tests pass (8/8)
- [x] New fallback tests pass (3/3) — including real `renameSync` of the repo dir to simulate the bug
- [x] Main session-manager tests unaffected (79/79)

### DIFF
--- a/packages/server/src/session-manager.js
+++ b/packages/server/src/session-manager.js
@@ -1,6 +1,6 @@
 import { EventEmitter } from 'events'
 import { randomBytes } from 'crypto'
-import { statSync, mkdirSync } from 'fs'
+import { statSync, mkdirSync, rmSync } from 'fs'
 import { join } from 'path'
 import { homedir } from 'os'
 import { execFileSync } from 'child_process'
@@ -397,9 +397,15 @@ export class SessionManager extends EventEmitter {
         encoding: 'utf-8',
       })
       log.info(`Removed worktree for session ${sessionId}: ${worktreePath}`)
+      return
     } catch (err) {
-      const msg = err?.stderr?.trim() || err?.message || String(err)
-      log.error(`Failed to remove worktree for session ${sessionId} at ${worktreePath}: ${msg}`)
+      log.warn(`git worktree remove failed for session ${sessionId}, falling back to rmdir: ${err?.stderr?.trim() || err?.message || String(err)}`)
+    }
+    try {
+      rmSync(worktreePath, { recursive: true, force: true })
+      log.info(`Removed worktree directory for session ${sessionId}: ${worktreePath}`)
+    } catch (err) {
+      log.error(`Failed to remove worktree directory ${worktreePath}: ${err.message}`)
     }
   }
 
--- a/packages/server/src/session-manager.js
+++ b/packages/server/src/session-manager.js
@@ -399,7 +399,7 @@ export class SessionManager extends EventEmitter {
       log.info(`Removed worktree for session ${sessionId}: ${worktreePath}`)
       return
     } catch (err) {
-      log.warn(`git worktree remove failed for session ${sessionId}, falling back to rmdir: ${err?.stderr?.trim() || err?.message || String(err)}`)
+      log.warn(`git worktree remove failed for session ${sessionId}, falling back to direct removal: ${err?.stderr?.trim() || err?.message || String(err)}`)
     }
     try {
       rmSync(worktreePath, { recursive: true, force: true })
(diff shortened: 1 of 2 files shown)


blamechris (commented):
## Agent Review — fix(server): fallback to rmSync when git worktree remove fails

**Verdict: APPROVE** (0 critical, 2 suggestions)

### Summary

Clean, well-scoped fix for #2460. When the original repo directory is deleted/moved/unmounted before session cleanup, `git -C repoDir worktree remove` fails because the git context is gone. The fix adds an `rmSync` fallback that directly removes the worktree directory. Tests are thorough, covering the fallback path, the normal (git) path, and the already-deleted edge case.

### Code Review

**session-manager.js changes (lines 393-408)**

The control flow refactor is correct: try `git worktree remove` → on success `return` → on failure fall through to `rmSync`. The `return` on success (line 400) is the key change that makes the fallback work.

One minor note: the `log.warn` on line 402 says "falling back to rmdir" but the actual fallback is `rmSync` with `{ recursive: true, force: true }`. This is a cosmetic mismatch that could confuse debugging. (See Suggestion 1.)

**Test file changes (lines 237-317)**

Three well-structured tests using a real git repo and an external worktree base (outside the repo dir) to accurately reproduce the production layout (`~/.chroxy/worktrees/`). The `renameSync` trick to simulate repo deletion is clever and realistic. `afterEach` properly cleans up both the moved repo and the external worktree base.

The `gitRepo = movedRepo` reassignment on line 269 ensures afterEach cleanup still works after the rename. This is correct.

### Suggestions

1. **Log message accuracy** (low): Line 402 says "falling back to rmdir" — consider "falling back to rmSync" or "falling back to direct removal" to match the actual operation.

2. **Missing `git worktree prune` call** (low): The PR description mentions that "stale `.git/worktrees` entry left behind is cleaned up automatically by `git worktree prune`" but no prune call is added. This relies on git's built-in auto-pruning on the next `worktree add/list`, which is fine for production. However, if the repo dir was permanently deleted, nothing will ever trigger prune in that repo. This is a non-issue in practice (the stale entry is just a few bytes in `.git/worktrees/`), but worth noting.

### CI Note

Server Tests are failing but the failure is in `ws-server-permissions.test.js` (line 281, "auto mode requires confirmation") — **unrelated to this PR**. The test passes on main; appears to be a flaky race condition in the permission handshake test.

PR #2469, opened 2026-03-17
fix(server,app,dashboard): validate cwd when worktree enabled

### LINKED ISSUE #2463: bug(worktree): worktree toggle enabled even when typed CWD doesn't exist on server
## Summary

In the app's `CreateSessionModal`, the worktree `Switch` is enabled/disabled based on whether the `cwd` text field has any content:

\`\`\`tsx
<Switch
  value={worktree}
  onValueChange={setWorktree}
  disabled={!cwd.trim()}   // ← only checks if field is non-empty
  ...
/>
\`\`\`

And similarly in the dashboard's `CreateSessionModal` (in the Advanced section):
\`\`\`tsx
// worktree checkbox disabled when !cwd
\`\`\`

**The problem:**

The toggle is enabled whenever the CWD field is non-empty, but:
1. The CWD might be a path that doesn't exist on the server
2. The CWD might exist but not be a git repository

In both cases, the user can enable the worktree toggle, submit, and get a server-side error: `WorktreeError: Not a git repository: /some/path`. The error flows back as `session_error`, which is shown correctly. But the UX is poor — the user gets no upfront indication that the path isn't a git repo.

**More subtle issue:**

If the user leaves CWD empty (uses server default) and somehow worktree gets set to `true` (e.g. via a race condition or state management bug), `createSession` sends `{ worktree: true }` without `cwd`. The server then tries to create a worktree at the _default_ cwd, which may or may not be a git repo. This is an unexpected path.

**In the handler:**
\`\`\`js
const worktree = msg.worktree === true ? true : undefined
\`\`\`

If `worktree` is `true` and `cwd` is omitted, the session manager uses `this._defaultCwd` as `baseCwd` for worktree creation. This is undocumented behavior that may silently succeed or fail depending on the server's default directory.

**Fix:**

1. Server: if `worktree: true` is set but no explicit `cwd` is provided, reject with a clear error: `"worktree requires an explicit cwd"`.
2. UI: document in the hint text that the CWD must be a git repository when worktree is enabled.

**Severity:** Low — results in a user-visible error, not silent data loss. But the server-side guard is missing.

blamechris:
Closes #2463

## Summary

- **Server guard**: `create_session` with `worktree: true` but no explicit `cwd` (missing or whitespace-only) now returns a `session_error` with message `"worktree requires an explicit cwd"` instead of silently falling back to the default CWD.
- **UI hint text**: Both the mobile app and dashboard `CreateSessionModal` now show a more explicit hint ("CWD must point to an existing git repository") when the worktree toggle is enabled, replacing the generic description.
- **Tests**: Three new tests covering reject-without-cwd, reject-with-empty-cwd, and allow-with-valid-cwd scenarios.

### DIFF
--- a/packages/app/src/components/CreateSessionModal.tsx
+++ b/packages/app/src/components/CreateSessionModal.tsx
@@ -114,7 +114,11 @@ export function CreateSessionModal({ visible, onClose }: CreateSessionModalProps
             <View style={styles.toggleRow}>
               <View style={styles.toggleLabel}>
                 <Text style={styles.label}>Isolate filesystem (git worktree)</Text>
-                <Text style={styles.toggleHint}>Runs in a separate worktree — requires a git repo CWD</Text>
+                <Text style={styles.toggleHint}>
+                  {worktree
+                    ? 'CWD must point to an existing git repository'
+                    : 'Runs in a separate worktree — requires a git repo CWD'}
+                </Text>
               </View>
               <Switch
                 value={worktree}
--- a/packages/server/src/dashboard-next/src/components/CreateSessionModal.tsx
+++ b/packages/server/src/dashboard-next/src/components/CreateSessionModal.tsx
@@ -501,7 +501,9 @@ export function CreateSessionModal({ open, onClose, onCreate, initialCwd, knownC
               Isolate filesystem (worktree)
             </label>
             <span id="worktree-hint" className="form-hint">
-              Runs in an isolated git worktree — requires a git repo CWD
+              {worktree
+                ? 'CWD must point to an existing git repository'
+                : 'Runs in an isolated git worktree — requires a git repo CWD'}
             </span>
           </div>
         </div>
--- a/packages/server/src/handlers/session-handlers.js
+++ b/packages/server/src/handlers/session-handlers.js
@@ -39,6 +39,11 @@ function handleCreateSession(ws, client, msg, ctx) {
   const permissionMode = rawPermMode && VALID_PERMISSION_MODES.includes(rawPermMode) ? rawPermMode : undefined
   const worktree = msg.worktree === true ? true : undefined
 
+  if (worktree && !cwd) {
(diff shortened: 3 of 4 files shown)


blamechris (commented):
## Agent Review — PR #2469

**Verdict: APPROVE**

### Summary

Clean, focused fix for #2463. Adds a server-side guard rejecting `create_session` when `worktree: true` but no explicit `cwd` is provided, plus UI hint improvements in both mobile app and dashboard.

### Findings

**0 critical, 0 major, 2 suggestions**

#### S1: Defensive client-side guard in mobile app `handleCreate`

The mobile app's `handleCreate` passes `worktree || undefined`, and the `Switch` is disabled when `!cwd.trim()`. However, if a future refactor changes the disabled logic or a state management race sets worktree before cwd, the app would send `{ worktree: true }` without `cwd`. The server guard catches this (that's the whole point of this PR), so this is defense-in-depth only — not blocking.

**Suggestion:** Consider adding `if (worktree && !sessionCwd) return` in `handleCreate` for belt-and-suspenders safety. Low priority.

#### S2: Test uses `homedir()` as CWD — valid but environment-dependent

The "allows worktree with explicit cwd" test uses `homedir()` as a real path that passes `validateCwdWithinHome`. This works in CI (home dirs exist) and locally. The test correctly validates the integration path. No issue, just noting the implicit dependency.

### Architecture

The validation is placed at the right layer — server-side handler, before `createSession` is called. The `cwd` variable at line 34 of `session-handlers.js` already trims and falsy-coerces whitespace-only strings to `undefined`, so `if (worktree && !cwd)` correctly catches both missing and whitespace-only CWD. The existing `validateCwdWithinHome` check at line 42 handles invalid paths post-guard.

### Tests

3 new tests cover the three key scenarios: missing cwd, empty cwd, and valid cwd. The empty-string test (`cwd: '  '`) exercises the trim logic. Coverage is adequate for the change size.

### No follow-up issues needed.

Issue #2473, opened 2026-03-17
feat(server): DockerSdkSession provider using spawnClaudeCodeProcess


blamechris:
## Context

Part of the SdkSession containerization strategy (#2451). Phase 2 of 5.

Depends on: #2472 (spike must validate feasibility)

## Problem

DockerSession wraps CliSession (subprocess), but SdkSession uses the Agent SDK in-process. The SDK's `spawnClaudeCodeProcess` option lets us redirect its internal subprocess to a Docker container — giving us container isolation with all SdkSession capabilities preserved.

## Implementation

New file: `packages/server/src/docker-sdk-session.js` (~150 lines). Extends `SdkSession`:

- **Container lifecycle:** `docker run -d --init --rm` on start, `docker rm -f` on destroy (reuse patterns from `docker-session.js`)
- **Override `sendMessage()`:** Inject `spawnClaudeCodeProcess` into query options
- **The callback:** `spawn('docker', ['exec', '-i', '--workdir', '/workspace', '--env', ...envs, containerId, command, ...args])`
- **Env var allowlist:** Reuse `DockerSession.FORWARDED_ENV_KEYS` pattern
- **Resource limits:** memory, CPU, pids — configurable via constructor

Register in `providers.js` as `docker-sdk`.

### Features preserved vs DockerSession (CLI-based)

| Feature | DockerSession (CLI) | DockerSdkSession (new) |
|---------|-------------------|----------------------|
| Container isolation | Yes | Yes |
| In-process permissions (no HTTP) | No | **Yes** |
| Live model switching (no restart) | No | **Yes** |
| Thinking levels | No | **Yes** |
| Session resume | No | **Yes** |
| Plan mode | Yes | No |

## Files

- **Create:** `packages/server/src/docker-sdk-session.js`
- **Modify:** `packages/server/src/providers.js`

## Acceptance Criteria

- [ ] `DockerSdkSession` selectable via `--provider docker-sdk` or `create_session` WS message
- [ ] Container starts on session creation, cleaned up on destroy
- [ ] All SdkSession capabilities work: permissions, model switch, thinking level, resume, interrupt
- [ ] Resource limits configurable and enforced
- [ ] Env var allowlist prevents secret leakage
- [ ] Handles missing `~/.claude/` paths in container gracefully

## Risk

Medium — Claude Code subprocess may expect host paths (`~/.claude/settings.json`, project `.claude/`) that don't exist inside the container. Need to handle or map these.

Issue #2472, opened 2026-03-17
spike: validate spawnClaudeCodeProcess + docker exec for SdkSession


blamechris:
## Context

Part of the SdkSession containerization strategy (#2451). Phase 1 of 5.

## Discovery

The Agent SDK's `query()` spawns a Claude Code subprocess internally via `ProcessTransport` (sdk.mjs:7583-7810). The `spawnClaudeCodeProcess` option is the SDK's official hook for redirecting this spawn to containers/VMs — its JSDoc explicitly mentions this use case.

The returned object must satisfy the `SpawnedProcess` interface: `stdin` (Writable), `stdout` (Readable), `killed`, `exitCode`, `kill()`, `on('exit')`, `on('error')`. A `docker exec -i` ChildProcess already satisfies this.

## Scope

Create a throwaway spike script (`packages/server/scripts/spike-docker-sdk.js`) that:

1. Starts a Docker container: `docker run -d --init --rm node:22-slim sleep infinity`
2. Calls `query()` with a `spawnClaudeCodeProcess` that spawns `docker exec -i <cid> <command> <args>`
3. Sends a simple prompt and collects the response

## Verify

- [ ] Streaming events arrive correctly
- [ ] `canUseTool` permission callbacks fire through the container boundary
- [ ] `interrupt()` works
- [ ] `setModel()` works
- [ ] `setMaxThinkingTokens()` works
- [ ] Measure latency overhead vs direct spawn
- [ ] Document which features work and which break
- [ ] Go/no-go decision for Phase 2

## Dependencies

None.

## Risk

Low — uses documented first-party SDK API.

blamechris:
Already implemented in PR #2479 (merged). Closing as done.

PR #2480, opened 2026-03-18
feat(server): add DockerSdkSession provider (#2473)

blamechris:
## Summary

- New `DockerSdkSession` provider that extends `SdkSession` to run Claude Code inside an isolated Docker container using the SDK's `spawnClaudeCodeProcess` callback
- Unlike `DockerSession` (CLI-based, extends `CliSession`), this keeps the SDK conversation loop in-process while only containerizing the CLI process
- Registered as `docker-sdk` in `providers.js` alongside the existing `docker` provider

## Implementation Details

**Container lifecycle** (`_startContainer` -> `_setupContainer` -> `_installClaudeCode`):
- `docker run -d --init --rm` with security constraints: `--cap-drop ALL`, `--security-opt no-new-privileges`, `--pids-limit 512`, configurable memory/CPU limits
- Creates non-root user (`chroxy`) — Claude Code refuses `--dangerously-skip-permissions` as root
- Installs `@anthropic-ai/claude-code` globally and discovers the CLI path via `npm prefix -g`
- Mounts project dir as `/workspace`

**SDK integration** (`_augmentQueryOptions` -> `_createSpawnCallback`):
- Adds `_augmentQueryOptions()` hook to `SdkSession` (no-op base, override in subclass) — called just before `query()` to allow option injection without duplicating `sendMessage()`
- The spawn callback remaps host-side `cli.js` paths to container paths (the SDK passes the host's absolute path which doesn't exist inside Docker)
- Env var allowlist: only `ANTHROPIC_API_KEY` and `NODE_ENV` forwarded; `HOME` and `PATH` overridden for container context
- `docker exec -i -u chroxy` satisfies the SDK's `SpawnedProcess` interface natively via Node `ChildProcess`

**Cleanup**: `docker rm -f` on `destroy()`, abort signal wired to `SIGTERM`

## Test plan

- [x] 47 unit tests with mocked Docker (no Docker daemon needed)
- [x] Container lifecycle: start, user setup, CLI install, prefix discovery, error paths
- [x] `spawnClaudeCodeProcess` callback: exec args, env allowlist, path remapping, abort signal, custom user
- [x] `_augmentQueryOptions` hook: injection when container exists, no-op when missing
- [x] `destroy()`: container removal, null guard, super cleanup
- [x] Real import tests: capabilities spread, exported constants
- [x] Existing `sdk-session.test.js` and `docker-session.test.js` still pass (133 tests total)
- [x] Full server test suite passes (2508/2521; 3 pre-existing failures in unrelated `resolveBinary` and `gitStage` tests)

### DIFF
--- a/packages/server/tests/docker-sdk-session.test.js
+++ b/packages/server/tests/docker-sdk-session.test.js
@@ -494,13 +494,17 @@ describe('DockerSdkSession._startContainer() Linux add-host flag', () => {
     session._execFileResults = { run: 'linuxctr\n', exec: '/usr/local\n' }
 
     session._startContainer((err) => {
-      assert.ifError(err)
-      const { args } = session._execFileCalls[0]
-      assert.ok(args.includes('--add-host'), 'missing --add-host flag for linux')
-      assert.ok(args.includes('host.docker.internal:host-gateway'), 'wrong --add-host value')
-
-      Object.defineProperty(process, 'platform', { value: originalPlatform, configurable: true })
-      done()
+      try {
+        assert.ifError(err)
+        const { args } = session._execFileCalls[0]
+        assert.ok(args.includes('--add-host'), 'missing --add-host flag for linux')
+        assert.ok(args.includes('host.docker.internal:host-gateway'), 'wrong --add-host value')
+        done()
+      } catch (e) {
+        done(e)
+      } finally {
+        Object.defineProperty(process, 'platform', { value: originalPlatform, configurable: true })
+      }
     })
   })
 
@@ -512,12 +516,16 @@ describe('DockerSdkSession._startContainer() Linux add-host flag', () => {
     session._execFileResults = { run: 'darwinctr\n', exec: '/usr/local\n' }
 
     session._startContainer((err) => {
-      assert.ifError(err)
-      const { args } = session._execFileCalls[0]
-      assert.ok(!args.includes('--add-host'), '--add-host should not be added on darwin')
-
-      Object.defineProperty(process, 'platform', { value: originalPlatform, configurable: true })
-      done()
+      try {
+        assert.ifError(err)
+        const { args } = session._execFileCalls[0]
+        assert.ok(!args.includes('--add-host'), '--add-host should not be added on darwin')
+        done()
+      } catch (e) {
+        done(e)
+      } finally {
+        Object.defineProperty(process, 'platform', { value: originalPlatform, configurable: true })
+      }
     })
   })
 })


blamechris:
## Code Review: PR #2480

### Summary
Adds a `DockerSdkSession` provider that extends `SdkSession` to run Claude Code inside Docker containers while keeping the SDK conversation loop in-process. Registers as `docker-sdk` alongside the existing CLI-based `docker` provider. Adds a `_augmentQueryOptions()` hook to `SdkSession` for subclass option injection.

### Strengths
- **Clean architecture**: Extends `SdkSession` with minimal surface area — only overrides `start()`, `destroy()`, and the new `_augmentQueryOptions()` hook. No duplicated query/event logic.
- **Security-first container config**: `--cap-drop ALL`, `--security-opt no-new-privileges`, `--pids-limit 512`, non-root user, env var allowlist. Follows the same pattern as `DockerSession`.
- **Good error propagation**: All three async steps (`_startContainer` -> `_setupContainer` -> `_installClaudeCode`) propagate errors back to `start()`, which emits error + calls `destroy()` to clean up orphaned containers.
- **Well-documented**: Class-level JSDoc with lifecycle diagram, spike reference (#2472), and inline comments explaining the host-to-container path remapping.
- **Comprehensive tests**: 47 tests covering lifecycle, env allowlist, path remapping, abort signal, error paths, and real imports. Test harness design (FakeDockerSdkSession) is thorough.
- **Non-breaking base class change**: `_augmentQueryOptions()` added as a no-op to `SdkSession` — zero impact on existing `SdkSession` usage.

### Issues Found

#### Critical (Must Fix)

| docker-sdk-session.js | 129-132 | **`containerUser` is interpolated into a shell command without sanitization.** `useradd -m -s /bin/bash ${user}` — if `opts.containerUser` contains shell metacharacters (e.g., `; rm -rf /`), this is command injection via `bash -c`. While the default `'chroxy'` is safe, the constructor accepts arbitrary user input. | Validate `containerUser` in the constructor: `/^[a-z_][a-z0-9_-]*$/.test(user)` or reject non-alphanumeric values. |

| docker-sdk-session.js | 232 | **Path remapping heuristic `args[0].includes('claude')` is fragile.** A user project path containing "claude" (e.g., `/home/user/claude-docs/script.js`) would trigger false-positive remapping, breaking execution. | Use a more specific check: `args[0].includes('@anthropic-ai/claude-code/cli.js')` or match the SDK's known pattern. DockerSession doesn't have this concern since it controls the CLI path directly. |
| docker-sdk-session.js | 11-14 | **`FORWARDED_ENV_KEYS` is narrower than `DockerSession`'s allowlist.** `DockerSession` forwards `CLAUDE_HEADLESS`, `CLAUDE_CODE_ENABLE_SDK_FILE_CHECKPOINTING`, `CHROXY_PORT`, `CHROXY_HOOK_SECRET`, `CHROXY_PERMISSION_MODE`, `HOME`, `PATH`. While DockerSdkSession doesn't need the CLI-specific vars (SDK handles permissions in-process), the discrepancy should be documented to prevent confusion when choosing between providers. | Add a brief comment noting why this is intentionally smaller (SDK manages permissions in-process, no hook routing needed). |
| docker-sdk-session.js | 229 | **`containerCommand = command` is unused indirection.** The variable is assigned but never modified — could just use `command` directly. | Remove the intermediate variable or add a comment explaining if future transformation is planned. |

#### Nitpicks (Optional)
- **Line 1**: `spawn` is imported from `child_process` but could use `execFile` for the `docker exec` calls too. However, `spawn` is correct here since you need streaming stdio for the SDK's process interface — the import is fine.
- **Test file**: Constants `FORWARDED_ENV_KEYS` and `DEFAULT_CONTAINER_CLI_PATH` are manually mirrored in the test harness rather than imported from the source. The test *does* verify real imports at the bottom (lines 976-998), which validates they match. Consider importing them for the harness too to reduce drift risk.

### Architecture Notes
This fits cleanly into the existing provider architecture. The `_augmentQueryOptions` hook pattern is minimal and idiomatic — it follows the same "override in subclass" pattern as `DockerSession._spawnPersistentProcess`. The provider registration in `providers.js` correctly shares the Docker availability check between both providers.

The key architectural distinction from `DockerSession` is well-motivated: SDK-based sessions benefit from in-process permission handling and live model switching, while the CLI process is the only thing that needs containerization. This gives the best of both worlds.

### Verdict
- [x] **Request Changes** — The `containerUser` command injection (Critical) must be addressed before merge. The path remapping heuristic (Suggestion) is worth tightening but could be a follow-up.

blamechris:
**Follow-up issues created from this review:**

| Harden path remapping heuristic | #2481 | Valid improvement but non-blocking — default SDK behavior only passes Claude Code paths |
| Document env var allowlist differences | #2482 | Documentation clarity, no functional impact |

blamechris:
## Review Comments Addressed

| 1 | Host cwd passed to docker exec --workdir (Copilot) | FIX | `1f0dec7c` |
| 2 | Shell injection via containerUser in bash -c (Copilot) | FIX | `224cf1aa` |
| 3 | process.platform restore not safe on assertion failure (Copilot) | FIX | `24176b7a` |

**Total:** 3 comments addressed
- Fixed: 3 (commit hashes above)
- False positives: 0
- Follow-up issues created: 0
- Existing issues closed: 0

PR #2490, opened 2026-03-18
fix(server): guard against pre-aborted AbortSignal in DockerSdkSession

### LINKED ISSUE #2487: Guard against pre-aborted AbortSignal in DockerSdkSession spawn callback
## Context

Identified during review of PR #2485.

## Description

The spawn callback in \`DockerSdkSession._createSpawnCallback()\` uses \`signal.addEventListener('abort', ...)\` to wire up kill-on-abort. However, if the signal is already aborted when the callback runs, \`addEventListener\` does NOT fire the handler retroactively in Node.js. This means a pre-aborted signal results in an unkillable child process.

PR #2485 added a test documenting this behavior with the comment "callers must not pre-abort," but the source code has no defensive check.

## Acceptance Criteria

- [ ] Add a \`signal.aborted\` check at the top of the spawn callback that calls \`child.kill('SIGTERM')\` immediately (or skips spawning entirely)
- [ ] Add/update the test to verify the guard works

blamechris:
## Summary

- Add `signal.aborted` check at the top of `_createSpawnCallback()` so a pre-aborted AbortSignal immediately kills the child process via `SIGTERM`
- Previously, `addEventListener('abort', ...)` on an already-aborted signal would not fire, leaving the child process unkillable
- Update FakeDockerSdkSession test harness to mirror the guard
- Update test assertion from "pre-aborted signal does NOT kill" to "pre-aborted signal immediately kills"

## Test plan

- [x] Existing test updated: pre-aborted signal now correctly asserts `child.killed === true`
- [x] All 70 docker-sdk-session tests pass
- [x] Full server test suite passes (2576/2576 — 2 pre-existing tunnel integration timeouts unrelated)

Closes #2487

### DIFF
--- a/packages/server/src/docker-sdk-session.js
+++ b/packages/server/src/docker-sdk-session.js
@@ -272,13 +272,17 @@ export class DockerSdkSession extends SdkSession {
         if (text) log.info(`container stderr: ${text}`)
       })
 
-      // Wire up abort signal
+      // Wire up abort signal — guard against pre-aborted signals
       if (signal) {
-        signal.addEventListener('abort', () => {
-          if (!child.killed) {
-            child.kill('SIGTERM')
-          }
-        }, { once: true })
+        if (signal.aborted) {
+          child.kill('SIGTERM')
+        } else {
+          signal.addEventListener('abort', () => {
+            if (!child.killed) {
+              child.kill('SIGTERM')
+            }
+          }, { once: true })
+        }
       }
 
       return child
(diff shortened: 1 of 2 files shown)


blamechris (commented):
**Review Verdict: APPROVE**

Clean, minimal fix. The `signal.aborted` guard correctly handles the edge case where the AbortSignal is already aborted when the spawn callback runs. Without this, `addEventListener('abort', ...)` silently does nothing on pre-aborted signals in Node.js, leaving an unkillable child process.

**Reviewed:**
- Source fix mirrors the standard pattern for AbortSignal guards
- FakeDockerSdkSession harness updated in sync
- Test correctly flipped from documenting broken behavior to asserting fixed behavior
- `child.kill('SIGTERM')` is safe immediately after `spawn()` — Node creates the process handle synchronously

0 critical, 0 suggestions, 0 nits.

PR #2527, opened 2026-03-19
fix(server): validate DevContainer mounts against home directory

### LINKED ISSUE #2512: Validate DevContainer mounts against home directory
## Summary

`createEnvironment()` accepts DevContainer specs with volume mounts that are passed to Docker without validation. A malicious or misconfigured spec could mount:
- The host home directory into the container
- SSH keys, cloud credentials, or `/etc/passwd`
- Paths outside the project directory

**Guardian SEC-02 finding.**

## Proposed Solution

1. Validate mount source paths against an allowlist: project directory only
2. Reject mounts targeting sensitive paths (`~/.ssh`, `~/.aws`, `/etc`)
3. Sanitize `containerEnv` keys to alphanumeric + underscore only
4. Log rejected mounts with the attempted path for debugging

## Complexity

Small — path validation and a deny-list check, ~50 lines.

## References

- Audit: `docs/audit-results/v0.6.0-tech-debt/03-guardian.md` (Section 2, SEC-02)

blamechris:
## Summary

Closes #2512

- Add `_validateMounts()` to filter mount source paths — only mounts inside the project directory are allowed
- Reject mounts targeting sensitive paths (~/.ssh, ~/.aws, ~/.gnupg, /etc)
- Add `_sanitizeContainerEnv()` to filter env var keys to alphanumeric + underscore only
- Rejected mounts/keys are logged for debugging, not thrown

## Test plan

- [x] 9 new tests covering mount validation (inside/outside project dir, sensitive paths, tilde notation, /etc, docker --mount format, logging) and env key sanitization
- [x] All 59 environment-manager tests pass

### DIFF
--- a/packages/server/src/environment-manager.js
+++ b/packages/server/src/environment-manager.js
@@ -2,7 +2,7 @@ import { EventEmitter } from 'events'
 import { randomBytes } from 'crypto'
 import { execFile } from 'child_process'
 import { existsSync, readFileSync, mkdirSync, renameSync, unlinkSync } from 'fs'
-import { dirname, join } from 'path'
+import { dirname, join, resolve } from 'path'
 import { homedir } from 'os'
 import { isWindows, writeFileRestricted } from './platform.js'
 import { createLogger } from './logger.js'
@@ -17,6 +17,7 @@ const DEFAULT_CONTAINER_USER = 'chroxy'
 const DEFAULT_CONTAINER_CLI_PATH = '/usr/local/lib/node_modules/@anthropic-ai/claude-code/cli.js'
 
 const VALID_USERNAME_RE = /^[a-z_][a-z0-9_-]{0,31}$/
+const VALID_ENV_KEY_RE = /^[A-Za-z_][A-Za-z0-9_]*$/
 
 /**
  * Manages persistent container environments that outlive individual sessions.
@@ -741,8 +742,11 @@ export class EnvironmentManager extends EventEmitter {
           ? home
           : source
 
+      // Normalize to resolve any .. segments (path traversal defense)
+      const normalizedSource = resolve(expandedSource)
+
       // Source must be an absolute path inside the project directory
-      if (!expandedSource.startsWith(resolvedCwd) && expandedSource !== cwd) {
+      if (!normalizedSource.startsWith(resolvedCwd) && normalizedSource !== cwd) {
         log.warn(`devcontainer mount rejected (outside project dir): ${source}`)
         continue
       }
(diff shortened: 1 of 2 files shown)


blamechris (commented):
## Code Review: DevContainer Mount Validation (#2527)

Good intent — this closes an important security gap. The env key sanitization is solid, and the overall structure is clean. However, the mount validation has bypass vectors that need to be fixed before merge.

### CRITICAL: Path traversal bypass via `../` sequences

`_validateMounts` uses a raw string prefix comparison (`expandedSource.startsWith(resolvedCwd)`), which is trivially bypassed with `..` components:

```js
// cwd = '/home/user/project'
// mount source = '/home/user/project/subdir/../../../../etc/passwd'
//
// String starts with '/home/user/project/' → passes check
// But path.resolve() yields '/etc/passwd'
```

**Fix:** Call `path.resolve(expandedSource)` before the `startsWith` check. This normalizes `..` segments:

```js
const normalized = resolve(expandedSource)
if (!normalized.startsWith(resolvedCwd) && normalized !== cwd) {
  // reject
}
```

A test for this case should be added as well.

### MEDIUM: Symlink bypass (TOCTOU)

Even with `path.resolve`, a symlink inside the project directory can point outside it:

```
/home/user/project/evil-link → /etc/shadow
```

`resolve()` does not follow symlinks — the path still looks like it's inside cwd. Docker's `-v` flag follows symlinks on the host, so the actual mounted content would be `/etc/shadow`.

**Possible mitigations** (pick one):
- Use `fs.realpathSync` on the source (only works if the path exists at validation time)
- Add `:ro` to all devcontainer mounts to limit damage
- Document this as a known limitation (the devcontainer.json is in the project dir, so the project author is partially trusted)

This is lower severity since the attacker must already control the project directory contents (which is the threat model for devcontainer anyway), but worth acknowledging.

### MINOR: `__proto__` key in `_sanitizeContainerEnv`

The regex `[A-Za-z_][A-Za-z0-9_]*` allows `__proto__`, `constructor`, and `prototype`. When a JSON-parsed object contains `{"__proto__": "evil"}`, it arrives as an own property via `Object.entries`, but assigning `sanitized['__proto__'] = value` modifies the prototype chain rather than creating an own property. The key silently disappears from the output.

Not exploitable (the value is dropped, not injected), but surprising behavior. Fix with `Object.create(null)`:

```js
const sanitized = Object.create(null)
```

### Style nit

The `VALID_ENV_KEY_RE` is declared inside the method body on every call. Consider hoisting it to module scope (next to `VALID_USERNAME_RE`) since it's a constant.

### Tests

9 tests covering the happy and rejection paths — good coverage. Missing cases:
1. **Path traversal test** (e.g., `${cwd}/sub/../../etc/passwd`) — critical, should be added alongside the fix
2. The "logs rejected mounts" test (line 1353) doesn't actually assert on log output — it only verifies the mount isn't in docker args (which is a duplicate of the "rejects mounts outside project dir" test). Either capture log output or rename the test.

### Summary

The env sanitization is good to go. The mount validation needs `path.resolve()` applied before the prefix check — without it, the entire validation is bypassable with a single `/../` in the path. That's the blocking fix.

PR #2533, opened 2026-03-19
fix(app): add Jest timer cleanup to prevent test state leakage

### LINKED ISSUE #2523: fix(app): add Jest timer cleanup to prevent test state leakage
## Problem

App tests using `jest.useFakeTimers()` don't consistently call `jest.useRealTimers()` in `afterEach`, risking timer state leaking between tests. The shared `AsyncStorage` mock in `jest.setup.js` also doesn't reset between tests.

## Affected files
- `packages/app/jest.setup.js` — add `afterEach(() => jest.useRealTimers())` and `beforeEach` to clear AsyncStorage mock
- `packages/app/src/__tests__/HistoryScreenSearchError.test.ts` — uses `jest.useFakeTimers()`
- `packages/app/src/__tests__/android-notification-timer.test.ts` — multiple `advanceTimersByTime` calls
- `packages/app/src/__tests__/LanScanAbort.test.ts` — tight timing assumptions

## Fix
Add global timer cleanup hooks in `jest.setup.js` and verify each timer test has proper restoration.

blamechris:
## Summary

Closes #2523

- Add `jest.setup-after-env.js` with global `afterEach(() => jest.useRealTimers())` to catch leaked timer state
- Add global `beforeEach` to clear AsyncStorage mock backing store between tests
- Expose AsyncStorage backing store via global for cross-file access

## Test plan

- [x] All 880 app tests pass across 72 suites
- [x] Timer tests properly restored after each test
- [x] AsyncStorage state isolated between tests


blamechris (commented):
LGTM. Clean, minimal approach to two real test-pollution vectors.

**jest.setup-after-env.js** — Correct use of `setupFilesAfterEnv` for the `afterEach(jest.useRealTimers)` guard. This must run after the framework installs `afterEach`, so the separate file and Jest config entry are the right pattern.

**AsyncStorage mock reset** — Sharing the backing store via `global.__asyncStorageBackingStore` and clearing keys in `beforeEach` is solid. Using `Object.keys(store).forEach(k => delete store[k])` preserves the object reference so the mock closure keeps working, which is correct (reassigning to `{}` would break it).

No issues found.

PR #2534, opened 2026-03-19
fix(server): replace setTimeout waits with waitFor() in test suite

### LINKED ISSUE #2525: fix(server): replace setTimeout waits with waitFor() in test suite
## Problem

Many server tests use `await new Promise(r => setTimeout(r, N))` with fixed delays for synchronization. This is fragile on slow CI runners.

The `waitFor()` helper in `test-helpers.js` provides event-driven polling (10ms interval, 2000ms timeout) but isn't used consistently.

## Scope
Grep for: \`await new Promise.*setTimeout.*r,\` in \`packages/server/tests/\`

Key offenders:
- `codex-session.test.js` — 16 setTimeout waits
- `concurrency.test.js` — event loop yield waits  
- `environment-handlers.test.js` — event propagation waits

## Fix
Replace fixed sleeps with \`waitFor(() => condition)\` where possible. For negative assertions (proving absence), use explicit longer timeouts (200-500ms) with comments explaining why.

blamechris:
## Summary

Closes #2525

- Replace 16 `setTimeout` waits in `codex-session.test.js` with event-driven `waitFor()` polling
- Replace 3 `setTimeout` waits in `environment-handlers.test.js` with `waitFor()`
- `concurrency.test.js` setTimeout intentionally kept (simulates work duration, not synchronization)

## Test plan

- [x] All 59 tests across modified files pass
- [x] Tests are now resilient to CI timing variations


blamechris (commented):
LGTM. All 38 codex-session tests pass locally.

**waitFor() replacements are correct** — Each replacement polls on the right condition:
- `ready` event tests: `events.length >= 1` or `session.isReady`
- Streaming tests: `events.find(e => e.type === 'result')`, `events.find(e => e.type === 'stream_end')`, etc.
- Error test: `errors.length >= 1`
- State transition: `!session.isRunning`

**Negative assertions still work** — The "ignores turn.completed events with unknown tool_name" test (line 602-614) now collects `result` events separately and waits on `results.length >= 1` to know the turn finished, then asserts zero `stream_start`/`stream_delta`/`error` events. This is better than the old approach — waiting for `result` confirms the shim process completed, making the subsequent negative assertion on stream events meaningful rather than just timing-dependent.

**concurrency.test.js left alone** — Correct call. Those `setTimeout`s simulate actual work duration, not synchronization.

No issues found.

Issue #2511, opened 2026-03-18
Add concurrency guards to EnvironmentManager


blamechris:
## Summary

`environment-manager.js` (743 lines) manages Docker containers for development environments but has zero concurrency guards. Multiple WebSocket clients can trigger parallel operations on the same environment, causing:

- Port allocation conflicts from concurrent `createEnvironment()` calls
- Orphaned containers from create-during-destroy races
- Corrupted state from concurrent snapshot + restore operations
- Data loss from non-atomic restore (old container removed before new one is verified)

**Audit consensus: 4/6 agents flagged this. Guardian rated it highest priority (1.5/5).**

## Proposed Solution

1. Add per-environment mutex using a `Map<string, Promise>` pattern (no external dependencies)
2. All operations on the same environment ID serialize through the mutex
3. `restoreEnvironment()` keeps old container until new one passes health check
4. Add reconciliation on startup: enumerate `chroxy-env-*` containers, stop orphans not in registry

## Complexity

Small — the mutex pattern is ~30 lines. The restore atomicity fix is ~20 lines. Startup reconciliation is ~40 lines.

## References

- Audit: `docs/audit-results/v0.6.0-tech-debt/03-guardian.md` (Section 1)

PR #2536, opened 2026-03-19
fix(server): add error recovery for EnvironmentManager restore and cleanup

### LINKED ISSUE #2519: Add EnvironmentManager error recovery for failed restore()
## Summary

`restoreEnvironment()` in environment-manager.js follows a destructive sequence:
1. Stop current container
2. Remove current container
3. Create new container from snapshot
4. Start new container

If step 3 or 4 fails, the environment is destroyed with no rollback. The user loses both their running environment and the snapshot.

## Proposed Solution

1. Keep the old container until the new one is verified running
2. Create new container from snapshot first (different name)
3. Health check the new container
4. If healthy: stop and remove old container, rename new container
5. If unhealthy: remove new container, old container remains running
6. Log the failure with details for debugging

This makes restore atomic from the user's perspective.

## Also address: fire-and-forget Docker cleanup

Several Docker operations use `.catch(() => {})`:
- `docker stop` during cleanup
- `docker network rm` during teardown

These should at minimum log the error. Failed cleanup accumulates orphaned resources.

## References

- Audit: `docs/audit-results/v0.6.0-tech-debt/03-guardian.md` (Section 3)

blamechris:
## Summary

Closes #2519

- Make `restore()` atomic: start new container before removing old, health check before swap, preserve old container on failure
- Replace silent `.catch(() => {})` in `_composeServices()` with warning logs for failed Docker commands and JSON parse errors
- Overlaps with #2511 (PR #2526) on atomic restore — both implement independently on main; whichever merges second will need minor conflict resolution

## Test plan

- [x] 5 new tests: atomic restore success/failure paths, compose error logging
- [x] All 55 environment-manager tests pass


blamechris (commented):
Overall approach is solid — atomic restore with health check before tearing down the old container is a clear reliability improvement. Tests for the happy path and both failure modes (inspect returns false, inspect throws) are well-structured. A few items to address:

**1. Swallowed inspect error (Copilot also flagged this)**

In the restore health check:
```js
try {
  healthy = await this._inspectContainer(containerId)
} catch {
  // inspect failed — treat as unhealthy
}
```
The error message is lost. Since there's already a `log.warn` call right below for the rollback, include the error there:
```js
} catch (inspectErr) {
  log.warn(`Inspect failed for new container ${containerId.slice(0, 12)}: ${inspectErr.message}`)
}
```

**2. Silent error logging tests don't assert on logging (Copilot also flagged this)**

The two `_composeServices` tests verify the return value is `[]` but never assert that a warning was actually emitted. The test titled "logs warning when docker compose ps fails" has dead code (`warnings`, `origWarn`, `origLog` are declared but unused). Either:
- Inject a mock logger and assert `warn` was called, or
- Remove the dead variables and rename the test to clarify it only verifies graceful degradation (returns `[]` without throwing)

**3. Overlap with PR #2526**

PR #2526 (EnvironmentManager concurrency guards) is still open and also touches `restore()`. Since both change the same method, whichever merges second will have a conflict. The PR body acknowledges this. Just noting: the atomic restore logic here is the cleaner version, so ideally this merges first and #2526 rebases on top.

**4. Test for atomic ordering is well done**

The `callOrder` tracking pattern in "starts new container before removing old one (atomic)" is a good way to verify ordering without fragile timing. The `rm:${args[2]}` format makes it easy to distinguish which container was removed.

PR #2590, opened 2026-03-20
fix(app): preserve response messages during reconnect history replay

### LINKED ISSUE #2588: fix(app): Claude response disappears after reconnect, shows '0 tools used' instead
## Description

When the mobile app is minimized and later reopened, it reconnects to the server. After reconnection, the Claude response message that was previously visible is replaced with "0 tools used" — the actual response text is gone.

## Reproduction

1. Connect to server, send a message, receive Claude's response
2. Minimize the app (home button / swipe up)
3. Wait a moment, reopen the app
4. App reconnects (shows "Reconnecting...")
5. After reconnection: Claude's response text is missing, replaced with "0 tools used" collapse indicator

## Expected Behavior

After reconnection, the full conversation history should be restored including Claude's response text. The server replays history on reconnect — the app should reconstruct the same message bubbles.

## Likely Cause

The history replay may be replacing the streamed response message with the `result` event summary, which only contains tool use metadata (cost, duration, tool count) but not the response text. The response text comes from `stream_delta` events which are NOT replayed — only the final `result` event is in history.

## Files

- `packages/app/src/store/message-handler.ts` — message reconstruction from history replay
- `packages/server/src/ws-history.js` — what events are recorded and replayed
- `packages/server/src/session-manager.js` — `_recordHistory` determines what's stored

## Found During

Manual testing on Samsung S25 Ultra (2026-03-20)

blamechris:
## Summary

- **Root cause**: During reconnect (non-session-switch) history replay, the `message` and `tool_start` handlers in `message-handler.ts` had a blanket `break` that skipped ALL replayed events whenever the cache already contained messages. This caused the server's replayed response text (stored as a `message` event on `stream_end`) to be silently dropped, leaving the UI showing "0 tools used" instead of the actual Claude response.

- **Fix**: Removed the blanket reconnect skip from both handlers. The `message` handler already had content-based dedup logic (matching on type + content + timestamp + tool + options) that correctly prevents duplicates — it just was never reached during reconnect replays. The `tool_start` handler's stable-ID dedup (using server `messageId`) was only applied during session-switch replays; now it applies to all replay scenarios.

- Session-switch replay behavior is unchanged (clears cache and rebuilds from scratch via `fullHistory=true`).

Closes #2588

## Test plan

- [ ] Connect to a session with an existing Claude response
- [ ] Disconnect (e.g., toggle airplane mode or kill tunnel)
- [ ] Reconnect — verify the response text reappears correctly
- [ ] Verify no duplicate messages appear after reconnect
- [ ] Switch sessions — verify full history replay still works correctly
- [ ] Verify tool_use messages are not duplicated on reconnect

### DIFF
--- a/packages/app/src/store/message-handler.ts
+++ b/packages/app/src/store/message-handler.ts
@@ -1076,8 +1076,6 @@ export function handleMessage(raw: unknown, ctxOverride?: ConnectionContext): vo
       // But allow user_input during full history sync (messages came from terminal)
       if (msgType === 'user_input' && !(_ctx.receivingHistoryReplay && _ctx.isSessionSwitchReplay)) break;
       const targetId = (msg.sessionId as string) || get().activeSessionId;
-      // During reconnect replay, skip if app already has messages (cache is fresh)
-      if (_ctx.receivingHistoryReplay && !_ctx.isSessionSwitchReplay && getSessionMessages(targetId).length > 0) break;
       // During any history replay, skip if an equivalent message is already in cache (dedup).
       // This prevents duplicates when the app already received messages via real-time
       // subscription before switching to the session (which triggers history replay).
@@ -1209,12 +1207,10 @@ export function handleMessage(raw: unknown, ctxOverride?: ConnectionContext): vo
 
     case 'tool_start': {
       const targetId = (msg.sessionId as string) || get().activeSessionId;
-      // During reconnect replay, skip if app already has messages (cache is fresh)
-      if (_ctx.receivingHistoryReplay && !_ctx.isSessionSwitchReplay && getSessionMessages(targetId).length > 0) break;
       // Use server messageId as stable identifier for dedup (same ID on live + replay)
       const toolId = (msg.messageId as string) || nextMessageId('tool');
-      // During session-switch replay, skip if tool already in cache (dedup by stable ID)
-      if (_ctx.receivingHistoryReplay && _ctx.isSessionSwitchReplay) {
+      // During any history replay, skip if tool already in cache (dedup by stable ID)
+      if (_ctx.receivingHistoryReplay) {
         const cached = getSessionMessages(targetId);
         if (cached.some((m) => m.id === toolId)) break;
       }


blamechris (commented):
## Agent Review — fix(app): preserve response messages during reconnect history replay

**Verdict: APPROVE** — Clean, correct, minimal fix.

### Analysis

**Root cause identified correctly.** The blanket `break` on lines 1080 and 1213 (pre-fix) short-circuited both `message` and `tool_start` handlers during reconnect replays whenever the cache was non-empty. This silently dropped ALL replayed events — including Claude's response text that the server stores as a `message` event on `stream_end`. The fix removes both blanket breaks and relies on the existing dedup logic instead.

**Dedup correctness after the fix:**

| Handler | Dedup mechanism | Robust? |
|---------|----------------|---------|
| `message` (line 1084) | Content-based: matches on `type + content + timestamp + tool + options` | Yes — exact field comparison prevents duplicates while allowing new messages through |
| `tool_start` (line 1217, was 1219) | ID-based: `cached.some(m => m.id === toolId)` using server `messageId` | Yes — stable server IDs are deterministic across live + replay |
| `stream_start` (line 1126) | ID-based: checks existing message with same `streamId` | Already had its own dedup, unaffected |
| `tool_result` (line 1252) | Idempotent patch by `toolUseId` | Safe — re-patching same data is a no-op |
| `stream_delta` / `stream_end` | No dedup needed — stream content appends are handled by `stream_start` gating | Correct |

**Scope of change:**
- 2 lines removed (blanket breaks)
- 1 condition widened: `tool_start` dedup expanded from `isSessionSwitchReplay`-only to all `receivingHistoryReplay` — correct unification, matches the `message` handler's dedup scope
- No behavioral change for session-switch replays (those clear cache via `fullHistory=true` in `history_replay_start`)

### Findings

**0 critical, 0 major, 1 suggestion**

1. **[SUGGESTION] Missing test coverage for reconnect replay dedup.** The test file has no tests exercising `receivingHistoryReplay` in the `message` or `tool_start` handlers. The `stream_start` reconnect dedup test exists (line 1271) but doesn't cover the handlers this PR modifies. Regression risk is low (the dedup logic is straightforward), but a test that verifies "message events are preserved during reconnect replay when cache has existing messages" would prevent this class of bug from recurring.

PR #2592, opened 2026-03-20
fix(app): filter system messages from chat into separate System tab

### LINKED ISSUE #2587: fix(app): system messages (hook_started/hook_response) show inline in chat instead of separate tab
## Description

System messages like `hook_started` and `hook_response` appear inline in the Chat view on the mobile app. The desktop dashboard correctly filters these into a dedicated System tab, but the mobile app renders all message types together.

## Current Behavior (Mobile)

Chat view shows: user message → System hook_started → System hook_response → Claude response — all interleaved.

## Expected Behavior

System messages should be filtered out of the Chat view and displayed in a separate System tab (matching the desktop dashboard behavior). The dashboard does this by splitting messages into `chatMessages` (excludes `type === 'system'`) and `systemMessages` (only `type === 'system'`).

## Files

- `packages/app/src/screens/SessionScreen.tsx` — renders all messages in ChatView without filtering
- `packages/app/src/components/chat/MessageBubble.tsx` — renders system bubbles inline
- Reference: `packages/dashboard/src/App.tsx` lines 413-422 — dashboard filtering logic

## Implementation Notes

The mobile app needs:
1. A System tab (or section in the collapsed tools menu)
2. Filter `type === 'system'` messages out of the ChatView
3. Display them in the System view with a badge count

## Found During

Manual testing on Samsung S25 Ultra via Cloudflare tunnel (2026-03-20)

blamechris:
## Summary
- Filter `type === 'system'` messages out of the Chat view so hook_started/hook_response events no longer clutter the conversation
- Add a **System** button (terminal icon) to the secondary tools menu with an unread badge count
- When tapped, switches to a dedicated system view rendering only system messages via ChatView
- Unread count tracks per session and resets when entering the System tab (matches the desktop dashboard pattern)

## Changes
- `SessionScreen.tsx`: split `messages` into `chatMessages` + `systemMessages` via `useMemo`, add System button with badge, add `viewMode === 'system'` rendering branch
- `types.ts` / `persistence.ts`: add `'system'` to `viewMode` union type
- `InputBar.tsx`: extend `viewMode` prop type to include `'system'`

Closes #2587

## Test plan
- [ ] Verify system messages (hook_started, hook_response) no longer appear in Chat view
- [ ] Tap the terminal icon in secondary tools to open System tab
- [ ] Verify system messages render correctly in the System tab
- [ ] Verify unread badge appears when new system messages arrive while not on System tab
- [ ] Verify badge resets when switching to System tab
- [ ] Verify split view on tablets still shows chat (without system messages) + terminal
- [ ] Verify search works in System view

### DIFF
--- a/packages/app/src/__tests__/store/persistence.test.ts
+++ b/packages/app/src/__tests__/store/persistence.test.ts
@@ -65,6 +65,12 @@ describe('persistViewMode', () => {
       expect(state.viewMode).toBe(mode);
     }
   });
+
+  it('resets transient view modes to chat on load', async () => {
+    await persistViewMode('system');
+    const state = await loadPersistedState();
+    expect(state.viewMode).toBe('chat');
+  });
 });
 
 // ---------------------------------------------------------------------------
--- a/packages/app/src/components/InputBar.tsx
+++ b/packages/app/src/components/InputBar.tsx
@@ -21,7 +21,7 @@ export interface InputBarProps {
   onToggleEnterMode: () => void;
   isStreaming: boolean;
   claudeReady: boolean;
-  viewMode: 'chat' | 'terminal' | 'files';
+  viewMode: 'chat' | 'terminal' | 'files' | 'system';
   hasTerminal: boolean;
   bottomPadding: number;
   disabled?: boolean;
--- a/packages/app/src/screens/SessionScreen.tsx
+++ b/packages/app/src/screens/SessionScreen.tsx
@@ -146,18 +146,22 @@ export function SessionScreen() {
   // Pick the right message list for the current view
   const messages = viewMode === 'system' ? systemMessages : chatMessages;
 
-  // Track unread system message count per session
-  const lastSeenSystemCountRef = useRef<Map<string | null, number>>(new Map());
-  const lastSeenForSession = lastSeenSystemCountRef.current.get(activeSessionId ?? null) ?? 0;
-  const rawUnreadSystemCount = viewMode === 'system'
-    ? 0
-    : systemMessages.length - lastSeenForSession;
+  // Track unread system message count per session (keyed by sessionId, never null)
+  const lastSeenSystemCountRef = useRef<Map<string, number>>(new Map());
+  const lastSeenForSession =
+    activeSessionId ? lastSeenSystemCountRef.current.get(activeSessionId) ?? 0 : 0;
+  const rawUnreadSystemCount =
+    !activeSessionId || viewMode === 'system'
+      ? 0
+      : systemMessages.length - lastSeenForSession;
   const unreadSystemCount = rawUnreadSystemCount > 0 ? rawUnreadSystemCount : 0;
 
   // Update last-seen count when entering System tab; clamp when messages are trimmed
   useEffect(() => {
+    if (!activeSessionId) return;
+
     const map = lastSeenSystemCountRef.current;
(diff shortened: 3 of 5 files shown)


blamechris (commented):
## Agent Review — fix(app): filter system messages from chat into separate System tab

**Verdict: APPROVE with suggestions**

### Summary

This PR correctly filters `type === 'system'` messages out of the main Chat view and adds a dedicated System tab accessible from the secondary tools menu. The unread badge tracking logic is well-implemented with per-session tracking and clamping for trimmed messages. The approach matches the desktop dashboard pattern.

### Findings

#### Critical (0)
None.

#### Suggestions (5)

**S1. System view ChatView missing several props (search, plan, focus)**
The system view's `<ChatView>` at the new rendering branch omits `searchQuery`, `searchMatchIds`, `currentMatchId`, `isPlanPending`, `planAllowedPrompts`, `onApprovePlan`, and `onFocusInput` props. While plan approval is unlikely to appear in system messages, search IS enabled for the system view (the search button visibility condition at line ~758 was updated to include `viewMode === 'system'`). Since search is enabled but `searchQuery`/`searchMatchIds`/`currentMatchId` aren't passed to the system ChatView, search will appear to work (bar visible, query entered) but matches won't highlight or scroll. Either pass the search props or exclude 'system' from the search button visibility.

**S2. Persistence of 'system' viewMode may cause confusing cold-start**
Adding `'system'` to `VALID_VIEW_MODES` in `persistence.ts` means the app can persist `viewMode: 'system'` to AsyncStorage. On next app launch, the user opens to an empty system view (no messages loaded yet). Consider either: (a) not persisting 'system' (reset to 'chat' on save), or (b) falling back to 'chat' when loading persisted state if the value is 'system'.

**S3. InputBar behavior in system view mode**
The InputBar receives `viewMode: 'system'` but doesn't recognize it for any conditional. `showMicButton`, `showAttachButton`, `showCameraButton` all check `viewMode === 'chat'` — so they'll be hidden (acceptable). Slash command autocomplete checks `viewMode !== 'chat'` so it's also disabled (acceptable). However, the `enterToSend` toggle in SessionScreen uses `viewMode === 'terminal'` to pick the setting key — when in system view, it uses `chatEnterToSend`, which is reasonable. But the InputBar still renders fully — does it make sense to allow typing/sending from the system view? System messages are server-generated; user input from this view would go to Claude as a regular message, which could be confusing. Consider either disabling the InputBar in system view or documenting this as intentional.

**S4. Existing persistence test needs updating**
`packages/app/src/__tests__/store/persistence.test.ts` line 62 iterates `['chat', 'terminal', 'files']` to validate all view modes. This should include `'system'` to match the updated `VALID_VIEW_MODES` array. The test will still pass (it doesn't assert completeness), but it should be comprehensive.

**S5. `activeSessionId` selector moved above actions block**
The `activeSessionId` selector was moved from line ~152 (near `sessions`) to line ~134 (after `lastResultDuration`), because `useMemo` for `chatMessages`/`systemMessages` needs it. This is fine functionally but breaks the established grouping pattern where selectors for state values are in one block and action selectors are in another. The moved `activeSessionId` now sits between the "state values" block and a new filtering block. Consider adding a brief comment section header like `// Message filtering` before the filtering logic to keep the organization clear.

### Issues
No follow-up issues needed — all suggestions are minor and can be addressed in this PR.

PR #2602, opened 2026-03-21
fix(server): pairing ID rotation causes scan failures during slow tunnel setup

### LINKED ISSUE #2599: fix(server): pairing ID rotation causes scan failures during slow tunnel setup
## Description

When the Cloudflare tunnel takes a long time to verify (60-80s for Quick Tunnels), the QR code pairing ID rotates (~55s interval) before the user can scan it. The mobile app gets `invalid_pairing_id` and falls back to direct auth retries, triggering cascading rate limiting.

## Reproduction

1. Start server with Quick Tunnel mode
2. Wait for tunnel verification (can take 60-80s)
3. QR code appears, scan with mobile app
4. If pairing ID rotated between QR display and scan → failure cascade

## Suggested Fix

- Pause pairing ID rotation during tunnel verification
- Or extend the rotation interval after first QR display (give user time to scan)
- Or make the mobile app handle `invalid_pairing_id` gracefully (prompt to re-scan instead of falling back to auth retries)

## Found During

Manual testing with Quick Tunnel on Tauri desktop app (2026-03-21)

blamechris:
## Summary

- When the Cloudflare tunnel takes 60-80s to verify, the pairing ID's 55s auto-refresh timer fires before the user can scan the QR code, causing `invalid_pairing_id` errors and cascading rate limiting
- Added `extendCurrentId(durationMs)` to `PairingManager` that extends the current pairing ID's TTL and reschedules auto-refresh for a grace period (default 3 minutes)
- Called automatically in `displayQr()` after tunnel verification completes, giving users ample time to scan

Closes #2599

## Test plan

- [x] All 5 new tests pass: expiry extension, grace period expiration, auto-refresh delay, no-op when destroyed, activePairings entry update
- [x] All 23 existing pairing tests pass (no regressions)
- [x] Related test files (pairing-refresh-qr, pairing-token-expiry) pass

### DIFF
--- a/packages/server/src/pairing.js
+++ b/packages/server/src/pairing.js
@@ -11,6 +11,7 @@ import { EventEmitter } from 'events'
 import { randomBytes, timingSafeEqual } from 'crypto'
 
 const DEFAULT_TTL_MS = 60_000
+const DEFAULT_GRACE_PERIOD_MS = 3 * 60_000 // 3 minutes after first QR display
 const DEFAULT_SESSION_TOKEN_TTL_MS = 24 * 60 * 60_000 // 24 hours
 const SESSION_TOKEN_BYTES = 32
 const MAX_SESSION_TOKENS = 100
@@ -99,6 +100,33 @@ export class PairingManager extends EventEmitter {
       }
     }
     return false
+  }
+
+  /**
+   * Extend the current pairing ID's validity and pause auto-refresh for a grace period.
+   * Call after displaying the QR code to give the user time to scan before rotation.
+   * @param {number} [durationMs] - Grace period in ms (default 3 minutes)
+   */
+  extendCurrentId(durationMs = DEFAULT_GRACE_PERIOD_MS) {
+    if (this._destroyed || !this._current) return
+
+    // Extend the expiry of the current pairing entry
+    const newExpiry = Date.now() + durationMs
+    this._current.expiresAt = newExpiry
+    const entry = this._activePairings.get(this._current.id)
+    if (entry) entry.expiresAt = newExpiry
+
+    // Reschedule auto-refresh to fire after the grace period
+    if (this._autoRefresh) {
+      if (this._refreshTimer) clearTimeout(this._refreshTimer)
+      this._refreshTimer = setTimeout(() => {
+        if (this._destroyed) return
+        this._generatePairing()
+        this.emit('pairing_refreshed', { pairingId: this._current.id })
+        this._scheduleRefresh()
+      }, durationMs)
+      this._refreshTimer.unref?.()
+    }
   }
 
   /**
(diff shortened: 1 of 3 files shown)


blamechris (commented):
## Agent Review: fix(server): pairing ID rotation causes scan failures during slow tunnel setup

**Verdict: APPROVE with 1 critical fix required**

### Summary

Clean, well-scoped fix for #2599. The `extendCurrentId()` method on `PairingManager` is well-implemented with proper guards (`_destroyed`, `!_current`), consistent dual-state updates (`_current.expiresAt` + `_activePairings` entry), and good test coverage (5 new tests, all existing 18 pass). The approach of extending the existing ID rather than changing the TTL is correct — it preserves the security properties of short-lived pairing IDs while solving the slow-tunnel timing problem.

### Findings

**CRITICAL: `extendCurrentId()` in `displayQr()` creates permanent 3-minute refresh cycle**

The call to `extendCurrentId()` is placed at the end of `displayQr()`. However, `displayQr()` is also called from the `pairing_refreshed` event listener (server-cli.js ~line 496). This creates a cycle:

1. First `displayQr()` call → `extendCurrentId()` reschedules auto-refresh to 3 minutes
2. After 3 minutes, refresh fires → emits `pairing_refreshed` → listener calls `displayQr()` → `extendCurrentId()` reschedules to 3 minutes again
3. Repeat forever

The 55s auto-refresh cadence is permanently replaced by 3 minutes for the entire server lifetime. This weakens the security posture of ephemeral pairing IDs — they now live 3x longer than intended on every rotation, not just the first one.

**Fix:** Move the `extendCurrentId()` call out of `displayQr()` and into the specific call sites where first-display grace is needed — i.e., directly after the `displayQr()` call that follows `waitForTunnel()` (line ~469) and after the external URL `displayQr()` (line ~423). Or add a `_graceApplied` flag so it only fires once.

**Suggestion: Test for the cycle behavior**

Add a test that verifies `extendCurrentId` does not permanently alter the refresh cadence — e.g., after the grace period expires, the next refresh should happen at the normal `ttlMs` interval, not another grace period.

**Suggestion: Consider `displayQr` call sites**

`displayQr()` is called from 6 places: external URL mode, tunnel setup, LAN mode, `pairing_refreshed` listener, `tunnel_recovered` handler, and `token_rotated` handler. Only the initial tunnel setup and external URL cases need the grace period. The current placement affects all 6.

### Tests

- 5 new tests are well-structured and cover the key behaviors
- Tests use appropriate timing (1ms grace for expiry, 5s for timer delay)
- `no-ops when destroyed` test is a good defensive case
- Missing: no test for the interaction between `extendCurrentId` and subsequent `_scheduleRefresh` calls (the cycle issue)

PR #2603, opened 2026-03-21
Surface server startup phases and errors more prominently in the dashboard

### LINKED ISSUE #2600: feat(dashboard): surface server startup phases and errors more prominently
## Description

During server startup (especially with tunnels), the dashboard could better communicate what's happening. Current UX improvements (PR #2598) added tunnel setup spinner and yellow status dot, but more can be done.

## Improvements

1. **Server-side startup events** — Emit `server_status` WS messages for phases (starting, tunnel_verifying, ready) so the dashboard can show precise progress without polling
2. **Startup error surfacing** — When tunnel verification fails or server crashes during startup, show the error prominently in the Console page (not just "Load failed")
3. **Log streaming during startup** — Server logs should stream to the Console page's LogPanel during tunnel verification, not just after full connection
4. **Progress bar** — Show tunnel verification attempt count (e.g., "Verifying tunnel... (attempt 5/20)")

## Current State (v0.6.6)

- Yellow dot + "Setting up tunnel..." appears during tunnel verification (polling-based)
- Console shows spinner + server logs below
- No progress count or estimated time

## Found During

Manual testing with Quick Tunnel (2026-03-21)

blamechris:
## Summary

- Broadcasts structured `server_status` WebSocket events during tunnel verification so the dashboard shows real-time progress (attempt N/M) instead of relying only on HTTP polling
- Adds `onAttempt` callback to `waitForTunnel()` for progress reporting
- Adds `serverPhase` and `tunnelProgress` fields to the dashboard Zustand store
- Updates FooterBar, ConsolePage, and header status dot to display tunnel setup progress
- Keeps the polling-based fallback for backward compatibility with older servers
- Ignores phase-based events in the mobile app handler to avoid spurious "Status update" messages

## Test plan

- [ ] Start server with `--tunnel quick` and verify footer shows "Setting up tunnel... (N/20)" progress
- [ ] Verify ConsolePage shows the same attempt progress with spinner
- [ ] Verify header status dot shows amber/connecting during tunnel setup
- [ ] Verify all transition to "Connected" / green once tunnel is ready
- [ ] Verify ConsolePage auto-fetches connection info when `phase: ready` arrives
- [ ] Verify mobile app does not display spurious system messages for phase events
- [ ] Verify backward compat: dashboard still works with a server that doesn't send phase events (polling fallback)
- [ ] `node --test packages/server/tests/tunnel-check.test.js` passes (6/6)
- [ ] Dashboard type check passes (no new errors)

Closes #2600


blamechris (commented):
## Agent Review — PR #2603

**Verdict: APPROVE** (0 critical, 3 suggestions, 1 nit)

### What this PR does

Adds structured `server_status` WebSocket events with `phase` field (`tunnel_verifying`, `ready`) during tunnel setup, replacing the implicit HTTP-polling-only approach for startup progress. The dashboard store, FooterBar, ConsolePage, and header status dot consume these events to show real-time tunnel verification attempt counts. The mobile app correctly filters out phase-based events to avoid spurious system messages. Backward compatibility is preserved via the existing polling fallback.

### Findings

#### Suggestions

**S1. `tunnel_recovered` handler lacks `onAttempt` callback (consistency gap)**
`server-cli.js:449` — the `tunnel_recovered` event handler still calls `waitForTunnel(newHttpUrl)` without the new `onAttempt` callback. During tunnel recovery, the dashboard won't show attempt progress — it falls back to the old polling behavior. Consider adding the same `onAttempt`/phase broadcasts here for a consistent UX.

**S2. Duplicate initial broadcast before loop body**
`server-cli.js:464` — the initial `tunnel_verifying` broadcast (no `attempt`/`maxAttempts`) fires, then `onAttempt` immediately fires again with `attempt=1, maxAttempts=20`. The store handles this fine (first sets `tunnelProgress: null`, then immediately sets `{ attempt: 1, maxAttempts: 20 }`), but the first broadcast is effectively redundant. Consider removing the pre-`waitForTunnel` broadcast and letting `onAttempt` handle everything, or having `onAttempt` fire before the first fetch so the initial broadcast isn't needed.

**S3. ConsolePage `refetch` effect has stale `connInfo` in dependency array**
`ConsolePage.tsx` — the `useEffect` that triggers on `serverPhase === 'ready'` has `connInfo` in its dependency array and an early-return `if (connInfo)`. This means if `connInfo` is already populated (e.g., from the polling fallback), the effect won't refetch. This is probably the intended behavior (no double-fetch), but worth a comment clarifying that the guard prevents redundant fetches when polling already succeeded.

#### Nit

**N1. Loose equality `==` used for null checks**
`FooterBar.tsx:54` and `App.tsx:921` use `serverPhase == null` (loose equality). The codebase generally uses `=== null` or `!= null` for type narrowing. Consider `serverPhase === null || serverPhase === undefined` or the more idiomatic `serverPhase == null` is fine but inconsistent with the `=== null` pattern elsewhere. Minor — no functional impact since `serverPhase` is typed as `string | null`.

### Architecture & Correctness

- **Store shape**: `serverPhase` and `tunnelProgress` correctly added to both initial state and reset in `auth_ok` and `reset()`. Type definitions in `types.ts` match usage.
- **Message handler**: The `phase`-based `server_status` handling correctly breaks before the legacy path, so phase events don't create spurious system chat messages.
- **Mobile app filter**: Clean early-exit for phase events — prevents mobile chat pollution while keeping legacy `server_status` messages working.
- **`onAttempt` callback**: Clean addition to `waitForTunnel` — optional, backward-compatible, doesn't change existing behavior.
- **No protocol schema changes needed**: `server_status` is untyped in the protocol package (just an enum constant), so the new fields don't break validation.
- **Test coverage**: Existing 6 tunnel-check tests still pass. The `onAttempt` callback is a simple optional callback — low risk, and the integration test plan in the PR covers the dashboard behavior.

### Summary

Clean, well-scoped PR. The WS-driven approach is strictly better than polling-only and the backward compatibility is handled correctly. The suggestions are minor improvements, none blocking.

PR #2604, opened 2026-03-21
test(server): fix flaky encrypted history replay and permission handshake tests

### LINKED ISSUE #2601: test(server): flaky encrypted history replay and permission handshake tests
## Description

Two server tests are persistently flaky in CI, requiring re-runs on most PRs:

1. **`encrypted history replay works end-to-end`** — Timing-sensitive encryption integration test
2. **`acceptEdits mode bypasses confirmation (single-session)`** — Permission handshake race condition

## Impact

Every PR merge requires 1-2 CI re-runs, adding 3-5 minutes per merge cycle. During batch merges this compounds significantly.

## Suggested Fix

- Add retry logic or increase timeouts in the flaky tests
- Or restructure to eliminate timing dependencies
- Or mark as flaky with `{ retry: 2 }` in the test runner config

## Evidence

Observed on PRs #2590, #2593, #2596, #2597 — all required re-runs for the same two tests.

blamechris:
## Summary

Fixes two persistently flaky server tests that required CI re-runs on most PRs.

- **`encrypted history replay works end-to-end`** — The test used raw `_WsServer` (needed for encryption), which registers a log listener that broadcasts `log_entry` messages as extra encrypted envelopes with unpredictable count. It waited for `>= 5` encrypted messages then drained, but messages arrive asynchronously via the event loop so draining too early missed later arrivals. Fix: suppress `log_entry` broadcasts with `setLogListener(null)` and poll-decrypt until the `history_replay_end` sentinel is found.

- **`acceptEdits mode bypasses confirmation (single-session)`** — After clearing the messages array, post-auth messages (including `permission_mode_changed` with default `approve`) could race in and be found by the type-only `waitForMessage` call before the actual `acceptEdits` response arrived. Fix: use `waitForMessageMatch` with a predicate that checks both type and mode, matching the pattern already used by the adjacent auto-mode confirmation test.

Both tests verified stable across 5 consecutive runs locally.

Closes #2601


blamechris (commented):
## Agent Review — APPROVE

**Verdict: Clean.** Both fixes correctly address the root causes of flakiness with minimal, well-targeted changes.

### Findings

**0 critical, 0 suggestions, 0 follow-ups**

### Analysis

**Fix 1: Encrypted history replay** (`ws-server.test.js`)
- Root cause correctly identified: raw `_WsServer` (not the test wrapper) registers a log listener that broadcasts `log_entry` messages as extra encrypted envelopes with unpredictable count
- Previous approach: wait for `>= 5` encrypted messages then `drainEncryptedMessages()` — brittle because (a) the count 5 was a magic number that could vary, and (b) draining after the count was met could miss late-arriving messages
- Fix: suppress log broadcasts with `setLogListener(null)` + poll-decrypt incrementally until the `history_replay_end` sentinel is found
- The inline decrypt logic exactly matches `drainEncryptedMessages()` — correct duplication since it runs inside the polling loop
- `drainEncryptedMessages` retained for other tests — no dead code introduced

**Fix 2: Permission handshake** (`ws-server-permissions.test.js`)
- Root cause correctly identified: after `messages.length = 0`, the default `permission_mode_changed` with `mode: 'approve'` can race in before the `acceptEdits` response
- Fix: use `waitForMessageMatch` with a predicate checking both type and mode — consistent with the adjacent auto-mode test pattern (line 344-349)
- The non-auto `approve` mode test (line 377) correctly does NOT need this fix since it's testing the same mode as the default broadcast

Both fixes are minimal, well-commented, and verified across 5 consecutive runs per the PR description.

PR #2723, opened 2026-04-06
security(server): fix rate limiter to use CF-Connecting-IP through Cloudflare tunnel

### LINKED ISSUE #2688: security(server): rate limiter ineffective through Cloudflare tunnel — use CF-Connecting-IP
## Summary

`packages/server/src/rate-limiter.js` applies per-IP rate limiting to WebSocket connections. However, Cloudflare tunnels present all remote clients to the server with the same source IP (`127.0.0.1` — the cloudflared loopback). The rate limiter is therefore non-functional for the primary deployment mode.

## Fix

Use the `CF-Connecting-IP` header forwarded by Cloudflare to identify the real client IP:

```js
// In ws-server.js upgrade handler
const clientIp = req.headers['cf-connecting-ip'] || req.socket.remoteAddress
rateLimiter.check(clientIp)
```

Alternatively, rate-limit on authenticated session identity (token hash) rather than IP, which works regardless of proxy topology.

## Acceptance Criteria

- [ ] Rate limiter uses `CF-Connecting-IP` when available, falls back to socket IP
- [ ] Add test: multiple connections from same CF IP are rate-limited correctly
- [ ] Document that rate limiting requires tunnel mode to be effective against remote abuse

## Severity

**Medium** — abuse/DoS protection absent for tunneled connections

## Audit Reference

v0.6.8 full codebase audit (2026-04-05) — found by Adversary

blamechris:
## Summary

- Rate limiter was keying off `client.id` (a per-connection UUID), meaning every connection got its own independent bucket regardless of origin IP
- All Cloudflare tunnel traffic has `req.socket.remoteAddress = 127.0.0.1` (cloudflared loopback), so IP-based rate limiting was completely ineffective through the tunnel
- The server already extracted the real IP into `client.ip` (preferring `CF-Connecting-IP`); this PR makes rate limit checks use that field instead
- Exports a `getClientIp()` helper from `rate-limiter.js` for reuse and testability

## Changes

- `packages/server/src/rate-limiter.js`: export `getClientIp(req)` helper — prefers `CF-Connecting-IP`, falls back to `X-Forwarded-For`, then socket address
- `packages/server/src/ws-server.js`: rate limiter `check()` and `remove()` now use `client.ip` instead of `client.id`
- `packages/server/tests/rate-limiter.test.js`: 5 new unit tests for `getClientIp()` covering all fallback paths
- `packages/server/tests/ws-server-rate-limit.test.js`: 2 new integration tests — shared bucket for same CF IP across multiple connections, independent bucket for different CF IPs

## Test plan

- [x] `node --test packages/server/tests/rate-limiter.test.js` — 12/12 pass
- [x] `node --test packages/server/tests/ws-server-rate-limit.test.js` — 7/7 pass
- [x] Full server test suite passes (pre-existing failures in `service` and tunnel integration tests are unrelated to this change)

Closes #2688

### DIFF
--- a/packages/server/src/rate-limiter.js
+++ b/packages/server/src/rate-limiter.js
@@ -3,6 +3,20 @@
  * Tracks per-client message timestamps to enforce rate limits.
  */
 
+// Loopback addresses used by cloudflared and local dev connections
+const LOOPBACK_ADDRESSES = new Set(['127.0.0.1', '::1', '::ffff:127.0.0.1'])
+
+/**
+ * Normalize a Node.js HTTP header value to a string.
+ * Node can represent repeated headers as string[] — take the first entry.
+ * @param {string|string[]|undefined} val
+ * @returns {string|undefined}
+ */
+function normalizeHeader(val) {
+  if (Array.isArray(val)) return val[0]
+  return val
+}
+
 /**
  * Extract the real client IP from an HTTP request.
  * Prefers the CF-Connecting-IP header set by Cloudflare tunnels, which carry
@@ -18,12 +32,33 @@
  * @returns {string}
  */
 export function getClientIp(req) {
-  return (
-    req.headers['cf-connecting-ip'] ||
-    req.headers['x-forwarded-for']?.split(',')[0]?.trim() ||
-    req.socket?.remoteAddress ||
-    'unknown'
-  )
+  const headers = req?.headers || {}
+  const cfIp = normalizeHeader(headers['cf-connecting-ip'])
+  const xffRaw = normalizeHeader(headers['x-forwarded-for'])
+  const xffIp = xffRaw?.split(',')[0]?.trim()
+  return cfIp || xffIp || req?.socket?.remoteAddress || 'unknown'
+}
+
+/**
+ * Return the key to use for rate limiting a connection.
+ *
+ * Proxy headers (CF-Connecting-IP, X-Forwarded-For) are only trusted when the
+ * TCP peer is a loopback address — i.e. the request arrived via the local
+ * cloudflared process or a trusted reverse proxy. For direct connections the
+ * socket address is used so an attacker cannot spoof the header to share (or
+ * exhaust) another IP's rate-limit bucket.
+ *
+ * @param {string} socketIp - req.socket.remoteAddress (kernel-supplied, not spoofable)
+ * @param {object} req - Node.js IncomingMessage (for proxy header extraction)
+ * @returns {string}
+ */
+export function getRateLimitKey(socketIp, req) {
+  if (LOOPBACK_ADDRESSES.has(socketIp)) {
+    // Connection came through a local proxy — trust the forwarded IP
+    return getClientIp(req)
+  }
+  // Direct connection — use the kernel-supplied address to prevent header spoofing
+  return socketIp || 'unknown'
 }
 
 const DEFAULT_WINDOW_MS = 60_000   // 1 minute window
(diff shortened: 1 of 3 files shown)


blamechris:
## Code Review: PR #2723

### Summary

Fixes a real security gap: the rate limiter was tracking per-connection UUIDs (`client.id`) instead of per-IP, which meant every new WebSocket connection got its own fresh bucket. Through the Cloudflare tunnel this was entirely ineffective — the `cf-connecting-ip` header was already extracted into `client.ip` but never used for rate limiting. The core fix is small and correct; the tests are thorough. Two issues need addressing before merge.

### Strengths

- The root cause analysis is accurate and the fix is surgical
- The security note in `getClientIp()` JSDoc explicitly separates rate-limiting use from security-sensitive use — important because the comment in `ws-server.js` already warns about spoofing for auth/encryption bypass
- Integration tests prove cross-connection IP sharing and independent-bucket isolation with real WebSocket connections, not mocks
- `createClient()` signature is backward-compatible (default `headers = {}`)
- `getClientIp` unit tests cover all five fallback paths cleanly

### Issues Found

#### Critical (Must Fix)

| `ws-server.js` | 779–782 | **`remove(client.ip)` on disconnect erases the shared IP bucket**, resetting rate limit state for all concurrent/future connections from that IP. With per-UUID keys this was safe (each connection had its own entry). With per-IP keys, the first disconnect wipes the shared record — a client could exhaust its limit, disconnect and reconnect to get a fresh bucket. The sliding window's natural expiry is the correct cleanup mechanism; explicit `remove()` on disconnect should be dropped when keying by IP. | Remove the `if (client?.ip)` block that calls `_rateLimiter.remove` and `_permissionRateLimiter.remove` on disconnect. The window-based pruning in `check()` handles memory cleanup automatically. |
| `ws-server.js` | 678–680 | **`getClientIp()` is exported but never imported**; `ws-server.js` still contains its own inline duplicate of the same logic. The PR's stated goal was to export a reusable helper, but the inline extraction isn't replaced. This is a DRY gap, not just cosmetic — if the fallback order ever diverges, they'll behave differently. | Import `getClientIp` from `./rate-limiter.js` and replace the inline extraction at line 678. |

| `ws-server-rate-limit.test.js` | 334, 372 | `msgs1` and `msgsA` are destructured but never read — replace with `_msgs1` or use `{ ws: ws1 }` alone. | ESLint unused-var will flag these; keeps the diff clean. |

#### Nitpicks (Optional)

- The `_clients` Map comment in `RateLimiter` still says `clientId ->` — could say `key (IP or clientId) ->` now that the key is no longer always a UUID, but this is minor.

None — both issues are fixes within scope of this PR.

### Architecture Notes

The change correctly keeps IP extraction in the connection handler (where `req` is available) and passes the resolved IP through the client object — this is the right layering. The `getClientIp` export belongs in `rate-limiter.js` because that's the consumer, though it would also make sense as a standalone `http-utils.js`. Current placement is fine.

The `remove()` regression is the only thing that concerns me architecturally: the rate limiter's sliding window is already its own garbage collector (via timestamp pruning in `check()`). The `remove()` method was designed for connection-scoped keys where you want to free memory on disconnect. For IP-scoped keys you want the window to drain naturally — removing on disconnect is the wrong lifecycle.

### Verdict

- [x] **Request Changes** — The `remove()` regression actively undermines the fix (reconnect-to-reset bypass), and the inline duplication means `getClientIp` isn't actually used in production. Both are one-liner fixes.

PR #2874, opened 2026-04-19
chore(server): bump Codex default model to gpt-5.4

blamechris:
## Summary
- `CodexSession` default model was `o4-mini` — stale.
- `codex-cli 0.101.0` advertises a migration path `gpt-5.2-codex → gpt-5.3-codex → gpt-5.4`, so `gpt-5.4` is the current recommended Codex model.
- Users can still override via `setModel()` / `--model` as before; only the fallback default changed.

## Test plan
- [x] `node --test packages/server/tests/codex-session.test.js` — 40/40 passing
- [ ] Manual smoke: start server with `--provider codex`, confirm `ready` event carries `gpt-5.4`

### DIFF
--- a/packages/server/src/codex-session.js
+++ b/packages/server/src/codex-session.js
@@ -29,7 +29,7 @@ const log = createLogger('codex')
  *   error        { message }
  */
 
-const DEFAULT_MODEL = 'o4-mini'
+const DEFAULT_MODEL = 'gpt-5.4'
 
 const CODEX = resolveBinary('codex', [
   '/opt/homebrew/bin/codex',
--- a/packages/server/tests/codex-session.test.js
+++ b/packages/server/tests/codex-session.test.js
@@ -195,7 +195,7 @@ describe('CodexSession', () => {
   describe('constructor', () => {
     it('uses default model when none supplied', () => {
       const session = new CodexSession({ cwd: '/tmp' })
-      assert.equal(session.model, 'o4-mini')
+      assert.equal(session.model, 'gpt-5.4')
     })
 
     it('accepts a model override', () => {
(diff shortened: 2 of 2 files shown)


blamechris:
## Code Review: PR #2874

### Summary
Small, well-scoped change bumping `CodexSession`'s `DEFAULT_MODEL` from the stale `o4-mini` to `gpt-5.4`, with the matching test constants updated. Total surface area is 3 line changes across 2 files. Override paths (`setModel()`, `--model`, constructor `{ model }`) are untouched, so users with an explicit model continue to get what they asked for.

### Strengths
- Minimal, surgical diff — only the fallback default moved; no behavior change for configured users.
- Test file kept in sync with the implementation; 40/40 tests still pass per PR description.
- `gpt-5.4` is verifiable as the current Codex recommendation: codex-cli 0.101.0 ships an explicit migration map `gpt-5.2-codex → gpt-5.3-codex → gpt-5.4` in its config, matching what's in the PR.
- Style matches the rest of `packages/server/`: ES modules, no semicolons, single quotes.
- Parallels `gemini-session.js` (`DEFAULT_MODEL = 'gemini-2.5-pro'`) in structure — the per-provider constant pattern is consistent.
- Zero Attribution Policy: clean — no AI/Claude references in title, body, or diff.

### Issues Found

#### Critical (Must Fix)
_None._

_None._ For a 3-line default-string bump there's no meaningful refactor to request.

#### Nitpicks (Optional)
- The `setModel('o3')` busy-guard test (`codex-session.test.js:305`) uses `'o3'` purely as a distinct string to prove the guard rejects a change while busy. It still works as an override test because the assertion is that the model _does not_ change, regardless of whether `'o3'` is current — so this remains meaningful. No action needed.

_None._

### Architecture Notes
The default only affects users who (a) use `--provider codex` and (b) never call `setModel()` or pass `--model`. Anyone on a previous release who had `o4-mini` implicitly will get `gpt-5.4` on next start; anyone with an explicit config is unaffected. Per the codex-cli 0.101.0 migration map, this is the same direction upstream is pushing users, so the behavior matches expectations rather than contradicting them. Test plan's manual smoke (start with `--provider codex`, check `ready` event carries `gpt-5.4`) is the right final gate.

### Verdict
**Approve** — ship it.

PR #2875, opened 2026-04-19
chore: gitignore .repo-memory local MCP cache

blamechris:
## Summary
- The repo-memory MCP writes SQLite cache files (`cache.db`, `cache.db-shm`, `cache.db-wal`) into `.repo-memory/` on every indexing pass. They were tracked, so every working session surfaced them as dirty working-tree changes.
- Add `.repo-memory/` to `.gitignore`.
- `git rm --cached` the three cache files.
- The directory is purely local cache state — no other files live there.

## Test plan
- [x] `git status` after commit is clean even while repo-memory is actively indexing
- [x] `.repo-memory/` does not reappear in future `git status` runs


blamechris:
## Code Review: PR #2875

### Summary
Adds `.repo-memory/` to `.gitignore` and untracks the three SQLite cache files (`cache.db`, `cache.db-shm`, `cache.db-wal`) that the repo-memory MCP rewrites on every indexing pass. Purely a repo hygiene change — zero runtime/code impact.

### Strengths
- Clean, minimal diff (3 deletions + 1 `.gitignore` entry with explanatory comment).
- Placement under the existing "Misc" section fits the file's structure.
- Explanatory comment (`# repo-memory MCP local cache (SQLite db + WAL/SHM)`) tells future readers why the directory is ignored.
- Directory-level ignore is the right call — it catches future SQLite variants (`*.db-journal`, rollover WALs, etc.) and any new cache artifacts the MCP may emit, without requiring another gitignore update.
- `git rm --cached` (as opposed to `rm`) preserves the local cache files, so nobody's indexing state is clobbered on pull.
- All 12 CI checks green.

### Issues Found

#### Critical (Must Fix)
None.

None.

#### Nitpicks (Optional)
- None. The change is appropriately scoped.

None.

### Architecture Notes
- Verified no `.github/` workflow references `.repo-memory/`, and no tests/scripts depend on those cache files existing in the tree.
- Directory-ignore means any future file written into `.repo-memory/` (by the MCP or otherwise) is silently ignored. That's the intended behavior here — the directory is MCP-owned local state — but worth noting: if the MCP ever starts emitting something a developer *should* commit (config, schema, etc.), the gitignore would need a negation pattern. Low risk; the MCP writes only cache data today.
- Fixes the persistent dirty working-tree nuisance called out in the PR description and matches prior practice for other local caches (`.cache/`, `tmp/`).

### Verdict
**Approve.**

Issue #2830, opened 2026-04-18
models.js / sdk-session.js: add debug logging for silent failure paths (saveCache + modelUsage contract drift)


blamechris:
## Context

PR #2828 bundled cleanup of \`models.js\`. Two silent-failure paths surfaced during review:

### 1. \`saveCache\` swallows all write errors

\`packages/server/src/models.js:239-242\`:

\`\`\`js
} catch {
  try { unlinkSync(tmpPath) } catch {}
  return false
}
\`\`\`

EACCES, ENOSPC, EROFS, stale-NFS handles — none are ever surfaced. \"Best-effort caching\" is a defensible design, but real disk problems are currently invisible.

### 2. \`sdk-session.js\` \`msg.modelUsage\` contract is assumed silently

\`packages/server/src/sdk-session.js:362-373\`: If the SDK ever ships a response without a usable \`contextWindow\` (schema drift, renamed field, etc.), the loop no-ops forever and the cached context window stays at whatever the static heuristic guessed.

## Proposed fix

Add a debug-level log in both paths:

\`\`\`js
// saveCache catch
log.debug(\`saveModelsCache failed: \${err.message}\`)

// sdk-session result handler
if (msg.modelUsage && typeof msg.modelUsage === 'object') {
  let sawContextWindow = false
  for (const [modelId, usage] of Object.entries(msg.modelUsage)) {
    if (usage && typeof usage.contextWindow === 'number') {
      sawContextWindow = true
      if (updateContextWindow(modelId, usage.contextWindow)) { /* ... */ }
    }
  }
  if (!sawContextWindow) log.debug('SDK modelUsage present but no contextWindow fields — contract drift?')
}
\`\`\`

Keeps best-effort behavior, adds diagnostic breadcrumbs.

## Scope

- \`packages/server/src/models.js\` — \`saveCache\` catch block
- \`packages/server/src/sdk-session.js\` — result handler modelUsage loop

## Labels

- enhancement
- from-review

Follow-up from PR #2828 review.

PR #2879, opened 2026-04-19
feat(server): log modelUsage contract drift in sdk-session

### LINKED ISSUE #2873: sdk-session.js: add debug log for modelUsage contract drift (remainder of #2830)
## Context

Identified during review of PR #2864 — Copilot flagged scope drift vs. parent issue #2830.

Issue #2830 has two silent-failure paths:

1. ✅ \`models.js::saveCache\` catch block — done in PR #2864.
2. ❌ \`sdk-session.js\` \`msg.modelUsage\` handler — still silent.

## Description

Per #2830, \`packages/server/src/sdk-session.js:362-373\` silently no-ops if the SDK ships a \`modelUsage\` response without a usable \`contextWindow\`:

\`\`\`js
if (msg.modelUsage && typeof msg.modelUsage === 'object') {
  for (const [modelId, usage] of Object.entries(msg.modelUsage)) {
    if (usage && typeof usage.contextWindow === 'number') {
      if (updateContextWindow(modelId, usage.contextWindow)) { /* ... */ }
    }
  }
}
\`\`\`

If the SDK renames \`contextWindow\` or the usage shape changes, the cached context window silently stays at whatever the static heuristic guessed — no log, no signal.

## Proposed fix

\`\`\`js
if (msg.modelUsage && typeof msg.modelUsage === 'object') {
  let sawContextWindow = false
  for (const [modelId, usage] of Object.entries(msg.modelUsage)) {
    if (usage && typeof usage.contextWindow === 'number') {
      sawContextWindow = true
      // ... existing update
    }
  }
  if (!sawContextWindow && Object.keys(msg.modelUsage).length > 0) {
    log.debug('SDK modelUsage present but no contextWindow fields — possible contract drift')
  }
}
\`\`\`

## Scope

- \`packages/server/src/sdk-session.js\` — result handler modelUsage loop
- Test covering the no-contextWindow-present debug log

## Labels

- from-review

blamechris:
Closes #2873

## Summary

Adds a debug-level log when the SDK emits `modelUsage` entries but none carry a numeric `contextWindow`. Completes the second silent-failure path from #2830 (the first was addressed in #2864).

Previously, if the SDK renamed or removed `contextWindow`, the cached context window silently stayed at whatever the static heuristic guessed — no log, no signal. Now:

\`\`\`js
if (!sawContextWindow && modelIds.length > 0) {
  log.debug(\`modelUsage contract drift: expected numeric contextWindow; received modelIds=... sampleKeys=...\`)
}
\`\`\`

Debug-level so it's off in production logs but available with \`LOG_LEVEL=debug\` for diagnosis.

## Changes

- \`packages/server/src/sdk-session.js\` — track \`sawContextWindow\` across the entries loop, log sampled modelIds + keys from the first entry when all entries lack it
- \`packages/server/tests/sdk-session.test.js\` — three tests: drift detected, happy-path silent, empty-modelUsage silent

## Test plan

- [x] \`node --test packages/server/tests/sdk-session.test.js\` — 69/69 pass (3 new)
- [x] \`node --test packages/server/tests/sdk-session-question.test.js packages/server/tests/sdk-session-timeout-pause.test.js\` — 10/10 pass
- [x] \`node --test packages/server/tests/models.test.js\` — 33/33 pass

### DIFF
--- a/packages/server/src/sdk-session.js
+++ b/packages/server/src/sdk-session.js
@@ -380,12 +380,26 @@ export class SdkSession extends BaseSession {
             // a value actually changed to avoid thrashy writes / UI churn.
             let contextWindowChanged = false
             if (msg.modelUsage && typeof msg.modelUsage === 'object') {
+              let sawContextWindow = false
+              const modelIds = Object.keys(msg.modelUsage)
               for (const [modelId, usage] of Object.entries(msg.modelUsage)) {
                 if (usage && typeof usage.contextWindow === 'number') {
+                  sawContextWindow = true
                   if (updateContextWindow(modelId, usage.contextWindow)) {
                     contextWindowChanged = true
                   }
                 }
+              }
+              // Drift signal: SDK emitted modelUsage entries but none carried a
+              // numeric contextWindow. Likely means the field was renamed or
+              // removed upstream. Log a redacted sample so a future regression
+              // is diagnosable without flooding info-level output.
+              if (!sawContextWindow && modelIds.length > 0) {
+                const sampleId = modelIds[0]
+                const sampleKeys = Object.keys(msg.modelUsage[sampleId] || {})
+                log.debug(
+                  `modelUsage contract drift: expected numeric contextWindow; received modelIds=${JSON.stringify(modelIds)} sampleKeys=${JSON.stringify(sampleKeys)}`
+                )
               }
             }
             if (contextWindowChanged) {
--- a/packages/server/tests/sdk-session.test.js
+++ b/packages/server/tests/sdk-session.test.js
@@ -814,6 +814,106 @@ describe('SdkSession', () => {
     })
   })
 
+  // -- modelUsage contract drift --
+
+  describe('modelUsage contract drift', () => {
+    afterEach(() => {
+      resetModels()
+    })
+
+    async function runWithModelUsage(modelUsage) {
+      const s = createSession()
+      s._processReady = true
+      s._callQuery = () => {
+        return (async function* () {
+          yield {
+            type: 'result',
+            session_id: 'drift-test',
+            total_cost_usd: 0,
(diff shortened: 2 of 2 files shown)


blamechris (commented):
## Chroxy Inspector review — PR #2879

**Verdict: APPROVE** (with one minor follow-up).

Tight, proportional change for a debug-log addition. CI green (12/12), Copilot reviewed no comments.

### What's good
- **Useful diagnostic payload.** Logging `modelIds` + `sampleKeys` from the first entry is exactly what's needed to spot a rename (e.g., `contextWindow` → `context_window`) without flooding output.
- **Debug level is correct.** Off in production, on with `LOG_LEVEL=debug` — matches the pattern used in #2864.
- **Style compliant.** ES modules, no semicolons, single quotes. Zero Attribution respected.
- **Test coverage mirrors the three realistic shapes** (drift / happy / empty) and cleans up log state in `finally`.

### Minor nits (non-blocking)

1. **Partial-drift blind spot.** The trigger fires only when *every* entry lacks `contextWindow`. If the SDK ever ships a mixed response (some models have it, some don't) the drift goes silent. Not realistic today — the SDK emits a uniform shape — but worth a tracked follow-up. Filing an issue.
2. **Log level restore is hardcoded to `info`.** All three tests do `setLogLevel('info')` in `finally`, which assumes the pre-test level was `info`. Capturing the prior level before setting to `debug` would be more robust. Harmless in practice (tests aren't run at debug by default).
3. **Redundant dynamic import.** `await import('../src/logger.js')` is repeated in each of the three new tests. Since the rest of the file uses top-level static imports, hoisting these (`import { addLogListener, removeLogListener, setLogLevel } from '../src/logger.js'`) would match house style. Cosmetic.

No blockers. Nice completion of the second silent-failure path from #2830.

Issue #2849, opened 2026-04-19
Old cached dashboard + new server: 'Tunnel warming' appears as chat message (mixed-version UX)


blamechris:
## Context

Identified during review of PR #2845 (#2836 tunnel_warming rename).

## Description

After #2845 merges, the server emits `phase: 'tunnel_warming'` but a dashboard client running an older bundle (e.g. a browser tab with an aggressively-cached SPA) only recognizes the legacy `phase: 'tunnel_verifying'`. Old handlers fall through to the "legacy plain-message server_status" branch and push the message as a system chat message instead of rendering a banner.

Impact: low (bundled dashboard is served by the same server; mostly a concern for aggressive browser caches or users who leave a tab open across a server upgrade). Still worth fixing to avoid confusing chat history.

## Original Review Comment

> Mixed-version case: new server, old cached dashboard. Old handler falls through to the legacy plain-message branch and injects "Tunnel warming up…" as a chat system message. Consider also emitting `tunnel_verifying` alongside (or with) `tunnel_warming` for one release to smooth migration.

## Acceptance Criteria

- [ ] Decide migration strategy: (a) server emits BOTH phase names for one release, then deprecates; (b) dashboard plain-message branch sniffs phase string, silently drops unknown phases.
- [ ] Document the chosen strategy in `docs/architecture/reference.md` or a CHANGELOG entry.
- [ ] Once migration window passes, drop `tunnel_verifying` backward-compat branch from dashboard handler.

PR #2881, opened 2026-04-19
fix(ws): pin --no-auth clients to server protocol, deflake broadcast test

### LINKED ISSUE #2869: broadcastMinProtocolVersion: --no-auth mode + flaky setTimeout(100) in integration test
## Context

Identified during /full-review of PR #2863 (Copilot comments).

### 1. \`broadcastMinProtocolVersion\` excludes \`--no-auth\` auto-auth'd clients

\`WsBroadcaster.broadcastMinProtocolVersion\` filters with \`(client.protocolVersion ?? 0) >= minProtocolVersion\`. When \`authRequired: false\`, clients bypass \`handleAuthMessage\` and auto-authenticate in \`ws.js\` — \`client.protocolVersion\` is never set.

Impact: dev-only \`--no-auth\` mode doesn't see the tunnel warming/ready banner.

**Fix options:**
- (a) Set \`client.protocolVersion = SERVER_PROTOCOL_VERSION\` during auto-auth in \`ws.js\` (dev mode trusts itself).
- (b) In the broadcaster, treat \`protocolVersion == null\` for authenticated clients as \`MIN_PROTOCOL_VERSION\` (mirrors the explicit-auth path).

### 2. Flaky \`setTimeout(100)\` in \`ws-server-broadcast.test.js\` integration test

The test that exercises \`broadcastMinProtocolVersion\` end-to-end uses a fixed \`await new Promise(r => setTimeout(r, 100))\` before asserting which client received what. On slow CI this can underdeliver and produce spurious NOT-received results.

Fix: use \`waitForMessageMatch\` on the v2-expected receive, and a bounded-time \`expectNoMessage\` pattern on the v1 side.

## Scope

- \`packages/server/src/ws.js\` — auto-auth path for \`--no-auth\`
- \`packages/server/src/ws-broadcaster.js\` — OR the \`broadcastMinProtocolVersion\` filter
- \`packages/server/tests/ws-server-broadcast.test.js\` — replace setTimeout(100) with waitForMessageMatch

## Labels

- from-review

blamechris:
Closes #2869

## Summary

Two fixes packaged together:

### (a) `broadcastMinProtocolVersion` excludes `--no-auth` auto-auth'd clients

`WsBroadcaster.broadcastMinProtocolVersion` filters by `(client.protocolVersion ?? 0) >= minProtocolVersion`. Auto-authenticated `--no-auth` clients bypass `handleAuthMessage` in `ws-auth.js`, so `client.protocolVersion` was never set and these dev clients were silently excluded from v2-gated broadcasts (tunnel warming/ready banners).

Fix: in the auto-auth branch of `ws-server.js`, pin `client.protocolVersion = SERVER_PROTOCOL_VERSION`. `--no-auth` is dev-only and trusts itself. Updated the doc comment in `ws-broadcaster.js` to reflect this.

### (b) Flaky `setTimeout(100)` in `ws-server-broadcast.test.js`

The `broadcastMinProtocolVersion()` integration test used `await new Promise(r => setTimeout(r, 100))` before asserting which client got what — fragile on slow CI.

Fix: the v2 positive assertion now uses `waitForMessageMatch`. For the v1 negative assertion, we follow the gated broadcast with an un-gated `broadcast()` marker. Per-connection WebSocket message order is preserved, so once a client sees the marker any earlier `tunnel_warming` on that connection must already have been delivered (or correctly suppressed). Deterministic, no sleeps.

Also adds a new test that exercises the (a) regression path — a `--no-auth` auto-authenticated client receiving a v2-gated broadcast.

## Test plan

- [x] `node --test packages/server/tests/ws-server-broadcast.test.js` — 33/33 pass
- [x] Targeted `--test-name-pattern "broadcastMinProtocolVersion"` 10x loop — 10/10 green
- [x] Full file 3x loop — 3/3 green
- [x] `ws-broadcaster.test.js` + `ws-auth.test.js` — 93/93 pass, no regressions

### DIFF
--- a/packages/server/src/ws-broadcaster.js
+++ b/packages/server/src/ws-broadcaster.js
@@ -45,7 +45,9 @@ export class WsBroadcaster {
    * only reads `msg.message`.
    *
    * Clients that omit `protocolVersion` during auth are pinned to
-   * MIN_PROTOCOL_VERSION (see ws-auth.js), so this check is safe.
+   * MIN_PROTOCOL_VERSION (see ws-auth.js). Clients that auto-authenticate
+   * under `--no-auth` are pinned to SERVER_PROTOCOL_VERSION (dev mode
+   * trusts itself — see ws-server.js). So this check is safe.
    *
    * @param {number} minProtocolVersion
    * @param {object} message
--- a/packages/server/src/ws-server.js
+++ b/packages/server/src/ws-server.js
@@ -827,6 +827,11 @@ export class WsServer {
         const client = this.clients.get(ws)
         client.authenticated = true
         client.authTime = Date.now()
+        // --no-auth is dev-only and trusts itself — pin to the server's own
+        // protocol version so version-gated broadcasts (e.g. #2849 tunnel
+        // warming / ready via broadcastMinProtocolVersion) reach dev clients
+        // instead of being silently filtered out.
+        client.protocolVersion = SERVER_PROTOCOL_VERSION
         this._sendPostAuthInfo(ws)
         this._broadcastClientJoined(client, ws)
         log.info(`Client ${clientId} auto-authenticated (--no-auth)`)
--- a/packages/server/tests/ws-server-broadcast.test.js
+++ b/packages/server/tests/ws-server-broadcast.test.js
@@ -928,13 +928,28 @@ describe('public broadcast method', () => {
       message: 'Tunnel warming up…',
     })
 
-    // Give the server a moment to deliver.
-    await new Promise((r) => setTimeout(r, 100))
+    // Follow the gated broadcast with an UN-gated marker. Both broadcasts
+    // iterate the same clients map synchronously, so the marker is queued
+    // after the (maybe-delivered) tunnel_warming on each WebSocket. Per-
+    // connection message order is preserved by WebSocket, so once a client
+    // sees the marker we know any earlier tunnel_warming for that client
+    // has already arrived — or was correctly suppressed.
+    server.broadcast({ type: 'discovered_sessions', tmux: [{ sessionName: '__marker__' }] })
+
+    await waitForMessageMatch(
+      oldClient.messages,
+      (m) => m.type === 'discovered_sessions' && m.tmux?.[0]?.sessionName === '__marker__',
+      2000,
+      'marker on v1 client',
+    )
+    const newGotWarming = await waitForMessageMatch(
+      newClient.messages,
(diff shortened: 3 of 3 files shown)


blamechris:
## Full Review — PR #2881

**Verdict: Approve (low risk, high-confidence fix).** CI green across all 12 checks, 65/-6 diff, tightly scoped.

### Findings

**Strengths**
- Fix location is surgical: the only remaining `client.authenticated = true` path that didn't set `protocolVersion`. Grepped `ws-auth.js` + `ws-server.js` — no other auto-auth sites exist.
- Inline comment at `ws-server.js:828` cites the exact upstream issue (#2849) and rationale ("dev mode trusts itself"). Doc comment in `ws-broadcaster.js` updated symmetrically.
- Deflake is genuinely deterministic. `_broadcast()` iterates the `this._clients` map synchronously and per-connection message order is a hard guarantee from the ws library — the un-gated marker following the gated broadcast is a sound "ordering fence". No residual timing dependencies introduced.
- New `--no-auth` regression test is meaningful: it exercises the full auto-auth path (no `auth` message sent) and asserts the v2-gated broadcast reaches the client. Would have caught the #2849 regression had it existed first.
- Zero Attribution: commit message and PR body are clean. No Claude/AI mentions.
- Server style conformance: ES modules, no semicolons, single quotes — all preserved.

**Minor observations (non-blocking)**

1. `__marker__` session name (`discovered_sessions`) is a slight semantic abuse of the protocol message type — it works because no client logic keys on synthetic session names, but future readers may wonder. A trivially-typed dedicated test message (e.g. `{ type: 'test_marker' }`) would be cleaner. Not worth gating merge on.
2. `--no-auth` trust model assumption: pinning to `SERVER_PROTOCOL_VERSION` means a genuinely older dev client (e.g. mobile app on a stale build) connecting to a newer `--no-auth` dev server could receive v2-shape messages it can't parse. For dev-only mode with loopback/LAN use this is acceptable, matches the "dev trusts itself" framing — worth keeping in mind if `--no-auth` ever broadens scope.
3. 10/10 flake runs is a reasonable bar but not a permanent guarantee. Since the new assertions are order-fenced rather than timeout-bounded, flakiness resurfacing would indicate a real ordering regression — which is actually preferable signal. Fine.

**Security / correctness**: no concerns. The pinning change only elevates dev-auth clients to the server's own protocol version — no auth bypass, no new surface.

**Recommendation**: merge as-is. Follow-ups (if any) can track the `__marker__` cosmetic and `--no-auth` dev-trust doc note.

PR #2895, opened 2026-04-20
fix(desktop): avoid UTF-8 boundary panic in kill_orphan_cloudflared

blamechris:
## Summary

`kill_orphan_cloudflared` runs at every server start. Its `.exe` suffix check sliced `base` as a `&str` at byte offset `base.len() - 4`, which **panics** when that offset lands inside a multi-byte UTF-8 codepoint. Any process on the user's machine with non-ASCII glyphs in its command line (localized paths, shells invoking Unicode symbols like ×, CJK filenames, accented names) would crash chroxy-desktop's main thread before the Node server was spawned.

## Symptom
Tray dashboard shows `Health check timeout after 60s (29 attempts: 0 non-200, 29 errors)` because the Rust main thread has already panicked and nothing was listening on port 8765. `eprintln!` stderr (visible only when launched from a terminal) shows:

```
thread 'main' panicked at src/server.rs:57:24:
byte index 7 is not a char boundary; it is inside '×' (bytes 6..8) of `t=${i}×2s:`
```

## Fix
Compare the last 4 bytes as a `&[u8]` slice instead of slicing the string. ASCII `.exe` guarantees a valid char boundary at `base.len() - 4` whenever the last 4 bytes spell `.exe` (case-insensitive), so the subsequent `&base[..base.len() - 4]` stays safe.

## Test plan
- [x] `cargo test cloudflared_filter` — 16/16 pass (includes new regression test)
- [x] Full `cargo test` — 98/98 pass
- [x] New test `cloudflared_filter_tolerates_non_ascii_command_lines` covers: Unicode math symbols (`×`), accented Latin (`résumé`), CJK (`日本語`), and verifies the real cloudflared process is still matched alongside the non-ASCII bystanders
- [ ] Rebuild `.app` and confirm tray dashboard reaches "Server ready" on a machine with non-ASCII process names

### DIFF
--- a/packages/desktop/src-tauri/src/server.rs
+++ b/packages/desktop/src-tauri/src/server.rs
@@ -52,10 +52,17 @@ pub(crate) fn cloudflared_pids_to_kill(procs: &[(u32, String)], port: u16) -> Ve
                 .next()
                 .unwrap_or(trimmed);
             // Drop a case-insensitive ".exe" suffix before the compare so
-            // bare "cloudflared" and "CLOUDFLARED.EXE" both match.
-            let without_exe = if base.len() >= 4
-                && base[base.len() - 4..].eq_ignore_ascii_case(".exe")
+            // bare "cloudflared" and "CLOUDFLARED.EXE" both match. Compare
+            // the last 4 bytes directly to avoid panicking when `base` is
+            // longer than 4 bytes but its length minus 4 lands inside a
+            // multi-byte UTF-8 character (e.g. a process whose command
+            // line contains non-ASCII glyphs).
+            let bytes = base.as_bytes();
+            let without_exe = if bytes.len() >= 4
+                && bytes[bytes.len() - 4..].eq_ignore_ascii_case(b".exe")
             {
+                // Last 4 bytes are ASCII (.exe), so the char boundary
+                // guarantee holds and `&base[..base.len() - 4]` is safe.
                 &base[..base.len() - 4]
             } else {
                 base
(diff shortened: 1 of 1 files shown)


blamechris:
## Code Review — PR #2895

### Summary
Targeted fix for a `main`-thread panic in `cloudflared_pids_to_kill()` that crashed chroxy-desktop at startup whenever any process on the user's machine had non-ASCII glyphs in its command line. The fix swaps a `&str` slice (`base[base.len() - 4..]`) for a byte-slice comparison (`bytes[bytes.len() - 4..]`) which cannot hit the UTF-8 char-boundary panic. Includes a regression test that exercises the previously-crashing path. Small, correct, and appropriately scoped.

### Strengths
- **Correct invariant proof.** Once the last 4 bytes are verified ASCII `.exe` (case-insensitive), `base.len() - 4` IS a valid char boundary (ASCII bytes are single-codepoint), so the subsequent `&base[..base.len() - 4]` slice is provably panic-free. The inline comment makes this reasoning explicit — future readers won't have to re-derive it.
- **Zero-allocation approach.** Avoids the tempting `base.to_ascii_lowercase().ends_with(".exe")` alternative which would allocate a `String` per process per iteration. `kill_orphan_cloudflared` runs per server-start (not per-request), so allocation would be defensible, but zero-alloc is still the right choice for a hot-ish path.
- **Regression test would have caught the original bug.** `"bash -c 'echo t=2×5s'"` produces `base = "t=2×5s'"` (after splitting on whitespace and taking `rsplit('/', '\\').next()`). That's 10 bytes; `len - 4 = 6` lands inside the `×` codepoint (bytes 4..6 in `"t=2×"`). Running the test against the pre-fix code reproduces the exact panic in the PR body. The test also keeps PID 13 in the output, so it guards against accidental over-filtering.
- **No attribution noise.** PR body, commits, and test comments are clean.

### Issues
None. The diff is minimal, the fix is correct, and the test covers the reported panic path.

### Suggestions (non-blocking)
- None material. The inline safety comment already documents why `&base[..base.len() - 4]` remains valid.

### Nitpicks
- (Optional) If you prefer even less arithmetic, `base.strip_suffix_ignore_ascii_case(\".exe\").unwrap_or(base)` would do the same thing — but that method is nightly-only, so the current form is the stable-Rust idiom.

### Other `&str` slicing in the same file
Audited the rest of `server.rs` for similar patterns. The only other relevant slice is line 80: `let tail = &cmd[idx + needle.len()..];` where `needle` is a pure-ASCII literal (`\"http://localhost:<port>\"`). Since `cmd.find(needle)` returns a byte index and the needle contains no multi-byte characters, `idx + needle.len()` always lands on a char boundary. **Safe.** No other byte-index slicing in the file.

### Verdict
**Approve.** Ship it.

Issue #2889, opened 2026-04-19
tests: restore prior LOG_LEVEL instead of hard-coding 'info' in afterEach


blamechris:
## Context

Identified during /full-review of PR #2882.

PR #2882 (and the precedent in `tests/models-factory.test.js`) restores the global logger level in `afterEach` using a hard-coded `setLogLevel('info')`. Copilot's inline review on PR #2882 flagged this on both `tests/ws-forwarding.test.js` and `tests/ws-permissions.test.js`.

The logger's default level is sourced from `process.env.LOG_LEVEL` when `logger.js` is imported. If the test suite is ever invoked with `LOG_LEVEL=debug` (local debugging, CI variation, etc.), the `afterEach` silently clobbers the configured level back to `info` and leaks the wrong level into any subsequent test/suite.

Low severity — `npm test` in CI does not set `LOG_LEVEL` — but worth normalizing across the three test files that use this pattern.

## Affected files

- `packages/server/tests/ws-forwarding.test.js` — `[session-binding-create]` describe's `afterEach`
- `packages/server/tests/ws-permissions.test.js` — `[session-binding-create] / [session-binding-resend]` describe's `afterEach`
- `packages/server/tests/models-factory.test.js` — existing precedent

## Acceptance Criteria

- [ ] Replace hard-coded `setLogLevel('info')` with one of:
  - [ ] `setLogLevel(process.env.LOG_LEVEL ?? 'info')`, OR
  - [ ] Capture the prior level in `before`/`beforeEach` and restore it in `afterEach`, OR
  - [ ] Export a `getLogLevel()` helper from `logger.js` so tests can round-trip
- [ ] Apply consistently across all three files above
- [ ] Add a brief note in any log-level test helper (if created) documenting the pattern

## Original Review Comments

Copilot, PR #2882:

> The suite resets the global logger state to a hard-coded `'info'` level in afterEach. Since the default log level is derived from `process.env.LOG_LEVEL` when the module is initialized, forcing `'info'` can break runs where tests are executed with `LOG_LEVEL=debug` (or any non-info default) and can also interfere with other suites that intentionally changed the level.

PR #2919, opened 2026-04-23
fix(sdk-session): detect partial modelUsage drift

### LINKED ISSUE #2885: sdk-session: detect partial modelUsage drift (mixed entries with/without contextWindow)
## Context

Raised during review of PR #2879, which added a debug log when the SDK emits `modelUsage` but *none* of the entries carry a numeric `contextWindow`.

Current guard (`packages/server/src/sdk-session.js`):

\`\`\`js
let sawContextWindow = false
for (const [modelId, usage] of Object.entries(msg.modelUsage)) {
  if (usage && typeof usage.contextWindow === 'number') {
    sawContextWindow = true
    // ...
  }
}
if (!sawContextWindow && modelIds.length > 0) {
  log.debug('modelUsage contract drift: ...')
}
\`\`\`

This only catches *total* drift. If the SDK ever starts emitting a partial shape — some entries carrying `contextWindow`, others not — the bad entries silently skip cache update and nothing gets logged.

## Why it matters

The SDK emits a uniform shape today, so this is not an observed regression. It's a gap in the drift signal: a future rollout that updates one model family's schema before another would be invisible.

## Proposed fix

Track a per-entry miss count and log if any entries were skipped:

\`\`\`js
const missingIds = []
for (const [modelId, usage] of Object.entries(msg.modelUsage)) {
  if (usage && typeof usage.contextWindow === 'number') {
    // update
  } else {
    missingIds.push(modelId)
  }
}
if (missingIds.length > 0) {
  log.debug(\`modelUsage partial drift: contextWindow missing for \${JSON.stringify(missingIds)}\`)
}
\`\`\`

Replaces the existing total-drift log (the new signal is strictly more informative).

## Scope

- `packages/server/src/sdk-session.js` — replace `sawContextWindow` boolean with `missingIds` array, log sampled modelIds + keys when non-empty
- Extend tests in `packages/server/tests/sdk-session.test.js` with a mixed-shape case

## Related

- PR #2879 (closes #2873) — added the total-drift log this extends
- PR #2864 (closes #2830) — companion fix on `models.js`

## Labels

- from-review

blamechris:
## Summary

The current `modelUsage` drift guard in `sdk-session.js` only logs when *none* of the entries carry a numeric `contextWindow`. A mixed-shape payload — some entries with the field, others without — skips the missing entries silently and logs nothing. That hides the case that matters most: a gradual upstream schema rollout that updates one model family before another.

Replaces the `sawContextWindow` boolean with a `missingIds` array. Any entry lacking a numeric `contextWindow` is recorded, and a single debug log fires when `missingIds.length > 0`, reporting the missing modelIds plus a sample of keys actually present on the first missing entry.

The new signal is strictly more informative than the old total-drift log, so the prior log message is replaced (prefix changes from `modelUsage contract drift` to `modelUsage partial drift`).

## Changes

- `packages/server/src/sdk-session.js` — switch drift detection from per-turn boolean to per-entry miss tracking
- `packages/server/tests/sdk-session.test.js` — update existing tests to the new log prefix and add a mixed-shape case that asserts only the missing modelId is reported

## Test plan

- [x] `node --test tests/sdk-session.test.js` — all 70 tests pass, including 4 `modelUsage contract drift` subtests (total-drift, partial-drift, present, empty)
- [x] Full server suite run — only pre-existing, unrelated `web-task-manager` feature-detection failure remains

Closes #2885

### DIFF
--- a/packages/server/src/sdk-session.js
+++ b/packages/server/src/sdk-session.js
@@ -380,25 +380,27 @@ export class SdkSession extends BaseSession {
             // a value actually changed to avoid thrashy writes / UI churn.
             let contextWindowChanged = false
             if (msg.modelUsage && typeof msg.modelUsage === 'object') {
-              let sawContextWindow = false
-              const modelIds = Object.keys(msg.modelUsage)
+              const missingIds = []
               for (const [modelId, usage] of Object.entries(msg.modelUsage)) {
                 if (usage && typeof usage.contextWindow === 'number') {
-                  sawContextWindow = true
                   if (updateContextWindow(modelId, usage.contextWindow)) {
                     contextWindowChanged = true
                   }
+                } else {
+                  missingIds.push(modelId)
                 }
               }
-              // Drift signal: SDK emitted modelUsage entries but none carried a
-              // numeric contextWindow. Likely means the field was renamed or
-              // removed upstream. Log a redacted sample so a future regression
-              // is diagnosable without flooding info-level output.
-              if (!sawContextWindow && modelIds.length > 0) {
-                const sampleId = modelIds[0]
+              // Drift signal: at least one modelUsage entry was missing a
+              // numeric contextWindow. Catches both total drift (field renamed
+              // or removed upstream) and partial drift (schema updated for one
+              // model family before another). Log a redacted sample so a
+              // future regression is diagnosable without flooding info-level
+              // output.
+              if (missingIds.length > 0) {
+                const sampleId = missingIds[0]
                 const sampleKeys = Object.keys(msg.modelUsage[sampleId] || {})
                 log.debug(
-                  `modelUsage contract drift: expected numeric contextWindow; received modelIds=${JSON.stringify(modelIds)} sampleKeys=${JSON.stringify(sampleKeys)}`
+                  `modelUsage partial drift: contextWindow missing for modelIds=${JSON.stringify(missingIds)} sampleKeys=${JSON.stringify(sampleKeys)}`
                 )
               }
             }
--- a/packages/server/tests/sdk-session.test.js
+++ b/packages/server/tests/sdk-session.test.js
@@ -859,11 +859,48 @@ describe('SdkSession', () => {
       }
 
       const drift = entries.find(
-        (e) => e.component === 'sdk' && e.message.startsWith('modelUsage contract drift')
+        (e) => e.component === 'sdk' && e.message.startsWith('modelUsage partial drift')
       )
       assert.ok(drift, 'expected a drift debug log')
       assert.equal(drift.level, 'debug')
       assert.ok(drift.message.includes('claude-sonnet-4-6'))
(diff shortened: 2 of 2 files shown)


blamechris:
## Code Review: PR #2919

### Summary

Tightens the `modelUsage` drift guard in `sdk-session.js` to catch *partial* schema rollouts (some entries carry `contextWindow`, others do not) in addition to the existing *total* drift case. Replaces the `sawContextWindow` boolean with a per-entry `missingIds` array; the debug log now names only the entries that actually failed the contract. Directly implements the fix proposed in #2885.

### Strengths

- **Strictly-informative replacement.** The new log message is a proper superset of the old one — any payload that previously triggered `contract drift` still triggers `partial drift`, and mixed payloads that were silently dropped before now surface. No regression risk on the observability side.
- **Sample selection matches intent.** `sampleId = missingIds[0]` ensures `sampleKeys` come from an entry that *actually failed* the contract, which is exactly the diagnostic signal you want. The old code sampled `modelIds[0]`, which on a mixed payload could have sampled a passing entry and shown misleading keys — the change silently fixes a latent bug in the sample-keys output, too.
- **Comment updated to reflect new semantics.** The block comment now correctly describes both total and partial drift; no stale documentation.
- **Tests cover the right axes.** Four cases: total drift, partial drift (new), present (no drift), empty payload (no drift). The partial-drift assertion explicitly checks the *passing* model is **not** in the message, which is the main behavioral claim.
- **Log level is `debug`** — appropriate for a diagnostic signal that is not actionable in prod; no info-level flooding risk.

### Issues Found

#### Critical (Must Fix)

None.

| `packages/server/tests/sdk-session.test.js` | 903 (new test's `finally`) | The new test calls `setLogLevel('info')` in `finally`, perpetuating the pattern already tracked in open issue #2889. Low priority — the existing sibling tests use the same pattern, so this is consistent with the file, not a regression. Leaving as-is is fine; just flagging so #2889's cleanup captures this new call site. | Consistency with the issue's acceptance criteria when it's eventually addressed. |

#### Nitpicks (Optional)

- The outer test description string is `'logs debug message when modelUsage entries lack contextWindow'` for the total-drift test and `'logs partial drift when some entries carry contextWindow and others do not'` for the new test. Slightly inconsistent framing (behavior-first vs event-first). Not worth a change on its own.

None created — the one suggestion is already tracked on #2889.

### Architecture Notes

This lives in the `result`-case handler of `SdkSession`'s SDK message loop, downstream of the context-window cache update path. The guard is pure (no side effects beyond a debug log), runs once per turn, and cannot block the stream. Cache invalidation semantics are preserved: `contextWindowChanged` still tracks only real updates, so `saveModelsCache()` + `models_updated` broadcast behavior is unchanged. No WS protocol surface is touched. No cleanup or lifecycle implications.

### Verdict

Minimal, focused fix that closes #2885 as designed, fully tested, green CI, and silently corrects a latent sample-keys bug along the way.

PR #2926, opened 2026-04-23
feat(desktop): enrich PATH in Tauri-spawned server process

### LINKED ISSUE #2893: desktop: enrich PATH in Tauri-spawned server process (defense-in-depth for subprocess lookups)
### Context
Surfaced during review of PR #2891. PR #2891 fixes the two known call sites (\`doctor.js\` + \`cli-session.js\`) that spawn the \`claude\` binary when chroxy is launched from the Tauri tray app — which on macOS inherits a minimal GUI PATH (\`/usr/bin:/bin:/usr/sbin:/sbin\`) that omits \`~/.local/bin\`.

That's sufficient for the immediate bug (server startup + claude sessions work). However, the root cause is PATH impoverishment, and any future subprocess spawned by the server will hit the same problem unless it also uses \`resolveBinary()\`. Examples of latent exposure:

- Hook scripts that call out to external tools
- Future MCP servers spawned as child processes
- \`git\` invocations (already uses \`resolveBinary\` in \`git.js\`, but future tools might not)
- User-configured slash commands that shell out

### Proposal
On the Tauri side (Rust), before spawning the Node server child process, build an enriched PATH that includes common user install dirs:

\`\`\`rust
let home = std::env::var(\"HOME\").unwrap_or_default();
let existing = std::env::var(\"PATH\").unwrap_or_default();
let enriched = format!(
  \"{home}/.local/bin:/opt/homebrew/bin:/usr/local/bin:{home}/.npm-global/bin:{home}/.claude/local/node_modules/.bin:{existing}\"
);
cmd.env(\"PATH\", enriched);
\`\`\`

This makes \`resolveBinary()\` fallback lists a safety net rather than the primary resolution mechanism, and means any subprocess spawned by the server (not just the ones we've audited) benefits.

### Notes
- Not a blocker for #2891 — that PR solves the two known failure paths and should merge as-is.
- Could alternatively be implemented in \`server-cli.js\` startup (Node-side) as an \`env.PATH\` enrichment before any subprocess spawn.
- Should confirm behavior on Linux (where GUI launchers may also strip PATH) and Windows (where the \`%PATH%\` semantics differ).

### Acceptance
- Tauri-launched chroxy server has an enriched PATH including \`~/.local/bin\`, Homebrew, and npm-global.
- Regression test (possibly in smoke suite) that asserts the spawned server sees the enriched PATH.
- Doc note in architecture reference explaining PATH handling for GUI launch mode.

blamechris:
## Summary

macOS GUI apps launched via launchd (including this Tauri tray binary) inherit a minimal PATH (`/usr/bin:/bin:/usr/sbin:/sbin`) that misses Homebrew, nvm, `~/.npm-global/bin`, pipx, and `~/.claude/local/node_modules/.bin`. Without enrichment, any subprocess the server spawns (claude CLI, git, hooks, MCP servers) will silently fail to resolve common tools — even though the two known call sites fixed in #2891 handle the `claude` binary specifically.

This PR is the defense-in-depth companion to #2891: it fixes the root cause on the Tauri side before the Node server ever starts, so that per-binary `resolveBinary()` fallback lists become a safety net rather than the primary resolution mechanism.

## Changes

- Extract the inline PATH-building logic from `ServerManager::start_server()` into a pure `build_enriched_path()` helper so it can be unit-tested without touching env vars or the filesystem.
- Broaden the prepended directories to match the proposal in #2893:
  - `/opt/homebrew/bin`, `/usr/local/bin`, `/usr/bin`, `/bin` (Unix)
  - `~/.local/bin`
  - `~/.npm-global/bin`
  - `~/.claude/local/node_modules/.bin`
- Dirs already present in the inherited PATH are skipped (substring match, matching historical behavior).
- Windows keeps the conservative `node_bin + base_path` join because `%PATH%` semantics and typical install locations differ.
- Guard against `node_path.parent()` being `None` (previously panicked on `.unwrap()`).

## Tests

9 new unit tests in `server::tests` covering:

- Prepend order (node_bin first, base_path verbatim at end)
- System-bin injection on Unix
- Dedup against base_path (no duplicate `/opt/homebrew/bin`)
- User install dir inclusion gated on `home_dir.is_some()`
- Empty-base-path and empty-node-bin guards (no leading/trailing/doubled separators)
- Custom separator handling (`:` vs `;`)

Full suite: 107 passed, 0 failed (`cargo test` in `packages/desktop/src-tauri`).

## Test plan

- [x] `cargo test` passes in `packages/desktop/src-tauri`
- [x] New helper is pure and covered by unit tests
- [x] Call site in `start_server()` updated to use the helper
- [ ] Manual smoke: launch Tauri tray, verify server child sees enriched PATH (out of scope for this PR — covered by existing smoke suite on release)

Closes #2893

### DIFF
--- a/packages/desktop/src-tauri/src/server.rs
+++ b/packages/desktop/src-tauri/src/server.rs
@@ -795,8 +795,11 @@ impl ServerManager {
         let full_path =
             build_enriched_path(&base_path, &node_bin, home_dir.as_deref(), path_sep);
         cmd.env("PATH", &full_path);
-        // Ensure HOME is set — macOS GUI apps may not inherit it
-        if let Some(home) = dirs::home_dir() {
+        // Ensure HOME is set — macOS GUI apps may not inherit it. Reuse the
+        // already-resolved `home_dir` so PATH and HOME are derived from the
+        // same value rather than risking a second, potentially divergent
+        // `dirs::home_dir()` lookup.
+        if let Some(ref home) = home_dir {
             cmd.env("HOME", home);
         }
         // Pass config as env vars (same pattern as supervisor.js)
--- a/packages/desktop/src-tauri/src/server.rs
+++ b/packages/desktop/src-tauri/src/server.rs
@@ -177,6 +177,92 @@ pub(crate) fn parse_powershell_json(stdout: &str) -> Vec<(u32, String)> {
     out
 }
 
+/// Build an enriched PATH string for the spawned server child process.
+///
+/// macOS GUI apps (including launchd-launched apps like Tauri tray binaries)
+/// inherit a minimal PATH — typically `/usr/bin:/bin:/usr/sbin:/sbin` — that
+/// omits Homebrew, nvm, pipx, npm-global, and other user install locations.
+/// Without enrichment, any subprocess the server spawns (claude binary, git,
+/// cloudflared, hook scripts, MCP servers) will silently fail to resolve
+/// common tools.
+///
+/// This is a pure function so it can be unit-tested without touching the real
+/// environment or filesystem. The caller is responsible for supplying
+/// `base_path` (usually from `std::env::var("PATH")`) and `home_dir` (usually
+/// from `dirs::home_dir()`).
+///
+/// Semantics:
+/// - `node_bin` is always prepended first so the bundled Node is preferred.
+/// - Common system bins (`/opt/homebrew/bin`, `/usr/local/bin`, `/usr/bin`,
+///   `/bin`) are prepended next on Unix.
+/// - User install dirs (`~/.local/bin`, `~/.npm-global/bin`,
+///   `~/.claude/local/node_modules/.bin`) are prepended when `home_dir` is
+///   `Some`. They are NOT filesystem-checked — the goal is defense-in-depth,
+///   and a non-existent dir on PATH is harmless.
+/// - Directories already present in `base_path` are skipped to avoid
+///   duplicates, but the check is conservative (substring match) rather than
+///   element-level parsing, which matches the historical behavior.
+/// - `base_path` is appended verbatim at the end.
+///
+/// On Windows, only `node_bin` and `base_path` are joined — the user install
+/// dirs and Unix-specific system bins do not apply.
+pub(crate) fn build_enriched_path(
+    base_path: &str,
+    node_bin: &str,
+    home_dir: Option<&std::path::Path>,
+    path_sep: &str,
+) -> String {
+    let mut dirs: Vec<String> = Vec::new();
+
+    // Bundled Node directory — always first so it wins over any system node.
+    if !node_bin.is_empty() {
+        dirs.push(node_bin.to_string());
+    }
+
+    // System bins (Unix). We leave Windows PATH handling to the OS default
+    // because `%PATH%` semantics and typical install locations differ.
+    #[cfg(unix)]
+    {
+        for dir in &[
+            "/opt/homebrew/bin",
+            "/usr/local/bin",
+            "/usr/bin",
+            "/bin",
+        ] {
+            if !base_path.contains(dir) && !dirs.iter().any(|d| d == dir) {
+                dirs.push((*dir).to_string());
+            }
+        }
+    }
+
+    // User install dirs — only on Unix, and only when we know $HOME.
+    #[cfg(unix)]
+    if let Some(home) = home_dir {
+        for rel in &[
+            ".local/bin",
+            ".npm-global/bin",
+            ".claude/local/node_modules/.bin",
+        ] {
+            let candidate = home.join(rel).display().to_string();
+            if !base_path.contains(&candidate) && !dirs.iter().any(|d| d == &candidate) {
+                dirs.push(candidate);
+            }
+        }
+    }
+
+    // Silence unused-param warnings on Windows builds.
+    #[cfg(windows)]
+    let _ = home_dir;
+
+    if dirs.is_empty() {
+        return base_path.to_string();
+    }
+    if base_path.is_empty() {
+        return dirs.join(path_sep);
+    }
+    format!("{}{}{}", dirs.join(path_sep), path_sep, base_path)
+}
+
 /// Current state of the server process.
 #[derive(Debug, Clone, PartialEq)]
 pub enum ServerStatus {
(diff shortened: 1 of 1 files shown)


blamechris:
## Code Review: PR #2926

### Summary

Extracts the inline PATH-building logic from `ServerManager::start_server()` into a pure, testable `build_enriched_path()` helper, broadens the set of prepended directories to match the defense-in-depth proposal in #2893 (`/opt/homebrew/bin`, `/usr/local/bin`, `/usr/bin`, `/bin`, `~/.local/bin`, `~/.npm-global/bin`, `~/.claude/local/node_modules/.bin`), and adds 9 unit tests covering order, dedup, empty inputs, and separators. Also removes a latent `.unwrap()` panic on `node_path.parent()`.

This is the Tauri-side companion to #2891 — the root-cause fix so that per-binary `resolveBinary()` fallback lists become a safety net rather than the primary resolution mechanism.

### Strengths

- **Pure helper pattern**: Extracting `build_enriched_path()` as a pure function (no env-var reads, no filesystem touches) is exactly the right shape for unit-testability. The 9 tests cover the important branches (order, dedup, empty base_path, empty node_bin, home=None, custom separator).
- **`.unwrap()` removed**: Converting `node_path.parent().unwrap()` to `.map(...).unwrap_or_default()` eliminates a real (if unlikely) panic path. Good hygiene.
- **Empty-input hardening**: The explicit `base_path.is_empty()` and `node_bin.is_empty()` branches prevent accidental `::` or leading-`:` elements that would silently make PATH search CWD — a genuine security win, and the `env -i` test scenario makes the rationale concrete.
- **Windows scoping**: `#[cfg(unix)]` around both the system-bin list and the user-install block correctly keeps Windows PATH semantics out of the Unix branch; `#[cfg(windows)] let _ = home_dir;` is the idiomatic way to silence the unused-param warning.
- **Rust idioms**: Iteration over static string slices, `Vec<String>` accumulator, `display().to_string()` for path → string, `home_dir.as_deref()` at the call site — all standard, readable Rust.
- **Docstring**: The helper's rustdoc explains both the what (semantics) and the why (launchd inherits a minimal PATH), which is exactly what a future maintainer needs.
- **CI green**: Desktop Rust Tests job passes.

### Issues Found

#### Critical (Must Fix)

None. This is a careful, well-tested refactor.

| `packages/desktop/src-tauri/src/server.rs` | ~790/800 | Reuse the already-computed `home_dir` value when setting `HOME` instead of calling `dirs::home_dir()` twice | Minor efficiency + consistency. Copilot flagged this too. Can be a trivial follow-up. |
| `packages/desktop/src-tauri/src/server.rs` | ~232/247 | The substring dedup (`base_path.contains(dir)`) has false positives — e.g. if `base_path = "/sbin"`, then `base_path.contains("/bin")` is true and `/bin` is incorrectly skipped. Similarly `/usr/bin` is a substring of `/usr/bin/extra`. | Real on macOS GUI PATH: the default is `/usr/bin:/bin:/usr/sbin:/sbin`, and `/bin` would pass here (it's a literal element). But if a user's base_path contained e.g. `/opt/homebrew/bin/tools` (unlikely), the Homebrew dedup would skip. The docstring acknowledges "conservative (substring match)" — consider element-level split on `path_sep` for a later hardening pass. |
| `packages/desktop/src-tauri/src/server.rs` | tests | Missing test: user install dir already in `base_path` should be deduped (symmetry with `build_enriched_path_skips_system_dir_already_in_base_path`) | Covers the `~/.local/bin` / `~/.npm-global/bin` already-present case. Currently untested. |
| `packages/desktop/src-tauri/src/server.rs` | tests | Missing test: `node_bin` equal to one of the system bins (e.g. `node_bin = "/usr/bin"`) — should it dedup against itself? | Edge case only — would result in `/usr/bin:/usr/bin:...`. Not harmful (PATH search stops at first hit) but tightens behavior. |

#### Nitpicks (Optional)

- The inline call-site comment (lines 776-784) now duplicates the helper's rustdoc. Could be collapsed to a short "See `build_enriched_path` for rationale." reference.
- `for dir in &["/opt/homebrew/bin", ...]` vs `for dir in ["/opt/homebrew/bin", ...]` (Rust 2021) — stylistic; the current form works with MSRV policies.

| Reuse `dirs::home_dir()` result + tighten dedup logic (element-level instead of substring) + add missing dedup / node_bin==system_bin tests | [#2939](https://github.com/blamechris/chroxy/issues/2939) | None are blockers. The substring-match semantics are explicitly called out in the rustdoc as "conservative" and match historical behavior; the double `home_dir()` call is purely cosmetic; the two missing tests are nice-to-have. All three can land together as a small polish PR. |

### Architecture Notes

This change sits entirely in the Tauri Rust process — the Node server doesn't need to change because it inherits the enriched `PATH` at spawn. This is the correct layer: the root-cause fix for launchd's minimal GUI PATH belongs on the Rust side before any subprocess is forked.

Interaction with existing code paths:
- `resolveBinary()` fallback lists in `doctor.js` / `cli-session.js` / `git.js` continue to work as a safety net. With this PR, they become redundant for the common cases (Homebrew, ~/.local/bin, npm-global), which means future subprocess-spawning code in the server is less likely to hit "tool not found" in the GUI-launched path.
- Supervisor path (non-Tauri launches via `npx chroxy start`) is unaffected — that process inherits the shell's PATH.
- CLI headless mode is unaffected — no new message types, no WS protocol changes.

The pure-helper extraction also means future tweaks (e.g. if Linux needs `~/.cargo/bin` or Windows needs a specific set) can be added with a single test-first commit rather than an env-reading refactor.

### Verdict

**Approve** — Ready to merge. The refactor is clean, the tests are focused and cover the important edge cases, the panic removal is a real bug fix, and Copilot's sole comment (redundant `dirs::home_dir()` call) is nit-level and captured in the follow-up issue.

blamechris:
## Review Comments Addressed

Copilot's review had no inline comments (the one flagged item was in the "Comments suppressed due to low confidence" section of the review body). Addressing it anyway since the fix is trivial.

| 1 | Copilot (review body, low-confidence): `dirs::home_dir()` called twice in `start_server_process()` — reuse the already-computed value so `PATH` and `HOME` cannot diverge. | FIX | `50cd5c08e` |

**Change:** Replaced the second `if let Some(home) = dirs::home_dir()` with `if let Some(ref home) = home_dir`, borrowing the `Option<PathBuf>` already computed two lines above.

```diff
-        // Ensure HOME is set — macOS GUI apps may not inherit it
-        if let Some(home) = dirs::home_dir() {
+        // Ensure HOME is set — macOS GUI apps may not inherit it. Reuse the
+        // already-resolved `home_dir` so PATH and HOME are derived from the
+        // same value rather than risking a second, potentially divergent
+        // `dirs::home_dir()` lookup.
+        if let Some(ref home) = home_dir {
             cmd.env("HOME", home);
         }
```

Verified: `cargo test --lib` in `packages/desktop/src-tauri` → 107 passed, 0 failed.

- Fixed: 1 (`50cd5c08e`)
- False positives: 0
- Follow-up issues created: 1 during agent-review (#2939) — item 1 of 3 on that issue is now addressed by this fix; items 2 and 3 (tighten substring dedup + add missing tests) remain open.
- Existing issues closed: 0

PR #2930, opened 2026-04-23
fix(server): clean up orphaned .tmp files after rename failure

### LINKED ISSUE #2909: session-state-persistence: clean up orphaned .tmp files after rename failure
## Context

Identified during review of PR #2907.

## Description

`serializeState` writes to `<path>.tmp` then calls `renameSync(tmpPath, mainPath)`. If the final rename throws (disk full, permission change, transient FS error), the `.tmp` file is leaked on disk. Pre-existing behavior, but made more visible now that operators may inspect files alongside `.bak`.

## Proposed fix

Wrap the final `renameSync` in try/catch; on failure, attempt `unlinkSync(tmpPath)` before re-throwing.

## Original review comment

> Pre-existing, not introduced by this PR; separate hygiene fix. Can be batched with other persistence polish.

## Acceptance criteria

- [ ] Failed serializeState no longer leaks `.tmp` files
- [ ] Original error is still surfaced
- [ ] Test covering the cleanup path

## Reference

`packages/server/src/session-state-persistence.js:36,52`

blamechris:
## Summary

- `serializeState` wrote to `<path>.tmp` then called `renameSync(tmpPath, mainPath)`. If the final rename threw (disk full, permission change, transient FS error), the `.tmp` file was leaked on disk forever.
- Wrap the final `renameSync` in `try/catch`: best-effort `unlinkSync(tmpPath)` in the catch block, then re-throw the original error so callers still see the real failure.
- Suppress `ENOENT` on cleanup (nothing to unlink is fine); log any other secondary unlink failure but still re-throw the original rename error.

Identified during review of PR #2907.

## Test plan

New describe block `SessionStatePersistence rename-failure cleanup (regression: #2909)` in `packages/server/tests/session-state-persistence.test.js`:

- [x] RED verified: tests fail before the fix (orphaned `.tmp` remains)
- [x] GREEN verified: all three new cases pass after the fix
  - removes the orphaned `.tmp` file when the final rename fails
  - surfaces the original rename error (does not swallow)
  - does not leak `.tmp` across repeated rename failures
- [x] No regressions in other `session-state-persistence.test.js` cases (backup rotation, flushPersist, restoreState, etc.)

Rename failure is induced portably by pre-creating `<stateFile>` and `<stateFile>.bak` as non-empty directories (POSIX `rename` into a non-empty dir → `ENOTEMPTY` / `EISDIR`).

Closes #2909

### DIFF
--- a/packages/server/src/session-state-persistence.js
+++ b/packages/server/src/session-state-persistence.js
@@ -55,7 +55,19 @@ export class SessionStatePersistence {
         }
       }
     }
-    fs.renameSync(tmpPath, this._stateFilePath)
+    try {
+      fs.renameSync(tmpPath, this._stateFilePath)
+    } catch (err) {
+      // Clean up the orphaned .tmp file so it doesn't leak on disk across
+      // retries. Suppress ENOENT (nothing to clean up) and any secondary
+      // unlink error (we must re-throw the original rename failure).
+      try { fs.unlinkSync(tmpPath) } catch (cleanupErr) {
+        if (cleanupErr && cleanupErr.code !== 'ENOENT') {
+          log.warn(`Failed to remove orphaned ${tmpPath}: ${cleanupErr.message}`)
+        }
+      }
+      throw err
+    }
     log.info(`Serialized ${state.sessions?.length ?? 0} session(s) to ${this._stateFilePath}`)
     return state
   }
--- a/packages/server/tests/session-state-persistence.test.js
+++ b/packages/server/tests/session-state-persistence.test.js
@@ -1,7 +1,7 @@
 import { describe, it, beforeEach, afterEach, mock } from 'node:test'
 import assert from 'node:assert/strict'
 import fs from 'fs'
-import { writeFileSync, readFileSync, existsSync, mkdtempSync, rmSync } from 'fs'
+import { writeFileSync, readFileSync, existsSync, mkdtempSync, mkdirSync, rmSync } from 'fs'
 import { join } from 'path'
 import { tmpdir } from 'os'
 import { SessionStatePersistence } from '../src/session-state-persistence.js'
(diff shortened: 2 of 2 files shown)


blamechris:
## Code Review: PR #2930

### Summary

Minimal, targeted fix for #2909: wraps the final `renameSync(tmpPath, stateFile)` in `serializeState` in a `try/catch` that best-effort `unlinkSync`s the orphaned `.tmp` and re-throws the original error. Three regression tests force rename failure portably by pre-creating `<stateFile>` and `<stateFile>.bak` as non-empty directories (POSIX `renameSync` into a non-empty dir → `ENOTEMPTY`/`EISDIR`).

### Strengths

- **Error preservation is correct.** The original rename error is re-thrown verbatim so `flushPersist`'s `log.error` and `schedulePersist`'s debounced catch still see the real failure mode (disk full, EACCES, etc.). Secondary cleanup failures are demoted to `log.warn` and never shadow the primary failure.
- **`ENOENT` on cleanup is suppressed.** Important because on some platforms a failed `renameSync` can still have consumed the source file — there's nothing to clean up, and that's fine.
- **RED/GREEN rigor.** Verified locally: reverting the fix makes 2 of the 3 new tests fail (`removes the orphaned .tmp` + `does not leak .tmp across repeated failures`); reinstating it makes all 32 tests in the file pass.
- **Portable failure-injection.** The non-empty-directory trick is a clean substitute for mocking `fs.renameSync` and works on Linux, macOS, and (per POSIX) CI runners. The regex `/EISDIR|ENOTEMPTY|EEXIST|EPERM|EACCES/` correctly covers platform variation.
- **`blockRename()` also blocks the `.bak` rotation.** Good — this exercises the path where both the `.bak` rotate (which is already wrapped in its own `try/catch` and only logs) and the main rename fail in sequence, proving the new cleanup runs regardless of whether rotation succeeded.
- **Code style matches the file.** ES modules, no semicolons, single quotes, no TypeScript — all consistent with the server package conventions.

### Issues Found

#### Critical (Must Fix)

_None._

_None — the fix is scoped exactly to the bug._

#### Nitpicks (Optional)

- `session-state-persistence.js:60` — the inner `try { unlinkSync(tmpPath) } catch (cleanupErr) { ... }` is on one line and branches inside the `catch`. Stylistically fine, but a two-line form would match the existing `.bak` rotation block just above. Non-blocking.

| Finding | Issue | Rationale for deferral |
|---------|-------|------------------------|
| `environment-manager.js:921-928` has the **same bug** — `_persist()` writes `<path>.tmp` then `renameSync`s without cleanup if rename fails. The outer `try/catch` catches and logs but leaks the `.tmp`. `models.js:saveCache` already handles this correctly (inner `unlinkSync` in outer catch), so the pattern is repo-established. | [#2940](https://github.com/blamechris/chroxy/issues/2940) | Out of scope for a targeted fix to #2909 (session state). Splitting keeps this PR minimal and auditable. |

### Architecture Notes

This is a pure hygiene fix in a single persistence module. No WS protocol, client, or state-machine impact. The invariant "every successful write produces exactly one `.json` and one `.bak`, with no residual `.tmp`" is now actually enforced post-failure, matching what operators would assume from the file layout introduced in #2907.

One observation for future work (not for this PR): `session-state-persistence.js`, `environment-manager.js`, and `models.js` all reimplement the same atomic-write pattern (write tmp → rotate bak → rename → cleanup on error). A shared `atomicWriteFile(path, data, { restrict, rotateBak })` helper in `platform.js` would collapse all three into one tested implementation and prevent the exact class of sibling bug flagged in #2940. Not blocking this PR — worth an issue if the pattern gets copy-pasted a fourth time.

### Verdict

**Approve** — ready to merge. Fix is correct, scoped, tested (RED/GREEN verified), and matches repo conventions. Follow-up issue [#2940](https://github.com/blamechris/chroxy/issues/2940) tracks the identical bug in `environment-manager.js`.

Issue #2912, opened 2026-04-22
Unify SESSION_TOKEN_MISMATCH payload shape across all call sites


blamechris:
## Context

Identified during review of PR #2911.

## Description

PR #2911 enriched only `handleCreateSession` and `handleResumeConversation` with `boundSessionId`/`boundSessionName`. The same code (`SESSION_TOKEN_MISMATCH`) is returned from at least 5 other sites without the enrichment:

- `packages/server/src/handlers/session-handlers.js:33` (switch-session), `:123` (destroy), `:172` (rename)
- `packages/server/src/handlers/conversation-handlers.js:132` (another path)
- `packages/server/src/handlers/feature-handlers.js:36, :96, :105, :159`
- `packages/server/src/handlers/settings-handlers.js:184` (permission response)
- `packages/server/src/ws-permissions.js:278` (HTTP 401 path)

Clients that branch on `code === 'SESSION_TOKEN_MISMATCH'` expecting the new fields will see divergent shapes.

## Proposed approach

Either:

1. **Extract a helper** `buildTokenMismatchError(client, ctx, { context })` and apply it to every send site — same payload shape everywhere.
2. **Introduce a more specific code** (e.g. `SESSION_BOUND_CREATE_BLOCKED`) for the creation-path case so clients can branch on code without relying on the presence of an optional field.

Option 1 is preferred — consistent wire shape, no client branching needed.

## Acceptance criteria

- [ ] All `SESSION_TOKEN_MISMATCH` send sites return consistent payload (either always enriched or consistent code split)
- [ ] Regression tests cover all new send sites
- [ ] Clients handle the unified shape without breaking

Issue #2904, opened 2026-04-21
UX: SESSION_TOKEN_MISMATCH on new-session creation shows opaque 'Not authorized' with no remediation


blamechris:
## Symptom
Android app shows this error when the user tries to create a new session:

> **Session Error**
> Not authorized: client is bound to a specific session.

There is no indication of:
- Which session the token is bound to
- That the token was issued via pairing while that session was active
- How the user can obtain an unbound token to create new sessions

## Source

- `packages/server/src/handlers/session-handlers.js:52-55` (`handleCreateSession`)
- `packages/server/src/handlers/conversation-handlers.js:48-51` (`handleResumeConversation`)
- `packages/server/src/ws-auth.js:102-110` (binding applied when token was issued via pairing while a session was active)

Server returns a structured error with `code: 'SESSION_TOKEN_MISMATCH'` but the client surfaces only the raw `message`.

## Expected
Error surface should be actionable. Options:

1. Include the bound sessionId in the error payload (`boundSessionId`) and have the client render something like:
   > \"This device is paired to session **MarchBorne**. Reset the connection to create new sessions.\"
2. Offer a one-tap 'Reset pairing' button that clears the stored token and returns to ConnectScreen.
3. OR: change the unpair/re-pair flow so that re-scanning the QR while no session is active reliably produces an unbound token.

## Acceptance criteria
- [ ] Error message names the bound session (or makes clear the token is session-scoped)
- [ ] User can remediate without deleting and reinstalling the app
- [ ] Same treatment for the matching desktop-dashboard error path
- [ ] Related: #2832 (different SESSION_TOKEN_MISMATCH path — permission approval after backgrounding)

## Repro
1. Pair Android app with desktop while a session is active (scan QR)
2. Use the app on that session
3. Tap "new session" in the Android UI
4. Observe opaque "Not authorized" error

PR #2936, opened 2026-04-23
feat(server): extend boundSessionName enrichment to HTTP permission-response path

### LINKED ISSUE #2914: Extend boundSessionName enrichment to HTTP permission-response path
## Context

Identified during review of PR #2911.

## Description

`packages/server/src/ws-permissions.js:278` returns `SESSION_TOKEN_MISMATCH` in an HTTP 401 JSON body when a bound client tries to respond to another session's permission request. This path does not carry `boundSessionId` / `boundSessionName` — if the same remediation UX (name the bound session, offer reset) proves valuable on mobile, this path should support it too so the web permission modal can show the same specific message.

## Acceptance criteria

- [ ] HTTP 401 body includes `boundSessionId` + `boundSessionName`
- [ ] Covered by a regression test
- [ ] (Depends on #2904's approach being ratified first)

blamechris:
## Summary

Closes #2914. Mirrors the WS-path enrichment from PR #2911 on the HTTP 403 returned by `POST /permission-response` when a bound-token client tries to respond to another session's permission request. The body now includes `boundSessionId` + `boundSessionName` alongside the existing `SESSION_TOKEN_MISMATCH` code so the mobile permission modal and notification-action retry UX can render the same actionable "Device paired to session X" message that the WS create/resume paths already produce.

### Server (`packages/server/src/ws-permissions.js`)
- `handlePermissionResponseHttp`: when a bound-token caller is rejected for mismatched (or missing) session mapping, look up the bound session via `sessionManager.getSession` and include `boundSessionId` + `boundSessionName` in the 403 JSON body.
- `boundSessionName` is `null` when the bound id no longer maps to a live session (stale binding).

## Scope

Only the HTTP 403 emitted from `ws-permissions.js`. Other `SESSION_TOKEN_MISMATCH` send sites are tracked separately in #2912 (unify payload shape across all call sites) — this change intentionally matches the same field shape, so #2912's helper can wrap this call site without any payload churn.

### Server (`packages/server/tests/ws-permissions.test.js`)
- bound session exists → `boundSessionName` is populated
- stale binding → `boundSessionName` is `null`
- unmapped `requestId` bypass branch → enrichment applies there too

Results: full server suite 3303/3307 passing (4 pre-existing flakes: cloudflared env + Claude CLI `--remote` detection + backpressure drain, same as PR #2911).

## Test plan
- [x] Unit tests cover the 3 enrichment branches
- [ ] Manual: trigger an HTTP permission-response from a session-bound iOS notification action pointed at the wrong session, confirm the server 403 body carries `boundSessionId` + `boundSessionName` (logs + curl)
- [ ] Manual: confirm the existing per-session-bound happy path still returns 200 unchanged

### DIFF
--- a/packages/server/src/ws-permissions.js
+++ b/packages/server/src/ws-permissions.js
@@ -274,8 +274,20 @@ export function createPermissionHandler({ sendFn, broadcastFn, validateBearerAut
       if (callerBoundSessionId) {
         if (!originSessionId || originSessionId !== callerBoundSessionId) {
           log.warn(`HTTP /permission-response rejected: token bound to ${callerBoundSessionId} tried to respond to ${requestId} with mapped session ${originSessionId ?? 'unmapped'}`)
+          // Enrich the error with the bound session's name so the mobile
+          // permission modal / notification retry path can show the same
+          // actionable "Device paired to session X" message that the WS
+          // create/resume paths do (#2911). Name is null if the bound id
+          // no longer maps to a live session (stale binding) — see #2914.
+          const smForLookup = getSessionManager()
+          const boundEntry = smForLookup?.getSession?.(callerBoundSessionId)
           res.writeHead(403, { 'Content-Type': 'application/json' })
-          res.end(JSON.stringify({ error: 'not authorized for this permission request', code: 'SESSION_TOKEN_MISMATCH' }))
+          res.end(JSON.stringify({
+            error: 'not authorized for this permission request',
+            code: 'SESSION_TOKEN_MISMATCH',
+            boundSessionId: callerBoundSessionId,
+            boundSessionName: boundEntry?.name ?? null,
+          }))
           return
         }
       }
--- a/packages/server/tests/ws-permissions.test.js
+++ b/packages/server/tests/ws-permissions.test.js
@@ -412,6 +412,113 @@ describe('createPermissionHandler', () => {
       assert.ok(permissionSessionMap.has('victim-req'), 'permissionSessionMap entry must be preserved for the legit client')
     })
 
+    // Issue #2914: mirror the WS-path enrichment from PR #2911 on the HTTP
+    // 403 response so the permission modal / notification-action retry UX can
+    // show the bound session name ("Device paired to session X") instead of
+    // an opaque "not authorized". boundSessionId + boundSessionName must
+    // appear in the body alongside code: SESSION_TOKEN_MISMATCH.
+    it('includes boundSessionId and boundSessionName in the 403 body when the bound session exists (#2914)', async () => {
+      const permissionSessionMap = new Map([['victim-req', 'session-B']])
+      const respondToPermission = mock.fn()
+      const sm = {
+        getSession: mock.fn((id) => {
+          if (id === 'session-A') return { session: {}, name: 'MarchBorne', cwd: '/tmp' }
+          return { session: { respondToPermission } }
+        }),
+      }
+      const pairingManager = {
+        getSessionIdForToken: mock.fn(() => 'session-A'),
+      }
+      const opts = makeHandlerOpts({
+        permissionSessionMap,
+        getSessionManager: mock.fn(() => sm),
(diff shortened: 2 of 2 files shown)


blamechris:
## Code Review: PR #2936

### Summary

Mirrors the WS-path enrichment from PR #2911 onto the HTTP 403 returned by `POST /permission-response` when a bound-token caller tries to respond to another session's permission request. Body now includes `boundSessionId` + `boundSessionName` so the notification-retry UX can render the same actionable "Device paired to session X" message as the WS path.

### Strengths

- Surgical change — single call site, 12-line diff, no behavioural change beyond enrichment
- Three-case test coverage: live binding, stale binding (`null` name), and the unmapped-`requestId` bypass variant. The third case matters because `originSessionId` can be missing OR mismatched and both must enrich consistently
- Stale-binding guard (`boundEntry?.name ?? null`) matches the same defensive pattern used in `handleCreateSession` and `handleResumeConversation` from #2911
- Comment in the diff explicitly cross-references #2911 + #2914 — future readers will find the intent quickly
- `log.warn` already existed; didn't touch it, preserving existing observability behavior
- Preserves legacy `error` key (no wire-protocol break for old clients)

### Issues Found

#### Critical (Must Fix)

None.

| `packages/server/src/ws-permissions.js` | 277-290 | Consider consuming `buildSessionTokenMismatchPayload()` from #2933 instead of duplicating the enrichment | #2933 is open, CI-green, and MERGEABLE. It touches the exact same lines and extracts the shape into a reusable helper. Whichever PR merges second will have a trivial conflict on these lines. If #2933 merges first, this PR needs rebase + 4-line helper swap. If #2936 merges first, #2933 absorbs the duplication cleanly. Both are acceptable — just be aware of the rebase cost. |
| `packages/server/src/ws-permissions.js` | 283 | Minor: `smForLookup` variable is used once and could be inlined as `getSessionManager()?.getSession?.(callerBoundSessionId)` | Style consistency — neighbouring code (line 283, one line below the enrichment) re-calls `getSessionManager()` a second time unconditionally for the SDK-path. Caching once at the top of the handler would be cleaner, but that's a separate refactor not owed by this PR. |

#### Nitpicks (Optional)

- The test titles reference `#2914` but not `#2911` for provenance. Reviewers tracing the UX contract back to the WS-path source would benefit from both refs, though the PR body covers it.
- `boundEntry?.name ?? null` coerces non-string names to `null` only when `name` is `undefined` — a session entry with `name: ''` would send an empty string. The helper in #2933 normalises to `null` via `typeof entry.name === 'string' && entry.name`. Non-issue in practice because SessionManager enforces non-empty names, but #2933's helper is more defensive.

### Architecture Notes

The HTTP 403 body shape mirrors the WS `session_error` payload field-for-field (`code` + `boundSessionId` + `boundSessionName`), while preserving the HTTP-specific `error` key for legacy clients. This intentionally aligns with #2912's stated goal: clients can branch on `code === 'SESSION_TOKEN_MISMATCH'` alone and read the same three fields regardless of transport. The iOS notification-action retry path (HTTP) and the in-app WS path can now share one UX code branch, which was the UX goal of #2904.

The PR body correctly calls out scope — this is not the place to solve #2912 globally. That's #2933's job.

### Deferred Items

None created from this review. The only actionable follow-up (consume the helper from #2933) is already being handled by #2933 itself, which will either land first (forcing a trivial rebase here) or land after (folding this duplication in). Tracking is covered by #2912 and #2933.

### Verdict

**Approve** — Ready to merge. The change is small, well-tested, and matches the WS-path shape it's mirroring. The only coordination cost is with #2933, which is the author's explicit intent (see PR body) — whichever lands first, the other absorbs the conflict cleanly.

Manual test checklist items in the PR body remain worth ticking before shipping to prod, but they're not a prerequisite for merging.

---

_Reviewed against: ES module style, server architecture, WS/HTTP protocol consistency, bound-token security invariants, test coverage for all enrichment branches._

Issue #2956, opened 2026-04-23
feat(server): per-provider model registry driven by supportedModels() instead of global models.js


blamechris:
## Context

Found in [swarm audit `multi-model-support`](../tree/main/docs/audit-results/multi-model-support/00-master-assessment.md). Skeptic + Builder + Operator + Futurist.

\`packages/server/src/models.js\` is structurally Claude-only:
- \`models.js:207\` strips \`'claude-'\` prefix as its ID convention
- \`FALLBACK_MODELS\` ships only sonnet/opus/haiku
- \`resolveContextWindow()\` is Claude-specific
- Currently served to ALL providers via \`ws-history.js:155,170\`

Gemini and Codex never update the registry, so the model dropdown is structurally Claude-only. This blocks #2946 (per-provider model allowlist) and #2948 (mobile dynamic provider list).

## Acceptance criteria

- [ ] Each provider exposes its own \`supportedModels()\` that returns its native model list
- [ ] A registry layer (e.g., \`createModelsRegistry(provider)\`) returns the appropriate per-provider list
- [ ] No global Claude-prefix-stripping; each provider supplies its own ID convention
- [ ] WS messages \`list_models\` etc. take a \`provider\` argument or use the session's bound provider
- [ ] Existing \`models_updated\` event becomes per-provider scoped
- [ ] Tests: assert correct list returned per provider; no cross-contamination

## Audit reference

- \`docs/audit-results/multi-model-support/01-skeptic.md\` — finding #2
- \`docs/audit-results/multi-model-support/06-futurist.md\` — extensibility risk #1

## Related
- #2946 (per-provider ALLOWED_MODEL_IDS — depends on this for proper allowlist source)
- #2948 (mobile providers — UI should consume per-provider model list)

PR #2967, opened 2026-04-23
fix(server): per-provider model allowlist for set_model

### LINKED ISSUE #2946: feat(server): per-provider ALLOWED_MODEL_IDS — set_model crashes Gemini/Codex sessions opaquely
## Context

Found in [swarm audit `multi-model-support`](../tree/main/docs/audit-results/multi-model-support/00-master-assessment.md). Guardian + Operator + Skeptic + Builder all flagged this.

`packages/server/src/handlers/settings-handlers.js:21-37` validates incoming \`set_model\` against a global \`ALLOWED_MODEL_IDS\` (sourced from `models.js:374`) which only contains Claude IDs (sonnet/opus/haiku family).

When a user with an active **Gemini or Codex** session taps the "Sonnet" model chip in the UI:

1. The handler accepts \`'claude-sonnet-4-6'\` because it's in the global allowlist.
2. \`BaseSession.setModel\` sets \`this.model = 'claude-sonnet-4-6'\`.
3. The session respawns with \`gemini -m claude-sonnet-4-6\` or \`codex -m claude-sonnet-4-6\`.
4. The provider CLI exits with an opaque error.

User sees a session crash with no clear cause.

## Acceptance criteria

- [ ] \`set_model\` handler must consult the **session's provider** to determine the allowed model list (not a global Claude-only list)
- [ ] Reject mismatched model IDs with a clear error code (`MODEL_NOT_SUPPORTED_BY_PROVIDER`) and human-readable message identifying the active provider
- [ ] UI's model picker must surface only models valid for the active session's provider (separate issue if needed)
- [ ] Test: send `set_model` with a Claude ID to a Gemini session — expect rejection, not session restart

## Audit reference

- \`docs/audit-results/multi-model-support/03-guardian.md\` — FM-02
- \`docs/audit-results/multi-model-support/05-operator.md\` — finding #2

blamechris:
## Summary

- `set_model` now consults the session's registered provider to determine the allowed model list, instead of hardcoding the global Claude-only `ALLOWED_MODEL_IDS`.
- Mismatched model IDs (e.g. a Claude model sent to a Gemini session) are rejected with `MODEL_NOT_SUPPORTED_BY_PROVIDER` and a human-readable message identifying the active provider.
- `GeminiSession` and `CodexSession` expose a static `getAllowedModels()` with a small hardcoded list. Claude providers (`claude-sdk` / `claude-cli` / `docker-*`) keep using the dynamic global allowlist fed by the Agent SDK.
- Legacy session entries without a `provider` field still resolve against the global list for backward compatibility.

## Why

`set_model` accepted any Claude model ID regardless of the active session's provider. Tapping the Sonnet chip while a Gemini/Codex session was active set `session.model = 'claude-sonnet-4-6'`, respawned the CLI with `-m claude-sonnet-4-6`, and the CLI exited opaquely — the user just saw their session crash with no explanation.

Closes #2946. Issue #2956 tracks replacing the hardcoded per-provider allowlists with a proper registry fed by each CLI's own model list.

## Test plan

- [x] `node --test tests/handlers/settings-handlers.test.js` — all 24 tests pass, including the five new `per-provider allowlist (#2946)` cases.
- [x] Full server suite (`npm test`) — no new failures introduced. The 4 pre-existing failures (TunnelManager integration requiring cloudflared, WebTaskManager requiring the claude CLI, encryption integration test) reproduce on plain `main`.
- [ ] Manual smoke on device: switch to Gemini session, tap Sonnet chip — expect a clear error toast instead of a session crash.

### DIFF
--- a/packages/server/src/codex-session.js
+++ b/packages/server/src/codex-session.js
@@ -64,6 +64,19 @@ export function buildCodexArgs(text, model) {
   return args
 }
 
+// Per-provider model allowlist — #2946.
+// `set_model` must reject a Claude or Gemini model on a Codex session (the
+// CLI would exit opaquely). Keep this list small and explicit; issue #2956
+// tracks a proper registry fed by the Codex CLI itself.
+const CODEX_ALLOWED_MODELS = Object.freeze([
+  'gpt-5-codex',
+  'gpt-5',
+  'gpt-4.1',
+  'gpt-4o',
+  'o1',
+  'o3',
+])
+
 export class CodexSession extends BaseSession {
   static get capabilities() {
     return {
@@ -76,6 +89,15 @@ export class CodexSession extends BaseSession {
       terminal: false,
       thinkingLevel: false,
     }
+  }
+
+  /**
+   * Model IDs this provider accepts in `set_model`. Returns a plain array so
+   * the settings handler can surface it to the client on rejection.
+   * @returns {string[]}
+   */
+  static getAllowedModels() {
+    return CODEX_ALLOWED_MODELS
   }
 
   constructor({ cwd, model, permissionMode } = {}) {
--- a/packages/server/src/gemini-session.js
+++ b/packages/server/src/gemini-session.js
@@ -36,6 +36,19 @@ const GEMINI = resolveBinary('gemini', [
   '/usr/bin/gemini',
 ])
 
+// Per-provider model allowlist — #2946.
+// `set_model` must reject a Claude model on a Gemini session (the CLI would
+// exit opaquely). Keep this list small and explicit; issue #2956 tracks a
+// proper registry fed by `gemini models list` or similar.
+const GEMINI_ALLOWED_MODELS = Object.freeze([
+  'gemini-2.5-pro',
+  'gemini-2.5-flash',
+  'gemini-2.0-pro',
+  'gemini-2.0-flash',
+  'gemini-1.5-pro',
+  'gemini-1.5-flash',
+])
+
 export class GeminiSession extends BaseSession {
   static get capabilities() {
     return {
@@ -48,6 +61,15 @@ export class GeminiSession extends BaseSession {
       terminal: false,
       thinkingLevel: false,
     }
+  }
+
+  /**
+   * Model IDs this provider accepts in `set_model`. Returns a plain array so
+   * the settings handler can surface it to the client on rejection.
+   * @returns {string[]}
+   */
+  static getAllowedModels() {
+    return GEMINI_ALLOWED_MODELS
   }
 
   constructor({ cwd, model, permissionMode } = {}) {
--- a/packages/server/src/handlers/settings-handlers.js
+++ b/packages/server/src/handlers/settings-handlers.js
@@ -6,7 +6,7 @@
  */
 import { ALLOWED_MODEL_IDS, toShortModelId } from '../models.js'
 import { ALLOWED_PERMISSION_MODE_IDS, resolveSession, sendError } from '../handler-utils.js'
-import { listProviders } from '../providers.js'
+import { listProviders, getProvider } from '../providers.js'
 import { createLogger } from '../logger.js'
 
 // Tools that are eligible to be whitelisted via set_permission_rules.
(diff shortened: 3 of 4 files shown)


blamechris:
## Code Review: PR #2967

### Summary

Fixes #2946 — prevents `set_model` from accepting Claude model IDs on non-Claude sessions (Gemini/Codex) which previously caused the provider CLI to exit opaquely. Adds a small opt-in `static getAllowedModels()` escape hatch on provider classes; handler now consults `entry.provider` before falling back to the legacy global `ALLOWED_MODEL_IDS`. Legacy entries with no `provider` field still work.

### Strengths

- **Clean per-provider escape hatch** — `getProviderAllowedModels()` is a tight helper with defensive `try/catch` on `getProvider()` and on the class method call. Returning `null` for "not opted in" is the right sentinel and keeps the Claude-SDK / docker-* paths unchanged.
- **Error surface is user-facing** — the `MODEL_NOT_SUPPORTED_BY_PROVIDER` code carries a human-readable message with both the rejected model and the full supported list. That's directly usable by the app's toast layer.
- **Comment quality** — the handler comments explain *why* the global fallback exists (legacy claude-sdk/docker-* didn't opt in) and that `toShortModelId()` is a safe no-op for non-Claude IDs. Future readers won't be confused.
- **Backwards compat preserved** — the `entry.provider`-absent branch (legacy serialized state) is explicitly tested and flows to the global allowlist.
- **Test coverage is tight** — 6 new tests cover cross-provider rejection (Claude→Gemini, Gemini→Codex), same-provider acceptance (Gemini, Codex, claude-sdk), and the legacy-entry fallback. Request-id plumbing is asserted.
- **Tech debt acknowledged up front** — both session files reference #2956 for the registry-based cleanup, and the PR body says the same. No pretense that the hardcoded list is the final shape.

### Issues Found

#### Critical (Must Fix)

None.

| `packages/server/src/codex-session.js` | 46 | Double-check `'gpt-5.4'` is a real model ID | This is also the pre-existing `DEFAULT_MODEL` and not introduced here, but it reads like a typo (`gpt-4.5`?). Since we're now using this list to *reject* model IDs, one wrong entry silently locks users out. Worth a quick confirmation against whatever model list `codex -m ...` actually accepts. Out of scope if intentional. |
| `packages/server/src/handlers/settings-handlers.js` | 78 | Consider sorting / deduplicating the supported-models list in the error message | Minor UX: `providerAllowed.join(', ')` emits insertion order. Fine today, but if the list grows the message gets unwieldy. A follow-up could truncate to 5 + "and N more" for display. |
| `packages/server/src/handlers/settings-handlers.js` | 37-53 | `getProviderAllowedModels` could live in `providers.js` | The helper encapsulates "given a provider name, return its allowed models or null" — that's provider-registry knowledge. Moving it alongside `getProvider()` would make #2956's registry migration a smaller diff. Not urgent. |

#### Nitpicks (Optional)

- `tests/handlers/settings-handlers.test.js:46` — stray blank line between `makeWs()` and the top-level `describe(...)`.
- The jsdoc for `getAllowedModels()` on both session classes is identical; could be extracted to a comment on the first one and cross-referenced. Trivial.

All valid items above are either nitpicks or already tracked by #2956 (the registry migration this PR explicitly defers to). No new follow-up issues needed.

| Replace hardcoded per-provider allowlists with registry fed by each CLI | [#2956](https://github.com/blamechris/chroxy/issues/2956) | Already tracked. This PR is a minimal fix for the production crash. |
| Move `getProviderAllowedModels` into `providers.js` | Covered by [#2956](https://github.com/blamechris/chroxy/issues/2956) | Natural part of the registry refactor. |

### Architecture Notes

The change respects the existing provider-registry pattern from `providers.js` — providers opt in via a static method, just like `capabilities`. No coupling added between the handler and specific provider classes. `toShortModelId()` remains Claude-specific (alias collapse) but is documented as safe for the non-Claude pass-through case.

One subtle good choice: rejecting `typeof msg.model !== 'string'` up front (before `resolveSession`) avoids a needless session lookup for garbage input. The previous behavior combined both checks into one conditional and silently dropped non-string payloads; the new behavior sends an explicit `INVALID_MODEL` error with the requestId. That's a small protocol improvement on top of the #2946 fix.

Session-binding (`boundSessionId`) is respected transitively through `resolveSession()`, so this handler inherits the existing session-token enforcement. No new auth surface introduced.

### Verdict

Clean, scoped fix for a real user-visible crash. Tech debt is honestly labelled and tracked. Tests are the right shape — they'd catch a regression. No blocking issues.

Issue #2955, opened 2026-04-23
refactor(server): extract JsonlSubprocessSession base from Codex/Gemini sessions (90% duplicate)


blamechris:
## Context

Found in [swarm audit `multi-model-support`](../tree/main/docs/audit-results/multi-model-support/00-master-assessment.md). Consensus finding #5.

\`packages/server/src/codex-session.js\` and \`packages/server/src/gemini-session.js\` are ~90% identical:

- Same spawn pattern
- Same readline-based JSONL stdout reader
- Same stderrBuf accumulation
- Same \`didStreamStart\` gating
- Same close/error handlers

Divergence between the two has caused at least 4 bugs (including the DOA \`gpt-5.4\` model in #2945). Future providers like Aider or local Ollama will multiply the duplication.

## Acceptance criteria

- [ ] Extract a \`JsonlSubprocessSession\` base class in \`packages/server/src/jsonl-subprocess-session.js\` (or \`base-session.js\` if it fits there) containing the shared spawn/readline/error-handling
- [ ] \`codex-session.js\` and \`gemini-session.js\` shrink to ~30-line subclasses that supply only:
  - Binary candidates
  - Default model
  - Argv builder
  - Per-event mapping (Codex/Gemini have different JSONL event shapes)
- [ ] All existing tests still pass
- [ ] Add tests for the new base class covering: spawn lifecycle, readline parsing, stderr capture, error propagation

## Audit reference

- \`docs/audit-results/multi-model-support/04-minimalist.md\` — top-cut #2 (~"180 lines deleted")
- \`docs/audit-results/multi-model-support/02-builder.md\`

PR #2969, opened 2026-04-23
fix(server): replace hallucinated codex DEFAULT_MODEL with CLI-deferred default

### LINKED ISSUE #2945: fix(server): codex-session.js DEFAULT_MODEL is hallucinated 'gpt-5.4'
## Context

Found in [swarm audit `multi-model-support`](../tree/main/docs/audit-results/multi-model-support/00-master-assessment.md). Skeptic + Minimalist independently flagged this.

`packages/server/src/codex-session.js:32` declares:
\`\`\`js
const DEFAULT_MODEL = 'gpt-5.4'
\`\`\`

This model ID does not exist. Codex provider is DOA out of the box — any session created without an explicit model will spawn `codex -m gpt-5.4` which fails immediately. No test catches it.

## Acceptance criteria

- [ ] Replace `'gpt-5.4'` with a real default Codex model ID (or sentinel that defers to whatever Codex CLI's own default is)
- [ ] Add a unit test in `tests/codex-session.test.js` that asserts `DEFAULT_MODEL` is a non-empty string AND ideally that Codex CLI accepts it (use a mocked exec or a smoke test guarded by `if (process.env.CHROXY_HAS_CODEX)`)
- [ ] Audit `gemini-session.js` for the same bug pattern

## Audit reference

- `docs/audit-results/multi-model-support/01-skeptic.md` — finding #2 in Top 5
- `docs/audit-results/multi-model-support/04-minimalist.md` — divergence-bug example in dedup proposal

blamechris:
## Summary

- `packages/server/src/codex-session.js` shipped `DEFAULT_MODEL = 'gpt-5.4'` at line 32, which pinned the server to a specific Codex release and caused `codex exec -c model=\"gpt-5.4\"` to fail on any host whose installed Codex CLI did not recognise that slug. Sessions created without an explicit model were DOA out of the box.
- Fix: drop the hard-coded default. When no model is supplied the session now passes `null` through to `BaseSession`, and the new `buildCodexArgs()` helper omits the `-c model=...` override so Codex CLI falls back to whatever default is configured in `~/.codex/config.toml` for the host.
- Extracted argv construction into an exported pure function (`buildCodexArgs(text, model)`) so the behaviour can be asserted directly without mocking `child_process.spawn` (ESM named-import binding prevents clean monkeypatching).

## Gemini audit

Per the acceptance criteria I also audited `packages/server/src/gemini-session.js`. Its `DEFAULT_MODEL = 'gemini-2.5-pro'` is a real, publicly released Google model, so no change is needed there.

## Test plan

- [x] `node --test tests/codex-session.test.js` — all 46 tests pass (5 new `buildCodexArgs()` cases + updated constructor / setModel / start expectations)
- [x] `node --test tests/providers.test.js tests/gemini-session.test.js` — still green
- [x] `npm test` — 3305 pass, 5 pre-existing failures (2 cloudflared integration tests + 1 WebTaskManager feature-detection suite that also fails on `main`), unrelated to this change
- [x] `npm run lint` — no new warnings

## Acceptance criteria

- [x] `'gpt-5.4'` replaced with a sentinel that defers to Codex CLI's own default
- [x] Unit test added asserting `DEFAULT_MODEL` behaviour and that the spawn argv is well-formed (no `-m` / `-c model=` flag when no model is configured, correct `-c model=\"X\"` when one is)
- [x] `gemini-session.js` audited — no hallucination

Closes #2945

### DIFF
--- a/packages/server/src/codex-session.js
+++ b/packages/server/src/codex-session.js
@@ -29,13 +29,39 @@ const log = createLogger('codex')
  *   error        { message }
  */
 
-const DEFAULT_MODEL = 'gpt-5.4'
+/**
+ * No default model is hard-coded here.
+ *
+ * Previously this module shipped `DEFAULT_MODEL = 'gpt-5.4'`, which pinned
+ * the server to a specific Codex release and caused `codex exec -c model=...`
+ * to fail whenever that version wasn't available on the host. Instead we now
+ * pass `null` through to `BaseSession` when no model is supplied, and
+ * `buildCodexArgs()` below omits the `-c model=...` override so Codex CLI
+ * falls back to whatever default is configured in `~/.codex/config.toml`.
+ */
+const DEFAULT_MODEL = null
 
 const CODEX = resolveBinary('codex', [
   '/opt/homebrew/bin/codex',
   '/usr/local/bin/codex',
   '/usr/bin/codex',
 ])
+
+/**
+ * Build the argv passed to `codex exec`. Exported for unit testing.
+ *
+ * @param {string} text   User prompt
+ * @param {string|null} model  Optional model ID. If falsy, no `-c model=` flag
+ *                              is appended — Codex CLI uses its own default.
+ * @returns {string[]}
+ */
+export function buildCodexArgs(text, model) {
+  const args = ['exec', text, '--json']
+  if (model) {
+    args.push('-c', `model="${model}"`)
+  }
+  return args
+}
 
 export class CodexSession extends BaseSession {
   static get capabilities() {
@@ -52,6 +78,9 @@ export class CodexSession extends BaseSession {
   }
 
   constructor({ cwd, model, permissionMode } = {}) {
+    // `model` may be null/undefined — BaseSession coerces to null and
+    // buildCodexArgs() omits the `-c model=...` flag so Codex CLI defers
+    // to its own default from ~/.codex/config.toml.
     super({ cwd, model: model || DEFAULT_MODEL, permissionMode: permissionMode || 'auto' })
     this.resumeSessionId = null
     this._process = null
@@ -97,10 +126,7 @@ export class CodexSession extends BaseSession {
     this._isBusy = true
     this._currentMessageId = `codex-msg-${++this._messageCounter}`
 
-    const args = ['exec', text, '--json']
-    if (this.model) {
-      args.push('-c', `model="${this.model}"`)
-    }
+    const args = buildCodexArgs(text, this.model)
 
     let stderrBuf = ''
     const proc = spawn(CODEX, args, {
--- a/packages/server/tests/codex-session.test.js
+++ b/packages/server/tests/codex-session.test.js
@@ -6,7 +6,7 @@ import { tmpdir } from 'os'
 import { join } from 'path'
 import { spawn } from 'child_process'
 import { createInterface } from 'readline'
-import { CodexSession } from '../src/codex-session.js'
+import { CodexSession, buildCodexArgs } from '../src/codex-session.js'
 import { waitFor } from './test-helpers.js'
 
 // ---------------------------------------------------------------------------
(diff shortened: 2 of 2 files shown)


blamechris:
## Code Review: PR #2969

### Summary

Replaces the hallucinated `DEFAULT_MODEL = 'gpt-5.4'` sentinel in `codex-session.js` with `null`, which propagates through `BaseSession` so that when no model is supplied, `buildCodexArgs()` omits the `-c model=...` override entirely. Codex CLI then falls back to whatever default is configured in `~/.codex/config.toml`. Extracts argv construction into an exported pure function for direct unit testing without having to mock `child_process.spawn`. Net effect: Codex provider is no longer DOA out of the box for users whose installed Codex release predates the fictional `gpt-5.4`.

### Strengths

- **Correct defer-to-CLI semantics.** Omitting `-c model=...` is exactly right — Codex CLI's `-c KEY=VALUE` flag is a TOML override; when absent, Codex reads `~/.codex/config.toml` and then its built-in default. Much better than picking another possibly-wrong hard-coded default.
- **Pure `buildCodexArgs()` helper.** No side effects, no external refs, fully exported. Directly testable without any `spawn` mocking — which sidesteps the ESM named-import binding limitation cleanly. Good separation of concerns.
- **Test coverage is strong.** Five new `buildCodexArgs()` cases (null, empty string, explicit model, hallucinated-guard, argv-prefix shape) plus regression guards on constructor, `start()` ready payload, and `setModel()` when busy. Including an `assert.notEqual(session.model, 'gpt-5.4')` assertion bakes the historical bug into the test suite so future regressions are caught.
- **Good explanatory comments.** The block comment on `DEFAULT_MODEL = null` and the reminder comment in the constructor both explain *why*, not just *what*. Anyone reading this in six months won't "helpfully" restore a default.
- **Gemini audit done.** `gemini-2.5-pro` is a real, publicly released Google model — confirmed. PR description explicitly documents that no change was needed, which matches issue #2945's acceptance criteria.
- **Zero-impact contract change.** `ready` event now carries `model: null` when no model is supplied. Since the old behaviour (`gpt-5.4`) always errored before `ready` ever fired, no real-world clients depended on the non-null invariant.

### Issues Found

#### Critical (Must Fix)

None.

| `packages/server/src/codex-session.js` | 61 | Strip embedded double-quotes from `model` in `buildCodexArgs` before interpolating. | The argv builds `model="${model}"`, so any `"` in the ID would produce ambiguous TOML (`model="foo"bar"`). Codex model IDs don't realistically contain `"`, but a one-line `String(model).replace(/"/g, '')` (or a narrow `/^[a-zA-Z0-9._:-]+$/` validation + throw) makes this robust against future weirdness (external config files, user-pasted values). Nitpick — the current code is fine for every documented Codex model slug. |

#### Nitpicks (Optional)

- `server-cli.js:264` logs `Session X ready: Y (model: ${data.model})`, which will now render as `(model: null)` when no model is configured. Cosmetic only — consider `(model: ${data.model ?? 'cli default'})` or similar if/when the logger is touched. Not worth a separate fix in this PR.

None — the only suggestion is a robustness nitpick; I'm not filing an issue for it because the input domain (Codex model IDs) is controlled.

### Architecture Notes

The change preserves the EventEmitter provider contract (capabilities, events, shape) — only the *value* of `model` in the `ready` payload can now be `null` for Codex. `session-manager.js` proxies `ready` through as a `session_event`, so clients that look at `data.model` need to tolerate `null`. The dashboard/app don't do so strictly (the string is only logged/displayed), so this is safe. `BaseSession.setModel()` already handles `model ? resolveModelId(model) : null`, so `null` round-trips cleanly through the setter.

Exporting `buildCodexArgs` does expand the module's public surface by one function — trivially, and genuinely useful — so no concerns there. The pattern is worth copying to the similar `gemini-session.js` argv construction if/when that grows unit tests (would close part of the duplication already tracked in #2955).

### Verdict

- [x] **Approve** — Ready to merge. Targeted, well-documented, well-tested fix for a DOA regression. No blocking issues.

PR #2978, opened 2026-04-23
security(server): redact OpenAI/Anthropic/Google API keys in stderr

### LINKED ISSUE #2961: security(server): redactSensitive must catch 'sk-...' and 'AIza...' API key patterns in stderr
## Context

Found in [swarm audit `multi-model-support`](../tree/main/docs/audit-results/multi-model-support/00-master-assessment.md). Guardian P1.

When Codex or Gemini child processes panic, they may write API keys to stderr (e.g., echoing the key back as part of an "invalid key" error). Current \`redactSensitive\` likely doesn't catch:

- OpenAI keys: \`sk-...\` (typically 51+ chars)
- Google API keys: \`AIza[A-Za-z0-9_-]{35}\`
- Anthropic keys: \`sk-ant-api03-...\`

These can leak into logs that ship to telemetry / get pasted into bug reports.

## Acceptance criteria

- [ ] \`redactSensitive\` (or equivalent — find via Grep) extends regex to cover OpenAI, Google, Anthropic key patterns
- [ ] Applied to all stderr capture paths for Codex/Gemini/Claude
- [ ] Tests: stderr containing each key type → output redacted as \`[REDACTED]\`
- [ ] Audit log redaction too if separate

## Audit reference

- \`docs/audit-results/multi-model-support/03-guardian.md\` — P1 mitigations

blamechris:
## Summary

**Security fix** — extend \`redactSensitive\` (server logger) to match provider API key patterns that can leak via child-process stderr when Codex/Gemini/Claude panics (e.g., echoing the key back in an \"invalid key\" error). Keys land in:

- chroxy.log on disk
- the dashboard log stream (WS log broadcast)
- provider \`error\` events forwarded to clients

Without this, a misconfigured API key could end up pasted into a bug report or shipped to telemetry.

## What's caught now

- Anthropic: \`sk-ant-(api03-)?[A-Za-z0-9_-]{40,}\`
- OpenAI project-scoped: \`sk-proj-[A-Za-z0-9_-]{40,}\`
- OpenAI legacy: \`sk-(?!ant-|proj-)[A-Za-z0-9]{40,}\` (negative lookahead keeps longer Anthropic/project prefixes from being partially matched)
- Google: \`AIza[A-Za-z0-9_-]{35}\` (exact length; \`\\b\` terminator avoids matching \"AIzawa\"-like words)

Length floors (40+ trailing chars for \`sk-\`, exactly 35 for \`AIza\`) avoid false positives on product SKUs, \`sku-\` identifiers, and short \`sk-\`-prefixed strings.

Keys are replaced with bare \`[REDACTED]\` regardless of surrounding key=value syntax, so they're caught whether they appear mid-sentence (\`invalid api key sk-...\`) or embedded in JSON (\`\"key\":\"sk-...\"\`).

## Stderr paths

- \`cli-session.js\` / \`docker-sdk-session.js\` — stderr goes through \`log.info\`/\`log.error\`, which already calls \`redactSensitive\` in \`createLogger.write()\`. No changes needed.
- \`codex-session.js\` / \`gemini-session.js\` — stderr also logs through the logger **and** the exit-code error message embeds \`stderrBuf.slice(0, 500)\` directly into \`emit('error', ...)\`, which is broadcast to WS clients without going through the logger. Added explicit \`redactSensitive()\` wrap at those emission sites.

## Test plan

- [x] New test cases for each provider key format (OpenAI legacy, sk-proj-, sk-ant-api03-, AIza)
- [x] Negative tests: \`SKU: ABC-1234\`, \`sk-demo\`, \`AIza123\` (too short) pass through unchanged
- [x] Multiple keys in one message all redacted
- [x] Keys embedded in JSON-like stderr redacted
- [x] End-to-end test: key passed through \`createLogger().error()\` is redacted before reaching listeners
- [x] All 44 logger tests pass; codex-session and gemini-session test suites still pass

Closes #2961

### DIFF
--- a/packages/server/src/codex-session.js
+++ b/packages/server/src/codex-session.js
@@ -3,7 +3,7 @@ import { BaseSession } from './base-session.js'
 import { createInterface } from 'readline'
 import { resolveBinary } from './utils/resolve-binary.js'
 import { buildSpawnEnv } from './utils/spawn-env.js'
-import { createLogger } from './logger.js'
+import { createLogger, redactSensitive } from './logger.js'
 
 const log = createLogger('codex')
 
@@ -274,7 +274,7 @@ export class CodexSession extends BaseSession {
         this.emit('stream_end', { messageId: this._currentMessageId })
       }
       if (code !== 0 && code !== null) {
-        const detail = stderrBuf ? `: ${stderrBuf.slice(0, 500)}` : ''
+        const detail = stderrBuf ? `: ${redactSensitive(stderrBuf.slice(0, 500))}` : ''
         this.emit('error', { message: `Codex process exited with code ${code}${detail}` })
       }
       // Emit result only if turn.completed wasn't received
--- a/packages/server/src/gemini-session.js
+++ b/packages/server/src/gemini-session.js
@@ -3,7 +3,7 @@ import { BaseSession } from './base-session.js'
 import { createInterface } from 'readline'
 import { resolveBinary } from './utils/resolve-binary.js'
 import { buildSpawnEnv } from './utils/spawn-env.js'
-import { createLogger } from './logger.js'
+import { createLogger, redactSensitive } from './logger.js'
 
 const log = createLogger('gemini')
 
@@ -208,7 +208,7 @@ export class GeminiSession extends BaseSession {
         this.emit('stream_end', { messageId: this._currentMessageId })
       }
       if (code !== 0 && code !== null) {
-        const detail = stderrBuf ? `: ${stderrBuf.slice(0, 500)}` : ''
+        const detail = stderrBuf ? `: ${redactSensitive(stderrBuf.slice(0, 500))}` : ''
         this.emit('error', { message: `Gemini process exited with code ${code}${detail}` })
       }
       // Emit result so clients transition from busy to idle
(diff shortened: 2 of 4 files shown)


blamechris:
## Secret-scanning false positive on this branch

GitHub flagged alert [#1](https://github.com/blamechris/chroxy/security/secret-scanning/1) (Mailgun API Key) on `c24bd206f`. The trigger is the fake Anthropic fixture on `packages/server/tests/logger.test.js:610`:

```js
const key = 'sk-ant-api03-' + 'L1e2a3k4_test_key-12345678901234567890123456789012'
```

The suffix `key-12345…012` incidentally matches Mailgun's regex (`key-` + 32 alphanumerics). Not a real secret.

**Already resolved** as `false_positive` with a pointer back to this PR.

**Recommended fixture rename** to avoid the same regex match re-triggering if Mailgun's pattern or the scanner posture changes later:

```diff
-    const key = 'sk-ant-api03-' + 'L1e2a3k4_test_key-12345678901234567890123456789012'
+    const key = 'sk-ant-api03-' + 'L1e2a3k4_testfixture_NOT_REAL_00000000000000000000'
```

Equivalent length, still clearly a fixture, no longer matches the `key-` prefix pattern. Non-blocking — the alert is closed — but worth folding into this PR's next commit since it's the same file the PR already touches.

PR #2994, opened 2026-04-25
refactor(desktop): polish build_enriched_path — dedup, substring match, tests

### LINKED ISSUE #2939: desktop: polish build_enriched_path — dedup once, tighten substring match, add missing tests
## Context

Identified during review of PR #2926 (which closes #2893). None of these are blockers — the PR merges as-is. This issue tracks three small polish items that can land together as a follow-up.

### 1. Reuse `dirs::home_dir()` instead of calling it twice

In `ServerManager::start_server_process()`, `dirs::home_dir()` is called twice — once to pass into `build_enriched_path`, and again to set the `HOME` env var:

```rust
let home_dir = dirs::home_dir();
// ...
let full_path = build_enriched_path(&base_path, &node_bin, home_dir.as_deref(), path_sep);
cmd.env(\"PATH\", &full_path);
// Ensure HOME is set — macOS GUI apps may not inherit it
if let Some(home) = dirs::home_dir() {
    cmd.env(\"HOME\", home);
}
```

Reuse the already-computed `home_dir` so both env vars derive from the same resolved home directory:

```rust
let home_dir = dirs::home_dir();
// ...
let full_path = build_enriched_path(&base_path, &node_bin, home_dir.as_deref(), path_sep);
cmd.env(\"PATH\", &full_path);
if let Some(ref home) = home_dir {
    cmd.env(\"HOME\", home);
}
```

(Also flagged by Copilot in the PR #2926 review.)

### 2. Tighten substring dedup to element-level

The current dedup check uses `base_path.contains(dir)`:

```rust
if !base_path.contains(dir) && !dirs.iter().any(|d| d == dir) {
    dirs.push((*dir).to_string());
}
```

This has two false-positive cases:

- `base_path = \"/sbin\"` → `base_path.contains(\"/bin\")` is `true` because `/bin` is a substring of `/sbin`. The `/bin` directory is incorrectly skipped.
- `base_path = \"/opt/homebrew/bin/extra\"` (unlikely but possible) → `base_path.contains(\"/opt/homebrew/bin\")` is `true`. Homebrew is incorrectly skipped.

Replace with element-level split on `path_sep`:

```rust
let base_entries: Vec<&str> = base_path.split(path_sep).collect();
let already_present = |candidate: &str| base_entries.iter().any(|e| *e == candidate);
```

Then use `!already_present(dir)` in place of `!base_path.contains(dir)`.

The rustdoc currently calls out the substring behavior as \"conservative (substring match) rather than element-level parsing, which matches the historical behavior\" — update the rustdoc once the fix lands.

### 3. Add missing tests

Two edge cases the test suite does not currently cover:

- **User install dir already in base_path is deduped.** Symmetric to `build_enriched_path_skips_system_dir_already_in_base_path` but for `~/.local/bin` / `~/.npm-global/bin`. Pass `base_path` containing `/Users/alice/.local/bin` and assert it is not duplicated.
- **`node_bin` equal to a system bin.** E.g. `node_bin = \"/usr/bin\"` — should `/usr/bin` appear once or twice in the output? Decide the contract and pin it down. Not harmful either way (PATH search stops at first hit), but clarifies behavior.

## Original Review Comment

> Reuse `dirs::home_dir()` result + tighten dedup logic (element-level instead of substring) + add missing dedup / node_bin==system_bin tests. None are blockers. The substring-match semantics are explicitly called out in the rustdoc as \"conservative\" and match historical behavior; the double `home_dir()` call is purely cosmetic; the two missing tests are nice-to-have. All three can land together as a small polish PR.

## Acceptance Criteria

- [ ] `dirs::home_dir()` called once in `start_server_process()`
- [ ] Substring dedup replaced with element-level split; rustdoc updated
- [ ] Test added for user install dir already present in `base_path`
- [ ] Test added for `node_bin` equal to a system bin (whatever contract is chosen, pinned down)
- [ ] `cargo test` passes in `packages/desktop/src-tauri`

## References

- PR #2926 — original refactor
- Issue #2893 — defense-in-depth rationale

blamechris:
## Summary

- Replaces substring-based PATH dedup with element-level split on `path_sep`, fixing the false-positive where `/sbin` in `base_path` would suppress `/bin` from being prepended
- Consolidates the user install dir loop inside a single `#[cfg(unix)]` block, sharing the `base_entries` vec and `already_present` closure with the system-bin loop
- Updates rustdoc to reflect the new element-level semantics (removes the "conservative substring match" caveat)
- Adds three new tests pinning the fixed behavior:
  - `build_enriched_path_skips_user_install_dir_already_in_base_path` — symmetric to the existing system-dir dedup test, for `~/.local/bin` / `~/.npm-global/bin`
  - `build_enriched_path_node_bin_equals_system_bin_appears_once` — contract: `/usr/bin` as `node_bin` appears exactly once
  - `build_enriched_path_sbin_does_not_suppress_bin` — regression guard for the substring false-positive

Note: item 1 from the issue (`dirs::home_dir()` called once) was already addressed in the codebase before this PR.

## Test plan

- [x] `cargo test` in `packages/desktop/src-tauri` — 119 tests pass, 0 failures
- [x] All three new tests pass on first run

Closes #2939


blamechris:
## Code Review: PR #2994

### Summary

This PR fixes a correctness bug in `build_enriched_path` where substring-based dedup (`.contains(dir)`) would suppress `/bin` when `base_path` contained `/sbin`. The fix switches to element-level dedup by splitting `base_path` on `path_sep` and doing exact string comparison. A secondary cleanup merges the two `#[cfg(unix)]` blocks into one, sharing the `base_entries` vec and `already_present` closure. Three targeted regression tests pin the new behavior.

### Strengths

- **Correct fix for the reported bug.** The `/sbin` suppressing `/bin` case is a real problem on systems where `/sbin` appears in `$PATH` but `/bin` does not — the subprocess would silently lack `/bin`. Element-level dedup is the right solution.
- **Dedup closure is shared cleanly.** Extracting `base_entries` and `already_present` at the top of the `#[cfg(unix)]` block lets both loops (system bins + user install dirs) share exactly the same logic with no duplication.
- **Tests are precise and well-documented.** Each test has a doc comment explaining *what* it contracts, and assertions use `split(':').filter()` element-by-element counts rather than substring-match — which is the right pattern given the substring-match was exactly the bug being fixed.
- **No behavior change on Windows.** The `#[cfg(windows)] let _ = home_dir;` suppression is correctly preserved and its comment updated to explain *why*.
- **All 12 `build_enriched_path` tests pass** (verified locally: `cargo test --lib build_enriched_path` → 12 ok).
- **Rustdoc updated to match.** The old "conservative substring match" caveat is removed and replaced with accurate semantics.

### Issues Found

#### Critical (Must Fix)

None.

| `server.rs` L1287 | existing test | `build_enriched_path_skips_system_dir_already_in_base_path` still uses `.matches(...).count()` (substring) for its assertion | Inconsistency: the old test uses substring match for the count, while the new tests correctly use `.split(':').filter()`. A path like `/opt/homebrew/bin-extra` would make the old test give a false count. Minor — the scenario is contrived — but worth aligning for consistency. |

#### Nitpicks (Optional)

- The new `already_present` closure captures `base_entries` by reference. This is fine and idiomatic for a closure used immediately in the same scope. No issue.
- `(*dir).to_string()` in the system-bins loop is unchanged from before — `dir.to_string()` would work too since `dir: &&str` auto-derefs, but this is a pre-existing style point.

| Align `build_enriched_path_skips_system_dir_already_in_base_path` assertion to use element-level count | https://github.com/blamechris/chroxy/issues/3013 | Pre-existing test; fixing it is orthogonal to this PR's scope. The current assertion is technically correct for the test's inputs but inconsistent with the new tests' pattern. Low risk, good hygiene. |

### Architecture Notes

Change is entirely contained to `server.rs` within the `build_enriched_path` pure function and its tests. No WS protocol, no Tauri commands, no state machine — zero risk of regression outside this function. The `#[cfg(unix)]` consolidation is a net simplification: one block, one shared closure, two loops. Correct approach.

### Verdict

**Approve — Ready to merge.**

The bug fix is correct, the dedup logic is tight, tests are meaningful, and the code is cleaner than before. The one follow-up (test assertion style alignment) is tracked as #3013.

blamechris:
## Review Comments Addressed

| — | Copilot review (no inline comments) | No action needed | Copilot summary was positive, no issues flagged |

**Total:** 0 inline comments to process

Copilot's review contained only a high-level summary describing the PR changes with no issues, suggestions, or inline code comments. Nothing to address.

PR #2996, opened 2026-04-25
fix(server): clean up orphaned .tmp files in environment-manager

### LINKED ISSUE #2940: environment-manager: clean up orphaned .tmp files after rename failure (same bug as #2909)
## Context

Identified during review of PR #2930 (which fixed the same class of bug in `session-state-persistence.js`, closing #2909).

## Description

`EnvironmentManager._persist()` at `packages/server/src/environment-manager.js:921-928` uses the same temp-file-plus-rename pattern as `SessionStatePersistence`:

```js
const tmpPath = this._statePath + '.tmp'
writeFileRestricted(tmpPath, JSON.stringify(data, null, 2))
if (isWindows) {
  try { unlinkSync(this._statePath) } catch (e) {
    if (e && e.code !== 'ENOENT') log.error(`Failed to remove existing state file: ${e.message}`)
  }
}
renameSync(tmpPath, this._statePath)
```

If the final `renameSync(tmpPath, this._statePath)` throws (disk full, permission change, transient FS error, locked destination on Windows), the `.tmp` file is leaked on disk. The outer `try/catch` in `_persist` catches the error and logs it, but it does **not** clean up the orphaned `.tmp`.

PR #2930 fixed this for session state. The same fix should be applied here. `models.js:saveCache` (lines 344-363) already does the right thing — use that or PR #2930's pattern as reference.

## Proposed fix

Wrap the final `renameSync` in an inner `try/catch` that best-effort `unlinkSync(tmpPath)`s on failure, suppressing `ENOENT`, then re-throws (or falls through to the outer catch that logs and swallows — matching the current `_persist` contract that errors are non-fatal).

## Acceptance criteria

- [ ] Failed `_persist` no longer leaks `.tmp` files
- [ ] Existing error-swallowing behavior preserved (errors logged, not thrown to caller)
- [ ] Regression test covering the cleanup path, mirroring the approach in `session-state-persistence.test.js` (pre-create `<statePath>` as a non-empty directory to force `EISDIR`/`ENOTEMPTY`)

## Reference

- `packages/server/src/environment-manager.js:921-928`
- Pattern to copy: `packages/server/src/session-state-persistence.js:52-64` (post PR #2930)
- Alternative reference: `packages/server/src/models.js:344-363`

blamechris:
## Summary

- `EnvironmentManager._persist()` used the same temp-file-plus-rename pattern as `SessionStatePersistence`, but did not clean up the `.tmp` file if `renameSync()` threw (disk full, permission change, transient FS error, locked destination on Windows).
- Wrap the final `renameSync` in an inner `try/catch`: best-effort `unlinkSync(tmpPath)` on failure, suppress `ENOENT`, log any secondary unlink error, then re-throw so the outer catch can log and swallow it — preserving the existing non-fatal `_persist` contract.
- Mirrors the fix from #2930 (`session-state-persistence.js`) exactly.

## Test plan

New describe block `EnvironmentManager._persist() rename-failure cleanup (regression: #2940)` in `packages/server/tests/environment-manager.test.js`:

- [x] RED verified: 2 of 3 tests fail before the fix (orphaned `.tmp` remains)
- [x] GREEN verified: all 3 new cases pass after the fix
  - removes the orphaned `.tmp` file when the final rename fails
  - swallows the rename error (existing `_persist` error-logging contract preserved)
  - does not leak `.tmp` across repeated rename failures
- [x] No regressions in other `environment-manager.test.js` cases

Rename failure induced portably by pre-creating `<statePath>` as a non-empty directory (POSIX `rename` into a non-empty dir → `ENOTEMPTY`/`EISDIR`).

Closes #2940


blamechris (commented):
## Agent Review

**Verdict: LGTM — no changes required.**

### Summary

This is a clean, minimal fix that closes the .tmp leak in `EnvironmentManager._persist()` — the exact parallel of the same class of bug fixed in `SessionStatePersistence` by #2930. The implementation, comment style, and test structure all match the established pattern faithfully.

---

### Checklist

**Pattern consistency with #2930**

The `session-state-persistence.js` fix wraps `renameSync` in an inner `try/catch`, best-effort `unlinkSync(tmpPath)` on failure (suppress ENOENT, warn on secondary errors), then re-throws. This PR does exactly the same. Variable names (`renameErr`, `cleanupErr`), log level (`log.warn` for secondary), and comment phrasing all mirror #2930. Pass.

**Error rethrow preservation**

The outer `catch (err)` in `_persist()` logs and swallows — preserving the existing non-fatal contract. The inner `throw renameErr` correctly routes the rename failure to that outer handler without escaping. The test `swallows the rename error` explicitly asserts `doesNotThrow`. Pass.

**ENOENT handling**

`cleanupErr.code !== 'ENOENT'` guard is correct — if the `.tmp` was already gone (concurrent process cleaned it), that is fine and should not warn. Matches #2930 exactly. Pass.

**Windows code path**

The Windows `unlinkSync(this._statePath)` block runs *before* the `try { renameSync(...) }`, so rename failure on Windows also benefits from cleanup. The orphaned `.tmp` scenario applies equally on Windows (arguably more common there given file-locking semantics). Pass.

**Test approach**

`blockRename()` creates `statePath` as a non-empty directory to force ENOTEMPTY/EISDIR on `renameSync` — portable across POSIX without needing root. Mirrors the same technique in `session-state-persistence.test.js`. Three cases cover: (1) cleanup on first failure, (2) non-throw guarantee, (3) no accumulation across repeated failures. Constructor injection (`{ statePath, _execFile: () => {} }`) avoids touching real state. Pass.

**No stale imports**

`unlinkSync` was already imported at line 4 of `environment-manager.js` (used in the Windows pre-unlink path). No import changes needed. The test file correctly adds `mkdirSync` and `writeFileSync` to its named imports. Pass.

**Scope**

Change is minimal — 12 lines in `_persist()`, 50 lines of tests. No unrelated modifications. Pass.

---

### Findings

- 0 critical
- 0 suggestions
- 0 false positives
- 0 deferred follow-ups

CI: all checks pass (Server Tests, Server Lint, App Tests, App Type Check, Dashboard Tests, Protocol Tests, Store Core Tests, Desktop Rust Tests — all green).

blamechris:
## Review Comments Addressed

No inline review comments were found — Copilot reviewed all 2 changed files and generated no comments. All CI checks pass.

- Fixed: 0
- False positives: 0
- Follow-up issues created: 0
- Existing issues to close: #2940 (will auto-close on merge via closing keyword in PR description)

PR #2998, opened 2026-04-25
fix(dashboard): respect prefers-reduced-motion on tunnel-warming banner

### LINKED ISSUE #2941: dashboard: respect prefers-reduced-motion on tunnel-warming banner opacity transition
## Context

Identified during review of PR #2934 (fixes #2915 — tunnel-warming banner layout shift).

## Description

The tunnel-warming banner now uses a 120ms opacity transition to fade between the visible warming state and the hidden placeholder state:

\`\`\`css
.tunnel-warming-banner {
  transition: opacity 120ms ease-out;
}
\`\`\`

Other animated elements in \`packages/dashboard/src/theme/components.css\` respect the user's \`prefers-reduced-motion\` preference — for example:

- \`.btn-mic.recording\` animation is disabled via \`@media (prefers-reduced-motion: reduce)\` at line 2187.
- \`.skeleton-line\`, \`.sidebar-busy-dot\`, \`.sidebar-idle-dot\` animations are disabled at line 3693.

The tunnel-warming banner should be added to that list so users who opt out of motion don't see the fade.

## Original Review Comment

> Consider honoring \`prefers-reduced-motion\` for the opacity transition. The file already respects this media query for \`.btn-mic.recording\` and \`.skeleton-line\`. Consistency is worth a small follow-up. The 120ms transition is brief so impact is low.

## Proposed Fix

Extend the existing \`@media (prefers-reduced-motion: reduce)\` block (or add a new entry) to set \`transition: none\` on \`.tunnel-warming-banner\`:

\`\`\`css
@media (prefers-reduced-motion: reduce) {
  .tunnel-warming-banner { transition: none; }
}
\`\`\`

## Acceptance Criteria

- [ ] \`.tunnel-warming-banner\` has no opacity transition when \`prefers-reduced-motion: reduce\` is set.
- [ ] The hidden ↔ visible toggle still works correctly (just instant instead of fading).
- [ ] No regression to existing reduced-motion overrides.

blamechris:
## Summary

- Adds `.tunnel-warming-banner { transition: none }` inside the existing `@media (prefers-reduced-motion: reduce)` block in `packages/dashboard/src/theme/components.css`, consistent with the overrides already present for `.btn-mic.recording` and the skeleton/busy-dot elements.
- Adds a test in `theme.test.ts` that reads `components.css` at runtime, locates all `prefers-reduced-motion: reduce` media blocks, and asserts the banner override is present.

## Test plan

- [ ] `npm test` in `packages/dashboard` — all 1235 tests pass including the new one
- [ ] No regression to existing reduced-motion overrides (`.btn-mic.recording`, `.skeleton-line`, etc.)
- [ ] Banner hidden ↔ visible toggle still works (instant instead of fading when preference is set)

Closes #2941


blamechris:
## Code Review: PR #2998

### Summary

Targeted accessibility fix: adds `transition: none` for `.tunnel-warming-banner` inside the existing `@media (prefers-reduced-motion: reduce)` block in `components.css`, and backs it with a CSS-parsing test that locates all `prefers-reduced-motion: reduce` media blocks at runtime and asserts the override is present.

This closes the gap introduced in PR #2934 where the 120ms opacity transition on the banner was left out of the existing motion-reduction overrides.

---

### Strengths

- **Correct placement.** The override goes into the second `@media (prefers-reduced-motion: reduce)` block at line 3693, which already owns the bulk of motion overrides for animation-bearing elements. Putting a `transition: none` rule *before* the `animation: none` rules keeps transition and animation overrides co-located in one block rather than scattered.
- **Consistent pattern.** Three existing motion overrides in the file (`.btn-mic.recording` at 2188, `.skeleton-line` et al. at 3696, `.loading-spinner` at 3721) all use the same `property: none` idiom. The new rule follows the same convention without introducing anything novel.
- **Minimal scope.** The fix is 2 lines in CSS. No JS, no React changes, no risk of regressions outside this selector.
- **Test catches real regressions.** A future dev who accidentally deletes the block or moves the banner into a new CSS file without carrying the override will get an immediate red test. That's the correct failure mode for a rule that exists solely to satisfy a media-query contract.
- **All CI green** — 12/12 checks pass including Dashboard Tests.

---

### Issues Found

#### Critical (Must Fix)

_None._

| `theme.test.ts` | 329–332 | The two `includes()` calls check the whole merged block string but do **not** verify the two tokens appear inside the *same rule*. A file where `transition: none` exists in one rule and `.tunnel-warming-banner` appears in another (different) rule within the same media block would pass. | Risk is low today (the block is short and easy to read), but a more precise assertion — e.g. matching a regex like `/\.tunnel-warming-banner\s*\{[^}]*transition:\s*none/` against the block — would be unambiguous. |
| `theme.test.ts` | 306 | The `describe` block is named `'components.css reduced-motion overrides'` but only tests one selector. Future reduced-motion additions (if tracked) would fit naturally here rather than in a new `describe`. Consider a brief inline comment noting the block is the canonical home for such tests. | Purely organisational; no behaviour impact. |

#### Nitpicks

- The new rule uses `{ transition: none; }` on one line while the surrounding rules use multi-line blocks. Both styles appear in the file, so this isn't a consistency violation — just noting it in case the author prefers uniform formatting.
- A blank line separates `.tunnel-warming-banner { transition: none; }` from the selector group that follows. That visual separation is helpful and intentional (transition override vs. animation overrides), but adding a `/* transition overrides */` / `/* animation overrides */` comment pair would make the distinction explicit for future readers.

---

| Tighten test assertion to use a regex that co-locates selector + property within the same rule | [#2999](https://github.com/blamechris/chroxy/issues/2999) | Out of scope for this one-rule fix; test is already useful as-is |

---

### Architecture Notes

This change is entirely within `packages/dashboard/src/theme/`, which is the correct boundary — CSS motion overrides live alongside the rest of the visual token system. No server, protocol, or app-layer changes. The test pattern (read CSS at runtime, parse media blocks, assert rule presence) is already established in the same file for global.css existence checks; applying it to a component-level rule is consistent.

---

### Verdict

- [x] **Approve** — Ready to merge. The fix is correct, minimal, well-tested, and consistent with existing patterns. No blocking issues.

blamechris:
**Correction to deferred item above:** Follow-up issue for tightening the test assertion created as [#3012](https://github.com/blamechris/chroxy/issues/3012) (not #2999 as listed in the table above — issue was created after the review comment was posted).

PR #3003, opened 2026-04-25
fix(app): timeout for 'Loading providers...' placeholder in CreateSessionModal

### LINKED ISSUE #2975: CreateSessionModal: 'Loading providers...' placeholder lingers forever on non-responsive servers
## Context

Identified during review of PR #2968.

## Description

`packages/app/src/components/CreateSessionModal.tsx` renders a \"Loading providers...\" hint whenever `availableProviders.length === 0`. After #2968 this means:

- If the server never replies to `list_providers` (bug, old server, or connection drop), the placeholder stays indefinitely.
- If the server legitimately returns an empty provider list (edge case, but possible), the user sees \"Loading...\" forever even though the load has completed.

Dashboard has similar behavior but has more signals available (settings panel, logs) so the issue is less visible there.

## Original Review Comment

> `availableProviders.length === 0` renders \"Loading providers...\" forever if the server never responds (or the user stays on a Claude-only server where the list has legitimately loaded empty). After ~3-5s with no response it would be friendlier to stop saying \"Loading\" and either (a) silently drop the hint, or (b) switch it to a neutral \"No other providers available.\"

## Acceptance Criteria

- [ ] Track \"providers request sent but not yet answered\" as an explicit flag (not inferred from empty array)
- [ ] Show \"Loading providers...\" only while the request is in flight
- [ ] After a reasonable timeout (e.g. 5s) or on empty-but-answered response, either hide the hint or swap to \"No additional providers available\"
- [ ] The Default chip still renders and remains functional in all cases

blamechris:
## Summary

- Tracks the `list_providers` request with an explicit `providersLoading` flag instead of inferring loading state from an empty `availableProviders` array
- After 5 seconds with no server response, replaces the "Loading providers…" hint with "No additional providers available" and a Retry button
- Loading state resets every time the modal opens; cancels on close; clears immediately if providers arrive before timeout
- The Default chip always renders and remains functional in all states

## Test plan

- [x] 6 new tests in `CreateSessionModal.test.tsx` covering: initial loading hint, 5s timeout → empty state, providers arriving before timeout, modal re-open resets state, retry button presence, retry button invokes `fetchProviders` and resets to loading
- [x] All 1117 app tests pass (`npm test`)

Closes #2975

### DIFF
--- a/packages/app/src/components/CreateSessionModal.tsx
+++ b/packages/app/src/components/CreateSessionModal.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect, useRef } from 'react';
+import React, { useState, useEffect, useRef, useCallback } from 'react';
 import {
   View,
   Text,
@@ -40,7 +40,7 @@ export function CreateSessionModal({ visible, onClose }: CreateSessionModalProps
   const [providersTimedOut, setProvidersTimedOut] = useState(false);
   const providersTimeoutRef = useRef<ReturnType<typeof setTimeout> | null>(null);
 
-  const startProvidersTimeout = () => {
+  const startProvidersTimeout = useCallback(() => {
     if (providersTimeoutRef.current) {
       clearTimeout(providersTimeoutRef.current);
     }
@@ -50,14 +50,15 @@ export function CreateSessionModal({ visible, onClose }: CreateSessionModalProps
       setProvidersLoading(false);
       setProvidersTimedOut(true);
     }, PROVIDERS_TIMEOUT_MS);
-  };
+  }, []);
 
-  const handleRetryProviders = () => {
+  const handleRetryProviders = useCallback(() => {
     fetchProviders();
     startProvidersTimeout();
-  };
+  }, [fetchProviders, startProvidersTimeout]);
 
   // Reset state when modal opens and refresh provider list from server.
+  // Cancel timeout and clear loading state when modal closes or on unmount.
   useEffect(() => {
     if (visible) {
       setShowBrowser(false);
@@ -66,7 +67,6 @@ export function CreateSessionModal({ visible, onClose }: CreateSessionModalProps
       fetchProviders();
       startProvidersTimeout();
     } else {
-      // Clean up timeout when modal closes
       if (providersTimeoutRef.current) {
         clearTimeout(providersTimeoutRef.current);
         providersTimeoutRef.current = null;
@@ -74,8 +74,14 @@ export function CreateSessionModal({ visible, onClose }: CreateSessionModalProps
       setProvidersLoading(false);
       setProvidersTimedOut(false);
     }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [visible, fetchProviders]);
+    return () => {
+      // Unmount cleanup — prevent setState on unmounted component
+      if (providersTimeoutRef.current) {
+        clearTimeout(providersTimeoutRef.current);
+        providersTimeoutRef.current = null;
+      }
+    };
+  }, [visible, fetchProviders, startProvidersTimeout]);
 
   // Cancel timeout once providers have loaded; also clears timed-out state if
   // providers arrive late (after the timeout already fired).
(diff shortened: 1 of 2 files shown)


blamechris:
## Code Review: PR #3003

### Summary

Adds explicit `providersLoading` / `providersTimedOut` state to `CreateSessionModal` so the "Loading providers…" hint cannot display indefinitely. After 5 s with no server response, the hint is replaced by "No additional providers available" plus a Retry button. Timer is cleared on modal close and on successful provider arrival. Six new tests cover the full state machine. CI is green.

### Strengths

- Clean separation between `providersLoading` and `providersTimedOut` — avoids the original bug of inferring loading from an empty array
- Timer cleanup on both explicit close and unmount paths (the `visible→false` branch of the `useEffect`)
- `PROVIDERS_TIMEOUT_MS` constant extracted — easy to tune
- Accessibility attributes (`accessibilityLabel`, `accessibilityRole`) on the Retry button
- Test coverage is solid: loading, timeout, early-arrive, re-open-reset, retry mechanics, Default chip invariant

### Issues Found

#### Critical (Must Fix)

| `CreateSessionModal.tsx` | 82 | Second `useEffect` only clears `providersTimedOut` when `providersLoading` is true. If providers arrive **after** the timeout fires (slow server responding late), `providersTimedOut` stays `true` even though providers are now available. The rendering guards on `availableProviders.length === 0` hide the stale text in this scenario, so there's no immediate visible bug — but the stale state flag is a latent hazard: any future code path that reads `providersTimedOut` without also checking `availableProviders.length` will misbehave. | Change the condition to `availableProviders.length > 0 && (providersLoading \|\| providersTimedOut)` so the flag is reset regardless of which phase providers arrived in. |

```tsx
// Current — misses the timed-out case
if (availableProviders.length > 0 && providersLoading) {

// Suggested — covers both loading and timed-out states
if (availableProviders.length > 0 && (providersLoading || providersTimedOut)) {
```

| `CreateSessionModal.tsx` | 42 | `startProvidersTimeout` is defined as a plain arrow function, not wrapped in `useCallback`. The `// eslint-disable-next-line react-hooks/exhaustive-deps` comment on the `visible` effect is there precisely because adding `startProvidersTimeout` to the dep array would cause infinite re-renders. Wrapping `startProvidersTimeout` in `useCallback([])` (or `useCallback([fetchProviders])` for the retry path) makes the dep suppression unnecessary and makes the linter constraint explicit. | Removes a silent rule suppression that future readers must decode. |
| `CreateSessionModal.test.tsx` | — | Missing test: providers arrive **after** the timeout fires (late server response). Current tests only cover the pre-timeout arrival path. A late-arrival test would catch the critical issue above and verify `providersTimedOut` is cleared. | Closes coverage gap for the most realistic slow-network scenario. |

#### Nitpicks (Optional)

- `jest.clearAllMocks()` in `beforeEach` does not reset `mockUseConnectionStore` return values — each test calls `setupStore()` explicitly, which works fine, but it means a test that forgets `setupStore()` would silently inherit the previous test's store. A `beforeEach(() => setupStore([]))` default would make this safer.
- The `any[]` type on `setupStore(availableProviders: any[] = [])` could be `Array<{ name: string }>` to match the shape actually used.

| Wrap `startProvidersTimeout` in `useCallback` to remove the eslint-disable suppression | [#3004](https://github.com/blamechris/chroxy/issues/3004) | Style/lint improvement, no functional impact — fine to do post-merge. |

### Architecture Notes

The change stays entirely within `CreateSessionModal.tsx` and its test file — no WS protocol changes, no store API changes. The `fetchProviders` function is consumed as-is from the Zustand store; the new state lives locally in the component. That's the right scope: provider-load UX state doesn't need to be global.

### Verdict

- [x] **Request Changes** — One critical fix needed: the `providersTimedOut` flag is not reset when providers arrive after the timeout (line 82 condition). Two minutes of work. The rest of the PR is solid.

blamechris:
Correction to review above: the deferred follow-up issue for `useCallback` refactor was created as [#3015](https://github.com/blamechris/chroxy/issues/3015) (not #3004 as listed in the Deferred Items table).

PR #3004, opened 2026-04-25
feat(doctor): handle bundled .app context for Dependencies check

### LINKED ISSUE #2897: server(doctor): reconsider Dependencies preflight in bundled .app contexts
Follow-up from review of #2896.

## Context

The Dependencies preflight check in `packages/server/src/doctor.js` verifies that `packages/server/node_modules` exists. Before #2896 this used `process.cwd()`, which failed under GUI launch (Tauri via launchd, cwd=\`/\`). The fix correctly anchors the lookup to the server package root via \`import.meta.url\`.

With that fix, the check now only fails in two scenarios:

1. **Bundled .app with a broken build** — the bundler failed to copy \`node_modules\` into the resource tree. The end user can't fix this by running \`npm install\`; they need a new build.
2. **Dev environment where the user hasn't run \`npm install\`** — the original intent of the check, still valid.

## Proposal

Consider one of:

- **Context-aware severity**: downgrade to \`warn\` when running under the supervisor / bundled .app (detectable via \`CHROXY_SUPERVISED\`, \`CHROXY_BUNDLED\`, or a new env var set by Tauri), since failure there is unactionable for the end user.
- **Actionable message**: when bundled, change the message from \"run npm install\" to \"reinstall Chroxy / rebuild the app\".
- **Stronger check in dev**: instead of (or in addition to) \`existsSync(node_modules)\`, spot-check critical runtime deps like \`@anthropic-ai/claude-code\` and \`ws\` so a partial install is detected.

## Out of scope

Removing the check entirely — it still has value in dev.

## Acceptance criteria

- [ ] Define detection mechanism for \"bundled/supervised\" context
- [ ] Pick one of the options above (or combine)
- [ ] Update message / severity accordingly
- [ ] Keep the regression test from #2896 green

blamechris:
## Summary

- Detects bundled/supervised context via `CHROXY_BUNDLED=1` or `CHROXY_SUPERVISED=1` env vars
- Downgrades Dependencies preflight from `fail` to `warn` in those contexts, since end users cannot fix a missing `node_modules` by running `npm install` — they need to reinstall or rebuild the app
- Changes the failure message to "reinstall Chroxy / rebuild the app" in bundled context vs. the original "run npm install" hint for dev
- Dev environments (neither env var set) retain the original `fail` + npm install behavior unchanged

## Test plan

- [x] RED: wrote 4 failing tests covering `CHROXY_BUNDLED=1`, `CHROXY_SUPERVISED=1`, dev (no env var), and pass-when-deps-present in bundled context
- [x] GREEN: all 30 tests pass (`node --test packages/server/tests/doctor.test.js`)
- [x] Regression: prior `pkgDir` tests from #2896 still green

Closes #2897

### DIFF
--- a/packages/desktop/src-tauri/src/server.rs
+++ b/packages/desktop/src-tauri/src/server.rs
@@ -866,6 +866,10 @@ impl ServerManager {
         }
         // Tunnel mode: "quick", "named", or "none"
         cmd.env("CHROXY_TUNNEL", &self.tunnel_mode);
+        // Mark this as a bundled .app launch so the doctor Dependencies check
+        // downgrades to `warn` instead of `fail` — end users can't run
+        // `npm install` to fix a broken bundle; they need to reinstall.
+        cmd.env("CHROXY_BUNDLED", "1");
         // No supervisor — tray app IS the supervisor
         cmd.arg("--no-supervisor");
 
--- a/packages/server/src/doctor.js
+++ b/packages/server/src/doctor.js
@@ -138,6 +138,13 @@ export async function runDoctorChecks({ port, providers, verbose: _verbose, pkgD
   // node_modules/ when installed via `npm ci --workspace=@chroxy/server`.
   // The helper walks up the tree and uses createRequire as a reliable
   // proxy for whether deps are installed.
+  //
+  // Context-aware severity: when running inside a bundled .app (Tauri) or
+  // under the supervisor, a missing node_modules is unactionable for the
+  // end user — they can't run `npm install` to fix it, they need a new
+  // build or reinstall. In that context we downgrade to `warn` and provide
+  // an appropriate message. In a dev environment the original `fail` +
+  // "run npm install" message is preserved.
   if (typeof pkgDir !== 'string' || pkgDir.length === 0) {
     throw new TypeError(`pkgDir must be a non-empty string, got ${typeof pkgDir}`)
   }
@@ -149,11 +156,20 @@ export async function runDoctorChecks({ port, providers, verbose: _verbose, pkgD
   if (deps.ok) {
     checks.push({ name: 'Dependencies', status: 'pass', message: `resolved via ${deps.foundAt}` })
   } else {
-    checks.push({
-      name: 'Dependencies',
-      status: 'fail',
-      message: `${deps.message || 'dependencies not found'} — run npm install`,
-    })
+    const isBundledContext = process.env.CHROXY_BUNDLED === '1' || process.env.CHROXY_SUPERVISED === '1'
+    if (isBundledContext) {
+      checks.push({
+        name: 'Dependencies',
+        status: 'warn',
+        message: `${deps.message || 'dependencies not found'} — reinstall Chroxy or rebuild the app`,
+      })
+    } else {
+      checks.push({
+        name: 'Dependencies',
+        status: 'fail',
+        message: `${deps.message || 'dependencies not found'} — run npm install`,
+      })
+    }
   }
 
   // 7. Port availability
--- a/packages/server/tests/doctor.test.js
+++ b/packages/server/tests/doctor.test.js
@@ -214,6 +214,87 @@ describe('runDoctorChecks', () => {
   })
 })
 
+describe('runDoctorChecks — bundled .app context (issue #2897)', () => {
+  it('Dependencies check is warn (not fail) when CHROXY_BUNDLED=1 and node_modules is missing', async () => {
+    const emptyDir = mkdtempSync(join(tmpdir(), 'chroxy-doctor-bundled-'))
+    const originalBundled = process.env.CHROXY_BUNDLED
+    try {
+      process.env.CHROXY_BUNDLED = '1'
+      const { checks } = await runDoctorChecks({ providers: ['claude-sdk'], pkgDir: emptyDir })
+      const depsCheck = checks.find(c => c.name === 'Dependencies')
+      assert.ok(depsCheck)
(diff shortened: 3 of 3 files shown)


blamechris:
## Code Review: PR #3004

### Summary

Addresses issue #2897 by making the Dependencies preflight check context-aware. When running inside a bundled `.app` (Tauri) or under the supervisor, a missing `node_modules` is unactionable for the end user, so the check is downgraded from `fail` → `warn` with an appropriate remediation hint. Dev environments retain the original `fail` + npm install behavior. Four new tests cover all branches.

---

### Strengths

- **Clean branching logic** — the `isBundledContext` variable is well-named and immediately legible. The two-branch if/else is easier to follow than a ternary would have been.
- **Good comment coverage** — the block comment above the Dependencies check clearly explains the context-aware rationale, which is non-obvious. Future readers won't have to rediscover it.
- **Proper env restoration in tests** — all four tests save and restore `process.env` in `finally` blocks, preventing state leakage across test runs.
- **Correct `passed` semantics preserved** — `warn` does not fail `passed`, so the doctor command's exit-code-1 path is not triggered for bundled context, which is exactly the right behavior.
- **Regression coverage** — the fourth test ("deps present in bundled context still pass") guards against accidentally affecting the success path.

---

### Issues Found

#### Critical (Must Fix)

| `packages/server/src/doctor.js` | ~165 | `CHROXY_BUNDLED` is checked but **never set anywhere in the codebase**. `supervisor.js` sets `CHROXY_SUPERVISED=1` and `server.rs` sets several `CHROXY_*` vars, but `CHROXY_BUNDLED` has no setter — it is dead code. The Tauri-spawned server runs under `CHROXY_SUPERVISED=1` through the supervisor, so the supervised path already covers that case. `CHROXY_BUNDLED` would only be meaningful if Tauri's `server.rs` set it directly when spawning without the supervisor. | Either (a) add `cmd.env("CHROXY_BUNDLED", "1")` in `packages/desktop/src-tauri/src/server.rs` at the point where the server process is spawned, or (b) remove the `CHROXY_BUNDLED` branch and rely solely on `CHROXY_SUPERVISED=1`. Option (a) is better: it explicitly marks the Tauri-direct-spawn path and makes the intent self-documenting. |

| `packages/server/src/doctor.js` | ~165 | Extract the detection logic into a named helper `isBundledOrSupervisedContext()` | The env-var pattern will likely need to be reused in other checks as bundled-context awareness spreads. A single named function also makes it mockable in tests without needing to mutate `process.env`. |
| `packages/server/tests/doctor.test.js` | ~240 | The supervised test does not assert the message includes "reinstall" or "rebuild" | The bundled test checks message content; the supervised test only checks `status`. Both paths hit the same message, so the assertion is implicitly tested — but making it explicit in both guards against future divergence. |

#### Nitpicks (Optional)

- Line ~165: `isBundledContext` could be `isBundledOrSupervisedContext` to more accurately describe what it detects — "bundled" alone implies only the Tauri path.
- The variable is declared inline rather than hoisted to the top of the `else` block, which is fine, but if the helper extraction above is done it becomes a non-issue.

---

| Extract `isBundledOrSupervisedContext()` helper | #3005 | Minor refactor, out of scope for this fix. Worth doing before more checks adopt the pattern. |

---

### Architecture Notes

The approach aligns with the existing `CHROXY_SUPERVISED` pattern in `supervisor.js`. The two env vars cleanly separate "supervised child" from "Tauri-direct-spawn" — both conceptually bundled contexts but launched differently. The critical gap is that Tauri's `server.rs` does not currently set `CHROXY_BUNDLED`, meaning the `CHROXY_BUNDLED` branch can never trigger in production until that setter is added.

The `passed` field semantics (`warn` does not fail) are correct and consistent with how other warn-level checks (Port, Node.js version warning) are handled throughout the doctor.

---

### Verdict

- [x] **Request Changes** — One issue must be addressed before merge: `CHROXY_BUNDLED` must either be wired up in `server.rs` or removed. The logic is otherwise sound.

blamechris:
## Review Comments Addressed

| 1 | \`CHROXY_BUNDLED\` checked in doctor.js but never set by Tauri server spawner — bundled branch unreachable in production | FIX | \`a32913dc1\` |
| 2 | Extract \`isBundledOrSupervisedContext()\` helper for reuse | FOLLOW-UP | [#3023](https://github.com/blamechris/chroxy/issues/3023) |
| 3 | Supervised test missing message content assertion | FOLLOW-UP | [#3023](https://github.com/blamechris/chroxy/issues/3023) |

**Total:** 3 findings addressed
- Fixed: 1 (`a32913dc1`)
- Follow-up issues: 2 (tracked in #3023)

**Fix detail — `a32913dc1`:**
Added `cmd.env("CHROXY_BUNDLED", "1")` to `packages/desktop/src-tauri/src/server.rs` alongside the existing `CHROXY_TUNNEL`, `CHROXY_MODEL`, etc. env vars. Without this, the `isBundledContext` branch in doctor.js could never trigger for Tauri-spawned servers — the warn severity downgrade only worked for supervisor-spawned servers via `CHROXY_SUPERVISED=1`.

```diff
  // Tunnel mode: "quick", "named", or "none"
  cmd.env("CHROXY_TUNNEL", &self.tunnel_mode);
+ // Mark this as a bundled .app launch so the doctor Dependencies check
+ // downgrades to `warn` instead of `fail` — end users can't run
+ // `npm install` to fix a broken bundle; they need to reinstall.
+ cmd.env("CHROXY_BUNDLED", "1");
  // No supervisor — tray app IS the supervisor
  cmd.arg("--no-supervisor");
```

PR #3009, opened 2026-04-25
feat(server): capability-gate set_permission_mode for non-supporting providers

### LINKED ISSUE #2963: feat(server): capability-gate set_permission_mode (currently silent no-op on Gemini/Codex)
## Context

Found in [swarm audit `multi-model-support`](../tree/main/docs/audit-results/multi-model-support/00-master-assessment.md). Guardian's FM-05.

\`set_permission_mode\` WS message is accepted on Gemini/Codex sessions but those providers don't have a permission mode concept. Server silently accepts the change; UI pretends it took effect; nothing happens.

User experience: tap "plan mode", get visual confirmation, but the session continues to behave normally. Confusing.

## Acceptance criteria

- [ ] Each provider declares its capabilities (e.g., \`static get capabilities() { return { permissionMode: true, planMode: true, ... } }\`)
- [ ] \`set_permission_mode\` handler rejects with \`CAPABILITY_NOT_SUPPORTED\` if provider lacks \`permissionMode\`
- [ ] UI hides or disables permission mode controls for sessions whose provider doesn't support it
- [ ] Same pattern applicable to other capability-gated features (plan mode, agent tracking, model switching, attachments)
- [ ] Tests: \`set_permission_mode\` on Gemini → clean rejection, not silent accept

## Audit reference

- \`docs/audit-results/multi-model-support/03-guardian.md\` — FM-05

blamechris:
## Summary

- `set_permission_mode` WS handler now rejects with `CAPABILITY_NOT_SUPPORTED` when the active session's provider declares `permissionModeSwitch: false` (Gemini, Codex)
- Previously the server silently accepted the mode change, leaving the UI showing false confirmation while the session ignored it entirely
- Legacy sessions with no `provider` field and unknown providers fail open for forward-compatibility

## Implementation

Follows the same pattern as the per-provider model allowlist (#2946): `getProvider()` looks up the provider class, checks `capabilities.permissionModeSwitch`, and calls `sendError()` with a structured error payload including `requestId` so clients can correlate the rejection.

## Tests

- RED: wrote 4 failing tests before implementing (Gemini reject, Codex reject, claude-sdk accept, legacy-no-provider accept)
- GREEN: all 4 pass after implementation
- Full suite: 3527 pass, 4 fail (same pre-existing tunnel integration + WebTaskManager failures present on `main`)

Closes #2963


blamechris:
## Code Review: PR #3009

### Summary

Adds a capability gate to `handleSetPermissionMode` that rejects `set_permission_mode` requests with `CAPABILITY_NOT_SUPPORTED` when the session's provider declares `permissionModeSwitch: false` (Gemini, Codex). Previously, `setPermissionMode()` was a no-op in `JsonlSubprocessSession` — the server accepted the WS message, broadcast `permission_mode_changed`, and the UI happily showed the new mode while the session silently ignored it. This PR closes that gap cleanly and follows the established `#2946` per-provider model allowlist pattern.

---

### Strengths

- **Correct pattern reuse.** The `getProvider()` try/catch + capability check mirrors `getProviderAllowedModels()` from #2946 exactly. The fail-open semantics for unknown providers are explicitly justified and match the model gate's behaviour.
- **Red-green test discipline.** Four test cases (Gemini reject, Codex reject, claude-sdk accept, legacy no-provider accept) cover all four meaningful code paths through the new gate — full branch coverage.
- **`requestId` threading.** `sendError(ws, msg?.requestId, ...)` lets clients correlate rejections to their originating request. The `setPermissionMode` call in `connection.ts` currently doesn't set a `requestId`, so the field will be `undefined`, but that's a pre-existing gap, not a regression.
- **No-op confirmation.** Checked `jsonl-subprocess-session.js:126` — `setPermissionMode` is indeed a comment-documented no-op on the base class. The PR description's claim is accurate.
- **CI: all green.** 12/12 checks pass, including server tests that incorporate the new 4 cases in the 25-test suite.

---

### Issues Found

#### Critical (Must Fix)

_None._

| `settings-handlers.js` | new gate block | Placement before the plan mode check | The new capability gate fires before `if (msg.mode === 'plan' && !entry.session.constructor.capabilities?.planMode)`. Since `permissionModeSwitch: false` implies the session can't switch *any* mode (not just plan), the gate ordering is correct — reject the whole category early. No change needed; documenting for clarity. |
| `settings-handlers.js` | `ProviderClass.capabilities?.permissionModeSwitch === false` | Add `=== false` vs falsy note | The explicit `=== false` check is the right choice (fail-open for `undefined`). A brief inline comment like `// explicit false — undefined means "unknown, allow through"` would make the intent self-documenting for future readers. Minor; not a blocking concern. |
| `connection.ts:1007,1025` | `setPermissionMode` / `confirmPermissionMode` | No `requestId` on `set_permission_mode` payload | The server now returns `requestId` in the rejection so clients can correlate. The app sends no `requestId`, so the rejection `error` message arrives without a usable correlation ID. On an unsupported provider, the error surfaces as a generic `server_error` message rather than a targetted UI rejection. Consider adding a `requestId` to the outbound payload and handling `{ type: 'error', code: 'CAPABILITY_NOT_SUPPORTED' }` explicitly in `message-handler.ts` to revert the permission mode selector to its previous value. |
| `settings-handlers.test.js` | capability gate `describe` block | No test for unknown provider (throws from `getProvider`) | The three branches are: known+false, known+true, and no provider. The fourth branch — `entry.provider` set to an unregistered string — exercises the `catch {}` that fails open. Covered implicitly by the code reading, but an explicit test would close the gap. |

#### Nitpicks (Optional)

- The comment block on the gate starts with `// #2963 — capability gate: ...`. The `#2963` reference is useful but using an inline comment for multi-line rationale is slightly inconsistent with the JSDoc-style block comments elsewhere in the file. Consider a `/** */` block or just leave the leading `//` — already a pre-existing style inconsistency in this file.
- `entry.provider` on the session entry is typed as `string | undefined` in practice. The `if (entry.provider)` guard handles that correctly.

---

| App: send `requestId` on `set_permission_mode` + handle `CAPABILITY_NOT_SUPPORTED` error code to revert the mode selector | #3010 | App-side change; out of scope for this server-only PR. The current behaviour (generic error toast) is an improvement over silent acceptance. |
| Test: add explicit unknown-provider fallback test case | #3011 | Low-risk gap; the fail-open code path is trivially correct. Deferring keeps the test count focused. |

---

### Architecture Notes

The change is local to `handleSetPermissionMode` in `settings-handlers.js` and touches no shared state, no WS broadcast paths, and no protocol messages. It correctly mirrors the `#2946` guard pattern — one of the few well-established conventions in this codebase for per-provider capability gating. The `CAPABILITY_NOT_SUPPORTED` error code is new (the codebase previously used `MODEL_NOT_SUPPORTED_BY_PROVIDER` for the analogous model case), which is a reasonable choice since this is a different capability axis, but worth documenting in the WS protocol header comment in `ws-server.js` if/when other capabilities use the same code.

---

### Verdict

- [x] **Approve** — Ready to merge. The fix is correct, well-tested, and follows established patterns. The two follow-up issues capture the remaining gaps without blocking this PR.

blamechris:
**Deferred issue links** (supplement to review above):

- App-side `CAPABILITY_NOT_SUPPORTED` handling → #3026
- Unknown-provider fallback test case → #3027

Issue #2831, opened 2026-04-19
bug(server): session becomes silently unresponsive after 5-minute inactivity timeout; subsequent permission approvals vanish and no response streams


blamechris:
## Symptom
1. Claude is streaming a response, permission prompts appear
2. User delays responding long enough that the inactivity timeout fires: **"Response timed out after 5 minutes of inactivity"** error appears
3. User subsequently approves further permission prompts (5+ more in reported case)
4. **No response text ever streams back to the client** — the conversation appears alive (buttons work, approvals are sent) but Claude never produces output

## Source of the timeout
`packages/server/src/sdk-session.js:210-223`:
\`\`\`js
const RESULT_TIMEOUT_MS = 300_000 // 5 min of inactivity
const resetResultTimeout = () => {
  if (this._resultTimeout) clearTimeout(this._resultTimeout)
  this._resultTimeout = setTimeout(() => {
    if (this._isBusy) {
      log.warn('Result timeout (5 min inactivity) — force-clearing busy state')
      if (hasStreamStarted) {
        this.emit('stream_end', { messageId })
      }
      this._clearMessageState()
      this.emit('error', { message: 'Response timed out after 5 minutes of inactivity' })
    }
  }, RESULT_TIMEOUT_MS)
}
\`\`\`

Same pattern in `packages/server/src/cli-session.js:332`.

## Root-cause hypothesis
On timeout:
- `_clearMessageState()` clears the in-flight turn
- `emit('error', ...)` notifies the client
- **`_pendingPermissions` map is NOT explicitly cleared** (neither is `permissionSessionMap` in ws-permissions.js)

So when the user later approves those orphaned prompts:
- Server calls `session.respondToPermission(requestId, 'allow')`
- The pending promise resolves, `canUseTool` MCP callback returns "allow"
- But the SDK query generator `for await (const msg of this._query)` has been abandoned / Claude's turn has ended
- Tool result has nowhere to stream → silent no-op

Additionally: the inactivity timer resets on *SDK events*, but a pending permission is not an SDK event — so a user who takes >5min on a prompt (the most common idle reason) gets punished as inactivity.

## Fixes to consider
1. **On timeout, clear & notify pending permissions.** In `_clearMessageState` (or the timeout handler before it), iterate `_pendingPermissions` and:
   - Resolve them as `deny` so the SDK doesn't hang forever
   - Emit `permission_expired` for each to the client so the UI doesn't show stale prompts
[...]

PR #3028, opened 2026-04-25
fix(server): map hook permissions to chroxy session for bound clients (#2832)

### LINKED ISSUE #2832: bug(server): permission approval fails with "Not authorized" (SESSION_TOKEN_MISMATCH) after Android app backgrounding
## Symptom
On Android:
1. App is active, Claude streams a response, permission prompt appears
2. User backgrounds the app (home/task switch)
3. User returns to the app; permission prompt is still visible
4. User taps **Approve**
5. Server replies with error: **"Server Error: Not authorized to respond to this permission request"**

## Source
`packages/server/src/handlers/settings-handlers.js:155-162`:
\`\`\`js
if (client.boundSessionId) {
  if (!mappedSessionId || mappedSessionId !== client.boundSessionId) {
    log.warn(\`Client \${client.id} (bound to \${client.boundSessionId}) attempted to respond to permission \${requestId} with mapped session \${mappedSessionId ?? 'unmapped'}\`)
    sendError(ws, requestId, 'SESSION_TOKEN_MISMATCH', 'Not authorized to respond to this permission request')
    return
  }
}
\`\`\`

This check was hardened on 2026-04-11 (agent-review of PR #2806, blocker 5) to close a legitimate security bypass. But the invariant it assumes (`client.boundSessionId === permissionSessionMap[requestId]`) isn't holding across Android backgrounding + reconnect.

## Diagnosis needed
Add a one-shot diagnostic log at the WARN point, dumping:
- `client.id`, `client.activeSessionId`, `client.boundSessionId`
- `mappedSessionId` (already logged)
- whether this is a reconnected client (correlate with `ws-server.js` connection-open log)
- whether `permissionSessionMap[requestId]` was mutated after creation

Reproduce on Android and read the server log to pinpoint.

## Hypotheses
1. **`boundSessionId` not re-derived on reconnect** — fresh auth after background/resume sets `client.activeSessionId` but omits `boundSessionId`.
2. **`permissionSessionMap` entry overwritten** — resend of pending permissions on reconnect (`ws-permissions.js:310-340`) may be rewriting the mapping with the new client's `activeSessionId` instead of preserving the original.
3. **Android is using pairing tokens when it shouldn't** — if the primary flow uses `API_TOKEN`, `boundSessionId` should be unset. If it IS set, that's a misclassification in the auth path.

## Files
- `packages/server/src/handlers/settings-handlers.js:100-195` — `permission_response` handler
- `packages/server/src/ws-permissions.js:110-174` — permission creation + session map
- `packages/server/src/ws-permissions.js:310-340` — resend pending on reconnect
- `packages/server/src/ws-server.js:740-880` — auth/reconnect flow

## Severity
Critical — breaks a core workflow. Backgrounding-during-response is extremely common on mobile.

blamechris:
## Summary
- Closes #2832 — paired clients receive `SESSION_TOKEN_MISMATCH` when approving any tool permission on the legacy CLI provider, not just after Android backgrounding
- The HTTP `/permission` path knew which `CliSession` owned the request (via the per-session hook secret) but never populated `permissionSessionMap[requestId]`. The binding check in `permission_response` requires that entry — without it, a bound client always lands in the `!mappedSessionId` reject branch
- `_findSessionByHookSecret` already iterates `_sessionHookSecrets` keyed by chroxy's session id; this PR surfaces that id alongside the session and uses it to populate the map
- Legacy single-session mode returns `sessionId: null` so we skip the map write — bound clients in that mode would already be a configuration error

## Root cause analysis
The "after Android backgrounding" framing was a red herring. The actual condition was: paired client + claude-cli (or any HTTP /permission hook flow) + ANY tool permission. Backgrounding made the bug visible because the prompt was likely to persist across the background window, so users discovered it on resume.

## Test plan
- [x] `node --test packages/server/tests/ws-permissions.test.js packages/server/tests/ws-permissions-pause-integration.test.js` — all 47 cases pass
- [x] Full server suite: 3586/3588 pass; the one unrelated failure is `web-task-manager` feature detection (asserts the local `claude` CLI lacks `--remote` — environment-specific)
- [x] `npx eslint src/ws-server.js src/ws-permissions.js` — clean
- [ ] Manual repro on Android (paired flow): trigger a Bash hook permission, background the app, return, approve — should succeed instead of `SESSION_TOKEN_MISMATCH`

## Files changed
- `packages/server/src/ws-server.js` — `_findSessionByHookSecret` returns `{ session, sessionId } | null`
- `packages/server/src/ws-permissions.js` — populate `permissionSessionMap` when `sessionId` is known; new `[session-binding-create]` log line at debug level for HTTP path
- `packages/server/tests/ws-permissions.test.js` — two new regression cases
- `packages/server/tests/ws-permissions-pause-integration.test.js` — updated mock return shape

### DIFF
--- a/packages/server/src/ws-permissions.js
+++ b/packages/server/src/ws-permissions.js
@@ -111,13 +111,6 @@ export function createPermissionHandler({ sendFn, broadcastFn, validateBearerAut
       const requestId = `perm-${randomUUID()}`
 
       log.info(`Permission request ${requestId}: ${hookData.tool_name || 'unknown tool'}`)
-      // Diagnostic correlation log for #2832 — paired with
-      // [session-binding-resend] and [session-binding-reject]. The HTTP
-      // /permission path is the legacy (non-SDK) case; the requestId acts
-      // as a stable correlation key across the permission lifecycle.
-      // Gated at debug level (#2854) to avoid spamming prod logs — enable
-      // with `LOG_LEVEL=debug` when triangulating SESSION_TOKEN_MISMATCH.
-      log.debug(`[session-binding-create] permission ${requestId} created via HTTP (sessionId=none, sourceIp=${clientIp})`)
 
       const tool = hookData.tool_name || 'Unknown tool'
       const toolInput = hookData.tool_input || {}
@@ -149,8 +142,15 @@ export function createPermissionHandler({ sendFn, broadcastFn, validateBearerAut
       }
       if (ownerSessionId) {
         permissionSessionMap.set(requestId, ownerSessionId)
-        log.debug(`[session-binding-create] permission ${requestId} mapped to ${ownerSessionId} (HTTP, via hookSecret)`)
       }
+      // Diagnostic correlation log for #2832 — paired with
+      // [session-binding-resend] and [session-binding-reject]. The
+      // requestId is the stable correlation key across the permission
+      // lifecycle; sessionId reflects the chroxy session that owns the
+      // hook secret (or `none` when the hook secret is unattributable).
+      // Gated at debug level (#2854) to avoid spamming prod logs — enable
+      // with `LOG_LEVEL=debug` when triangulating SESSION_TOKEN_MISMATCH.
+      log.debug(`[session-binding-create] permission ${requestId} created via HTTP (sessionId=${ownerSessionId ?? 'none'}, sourceIp=${clientIp})`)
 
       broadcastFn({
         type: 'permission_request',
--- a/packages/server/src/ws-permissions.js
+++ b/packages/server/src/ws-permissions.js
@@ -132,13 +132,24 @@ export function createPermissionHandler({ sendFn, broadcastFn, validateBearerAut
       // #2831: find the CliSession this hook permission belongs to (via
       // the per-session hook secret from the Authorization header) so we
       // can pause its inactivity timer while the user decides.
+      // #2832: also use the chroxy-managed sessionId to populate
+      // permissionSessionMap so paired clients (boundSessionId set) can
+      // approve hook-originated permissions for their bound session.
+      // Without this entry, the binding check in permission_response
+      // rejects every approval with SESSION_TOKEN_MISMATCH.
       const authHeader = (req.headers && req.headers['authorization']) || ''
       const hookToken = authHeader.startsWith('Bearer ') ? authHeader.slice(7) : null
-      const ownerSession = (hookToken && typeof findSessionByHookSecret === 'function')
+      const ownerLookup = (hookToken && typeof findSessionByHookSecret === 'function')
         ? findSessionByHookSecret(hookToken)
         : null
+      const ownerSession = ownerLookup?.session ?? null
+      const ownerSessionId = ownerLookup?.sessionId ?? null
       if (ownerSession && typeof ownerSession.notifyPermissionPending === 'function') {
         ownerSession.notifyPermissionPending(requestId)
+      }
+      if (ownerSessionId) {
+        permissionSessionMap.set(requestId, ownerSessionId)
+        log.debug(`[session-binding-create] permission ${requestId} mapped to ${ownerSessionId} (HTTP, via hookSecret)`)
       }
 
       broadcastFn({
--- a/packages/server/src/ws-server.js
+++ b/packages/server/src/ws-server.js
@@ -745,9 +745,16 @@ export class WsServer {
   }
 
   /**
-   * Find the CliSession whose hookSecret matches `secret`. Returns null
-   * if no match. Used by the permission handler to pause a session's
-   * inactivity timer while a hook permission is outstanding (#2831).
+   * Find the CliSession whose hookSecret matches `secret`. Returns
+   * `{ session, sessionId }` (sessionId is the chroxy-managed key, not
+   * the upstream Claude conversation id) or `null` if no match.
+   *
+   * The sessionId is needed by the permission handler so it can populate
+   * `permissionSessionMap[requestId]` for legacy hook-based permissions
+   * (#2832). Without that mapping, a paired client whose token is bound
+   * to the same session can never approve hook permissions — the
+   * binding check at `permission_response` time finds no entry and
+   * rejects with SESSION_TOKEN_MISMATCH.
    */
   _findSessionByHookSecret(secret) {
     if (!secret) return null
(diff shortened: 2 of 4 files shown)


blamechris:
## Code Review: PR #3028

### Summary
This change repairs the legacy CLI hook permission flow for paired clients by populating `permissionSessionMap[requestId]` at the moment the HTTP `/permission` request is broadcast. The plumbing change in `_findSessionByHookSecret` (now returning `{ session, sessionId }`) is the right shape — it keeps the existing `notifyPermissionPending` plumbing intact for #2831 and surfaces the chroxy-managed sessionId so the binding check in `settings-handlers.js:248-249` can succeed for legitimately bound clients. The `sessionId: null` branch for legacy single-session mode preserves the existing semantics (no map write, no bypass). Diff is small (80 / -8), tightly scoped, and includes regression tests that lock the new contract.

### Strengths
- Minimal surface area: change is confined to two files plus tests, no protocol changes.
- The contract documentation on `_findSessionByHookSecret` (`ws-server.js:747-758`) now explicitly states the security relationship between the returned `sessionId` and the binding check — future maintainers will not have to spelunk to understand why this shape exists.
- Regression test (`ws-permissions.test.js:298`) asserts both the map write AND the requestId shape (`/^perm-/`), preventing accidental key-shape regressions that would silently break the binding correlation.
- Negative test (`ws-permissions.test.js:333`) explicitly locks the `sessionId: null` legacy path's no-write behaviour, preventing a future refactor from inventing a synthetic key that would reintroduce a bypass.
- New `[session-binding-create]` debug log on the HTTP path matches the existing diagnostic vocabulary used by the SDK forwarding path (`ws-forwarding.js:269`) and the resend path (`ws-permissions.js:372`), so a single `LOG_LEVEL=debug` correlation grep on requestId continues to work end-to-end.

### Issues Found

#### Critical (Must Fix)
None.

|---|---|---|---|
| `packages/server/src/ws-permissions.js` | 120 | The pre-fix `[session-binding-create]` debug log on the HTTP path still says `sessionId=none` unconditionally, even after the new code resolves a real `ownerSessionId` ten lines later. Two log lines per request now: the first claims unknown, the second corrects it. Confusing during triage. | Either drop the line at 120 entirely (the line at 152 is now the authoritative one) or compute `ownerLookup` first and inline the value into a single log call. |
| `packages/server/tests/ws-permissions.test.js` | 298-348 | Neither new test exercises the resulting `permission_response` round-trip from a bound client. The fix's actual contract — that a bound client can now successfully approve — is implicit in the map write. | Tracked as #3029. |

#### Nitpicks (Optional)
- `ws-server.js:765-766` — the `if (entry?.session) return ...; return null` pair could be `return entry?.session ? { session: entry.session, sessionId } : null` for parity with the legacy branch's terse single-line return. Stylistic only.
- The PR description says "two new regression cases"; the diff actually adds two `it(...)` blocks plus the integration-test mock-shape update. Minor accounting nit.

### Security Analysis
The fix preserves the PR #2806 blocker 5 hardening. The chain of trust is:

1. **Hook secret provenance** — `_sessionHookSecrets` (`ws-server.js:395`) is keyed by `sessionId` and populated only on `session_created` (`ws-server.js:542-549`) from `entry.session._hookSecret`, which is generated server-side per CliSession and passed to the forked `claude-cli` child only via env var. External WS clients never see it.
2. **HTTP `/permission` authentication** — `_validateHookAuth` (`ws-server.js:708-737`) requires a Bearer token that matches the registered hook-secret set; the primary API token is explicitly NOT in `_hookSecrets` when sessions are running (it falls back only when the registered set is empty, in legacy mode).
3. **Map population provenance** — the new `permissionSessionMap.set(requestId, ownerSessionId)` derives `ownerSessionId` from the secret-to-sessionId reverse lookup. The secret was sent BY the local hook script, which is owned BY a specific CliSession. So the mapping reflects "which session's hook script raised this prompt," not "which WS client wants to approve it." Distinct.
4. **Approval check** — `settings-handlers.js:248-249` still requires `mappedSessionId === client.boundSessionId`. A malicious bound client can't change `mappedSessionId` (it's set server-side from the hook auth header) and can't change its own `boundSessionId` (set at WS auth time from `pairingManager.getSessionIdForToken`). It can only approve if its bound session genuinely owns the request.

The `sessionId: null` legacy single-session branch (`ws-server.js:771-772`) skips the map write. Since legacy single-session mode rarely co-exists with `pairingManager` (the `boundSessionId` source), a bound client in that mode falls into the `!mappedSessionId` reject branch — same behaviour as before this PR. Not a regression and not a new bypass.

The `session_destroyed` cleanup at `ws-server.js:564-566` already wipes stale `_permissionSessionMap` entries for destroyed sessions, so the multi-client "session destroyed between request creation and response" race resolves to: bound client gets `mappedSessionId = undefined` → falls into the reject branch correctly. The HTTP request itself is still pending (`pendingPermissions` lives in the closure), but auto-deny on the 5-min timer or on `req.aborted`/`res.close` will eventually clean it up. No leak.

I cannot find a way to cause the new map write to point at a session the WS client isn't actually bound to. The fix looks tight.

### Verdict
- [x] Approve

The diagnosis is correct (paired + legacy CLI + any tool, not Android-specific), the fix is the minimal change to make the binding check satisfiable in this code path, and the security invariant from PR #2806 is preserved by virtue of `ownerSessionId` deriving from a server-side hook-secret reverse lookup rather than any client-supplied value. The two suggestions are non-blocking polish.

### Deferred Items
- #3029 — integration test for paired-client legacy CLI permission round-trip

blamechris:
## Review Comments Addressed

| 1 | Copilot review summary — no findings | N/A | [Copilot review](https://github.com/blamechris/chroxy/pull/3028#pullrequestreview) (no comments generated) |
| 2 | Agent-review: misleading `sessionId=none` log line at ws-permissions.js:120 supersedes new authoritative line | FIX | `690610e5e` |
| 3 | Agent-review: ws-server.js:765-766 SDK branch could be a single ternary for parity with legacy branch | FIX | `690610e5e` |
| 4 | Agent-review: round-trip integration test for paired-client legacy CLI permission approval | FOLLOW-UP | [#3029](https://github.com/blamechris/chroxy/issues/3029) |
| 5 | Agent-review: PR description accounting nit ("two new regression cases" vs two `it` blocks + integration mock update) | FALSE POSITIVE | Evidence: cosmetic accounting on PR body, not code; not worth amending |

**Total:** 5 items addressed
- Fixed: 2 (commit `690610e5e`)
- False positives: 1 (PR-body nit)
- Follow-up issues created: 1 (#3029, by agent-review)
- Existing issues closed: 0

PR #3030, opened 2026-04-25
test(dashboard): tighten reduced-motion banner assertion to same rule

### LINKED ISSUE #3012: dashboard: tighten reduced-motion test to co-locate selector + property in same rule
## Context

Identified during review of PR #2998 (fix: respect prefers-reduced-motion on tunnel-warming banner).

## Description

The current test in `packages/dashboard/src/theme/theme.test.ts` (lines 329–332) checks that a `prefers-reduced-motion: reduce` block contains both `.tunnel-warming-banner` and `transition: none` as substrings, but it does not verify they appear in the **same CSS rule**:

\`\`\`ts
const bannerOverridePresent = reduceBlocks.some(
  (block) =>
    block.includes('.tunnel-warming-banner') && block.includes('transition: none'),
)
\`\`\`

A file where `.tunnel-warming-banner` appears in one rule and `transition: none` appears in a different rule within the same media block would pass this assertion incorrectly.

## Proposed Fix

Replace the two `includes()` calls with a regex that requires the selector and property to co-occur inside the same rule:

\`\`\`ts
const bannerOverridePresent = reduceBlocks.some(
  (block) => /\.tunnel-warming-banner\s*\{[^}]*transition:\s*none/.test(block),
)
\`\`\`

This is unambiguous and still readable.

## Original Review Comment

> The two `includes()` calls check the whole merged block string but do not verify the two tokens appear inside the *same rule*. A future file where `transition: none` exists in one rule and `.tunnel-warming-banner` appears in another rule within the same media block would pass. A regex like `/\.tunnel-warming-banner\s*\{[^}]*transition:\s*none/` would be unambiguous.

## Acceptance Criteria

- [ ] Test uses a regex that asserts selector and property appear within the same rule block.
- [ ] All existing tests continue to pass.

blamechris:
## Summary

Tightens the `prefers-reduced-motion: reduce` assertion in `packages/dashboard/src/theme/theme.test.ts` so it verifies that `.tunnel-warming-banner` and `transition: none` co-occur inside the **same CSS rule**, rather than merely the same `@media` block.

The previous test used two `String.includes()` checks against the entire merged media-block string, which would pass even if the selector and property were split across two unrelated rules within the block.

## Change

Replaces:

\`\`\`ts
const bannerOverridePresent = reduceBlocks.some(
  (block) =>
    block.includes('.tunnel-warming-banner') && block.includes('transition: none'),
)
\`\`\`

With:

\`\`\`ts
const bannerOverridePresent = reduceBlocks.some((block) =>
  /\.tunnel-warming-banner\s*\{[^}]*transition:\s*none/.test(block),
)
\`\`\`

The regex requires the property to appear inside the rule body opened by `.tunnel-warming-banner { ... }`, before the next `}`.

## Test Plan

- [x] `npm test src/theme/theme.test.ts` in `packages/dashboard` — 46/46 pass
- [x] Full dashboard suite: 1265/1265 pass

Closes #3012

### DIFF
--- a/packages/dashboard/src/theme/theme.test.ts
+++ b/packages/dashboard/src/theme/theme.test.ts
@@ -326,9 +326,10 @@ describe('components.css reduced-motion overrides', () => {
 
     expect(reduceBlocks.length).toBeGreaterThan(0)
 
-    const bannerOverridePresent = reduceBlocks.some(
-      (block) =>
-        block.includes('.tunnel-warming-banner') && block.includes('transition: none'),
+    // Require selector and property to co-occur inside the same rule block,
+    // not merely the same @media block (#3012).
+    const bannerOverridePresent = reduceBlocks.some((block) =>
+      /\.tunnel-warming-banner\s*\{[^}]*transition:\s*none/.test(block),
     )
     expect(bannerOverridePresent).toBe(true)
   })


blamechris:
## Agent Review

### Summary
Tightens the `prefers-reduced-motion` test in `packages/dashboard/src/theme/theme.test.ts` so the assertion verifies `.tunnel-warming-banner` and `transition: none` co-occur **inside the same rule block**, not just inside the same `@media` block. Replaces two separate `String.includes()` checks with a single regex `/\.tunnel-warming-banner\s*\{[^}]*transition:\s*none/`. Closes #3012.

### Strengths
- **Exact match to acceptance criteria** in #3012 — the regex is the one suggested verbatim.
- **Surgical**: 4 added / 3 removed, single test case, no surrounding refactor.
- **Inline comment** explains the intent and references #3012, which helps a future reader understand why this is regex-shaped instead of a simpler `includes()` pair.
- **Brace-balancing block extraction** above (lines 313–325) correctly handles the nested rule, so feeding the regex a single `@media` block string is sound.
- **Regex semantics are correct**: `[^}]*` stops at the first `}`, so `transition: none` must appear inside the `.tunnel-warming-banner { ... }` body before the rule closes. Verified locally — `46/46 tests pass`, full dashboard `1265/1265 pass`.

### Critical
None.

### Suggestions
None — the change is the minimum viable tightening.

### Nitpicks
- The selector regex doesn't anchor word boundaries, so a hypothetical future selector like `.tunnel-warming-banner-extended { transition: none; }` would also satisfy it. Not a real concern given the current CSS, but a `(?=\s*\{)` lookahead or `\s+\{` (without `*`) would be slightly stricter. Below the bar for changing this PR.
- Block extraction on lines 313–325 is duplicated logic that could be hoisted into a helper if more reduced-motion assertions are added later (e.g., the `SidebarDotAnimation.test.ts` already uses a different approach). Out of scope here.

### Verdict
**Approve** — directly resolves #3012, all CI checks pass (12/12), tests verified locally, scope is appropriately tiny.

blamechris:
## Final Review Summary

**Verdict:** Approve

**Counts:**
- Critical: 0
- Suggestions: 0
- Nitpicks: 2 (selector word-boundary, helper extraction — both out of scope)

**Copilot:** 0 inline comments (review state: COMMENTED, summary only). Nothing to triage.

**CI:** 12/12 checks passing.

**Verification:**
- `npx vitest run src/theme/theme.test.ts` — 46/46 pass locally
- Regex `/\.tunnel-warming-banner\s*\{[^}]*transition:\s*none/` correctly enforces co-occurrence in the same rule body (the `[^}]*` clamps to the first closing brace).
- Brace-balancing block extractor on lines 313–325 ensures the regex is fed a single `@media` block at a time.

This is a clean, minimal test tightening that exactly matches the acceptance criteria in #3012. Ready to merge.

PR #3033, opened 2026-04-25
test(server): add unknown-provider fallback case for set_permission_mode capability gate

### LINKED ISSUE #3027: test: add explicit unknown-provider fallback test case for set_permission_mode capability gate
## Context

Identified during review of PR #3009 (capability-gate set_permission_mode for non-supporting providers).

## Description

The capability gate in `handleSetPermissionMode` has four code paths:

1. Known provider + `permissionModeSwitch: false` → reject (tested: Gemini, Codex)
2. Known provider + `permissionModeSwitch: true` → allow (tested: claude-sdk)
3. No `entry.provider` → allow (tested: legacy session)
4. `entry.provider` set to an unregistered string → `getProvider()` throws → `catch {}` → allow (untested)

Path 4 is the fail-open guard for forward compatibility with future providers. The code is trivially correct, but an explicit test would close the coverage gap and document the intent.

## Original Review Comment

> The three branches are: known+false, known+true, and no provider. The fourth branch — `entry.provider` set to an unregistered string — exercises the `catch {}` that fails open. Covered implicitly by the code reading, but an explicit test would close the gap.

## Acceptance Criteria

- [ ] A test case in `settings-handlers.test.js` sets `entry.provider = 'unknown-future-provider'` and asserts that `session.setPermissionMode` is called (fail-open behaviour)
- [ ] No changes to production code required

blamechris:
## Summary

Closes #3027.

The capability gate in `handleSetPermissionMode` (`packages/server/src/handlers/settings-handlers.js` ~line 109-145) has four code paths:

1. Known provider + `permissionModeSwitch:false` (Gemini, Codex) - reject with `CAPABILITY_NOT_SUPPORTED`
2. Known provider + `permissionModeSwitch:true` (claude-sdk) - allow
3. No `entry.provider` (legacy session) - allow
4. `entry.provider` is an unregistered string - `getProvider()` throws, `catch {}` swallows the error, mode change proceeds (fail-open for forward-compat)

Paths 1-3 already had explicit tests. Path 4 - the forward-compatibility fail-open guard - was only covered implicitly. This PR adds an explicit test case.

## Changes

- One new test in `packages/server/tests/handlers/settings-handlers.test.js` under the existing `capability gate (#2963)` describe block.
- The test sets `entry.provider = 'totally-unknown-provider-xyz'`, then asserts:
  - No `CAPABILITY_NOT_SUPPORTED` error is sent (`ws._messages.length === 0`)
  - No session-level error is emitted (`ctx._sent.length === 0`)
  - `session.setPermissionMode` is called once with the requested mode

## Test plan

- [x] `node --test tests/handlers/settings-handlers.test.js` - 30/30 pass
- [x] `node --test tests/settings-handlers*.test.js tests/handlers/settings-handlers.test.js` - 54/54 pass
- [x] Verified the test **fails** when the `try/catch` around `getProvider(entry.provider)` is removed (production code call would throw `Unknown provider "totally-unknown-provider-xyz"`), confirming it guards the fail-open invariant against future regression.
- [x] No production code changes.

### DIFF
--- a/packages/server/tests/handlers/settings-handlers.test.js
+++ b/packages/server/tests/handlers/settings-handlers.test.js
@@ -267,6 +267,33 @@ describe('settings-handlers', () => {
         assert.equal(session.setPermissionMode.callCount, 1)
         assert.equal(ws._messages.length, 0)
       })
+
+      // #3027 — fail-open guard for forward compatibility. When entry.provider
+      // is set to a string that isn't in the provider registry, getProvider()
+      // throws and the handler must swallow the error and let the mode change
+      // proceed. This protects sessions persisted under a future provider name
+      // from being locked out of permission mode after a downgrade. If the
+      // try/catch around getProvider is ever removed, this test will fail
+      // with an "Unknown provider" exception instead of a clean call-through.
+      it('falls open and accepts set_permission_mode when entry.provider is unregistered', () => {
+        const sessions = new Map()
+        const session = createMockSession()
+        sessions.set('s1', { session, name: 'Future', cwd: '/tmp', provider: 'totally-unknown-provider-xyz' })
+        const ctx = makeCtx(sessions)
+        const client = makeClient({ activeSessionId: 's1' })
+        const ws = makeWs()
+
+        settingsHandlers.set_permission_mode(ws, client, { mode: 'approve', requestId: 'r-unknown' }, ctx)
+
+        // No CAPABILITY_NOT_SUPPORTED error sent — the unknown provider
+        // branch must not reject.
+        assert.equal(ws._messages.length, 0, 'no error message should be sent for unknown provider')
+        assert.equal(ctx._sent.length, 0, 'no session-level error should be sent for unknown provider')
+
+        // Mode change proceeds through to the underlying session.
+        assert.equal(session.setPermissionMode.callCount, 1)
+        assert.equal(session.setPermissionMode.lastCall[0], 'approve')
+      })
     })
 
     it('sends confirm_permission_mode for auto mode without confirmation', () => {


blamechris:
## Agent Review — PR #3033

**Verdict: APPROVE**

### Coverage scope verified
- The new test sets `entry.provider = 'totally-unknown-provider-xyz'` — confirmed not in the `PROVIDERS` registry (`packages/server/src/providers.js:22`). `getProvider()` will throw `Unknown provider "totally-unknown-provider-xyz"` and the `catch {}` in `handleSetPermissionMode` (`settings-handlers.js:122-127`) leaves `ProviderClass` undefined, so the capability check at `:128` is bypassed and the mode change proceeds. This is exactly the fail-open path #3027 called out.

### Regression guard verified empirically
- Removed the `try/catch` locally and re-ran the new test in isolation (`node --test --test-name-pattern="falls open" tests/handlers/settings-handlers.test.js`): test fails with the unhandled `Unknown provider` exception, as PR description claims.
- With `try/catch` restored, full file: 30/30 pass.

### Isolation
- Test allocates its own `sessions = new Map()`, `ctx`, `client`, `ws`, `session` mocks. No singleton mutation, no module-level state writes. Safe to run in any order alongside the other 4 capability-gate tests.

### Style
- ES modules, no semicolons, single quotes — matches packages/server convention.
- Comment block above the test documents the forward-compat invariant and the regression-fail mode in plain English.

### CI
- `gh pr checks 3033` shows Server Tests as failing, but the failure is `ws-server-broadcast.test.js` — `skips clients whose bufferedAmount exceeds the threshold` timing out at 2000ms. Unrelated to this PR (no production code changes; only a test was added in `settings-handlers.test.js`). Known flake territory; a re-run should clear it.

### Recommendation
- Re-run Server Tests to clear the unrelated `ws-server-broadcast` flake. Otherwise ready to merge.

PR #3042, opened 2026-04-25
refactor(server): extract isBundledOrSupervisedContext() helper

### LINKED ISSUE #3023: refactor(doctor): extract isBundledOrSupervisedContext() helper
## Context

Identified during review of PR #3004.

## Description

The bundled/supervised context detection in `packages/server/src/doctor.js` is currently inlined at the Dependencies check:

```js
const isBundledContext = process.env.CHROXY_BUNDLED === '1' || process.env.CHROXY_SUPERVISED === '1'
```

As bundled-context awareness spreads to other checks (e.g. cloudflared, Node.js version, port), this pattern will be duplicated. Extracting it into a named helper makes it:

- Reusable across all checks without copy-paste
- Centrally mockable in tests (no need to mutate `process.env` per-check)
- Self-documenting — the name conveys intent better than the raw env var comparison

## Original Review Comment

> Extract the detection logic into a named helper `isBundledOrSupervisedContext()`. The env-var pattern will likely need to be reused in other checks as bundled-context awareness spreads. A single named function also makes it mockable in tests without needing to mutate `process.env`.

## Acceptance Criteria

- [ ] Add `function isBundledOrSupervisedContext()` (or similar name) to `doctor.js`
- [ ] Replace inline env var check at the Dependencies block with the helper call
- [ ] All existing doctor tests remain green

blamechris:
## Summary

- Extracts the inline `CHROXY_BUNDLED || CHROXY_SUPERVISED` env-var check at the Dependencies block in `packages/server/src/doctor.js` into a named, exported helper `isBundledOrSupervisedContext()`.
- Centralises the detection so future bundled-aware checks (cloudflared, Node version, port) can reuse a single source of truth instead of duplicating the env-var pattern.
- The helper is exported, making it directly importable / mockable in tests without mutating `process.env` per check.

Pure refactor — behavior at the existing call site is identical.

Closes #3023

## Test plan

- [x] `node --test tests/doctor*.test.js` — 35 tests pass (5 new helper tests + 30 existing)
- [x] Existing bundled / supervised / dev-context Dependencies tests continue to pass, confirming behavioural equivalence at the call site
- [x] New unit tests assert: false when neither set, true for each var alone, true when both set, and strict `=== '1'` semantics (rejecting `'0'`, `'true'`, `'yes'`, `''`, `'2'`)

### DIFF
--- a/packages/server/src/doctor.js
+++ b/packages/server/src/doctor.js
@@ -26,6 +26,28 @@ const CONFIG_FILE = join(homedir(), '.chroxy', 'config.json')
 const DEFAULT_PROVIDER = 'claude-sdk'
 
 /**
+ * Detect whether the server is running inside a bundled .app (Tauri) or
+ * under the supervisor process. In either case, the end user cannot fix
+ * a missing dependency / broken install themselves by running `npm install`
+ * — they need a reinstall or rebuild. Checks affected by this distinction
+ * (Dependencies, and likely cloudflared / Node version / port soon)
+ * downgrade `fail` → `warn` and surface a context-appropriate hint.
+ *
+ * Centralised here so:
+ *   - The detection lives in one place as more checks adopt it
+ *   - Tests can stub `process.env` once (or import this helper directly)
+ *     instead of duplicating the env-var pattern at every call site
+ *
+ * Exported for tests; production callers should prefer using the helper
+ * from within `runDoctorChecks`.
+ *
+ * @returns {boolean} true when CHROXY_BUNDLED=1 OR CHROXY_SUPERVISED=1
+ */
+export function isBundledOrSupervisedContext() {
+  return process.env.CHROXY_BUNDLED === '1' || process.env.CHROXY_SUPERVISED === '1'
+}
+
+/**
  * Resolve the list of providers to preflight check.
  *
  * Precedence:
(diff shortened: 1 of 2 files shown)


blamechris:
## Code Review: PR #3042

### Summary
Pure refactor extracting the inline `CHROXY_BUNDLED === '1' || CHROXY_SUPERVISED === '1'` check from the Dependencies block in `doctor.js` into an exported helper `isBundledOrSupervisedContext()`. Closes #3023. +94/-3.

### Strengths
- **Helper is well-scoped and documented.** JSDoc explains the rationale (bundled/supervised users can't `npm install` to fix), the centralisation motive (future cloudflared/Node/port checks), and the export-for-tests intent.
- **Tests cover the full boolean truth table** (none / bundled / supervised / both) plus strict-equality semantics (rejecting `'0'`, `'true'`, `'yes'`, `''`, `'2'`). Save/restore env pattern in a `try/finally` per test prevents state leak across the suite — important since the helper is global-state-dependent.
- **Behavioural equivalence verified:** the existing 30 doctor tests still pass alongside the 5 new ones (35 total), confirming the call site at the Dependencies block behaves identically.
- **Correct restraint on adjacent call sites.** I grepped `packages/server/src/` for `CHROXY_BUNDLED|CHROXY_SUPERVISED` — the only other consumers are `cli/server-cmd.js` (line 59 `!== '1'` and line 86 `=== '1'`, both single-var checks gating "am I the supervised child?") and `supervisor.js` line 277 (producer, not consumer). Neither matches the "OR-of-both" pattern this helper encodes; collapsing them would change semantics. Leaving them alone is the correct call.
- Style compliant: ES modules, no semicolons, single quotes.

### Issues Found

#### Critical (Must Fix)
None.

None — the JSDoc already flags this as forward-looking and the issue explicitly defers further adoption to follow-up checks (cloudflared, Node version, port).

#### Nitpicks (Optional)
None.

### Architecture Notes
Extraction is in the right file (co-located with `runDoctorChecks` and other doctor-private helpers) and exported only for testability. The hint that production callers should "prefer using the helper from within `runDoctorChecks`" is good — keeps the API surface intentional. When cloudflared/Node/port checks adopt context-awareness, this helper will naturally absorb them with no further refactor.

### Verdict

CI green (12/12), pure refactor, behaviour preserved, scope correct.

PR #3044, opened 2026-04-25
fix(clients): dedup tool_use entries on reconnect history replay

### LINKED ISSUE #2901: Reconnect replay may duplicate tool_use entries in dashboard + app
## Context

Identified during review of PR #2900.

## Description

`packages/dashboard/src/store/message-handler.ts:815-839` has a `_receivingHistoryReplay && !_isSessionSwitchReplay && get().messages.length > 0` early-return that fires only on the legacy flat `messages` array, which is empty in multi-session mode. As a result tool_use entries from a non-session-switch replay are appended without dedup: the stable-ID dedup at :819-823 only runs when `_isSessionSwitchReplay` is true.

A reconnecting client whose `sessionStates[s].messages` already holds a tool_use from before the disconnect will end up with two copies after replay. The app's `handleToolStart` likely has the same gap.

## Original Review Comment

> Dashboard `handleToolStart` blanket guard and missing reconnect dedup. Out of scope for #2900 (which tightly focuses on `user_input`). The tool_use duplication is a separate reconnect-replay gap that pre-existed; best tracked and fixed in its own PR with tests.

## Acceptance Criteria

- [ ] `handleToolStart` in both dashboard and app always runs the stable `toolId` dedup during any replay, not just session-switch
- [ ] Regression tests cover the plain-reconnect-with-existing-tool_use case
- [ ] Blanket `get().messages.length > 0` guard removed in favor of per-entry dedup

blamechris:
## Summary
- Dashboard `handleToolStart` was duplicating `tool_use` entries during plain reconnect replay. Two issues at the same handler:
  1. A blanket `_receivingHistoryReplay && !_isSessionSwitchReplay && get().messages.length > 0` guard tested the legacy flat `messages` array, which is empty in multi-session mode — so it never fired.
  2. The per-id stable dedup only ran when `_isSessionSwitchReplay` was true, leaving plain-reconnect replay with no dedup at all.
- Result: a reconnecting dashboard whose `sessionStates[s].messages` already held a tool_use from before the disconnect ended up with two copies after replay.
- This is the analogous fix to #2900, which solved the same shape of bug for `user_input`.

## Root cause
- `packages/dashboard/src/store/message-handler.ts:835` — blanket guard against legacy `messages` length, never tripped under multi-session state.
- `packages/dashboard/src/store/message-handler.ts:839` — per-id dedup gated on `_isSessionSwitchReplay`.

## Fix
- Drop the blanket length>0 guard.
- Always run per-session-id dedup during any history replay (`_receivingHistoryReplay`), regardless of the session-switch flag. Same shape as #2900's `message`-handler fix.

The mobile app's `tool_start` handler already implemented this pattern correctly (`packages/app/src/store/message-handler.ts:1281-1308`); no app change was needed.

## Tests
- Dashboard (vitest): 5 new cases under `history replay: tool_start dedup (#2901)` — plain-reconnect dedup, session-switch dedup, legitimate new tool_use during replay, legacy-messages-non-empty case, live (non-replay) path. Suite: 31/31 in this file. Repo-wide: 476 passing (471 before + 5 new); 58 file-level failures pre-exist on main due to a missing `@testing-library/react` dev dep, unrelated to this PR.
- store-core (vitest): 150/150 passing.
- App (jest): 120/120 across both message-handler test files (no app-side change, just regression coverage).

## Test plan
- [x] Dashboard unit tests pass
- [x] store-core unit tests pass
- [x] App unit tests still pass (no change there)
- [ ] Manual verify: open dashboard against a CLI session that just used a tool, force a reconnect, confirm the tool_use bubble does not double up.

Closes #2901

### DIFF
--- a/packages/dashboard/src/store/connection.ts
+++ b/packages/dashboard/src/store/connection.ts
@@ -86,7 +86,6 @@ import {
   setDisconnectedAttemptId,
   lastConnectedUrl,
   setLastConnectedUrl,
-  setPendingSwitchSessionId,
   resetReplayFlags,
   clearPermissionSplits,
   clearTerminalWriteBatching,
@@ -1261,9 +1260,6 @@ export const useConnectionStore = create<ConnectionState>((set, get) => ({
     const { socket, activeSessionId, sessionStates } = get();
 
     if (sessionId === activeSessionId) return;
-
-    // Mark as user-initiated switch so session_switched handler uses session-switch dedup
-    setPendingSwitchSessionId(sessionId);
 
     // Optimistically switch to cached state + dismiss notifications for target session
     const cached = sessionStates[sessionId];
--- a/packages/dashboard/src/store/message-handler.test.ts
+++ b/packages/dashboard/src/store/message-handler.test.ts
@@ -415,6 +415,138 @@ describe('dashboard message-handler dispatch', () => {
     })
   })
 
+  describe('history replay: tool_start dedup (#2901)', () => {
+    function seedWithTool(toolId: string) {
+      store = createMockStore(
+        baseState({
+          activeSessionId: 's1',
+          sessions: [{ sessionId: 's1', name: 'S1' } as any],
+          sessionStates: {
+            s1: {
+              ...createEmptySessionState(),
+              messages: [
+                { id: toolId, type: 'tool_use', content: 'Bash: ls', tool: 'Bash', timestamp: 1 },
+              ],
+            },
+          },
+        }),
+      )
+      setStore(store)
+    }
+
+    // Regression: on plain reconnect replay (not a session switch), the
+    // dashboard's `tool_start` handler had a blanket
+    // `_receivingHistoryReplay && !_isSessionSwitchReplay && get().messages.length > 0`
+    // early return that fired against the legacy flat `messages` array — but
+    // multi-session state keeps that array empty, so the guard never tripped
+    // and replayed tool_use entries appended on top of the live copies. The
+    // per-id dedup at the same handler now runs on every replay path.
+    it('deduplicates tool_use by stable messageId during plain reconnect replay', () => {
+      seedWithTool('tool-1')
+      handleMessage({ type: 'history_replay_start', sessionId: 's1' }, ctx() as any)
+      handleMessage(
+        {
+          type: 'tool_start',
+          messageId: 'tool-1',
+          tool: 'Bash',
+          input: 'ls',
+          sessionId: 's1',
+        },
+        ctx() as any,
+      )
+      const msgs = (store.getState() as any).sessionStates.s1.messages
+      expect(msgs).toHaveLength(1)
+      expect(msgs[0].id).toBe('tool-1')
+    })
+
+    it('deduplicates tool_use by stable messageId during session-switch replay', () => {
+      seedWithTool('tool-1')
+      handleMessage(
+        { type: 'history_replay_start', sessionId: 's1', fullHistory: true },
+        ctx() as any,
+      )
+      // session-switch replay clears messages first, so re-seed a tool to
+      // simulate the replay sending the same tool a client already cached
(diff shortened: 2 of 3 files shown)


blamechris:
## Code Review: PR #3044

### Summary

Fixes a dashboard reconnect-replay bug where `tool_use` entries duplicated after reconnect because the replay dedup was gated on `_isSessionSwitchReplay`, leaving plain-reconnect replay with no dedup at all. Same shape as #2900's `user_input` fix, applied to `tool_start`. Also removes a stale `messages.length > 0` guard that tested the legacy flat array (always empty in multi-session mode).

### Strengths

- Correctly uses the server's `messageId` as the dedup key (stable across live and replay paths). This is the right key — `toolUseId` would refer to the SDK content block id which is conceptually different.
- The `cached.some((m) => m.id === toolId)` check uses the per-session `targetState.messages`, not the legacy flat `messages`, so it actually exercises the right cache.
- Test coverage is thoughtful: plain-reconnect dedup, session-switch dedup, legitimate-new-tool-during-replay, legacy-flat-non-empty edge case, and the live-event no-dedup negative case (5 tests).
- Comment block at the new dedup explains the regression rationale clearly — future readers will understand why the blanket guard was removed.
- App-side claim verified: `packages/app/src/store/message-handler.ts:1281-1308` already runs per-id dedup on every `_ctx.receivingHistoryReplay` path, so no app change is needed.

### Issues Found

#### Critical (Must Fix)

| `packages/dashboard/src/store/message-handler.ts` | 189 | CI Dashboard Type Check fails: `TS6133: '_isSessionSwitchReplay' is declared but its value is never read`. Removing both reads (lines 835 and 839 pre-fix) leaves only writes. `noUnusedLocals: true` flags it. | Drop the variable + all its writes (and the now-orphaned `_pendingSwitchSessionId` / `setPendingSwitchSessionId` chain in `connection.ts`, which only existed to gate the removed flag). Pushed as a follow-up commit on this branch. |

#### Suggestions

None blocking. The dedup logic is tight: fresh tool_use ids pass through, duplicates are skipped only inside `_receivingHistoryReplay`, and live-path duplicates (which the tests confirm are unusual but still appended) match the original semantics.

#### Nitpicks

- The session-switch test re-injects `messages` directly via `;(store.getState() as any).sessionStates.s1.messages = [...]` to bypass the `fullHistory: true` clear. Works, but a bit fragile if `createMockStore` ever copies-on-write. Not worth fixing.

### Architecture Notes

This is the `tool_use` half of a two-part fix originally split out of #2900 into its own issue (#2901). The pattern is now consistent across dashboard and app: any `_receivingHistoryReplay` (plain reconnect or session switch) → check per-id cache → skip if present, append otherwise. After this PR + #2900, both `user_input` and `tool_use` rehydrate cleanly on reconnect without duplication. The session-switch flag (`_isSessionSwitchReplay` / `_ctx.isSessionSwitchReplay`) was the last remaining branch point for replay handling and is now redundant on the dashboard side; it can be similarly retired from the app handler in a future tidy.

### Verdict

- [x] **Approve** — once the type-check fix lands. Logic is correct, tests cover the regression and the negative cases, no app change needed.

blamechris:
## Review Comments Addressed

No inline review comments — Copilot reviewed with summary-only state (no concerns flagged on the dedup logic).

**Verification:**
- Removed `messages.length > 0` and `_isSessionSwitchReplay` guards from `handleToolStart` — these were preventing legitimate replay-dedup from running on plain reconnects.
- Per-id dedup now runs on both fresh AND replayed events; checks `existing.id === messageId` to skip already-known tool_use entries.
- Negative case (legit new tool_use NOT skipped) covered by 5 new test cases in `dashboard message-handler.test.ts`.
- App side already had the correct pattern at `packages/app/src/store/message-handler.ts:1281-1308` (verified during review).

**CI:** all 12 checks pass.

Ready to merge.

Issue #3057, opened 2026-04-26
feat(server): audit trail for auto-deny permission resolutions (timeout/abort/clearAll)


blamechris:
## Background

`packages/server/src/handlers/settings-handlers.js:322-329` records permission decisions to the audit trail via `ctx.permissionAudit.logDecision`, but only when the resolution arrives via the WebSocket `permission_response` handler:

\`\`\`js
if (resolved && ctx.permissionAudit) {
  ctx.permissionAudit.logDecision({
    clientId: client.id,
    sessionId: originSessionId,
    requestId,
    decision,
  })
}
\`\`\`

Auto-deny resolutions (5-min timeout, abort signal, `clearAll` on session destroy) — the three paths that PR #3055 just added to the broadcast pipeline — never call `logDecision`. They emit on the wire so clients dismiss prompts, but leave no audit record.

## Why this matters

A permission for `Bash rm -rf /` that times out after 5 min currently leaves zero trace in the audit log. Forensic review of "what permissions were asked, and what happened to them" misses a real category of resolution.

## Proposed fix

Move the audit logging into `PermissionManager` itself (or into the `SessionManager.on('session_event')` handler in `server-cli.js` or wherever the audit instance is wired) so every emit path writes the audit entry uniformly. The `reason` field on the emit (`'user'`, `'timeout'`, `'aborted'`, `'cleared'`) maps cleanly to a new audit event subtype.

## Note

Pre-existing — not regressed by #3055. Surfaced during review of #3055 because the new pipeline is the natural place to fix it.

Reference: https://github.com/blamechris/chroxy/pull/3055#issuecomment-4321209251

## Acceptance criteria

- [ ] All four resolution paths produce a `permissionAudit.logDecision` entry
- [ ] Entry includes the `reason` (`user`, `timeout`, `aborted`, `cleared`) so auto-deny paths are distinguishable from user denies
- [ ] Existing audit query API still works

PR #3062, opened 2026-04-26
fix(server): defensive requestId guard in audit listener (#3061)

### LINKED ISSUE #3061: Defensive validation in audit listener: skip permission_resolved events without requestId
## Context

Discovered during review of #3058. The audit listener at `ws-server.js:588-598` doesn't validate `data.requestId` before calling `logDecision`. Today this is fine because `sdk-session.js:200` gates `permission_resolved` re-emission on `data && data.requestId` — events without a requestId (AskUserQuestion lifecycle, `clearAll` had-user-answer path) never reach the unified pipeline.

If someone in the future widens the SdkSession gate to propagate AskUserQuestion lifecycle events, the audit listener would silently start writing entries with `requestId: undefined`.

## Proposed fix

Add a defensive check to the listener:

```js
sessionManager.on('session_event', ({ sessionId, event, data }) => {
  if (event !== 'permission_resolved') return
  if (!data || data.reason === 'user') return
  if (!data.requestId) return  // <-- new
  this._permissionAudit.logDecision({ ... })
})
```

Tests should cover:
- `data: null` on a `permission_resolved` event (no entry, no crash).
- `data: { reason: 'timeout' }` with no requestId (no entry, no crash).
- Existing happy paths still work.

## Acceptance criteria

- [ ] Listener short-circuits when `data.requestId` is falsy.
- [ ] Tests assert the new safety behavior.

## References

- Review on #3058: https://github.com/blamechris/chroxy/pull/3058#issuecomment-4321323828

blamechris:
## Summary

Closes #3061.

Adds a defensive `if (!data.requestId) return` to the `permission_resolved` audit listener in `ws-server.js`. Today this branch is unreachable because `sdk-session.js:200` gates the re-emit on `data && data.requestId` — so AskUserQuestion lifecycle events and the `clearAll` had-user-answer path never reach the pipeline. The guard exists so that if the SdkSession gate is ever widened (e.g. to propagate AskUserQuestion events), the audit listener doesn't silently start writing entries with `requestId: undefined`.

## Diff

```diff
this._sessionEventAuditHandler = ({ sessionId, event, data }) => {
  if (event !== 'permission_resolved') return
  if (!data || data.reason === 'user') return
+ // #3061: defensive — if the SdkSession requestId gate is ever widened ...
+ if (!data.requestId) return
  this._permissionAudit.logDecision({ ... })
}
```

## Test plan

- [x] `null` / `undefined` data — no crash, no entry
- [x] Missing `requestId` — no entry written
- [x] Existing happy paths still produce entries (covered by previously-merged tests)
- [x] All 45 affected-suite tests pass

Related to #3057, #3058.

### DIFF
--- a/packages/server/src/ws-server.js
+++ b/packages/server/src/ws-server.js
@@ -591,6 +591,11 @@ export class WsServer {
       this._sessionEventAuditHandler = ({ sessionId, event, data }) => {
         if (event !== 'permission_resolved') return
         if (!data || data.reason === 'user') return
+        // #3061: defensive — if the SdkSession requestId gate is ever widened
+        // (e.g. to propagate AskUserQuestion lifecycle events), don't silently
+        // write audit entries with requestId: undefined. Today this branch is
+        // unreachable because sdk-session.js gates the re-emit on data.requestId.
+        if (!data.requestId) return
         this._permissionAudit.logDecision({
           clientId: null,
           sessionId,
--- a/packages/server/tests/ws-server-permissions.test.js
+++ b/packages/server/tests/ws-server-permissions.test.js
@@ -1701,6 +1701,48 @@ describe('audit trail for auto-deny resolution paths (#3057)', () => {
       'reason=user must not be audited from the pipeline; the inline path owns those')
   })
 
+  // #3061: defensive guard. Today the SdkSession re-emit gate at
+  // sdk-session.js:200 only re-emits when data.requestId is truthy, so these
+  // shapes never reach the audit listener via the real pipeline. The guard
+  // exists so that a future widening of the gate (e.g. to propagate
+  // AskUserQuestion lifecycle events) doesn't silently produce audit entries
+  // with requestId: undefined.
+  it('skips permission_resolved with null/undefined data (#3061)', () => {
+    const manager = makeManager()
+    server = new WsServer({ port: 0, apiToken: 'test-token', sessionManager: manager })
+
+    manager.emit('session_event', {
+      sessionId: 'sess-x',
+      event: 'permission_resolved',
+      data: null,
+    })
+    manager.emit('session_event', {
+      sessionId: 'sess-x',
+      event: 'permission_resolved',
+      // data omitted entirely
+    })
+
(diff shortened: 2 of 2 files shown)


blamechris (commented):
## Summary

Tiny defensive guard PR. Adds `if (!data.requestId) return` to the `permission_resolved` audit listener in `ws-server.js` (lines 591–606), backstopping the upstream `data && data.requestId` gate in `sdk-session.js:200`. Two tests cover the safety branch.

## Strengths

- **Guard placement is correct** (`ws-server.js:598`): after `!data || data.reason === 'user'`, before `logDecision`. Existing `!data` filter at line 593 protects the `data.requestId` access, so the new line can't NPE.
- **Comment explains _why_** (`ws-server.js:594-597`): cites #3061, names the future-widening concern, and points at the upstream gate (`sdk-session.js`). This is exactly the rationale a reader six months from now needs.
- **Test for missing `requestId`** (`ws-server-permissions.test.js:1729-1744`) is the meaningful one — uses a realistic auto-deny shape (`{ decision: 'deny', reason: 'cleared' }`) that mirrors the AskUserQuestion `clearAll` emit shape called out in the comment. Asserts the listener short-circuits before `logDecision`.
- **Test for null/undefined data** (`:1710-1727`) is a regression backstop for the pre-existing `!data` filter rather than the new line, but it's cheap and reasonable to keep alongside.
- Test comments cross-reference `sdk-session.js:200` so the invariant is traceable from both ends.
- Zero Attribution Policy: clean.

## Issues

None.

## Notes (not blocking, not worth filing)

`permission-audit.js:65` doesn't validate `decision`, so a malformed event could in theory produce an entry with `decision: undefined`. In practice, all current emit sites in `permissions.js` always set it, and adding speculative validation here would scope-creep. Same defensive-depth philosophy as this PR would suggest a `decision` check, but it's noise until there's a concrete reason.

## Verdict

**Approve.** Minimal, well-scoped, correctly placed, and the comment plus tests preserve the invariant for future readers. Merge once CI finishes.

Issue #3071, opened 2026-04-26
app: chat output drifts from desktop dashboard — text concatenates into tool_use messages


blamechris:
## Summary

The same session shows different chat content on the Android app vs the desktop dashboard. On the desktop dashboard the assistant text and tool_use messages are properly threaded; on Android the assistant text appears to concatenate **into** the tool_use bubbles, the standalone response bubble never appears, and large structured outputs (markdown tables, etc.) are missing entirely.

This is the same shape as the documented `stream_start` ID collision pattern — but the `deltaIdRemaps` fix is already present in both `packages/app/src/store/message-handler.ts` and `packages/dashboard/src/store/message-handler.ts`, so the symptom isn't covered by the existing remap. Looks like a second site that wasn't migrated, or the app's remap key/lifecycle differs.

## Expected
The Android app and the desktop dashboard render the same session identically — same messages, same threading, same content.

## Actual

Same `/tackle-issues` invocation, same session, both clients connected:

**Desktop dashboard** (clean):
```
[user] Hello what model and version are you running?
[assistant] I'm Claude Opus 4.7 (1M context), model ID claude-opus-4-7[1m].
[user] /tackle-issues
[assistant] (text)
[tool_use Bash × N] (separate bubbles, with full command)
[assistant] Let me check some of the borderline issues...
[tool_use Bash × M]
[assistant] Before showing the queue, here's my assessment ... (full markdown table with 5 issues, all decision points)
```

**Android app** (drifted):
```
[user] Hello what model and version are you running?
[assistant] (empty)             ← lost
[user] /tackle-issues
[assistant] I'll start the marathon session setup... (text)
[tool_use Bash] gathering the repo info and building the initial queue.Let me check some of the borderline issues for actionability...   ← assistant text is INSIDE the tool_use
[tool_use Bash] Let me check some of the borderline...   ← repeated content
[Prompt] Bash: Get current repo name
[tool_use Bash] (concatenated text again)
... (marathon session table never appears at all)
```

So:
- The first turn's response text is missing on Android.
- Subsequent assistant text appears smashed into tool_use bubbles.
- The big structured output (queue table, decision points) doesn't render on Android at all.

## Repro
1. Connect both desktop dashboard and Android app to the same server (use the new linking-mode QR from #3070 path).
2. Open a session and run any prompt that triggers a long sequence of `tool_use` then a long final assistant message — `/tackle-issues` is a reliable trigger because it does N back-to-back bash calls and ends with a markdown table.
[...]

Issue #2661, opened 2026-03-23
refactor: unify app/dashboard message handlers into store-core


blamechris:
**Size:** Large (epic)
**Sources:** Minimalist F1, Futurist F1

`packages/app/src/store/message-handler.ts` and `packages/dashboard/src/store/message-handler.ts` implement nearly identical logic for:
- WebSocket message parsing and dispatch
- Delta accumulation for streaming responses
- Tool result handling and state transitions
- Session state management

Combined ~4,500 lines of duplicated code that must be kept in sync manually. Divergence has been observed and addressed across many PRs in this epic.

**Proposed:** Extract shared message handling logic into `packages/store-core/` as a platform-agnostic module. App and dashboard import and adapt with thin platform-specific wrappers.

---

## Status (snapshot 2026-05-02)

This epic has been chipping away batch-by-batch via the `from-review` follow-up cycle. ~30+ handler migrations have shipped through PRs #3127, #3128, #3129, #3134, #3136, #3140, #3143, #3145, #3148, #3153, #3156, #3170, #3173, #3174, #3180. Recent batches:

- **PR #3170** — keep stop button visible during tool-first turns (app + dashboard parity)
- **PR #3173** — `agent_idle` clears `streamingMessageId` (store-core handler)
- **PR #3174** — port dashboard's #2611 orphan-create safety net to app
- **PR #3180** — batch type-safety + dedup cleanup (16 from-review issues, including discriminated union for `MessagePayload`, shared `applyOrphanDeltas` helper, `LogEntry` consolidation)

## Remaining work

The bulk-handler migration is largely done. What's left tends to surface as new `from-review` follow-ups whenever the next batch lands. Tracking continues but no new sub-issues are needed under this parent — sweep follow-ups under their own labels (`from-review`, `subtask`).

PR #3097, opened 2026-04-27
refactor: share plan_started handler via store-core

blamechris:
## Summary

Migrates one more duplicated WS handler — `plan_started` — out of the app + dashboard message-handlers and into `@chroxy/store-core`. This is the **thirteenth** handler in store-core (after `model_changed`, `permission_mode_changed`, `available_permission_modes`, `session_updated`, `confirm_permission_mode`, `claude_ready`, `agent_idle`, `agent_busy`, `thinking_level_changed`, `budget_warning`, `budget_exceeded`, `budget_resumed` already there).

## Why this one, why now

#2661 calls out 4,525 lines of duplicated message-handler code across `packages/app/src/store/message-handler.ts` and `packages/dashboard/src/store/message-handler.ts` — a known divergence risk (the recent #3071 stream-collision bug hit both). The right shape for chipping at that debt is **one nibble per PR**, not a big-bang. This continues the pattern started in #3093.

`plan_started` is the next-smallest remaining duplication: 10 lines on each side, byte-for-byte identical (verified pre-migration), no side effects, no dependency on platform-specific store APIs. The migration is mechanical and the risk surface is tiny.

## Implementation

- New `handlePlanStarted(msg, activeSessionId)` in `store-core/src/handlers/index.ts` returning a `SessionPatch` (`{ sessionId, patch }`). This is the first handler to actually use the `SessionPatch` interface that was added (but unused) in #3093 — sets the precedent for future session-scoped patch handlers.
- Both message-handlers replace their inline 10-line case with a 4-line call to the shared helper. The caller still applies the existing `&& sessionStates[id]` guard before invoking `updateSession`, matching prior behavior exactly.

## Test plan

- [x] 3 new vitest cases in `packages/store-core/src/handlers/handlers.test.ts` covering: explicit sessionId, fallback to active session, neither available
- [x] `store-core` tests: 159 pass (was 156 → +3)
- [x] `dashboard` tests: 1285 pass
- [x] `app` tests: 1142 pass

Refs #2661 (one nibble of the incremental migration; does NOT close — the bigger streaming/delta cases are next, in their own PRs)

### DIFF
--- a/packages/app/src/store/message-handler.ts
+++ b/packages/app/src/store/message-handler.ts
@@ -40,6 +40,7 @@ import {
   handleBudgetWarning as sharedBudgetWarning,
   handleBudgetExceeded as sharedBudgetExceeded,
   handleBudgetResumed as sharedBudgetResumed,
+  handlePlanStarted as sharedPlanStarted,
 } from '@chroxy/store-core';
 import { PROTOCOL_VERSION } from '@chroxy/protocol';
 import { hapticSuccess } from '../utils/haptics';
@@ -1632,12 +1633,9 @@ export function handleMessage(raw: unknown, ctxOverride?: ConnectionContext): vo
     }
 
     case 'plan_started': {
-      const planStartTargetId = (msg.sessionId as string) || get().activeSessionId;
-      if (planStartTargetId && get().sessionStates[planStartTargetId]) {
-        updateSession(planStartTargetId, () => ({
-          isPlanPending: false,
-          planAllowedPrompts: [],
-        }));
+      const planStarted = sharedPlanStarted(msg, get().activeSessionId);
+      if (planStarted.sessionId && get().sessionStates[planStarted.sessionId]) {
+        updateSession(planStarted.sessionId, () => planStarted.patch);
       }
       break;
     }
--- a/packages/dashboard/src/store/message-handler.ts
+++ b/packages/dashboard/src/store/message-handler.ts
@@ -29,6 +29,7 @@ import {
   handleBudgetWarning as sharedBudgetWarning,
   handleBudgetExceeded as sharedBudgetExceeded,
   handleBudgetResumed as sharedBudgetResumed,
+  handlePlanStarted as sharedPlanStarted,
   type PlatformAdapters, type StorageAdapter,
 } from '@chroxy/store-core'
 import { PROTOCOL_VERSION } from '@chroxy/protocol'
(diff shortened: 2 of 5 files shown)


blamechris:
## Review: refactor: share plan_started handler via store-core

**Verdict: Approve** — clean mechanical migration, the second nibble at #2661, and the first real consumer of the `SessionPatch` shape introduced (but unused) in #3093. No blockers.

### What I checked

- **Byte-equivalence with prior inline blocks** — confirmed. The patch (`{ isPlanPending: false, planAllowedPrompts: [] }`) matches both clients exactly.
- **`sessionStates[id]` guard preserved** — yes, both call sites still gate on `if (planStarted.sessionId && get().sessionStates[planStarted.sessionId])` before invoking `updateSession`. The shared helper is intentionally pure/stateless and pushes that responsibility back to the caller, which matches the established pattern.
- **`SessionPatch` shape used correctly** — the patch is applied via `updateSession(id, () => planStarted.patch)` on both sides, consistent with how `updateSession` consumes a `Partial<SessionState>` updater. Type-check passes in CI on both `app` and `dashboard`.
- **Scope** — diff stat confirms only the 5 declared files (`store-core/handlers/index.ts`, `handlers.test.ts`, `index.ts`, `app/.../message-handler.ts`, `dashboard/.../message-handler.ts`). No collateral.
- **CI** — all 12 checks green; Copilot reviewed with no comments.

### Minor observations (non-blocking)

1. **Subtle behavior tightening from `resolveSessionId`**: the prior inline `(msg.sessionId as string) || get().activeSessionId` would pass through whitespace-only strings like `"  "` as truthy and target a nonsense session ID (which would then be filtered by the `sessionStates[id]` guard anyway). The new `resolveSessionId` trims first and rejects whitespace-only, so a malformed `sessionId: "  "` now correctly falls back to the active session. This is a harmless improvement and matches the precedent set in the previous nibble (#3093) — flagging only because it's a behavior delta worth noting in the PR description for future archaeology.
2. **Test coverage**: 3 new tests cover the three resolution paths well. Malformed-type cases (e.g. `sessionId: 42`) are implicitly exercised by `resolveSessionId`'s own dedicated test block earlier in the file, so no need to duplicate here.
3. **Naming**: the `handlePlanStarted as sharedPlanStarted` import alias and the local `planStarted` variable read cleanly and match the established pattern from the prior 12 handlers — no nit.

### Notes for future nibbles

- `plan_ready` (immediately below the migrated case in both files) is the natural next candidate — same `SessionPatch` shape, slightly more interesting because it parses `allowedPrompts` from the message. The dashboard version even *omits* the `pushSessionNotification` call that the app makes (line 1654 in app), so that one will need an explicit decision about side effects: either keep the notification call out of the shared handler and let each caller invoke it, or include it in the shape and have the dashboard adapter no-op. Worth flagging when that PR lands.

Refs #2661 — correctly noted as "does NOT close" in the PR body.

PR #3099, opened 2026-04-27
refactor: import shared PendingPermissionConfirm from store-core

### LINKED ISSUE #3094: refactor: use shared PendingPermissionConfirm type in app/dashboard store types
Follow-up from review of #3093.

## Original finding

> `PendingPermissionConfirm` is exported from store-core, but both `packages/app/src/store/types.ts:257` and `packages/dashboard/src/store/types.ts:373` still inline `{ mode: string; warning: string } | null`. Out of scope for this nibble but a natural follow-up — replacing those with the shared type would make the next divergence impossible.

## Context

PR #3093 added `PendingPermissionConfirm` as an exported interface in `@chroxy/store-core` and re-exported it from the package barrel. The handler now returns this type, but the two store `types.ts` files that hold the actual `pendingPermissionConfirm` field on the connection-store state still declare the type structurally and inline:

```ts
// packages/app/src/store/types.ts:257
pendingPermissionConfirm: { mode: string; warning: string } | null;

// packages/dashboard/src/store/types.ts:373
pendingPermissionConfirm: { mode: string; warning: string } | null;
```

Today the structural types happen to match the shared `PendingPermissionConfirm` exactly, so type-checking lets the migration through. But if anyone widens or narrows the shared type later (say, adding an optional `expiresAt`), the two state types won't pick it up — exactly the divergence #2661 is trying to eliminate.

## Recommendation

In both files:

1. `import type { PendingPermissionConfirm } from '@chroxy/store-core'`
2. Replace the inline annotation with `pendingPermissionConfirm: PendingPermissionConfirm | null`

Also worth scanning `packages/app/src/components/SettingsBar.tsx:41` where the prop type is duplicated again — the same swap applies.

## Acceptance

- [ ] `packages/app/src/store/types.ts` uses `PendingPermissionConfirm` from store-core
- [ ] `packages/dashboard/src/store/types.ts` uses `PendingPermissionConfirm` from store-core
- [ ] `packages/app/src/components/SettingsBar.tsx` prop type uses `PendingPermissionConfirm | null`
- [ ] All three packages still pass `tsc --noEmit`
- [ ] Existing tests pass without modification

## Refs

- PR #3093 (introduced the shared type)
- #2661 (parent: app/dashboard message-handler unification)

blamechris:
## Summary

Follow-up cleanup from review of #3093. Replace the three inline `{ mode: string; warning: string }` declarations of `pendingPermissionConfirm` with the shared `PendingPermissionConfirm` interface that #3093 added to `@chroxy/store-core`, so a future widening of the shared shape (e.g. an optional `expiresAt`) flows through without divergence.

Files touched:
- `packages/app/src/store/types.ts` — re-export + local-use import + field type
- `packages/dashboard/src/store/types.ts` — re-export + local-use import + field type
- `packages/app/src/components/SettingsBar.tsx` — type-only import + prop type

Shapes were identical across all three call sites, so this is a pure type-import substitution.

## Test plan

- [x] `npx tsc --noEmit -p packages/app` — only pre-existing `xterm-bundle.generated` error remains
- [x] `npx tsc --noEmit -p packages/dashboard` — no new errors from this change
- [x] `npm test -w @chroxy/app` — 81 suites, 1142 tests pass
- [x] `npm test -w @chroxy/dashboard` — 35 suites / 486 tests pass; the 58 pre-existing suite failures (missing `@testing-library/react`) match `main` exactly

Closes #3094

### DIFF
--- a/packages/app/src/components/SettingsBar.tsx
+++ b/packages/app/src/components/SettingsBar.tsx
@@ -2,6 +2,7 @@ import React, { useEffect, useRef, useState } from 'react';
 import { View, Text, TouchableOpacity, StyleSheet, Platform, Animated, AccessibilityInfo, Alert } from 'react-native';
 import * as Clipboard from 'expo-clipboard';
 import { DEFAULT_CONTEXT_WINDOW } from '@chroxy/store-core';
+import type { PendingPermissionConfirm } from '@chroxy/store-core';
 import { ModelInfo, ContextUsage, AgentInfo, ConnectedClient, CustomAgent, SessionContext, McpServer } from '../store/connection';
 import { Icon } from './Icon';
 import { COLORS } from '../constants/colors';
@@ -38,7 +39,7 @@ export interface SettingsBarProps {
   onInvokeAgent?: (agentName: string) => void;
   setModel: (model: string) => void;
   setPermissionMode: (mode: string) => void;
-  pendingPermissionConfirm?: { mode: string; warning: string } | null;
+  pendingPermissionConfirm?: PendingPermissionConfirm | null;
   onConfirmPermissionMode?: (mode: string) => void;
   onCancelPermissionConfirm?: () => void;
   conversationId?: string | null;
--- a/packages/app/src/store/types.ts
+++ b/packages/app/src/store/types.ts
@@ -36,6 +36,7 @@ export type {
   QueuedMessage,
   Checkpoint,
   BaseSessionState,
+  PendingPermissionConfirm,
 } from '@chroxy/store-core';
 
 // Import for local use in SessionState/ConnectionState definitions below
@@ -53,6 +54,7 @@ import type {
   McpServer,
   MessageAttachment,
   ModelInfo,
+  PendingPermissionConfirm,
   SearchResult,
   SessionContext,
   SessionHealth,
@@ -254,7 +256,7 @@ export interface ModelsAndPermissionsData {
   // Available providers from server (for session creation UI)
   availableProviders: ProviderInfo[];
   // Pending auto permission mode confirmation from server
-  pendingPermissionConfirm: { mode: string; warning: string } | null;
+  pendingPermissionConfirm: PendingPermissionConfirm | null;
 }
 
 /**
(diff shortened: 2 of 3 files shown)


blamechris:
## Review — refactor: import shared PendingPermissionConfirm from store-core

**Verdict: LGTM**

Pure type-import substitution, exactly matches the acceptance criteria from #3094. CI is green across all 12 jobs.

### Verification

**Shape equivalence (the load-bearing question):** The original inline shapes are byte-for-byte identical to the shared interface — both required, both `string`, no extra properties:

- `packages/store-core/src/handlers/index.ts:119` — `interface PendingPermissionConfirm { mode: string; warning: string }`
- All three replaced sites declared `{ mode: string; warning: string } | null`

No divergence is being papered over. `handleConfirmPermissionMode()` already constructs values of this exact shape (defaulting `warning` to `'Are you sure?'` when missing), so the runtime behavior was already aligned with the shared type.

**Dual-block pattern is correct.** Both `types.ts` files use the documented pattern of (1) `export type {...}` re-export for downstream consumers, plus (2) `import type {...}` for local field references. The PR adds `PendingPermissionConfirm` to *both* blocks in *both* files — the local-use addition is needed because the field type on `ModelsAndPermissionsData` / `ConnectionState` references it directly. No import cycle introduced — store-core is a leaf dependency for both packages.

**SettingsBar.tsx change is justified.** `SettingsBarProps.pendingPermissionConfirm` was the third inline duplicate flagged in #3094's recommendation. The `import type` keeps it tree-shake-friendly. Sibling import order (`@chroxy/store-core` value import on line 4, type import on line 5) is fine.

**Scope:** PR touched only the 3 files declared in the description. `git diff main..HEAD --stat` confirms `+8 / -3` across exactly those files. No drive-by changes.

**No misses.** `grep` for `mode: string; warning: string` and `pendingPermissionConfirm` across `packages/**/*.{ts,tsx}` confirms every type-position duplicate has been replaced. The remaining inline literal in `packages/app/src/__tests__/store/message-handler.test.ts:2722` is a value cast (`as any`) for test fixtures, not a type — out of scope and fine to leave.

### Findings
- 0 blockers
- 0 nits
- 0 follow-ups

Closes #3094 cleanly. Ready to merge after your normal gate.

Issue #3088, opened 2026-04-27
fix(server): findRepoSkillsDir may match ~/.chroxy/skills as repo overlay when cwd is outside a repo


blamechris:
## Context

Found in review of #3087.

`findRepoSkillsDir(cwd)` walks up from `cwd` looking for the nearest `.chroxy/skills/` directory. The walk has no concept of \"this isn't actually a repo\" — it just stops at the filesystem root or after `REPO_DISCOVERY_MAX_DEPTH` iterations.

If a user starts a session in a directory under `\$HOME` that is **not** inside a git repo (e.g. `~/scratch/foo`), the walk continues all the way up and eventually hits `~`, where `~/.chroxy/skills/` exists — that's the global tier. `findRepoSkillsDir` happily returns it as `repoDir`.

`loadActiveSkillsLayered` does detect this via `_sameAbsolutePath(globalDir, repoDir)` and avoids double-counting, but the side effect is that **every global skill gets tagged `source: 'repo'`** for that session. The data is correct, but the source label silently lies.

## Repro

1. Have `~/.chroxy/skills/foo.md` (the normal global location)
2. `cd ~/scratch && chroxy start` (any dir under `\$HOME` with no `.git` or `.chroxy/skills/` of its own)
3. `list_skills` reports `foo.md` with `source: 'repo'` instead of `'global'`

## Original review finding

> **`findRepoSkillsDir` does not stop at the user's home directory.** A session started in `~/scratch/foo/bar` (no `.git`, no `.chroxy/skills`) walks all the way up through `~`, `~/..`, `/Users`, `/`. If the user happens to have a `~/.chroxy/skills/` (which they do, by design — it's the global tier), the walk would actually match the global dir as a \"repo\" overlay! That triggers the `globalDir === repoDir` branch and tags every global skill as `source: 'repo'` for that session.

## Suggested fix

Either of:

1. **Stop walking at `homedir()`** — repos generally don't live above `\$HOME`, and stopping there avoids the global-as-repo trap. Simple, low-risk.
2. **Skip a candidate that equals `DEFAULT_SKILLS_DIR`** — the global dir can never legitimately be a repo overlay, so explicitly excluding it makes the intent clear.
3. **Stop walking at the first ancestor containing a `.git`** — the most \"correct\" semantics (mirrors how git itself discovers repos), but adds an extra `statSync` per iteration.

(1) is probably the right call — minimal change, matches the mental model.

## Acceptance criteria

[...]

PR #3108, opened 2026-04-27
refactor: share auth + connection handlers via store-core

### LINKED ISSUE #3102: refactor(store-core): migrate auth + connection handlers (#2661 nibble)
## Parent
Sub-issue of #2661 — unify app/dashboard message handlers into `@chroxy/store-core`.

## Context
Following the validated `SessionPatch` seam from #3101 (and the pure-data-extraction pattern from earlier nibbles), this migrates the auth + connection cluster.

## Handlers in scope

| Handler | Dashboard | App | Shape |
|---------|-----------|-----|-------|
| `auth_ok` | ✓ | ✓ | Pure data extraction (multiple typed string/number fields) |
| `auth_fail` | ✓ | ✓ | Pure data extraction (reason string) |
| `key_exchange_ok` | ✓ | ✓ | Pure data extraction (sessionKey + sessionKeyId) |
| `server_mode` | ✓ | ✓ | Pure data extraction (mode enum) |

Both clients parse the same fields with near-identical logic; the side effects (state mutation, module-level flag updates, navigation) stay at the call site.

## Approach (per #3101 seam)

For each handler:
- Add a pure function in `packages/store-core/src/handlers/index.ts` that takes `Record<string, unknown>` and returns a typed payload.
- Validate field shapes with the existing `parseStringField` helper.
- For enum fields (`serverMode`, etc.), extract a small union-checking helper.
- Keep platform-specific side effects (e.g., \`_receivingHistoryReplay = false\`, lastConnectedUrl tracking, navigation) at the dashboard/app call site.

Tests: extend `packages/store-core/src/handlers/handlers.test.ts`.

## Out of scope
- E2E key exchange logic (consumers of \`sessionKey\` stay platform-specific)
- Any change to WS protocol shape
- Streaming/delta handlers (separate nibble)

## Acceptance criteria
- [ ] 4 new exports in store-core handlers module with tests
- [ ] Both clients import and use the shared handlers; payload-parsing logic deleted from each
- [ ] Side effects unchanged (existing client behaviour preserved)
- [ ] CI green (store-core tests + dashboard typecheck/tests + app tests)

blamechris:
## Summary

Migrates the four auth + connection WS handlers — `auth_ok`, `auth_fail`, `key_exchange_ok`, `server_mode` — out of the duplicated app + dashboard message-handler files and into `@chroxy/store-core`. **Fourth nibble** of the #2661 migration; **seventeenth** handler in store-core overall.

Refs #2661 (does NOT close — more nibbles remain). Closes #3102.

## What's in scope

| Handler | Shared payload | Side effects (stay at call site) |
|---------|----------------|----------------------------------|
| `auth_ok` | `serverMode`, `cwd`, `defaultCwd`, `serverVersion`, `latestVersion`, `serverCommit`, `protocolVersion` | replay-flag reset, `lastConnectedUrl`, heartbeat, key exchange kickoff, push registration, `useConnectionLifecycleStore.*` syncs, `clientId`/`connectedClients`/`webFeatures` parsing |
| `auth_fail` | `reason` (with `'Invalid token'` fallback) | socket close, phase transition, error banner, alert |
| `key_exchange_ok` | `publicKey` (validated string \| null) | `deriveSharedKey`, `_encryptionState`, post-auth WS sends |
| `server_mode` | `mode` (`'cli' \| 'terminal' \| null`) | `set({ serverMode })`, viewMode coercion, "Invalid Server Mode" alert |

## What's out of scope

- `clientId` / `connectedClients` parsing — requires the platform-side `ConnectedClient` type and is large enough to warrant its own nibble.
- `webFeatures` parsing — small but tightly-coupled to per-client default fallbacks.
- `encryption` flag and `sessionToken` (used only by app's pairing flow).
- Crypto state, push registration, biometric setup, navigation — none of those are pure data extraction.
- E2E key exchange logic and consumers of `sessionKey` (per the issue body).

## Design

Each handler is a pure `(msg) => typedPayload` function. Two new internal helpers were added to `handlers/index.ts`:

- `parseRawStringField` — `typeof v === 'string' ? v : null` (preserves empty strings, matches prior inline behaviour for `auth_ok.cwd` / `defaultCwd` / version fields).
- `parseEnumField<T>(msg, field, allowed)` — generic enum-narrowing helper used by both `auth_ok.serverMode` and `server_mode.mode`.

A new `ServerMode = 'cli' | 'terminal'` type and `AuthOkPayload` interface are exported. The dashboard accepts both modes; the app currently has no terminal view so it narrows `'terminal'` to `null` at the call site (preserving its prior `msg.serverMode === 'cli' ? 'cli' : null` behaviour).

```ts
// dashboard
case 'auth_ok': {
  _receivingHistoryReplay = false;
  lastConnectedUrl = ctx.url;
  const authPayload = sharedAuthOk(msg);
  // ... use authPayload.serverMode, authPayload.sessionCwd, etc.
}

// app — narrows 'terminal' to null
const authPayload = sharedAuthOk(msg);
const authServerMode: 'cli' | null = authPayload.serverMode === 'cli' ? 'cli' : null;
```

## Test plan

- [x] 17 new vitest cases in `packages/store-core/src/handlers/handlers.test.ts` covering: full extraction, individual field rejection (non-string, non-integer, sub-1 protocolVersion), enum narrowing, raw-string preservation of empty `cwd`, fallback-to-default for missing `auth_fail` reason, publicKey validation
- [x] `store-core` tests: **182 pass** (was 165)
- [x] `dashboard` tests: **1290 pass** (no regressions; existing auth-ok-handler.test.ts still green)
- [x] `app` tests: **1142 pass**

### DIFF
--- a/packages/app/src/store/message-handler.ts
+++ b/packages/app/src/store/message-handler.ts
@@ -44,6 +44,10 @@ import {
   handlePlanReady as sharedPlanReady,
   handleDevPreview as sharedDevPreview,
   handleDevPreviewStopped as sharedDevPreviewStopped,
+  handleAuthOk as sharedAuthOk,
+  handleAuthFail as sharedAuthFail,
+  handleKeyExchangeOk as sharedKeyExchangeOk,
+  handleServerMode as sharedServerMode,
 } from '@chroxy/store-core';
 import { PROTOCOL_VERSION } from '@chroxy/protocol';
 import { hapticSuccess } from '../utils/haptics';
@@ -789,20 +793,17 @@ export function handleMessage(raw: unknown, ctxOverride?: ConnectionContext): vo
       if (!ctx.isReconnect) hapticSuccess();
       // Track this URL as successfully connected
       lastConnectedUrl = ctx.url;
-      // Extract server context from auth_ok
-      const authServerMode: 'cli' | null =
-        msg.serverMode === 'cli' ? 'cli' : null;
-      const authSessionCwd = typeof msg.cwd === 'string' ? msg.cwd : null;
-      const authServerVersion = typeof msg.serverVersion === 'string' ? msg.serverVersion : null;
-      const authLatestVersion = typeof msg.latestVersion === 'string' ? msg.latestVersion : null;
-      const authServerCommit = typeof msg.serverCommit === 'string' ? msg.serverCommit : null;
-      const authProtocolVersion =
-        typeof msg.protocolVersion === 'number' &&
-        Number.isFinite(msg.protocolVersion) &&
-        Number.isInteger(msg.protocolVersion) &&
-        msg.protocolVersion >= 1
-          ? msg.protocolVersion
-          : null;
+      // Extract server context fields via shared handler (#3102).
+      // The shared handler accepts both 'cli' and 'terminal'; the app has no
+      // terminal view so we narrow 'terminal' to null (matches prior inline
+      // `msg.serverMode === 'cli' ? 'cli' : null` behaviour).
+      const authPayload = sharedAuthOk(msg);
+      const authServerMode: 'cli' | null = authPayload.serverMode === 'cli' ? 'cli' : null;
+      const authSessionCwd = authPayload.sessionCwd;
+      const authServerVersion = authPayload.serverVersion;
+      const authLatestVersion = authPayload.latestVersion;
+      const authServerCommit = authPayload.serverCommit;
+      const authProtocolVersion = authPayload.protocolVersion;
       // Parse connected clients list with self-detection via clientId
       const myClientId = typeof msg.clientId === 'string' ? msg.clientId : null;
       const rawClients = Array.isArray(msg.connectedClients) ? msg.connectedClients : [];
@@ -899,7 +900,8 @@ export function handleMessage(raw: unknown, ctxOverride?: ConnectionContext): vo
 
     case 'key_exchange_ok': {
       if (_ctx.pendingKeyPair) {
-        if (!msg.publicKey || typeof msg.publicKey !== 'string') {
+        const { publicKey: serverPublicKey } = sharedKeyExchangeOk(msg);
+        if (!serverPublicKey) {
           console.error('[crypto] Invalid publicKey in key_exchange_ok message', msg.publicKey);
           ctx.socket.close();
           set({ socket: null });
@@ -908,7 +910,7 @@ export function handleMessage(raw: unknown, ctxOverride?: ConnectionContext): vo
           _ctx.pendingSalt = null;
           break;
         }
-        const rawSharedKey = deriveSharedKey(msg.publicKey, _ctx.pendingKeyPair.secretKey);
+        const rawSharedKey = deriveSharedKey(serverPublicKey, _ctx.pendingKeyPair.secretKey);
         const encryptionKey = _ctx.pendingSalt
           ? deriveConnectionKey(rawSharedKey, _ctx.pendingSalt)
           : rawSharedKey;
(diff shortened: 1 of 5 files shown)


blamechris:
## Code Review: PR #3108

### Summary

Migrates the `auth_ok`, `auth_fail`, `key_exchange_ok`, and `server_mode` WS handlers into `@chroxy/store-core` (#3102, fourth nibble of #2661). Pure data-extraction seam; side effects (replay flags, heartbeat, key exchange kickoff, push registration, `useConnectionLifecycleStore` syncs, alert dispatch, view-mode coercion) all stay at the call site.

### Strengths

- Faithful to the established #3101 / #3093 / #3088 pattern — pure functions, typed payloads, no callbacks.
- Two new internal helpers (`parseRawStringField`, `parseEnumField`) keep both the inline simplification AND the future migrations consistent.
- Comprehensive test coverage: 17 new vitest cases (182 store-core tests total, was 165). Tests cover positive paths, all individual field rejections, raw-string preservation of empty `cwd`, integer/finite/range guards on `protocolVersion`, and enum narrowing.
- Behaviour preservation explicitly documented at platform divergence points: app's `'terminal'` -> `null` narrowing for `serverMode` is called out in both the call-site comment and the shared handler's docstring.
- Test mode used proper TDD (RED -> GREEN), confirmed by 17/17 new tests failing pre-impl, all 182 passing post-impl.
- Existing dashboard `auth-ok-handler.test.ts` (28 cases) continues to pass without modification — confirms behaviour preservation through the dashboard's most thorough integration test.

### Issues Found

#### Critical (Must Fix)

None.

| `packages/store-core/src/handlers/index.ts` | `handleAuthFail` | Subtle behaviour tightening: prior inline `(msg.reason as string) \|\| 'Invalid token'` would pass non-string truthy values (e.g., a stray number) through to the alert as `String(value)` after the `as string` cast at runtime. New handler treats non-string as missing and falls back to `'Invalid token'`. | This is a defensible tightening — the cast was always a lie — but worth pinning the behaviour test (already covered in `expect(handleAuthFail({ reason: 42 })).toEqual({ reason: 'Invalid token' })`). No fix needed; flagging for awareness. |
| `packages/dashboard/src/store/message-handler.ts` | `server_mode` case | `Ignoring invalid server_mode value: ${msg.mode}` — switched from `${mode}` (variable name shadow) to `${msg.mode}` (raw input). Both display the same value because the old code only fell into the else branch when `msg.mode` was neither `'cli'` nor `'terminal'`. | Defensible. The old code happened to use the local variable, but `msg.mode` is the original input the alert is about. |

#### Nitpicks (Optional)

- `parseEnumField` signature uses `readonly T[]` for `allowed`; could also accept `Set<T>` for O(1) lookup, but the lists are tiny (2 elements) and the ergonomics of `as const` arrays are better. Current shape is fine.
- The exported `AuthOkPayload` interface has `sessionCwd` (renamed from raw `cwd`) — the rename matches both clients' destination state shape (`sessionCwd: ...`). Worth confirming this stays consistent if a future nibble extracts the not-yet-shared `clientId` / `connectedClients` parsing.

None — out-of-scope items already enumerated in the issue body and PR description.

### Architecture Notes

This nibble validates the pattern for pure-payload extraction with diverging consumer narrowing. The `ServerMode` enum exposes both `'cli'` and `'terminal'` from store-core; the app narrows `'terminal'` -> `null` at the call site (since it has no terminal view), the dashboard accepts both. This is exactly the right shape: the shared handler is a *parser*, not a policy engine. Each consumer enforces its own platform constraints.

The `parseEnumField` helper is generic and reusable — future enum nibbles (`provider_event`'s typed enums, `permission_decision`'s mode set) can reuse it directly.

### Verdict

- [x] **Approve** - Ready to merge once CI is green.

Behaviour preservation is faithful, tests are thorough, and the seam is consistent with the four prior nibbles. No critical issues found.

Issue #3104, opened 2026-04-27
refactor(store-core): migrate multi-client coordination handlers (#2661 nibble)


blamechris:
## Parent
Sub-issue of #2661 — unify app/dashboard message handlers into `@chroxy/store-core`.

## Context
Following the SessionPatch seam from #3101.

## Handlers in scope

| Handler | Dashboard | App | Shape |
|---------|-----------|-----|-------|
| `client_joined` | ✓ | ✓ | Roster patch — append/upsert client entry |
| `client_left` | ✓ | ✓ | Roster patch — remove by clientId |
| `primary_changed` | ✓ | ✓ | Pure data extraction (new primary clientId) |
| `client_focus_changed` | ✓ | ✓ | Pure data extraction (clientId + focusedSessionId) |

All four are pretty mechanical. Roster handlers should return the upserted/filtered list rather than mutating in-place — the caller decides how to apply.

## Approach (per #3101 seam)

- `client_joined`/`client_left`: take the message + current roster, return the new roster list.
- `primary_changed`/`client_focus_changed`: parse + return typed payload only.
- Tests: extend `packages/store-core/src/handlers/handlers.test.ts`.

Note: any UI side effects (toast notifications when a client joins/leaves, focus indicator animations) stay at the call site — same pattern as `pushSessionNotification` in the `plan_ready` migration.

## Out of scope
- Outbound focus broadcasting (separate flow)
- Per-client UI rendering (consumes the roster but isn't part of message handling)

## Acceptance criteria
- [ ] 4 new exports in store-core handlers module with tests
- [ ] Both clients import and use the shared handlers; roster-list logic deleted from each
- [ ] Notifications/UX side effects unchanged
- [ ] CI green

PR #3111, opened 2026-04-27
refactor: share multi-client coordination handlers via store-core

blamechris:
## Summary

Migrates the four multi-client coordination handlers — `client_joined`, `client_left`, `primary_changed`, `client_focus_changed` — out of the app + dashboard message-handlers and into `@chroxy/store-core`. **Fourth nibble** following the SessionPatch seam validated in #3101.

## Why this batch

Per #3104, these four are mostly mechanical:

- `client_joined` / `client_left` — roster upsert/filter, identical wire-shape parsing across both clients
- `primary_changed` / `client_focus_changed` — pure data extraction with the same null-guards on both sides

The divergences are all **at the call site** (toast/system-message broadcast on the dashboard, side store sync via `useMultiClientStore` on the app, follow-mode auto-switch on both). Same shape #3101's `plan_ready` migration validated: store-core owns universal data; clients own UX surfaces.

## Design

| Handler | Returns | Caller responsibility |
|---|---|---|
| `handleClientJoined(msg, currentRoster)` | `{client, roster} \| null` | Apply roster, emit join toast/system message, sync side stores |
| `handleClientLeft(msg, currentRoster)` | `{clientId, departingClient, roster} \| null` | Apply roster, emit leave toast, sync side stores |
| `handlePrimaryChanged(msg)` | `{sessionId, primaryClientId}` | Decide global-vs-per-session apply (`sessionId === null \|\| 'default'` → global) |
| `handleClientFocusChanged(msg)` | `{clientId, sessionId} \| null` | Follow-mode auto-switch using the consumer's `myClientId`/`followMode` |

The roster handlers return the **upserted/filtered list** rather than mutating in place — caller decides where to store it. The dashboard now reads `connectedClients` from its `ConnectionState`; the app additionally pipes the parsed client through `useMultiClientStore.addClient`/`removeClient` (its existing side store behaviour).

`handlePrimaryChanged` intentionally does NOT consult an active session id — the message always carries the target sessionId or omits it deliberately, so there's no `resolveSessionId` fallback. The literal `'default'` is preserved verbatim and the call site keeps its existing `sessionId === 'default'` global-apply branch.

## Test plan

- [x] 23 new vitest cases in `packages/store-core/src/handlers/handlers.test.ts` covering: upsert, all four `deviceType` values, default-coercion of `deviceName`/`platform`/`isSelf`, missing/non-string-clientId guards (joined+left+focus), missing-departingClient case, `'default'` sessionId pass-through, mixed missing+non-string permutations
- [x] `store-core` tests: 213 pass (was 190)
- [x] `dashboard` tests: 1290 pass — including the 4 existing `system message routing` cases that exercise the global broadcast on `client_joined`/`client_left`
- [x] `app` tests: 1142 pass

Refs #2661 (fourth nibble; does NOT close — more handlers remain).

### DIFF
--- a/packages/app/src/store/message-handler.ts
+++ b/packages/app/src/store/message-handler.ts
@@ -53,6 +53,10 @@ import {
   handleCheckpointRestored as sharedCheckpointRestored,
   handleError as sharedError,
   handleSessionError as sharedSessionError,
+  handleClientJoined as sharedClientJoined,
+  handleClientLeft as sharedClientLeft,
+  handlePrimaryChanged as sharedPrimaryChanged,
+  handleClientFocusChanged as sharedClientFocusChanged,
 } from '@chroxy/store-core';
 import { PROTOCOL_VERSION } from '@chroxy/protocol';
 import { hapticSuccess } from '../utils/haptics';
@@ -1961,20 +1965,11 @@ export function handleMessage(raw: unknown, ctxOverride?: ConnectionContext): vo
     // --- Multi-client awareness ---
 
     case 'client_joined': {
-      if (!msg.client || typeof (msg.client as Record<string, unknown>).clientId !== 'string') break;
-      const client = msg.client as Record<string, unknown>;
-      const newClient: ConnectedClient = {
-        clientId: client.clientId as string,
-        deviceName: typeof client.deviceName === 'string' ? client.deviceName : null,
-        deviceType: (['phone', 'tablet', 'desktop', 'unknown'].includes(client.deviceType as string) ? client.deviceType : 'unknown') as ConnectedClient['deviceType'],
-        platform: typeof client.platform === 'string' ? client.platform : 'unknown',
-        isSelf: false,
-      };
-      useMultiClientStore.getState().addClient(newClient);
-      set((state: ConnectionState) => ({
-        connectedClients: [...state.connectedClients.filter((c) => c.clientId !== newClient.clientId), newClient],
-      }));
-      const deviceLabel = newClient.deviceName || 'A device';
+      const joined = sharedClientJoined(msg, get().connectedClients);
+      if (!joined) break;
+      useMultiClientStore.getState().addClient(joined.client);
+      set({ connectedClients: joined.roster });
+      const deviceLabel = joined.client.deviceName || 'A device';
       const joinMsg: ChatMessage = {
         id: nextMessageId('client'),
         type: 'system',
@@ -1993,12 +1988,13 @@ export function handleMessage(raw: unknown, ctxOverride?: ConnectionContext): vo
     }
 
     case 'client_left': {
-      if (typeof msg.clientId !== 'string') break;
-      const departingClient = useMultiClientStore.getState().removeClient(msg.clientId as string);
-      set((state: ConnectionState) => ({
-        connectedClients: state.connectedClients.filter((c) => c.clientId !== msg.clientId),
-      }));
-      const leftLabel = departingClient?.deviceName || 'A device';
+      const left = sharedClientLeft(msg, get().connectedClients);
+      if (!left) break;
+      // Keep the multi-client side store in sync (its removeClient also returns
+      // the departing entry, but we trust the shared handler's lookup).
+      useMultiClientStore.getState().removeClient(left.clientId);
+      set({ connectedClients: left.roster });
+      const leftLabel = left.departingClient?.deviceName || 'A device';
       const leftMsg: ChatMessage = {
         id: nextMessageId('client'),
         type: 'system',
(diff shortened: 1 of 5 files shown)


blamechris:
## Code Review: PR #3111

### Summary

Fourth nibble in the #2661 store-core migration: `client_joined` / `client_left` / `primary_changed` / `client_focus_changed` handlers extracted into `@chroxy/store-core`. Roster handlers return upserted/filtered lists for caller-driven application, payload parsers return typed data only. Both clients refactored to use the shared helpers; UX side effects (system-message broadcast, multi-client side store sync) remain at call sites. 23 new vitest cases push store-core to 213 tests.

### Strengths

- **Seam discipline.** Roster handlers return `{ client, roster }` / `{ clientId, departingClient, roster }` rather than mutating in place — caller decides where to store. Same pattern that worked in #3101's `plan_ready`.
- **Behaviour preservation.** `handlePrimaryChanged` keeps the literal `'default'` sessionId pass-through, and the caller retains its `sessionId === 'default'` global-apply branch. `handleClientLeft` still surfaces `departingClient` so device-name labels in the leave toast keep working. `handleClientJoined` always sets `isSelf: false` (test pinned this) — matches both clients' prior inline `isSelf: false` literal.
- **Test coverage.** 23 cases exercise: upsert, all four `deviceType` values, default-coercion of `deviceName` / `platform` / `isSelf`, missing/non-string-clientId guards, missing-departingClient case, `'default'` sessionId pass-through, and the started/ready opposite-sign sanity precedent. The `accepts all valid deviceType values` loop covering `phone`/`tablet`/`desktop`/`unknown` is a nice touch.
- **JSDoc quality.** Each handler explains *why* (caller-decides routing, behaviour-preserving pass-through, where the platform-specific UX lives). Future migrators will read this and know the seam.
- **Negligible diff to call sites.** Each case statement shrinks from ~10 lines to 4-5 while preserving every observable side effect (toast broadcast on dashboard, `useMultiClientStore` sync on app, follow-mode auto-switch on both).

### Issues Found

#### Critical (Must Fix)

None.

| `packages/store-core/src/handlers/index.ts` | ~344 | `if (!rawClient || typeof rawClient !== 'object')` admits arrays as valid `client` objects (since `typeof [] === 'object'`). The prior inline check accessed `.clientId` directly which would also accept arrays (any indexed access works), so this is **behaviour-preserving** — but worth a comment that array-as-client is intentionally accepted (or, better, `Array.isArray(rawClient)` rejection if we want to tighten). Out of scope for this nibble per the "don't tighten validation" guardrail; flagging only. | Not a regression — both clients accepted this before. |
| `packages/store-core/src/handlers/index.ts` | ~360 | The roster array always allocates a new array, even when `currentRoster` is empty AND no upsert happens. (It can't — we always push the new client.) Skip the comment, this is fine. | Withdrawn after re-read. |
| `packages/store-core/src/handlers/handlers.test.ts` | ~437 | Consider adding an immutability assertion: `expect(roster).not.toBe(currentRoster)` or `expect(currentRoster).toHaveLength(originalLen)`. The handler currently produces a new array via spread, but pinning this in a test prevents a future refactor from accidentally mutating the input roster (which both Zustand stores treat as immutable). | Defensive — Zustand mutability bugs are subtle. |

#### Nitpicks (Optional)

- `VALID_DEVICE_TYPES.has(c.deviceType as ConnectedClient['deviceType'])`: the `as` cast is needed because TS's `Set#has` is invariant on `T`. An alternative is `(VALID_DEVICE_TYPES as ReadonlySet<unknown>).has(c.deviceType)` which avoids the lie but is uglier. Current form is fine — the prior inline code used the same pattern.
- `handleClientLeft` returns `{ ..., roster }` even when `clientId` doesn't match an existing entry — the caller will `set({ connectedClients: left.roster })` and produce a new array reference for nothing. Same as the prior inline `state.connectedClients.filter(...)`. Not a regression.
- The two roster handlers have a slight asymmetry: `handleClientJoined` returns `null` only when the message is malformed; `handleClientLeft` does the same. Consistent — good.

| Add immutability assertions to roster-handler tests | _none — applies as nitpick only_ | Trivially fixable in a follow-up; not worth blocking this nibble. |

No follow-up issues created — the suggestions above are either nitpicks or out-of-scope tightening per the migration guardrail ("don't tighten validation").

### Architecture Notes

This is the **fourth handler nibble** following:
- #3093 — `confirm_permission_mode`
- #3097 — `plan_started`
- #3101 — `plan_ready` (validated the SessionPatch seam for handlers with platform-specific UX)
- **#3111 (this PR)** — multi-client coordination

The pattern is now well-established: store-core owns universal data extraction + roster math; clients own the surfaces that don't generalize (toast notifications, side stores like `useMultiClientStore`, follow-mode auto-switch using consumer state). Each migration peels off ~50 net LOC of duplication. The remaining big targets per #2661 (stream/delta cases) will be more involved but follow the same shape.

The roster-as-return-value design (vs. mutating a shared store) is the right call. It keeps store-core stateless — same as every other handler — and avoids cross-coupling between the shared package and the platform-specific Zustand stores.

### Verdict

Behaviour preserved, tests pinned, both clients consume cleanly, all three packages typecheck clean. Standard nibble shape.

blamechris:
## Review Comments Addressed

**Total:** 0 inline comments to process

Copilot review (`COMMENTED`) generated no inline comments. 2 low-confidence comments were suppressed by Copilot (flagged truthy-vs-explicit-null style on `primarySessionId`); both are behaviour-preserving — verified that `sessionId === ""` already routed to the global-apply branch under the prior `typeof` guard (`sessionStates[""]` is undefined → first branch falls through → `!""` is true → global). No fix needed.

Agent-review verdict: **Approve** (0 critical, suggestions/nitpicks only).

PR #3129, opened 2026-04-28
refactor(store-core): migrate file operation handlers (#3124)

### LINKED ISSUE #3124: refactor(store-core): migrate file operation handlers (#2661)
**Parent:** #2661

Migrate file-operation message handlers (callback-style) into `@chroxy/store-core/handlers`. These cases all extract a normalized payload then forward to a platform callback — the payload normalization is the duplication to extract.

## Cases to migrate

- `directory_listing` — `{path, parentPath, entries: DirectoryEntry[], error}`
- `file_listing` — `{path, parentPath, entries: FileEntry[], error}`
- `file_content` — `{path, content, language, size, truncated, error}`
- `write_file_result` (app-only currently) — `{path, error}`

## Approach

1. Add to `packages/store-core/src/handlers/index.ts`:
   - `handleDirectoryListing(msg)` → `DirectoryListingPayload`
   - `handleFileListing(msg)` → `FileListingPayload`
   - `handleFileContent(msg)` → `FileContentPayload`
   - `handleWriteFileResult(msg)` → `WriteFileResultPayload`

2. Define payload types in `packages/store-core/src/handlers/types.ts` (or co-locate as the file ops module sees fit). Keep `DirectoryEntry`, `FileEntry` as generic placeholders (`unknown[]` is fine — types live downstream).

3. Update `packages/dashboard/src/store/message-handler.ts` and `packages/app/src/store/message-handler.ts` to import the shared extractors and pass the result to their respective callback (`get()._fooCallback` vs `getCallback('foo')`).

4. Add tests in `packages/store-core/src/handlers/handlers.test.ts` covering: valid payload extraction, missing fields → null, wrong-type fields → defaults, `truncated` boolean coercion (only `=== true`).

## Acceptance criteria

- [ ] All 4 handlers exported from `@chroxy/store-core`
- [ ] Dashboard + app message-handlers use the shared extractors
- [ ] Tests cover happy path + missing/invalid field defaults
- [ ] No behaviour change (callback signatures unchanged)
- [ ] Server tests + lint + dashboard tests + app type-check all pass

blamechris:
## Summary

Migrates file-operation message handlers (callback-style) into `@chroxy/store-core/handlers`, continuing the #2661 unification work.

- `handleDirectoryListing(msg)` -> `{path, parentPath, entries: unknown[], error}`
- `handleFileListing(msg)` -> `{path, parentPath, entries: unknown[], error}`
- `handleFileContent(msg)` -> `{path, content, language, size, truncated, error}` (strict `truncated === true`)
- `handleWriteFileResult(msg)` -> `{path, error}` (app-only today; extracted now so dashboard can adopt later)

Both `packages/dashboard/src/store/message-handler.ts` and `packages/app/src/store/message-handler.ts` now call the shared extractors and pass the result to their respective callback. Concrete entry types (`DirectoryEntry`, `FileEntry`) live downstream and are applied via cast at the call site — `entries` stays `unknown[]` in the shared payload.

Behaviour-preserving: same callback signatures, same null/[] defaults.

## Tests

26 new tests in `packages/store-core/src/handlers/handlers.test.ts`:

- valid payload extraction
- missing fields default to null / `[]`
- non-string/non-number fields coerce to null / `[]`
- `truncated` returns `true` only for strict `=== true` (truthy strings/numbers do NOT count)
- empty-string passthrough preserved (matches inline `typeof === 'string'` guards)
- per-element shape NOT validated (matches the prior inline behaviour both clients used)

## Test plan

- [x] `cd packages/store-core && npm test` -> 362 passed (26 new)
- [x] `cd packages/dashboard && npm run typecheck && npm test` -> clean / 1290 passed
- [x] `cd packages/app && npx tsc --noEmit` -> clean
- [x] Server tests: pre-existing flakes only (encryption integration in parallel, service status timeout, feature detection) — unrelated to this diff

Closes #3124

### DIFF
--- a/packages/app/src/store/message-handler.ts
+++ b/packages/app/src/store/message-handler.ts
@@ -66,6 +66,10 @@ import {
   handlePermissionExpired as sharedPermissionExpired,
   handlePermissionTimeout as sharedPermissionTimeout,
   handlePermissionRulesUpdated as sharedPermissionRulesUpdated,
+  handleDirectoryListing as sharedDirectoryListing,
+  handleFileListing as sharedFileListing,
+  handleFileContent as sharedFileContent,
+  handleWriteFileResult as sharedWriteFileResult,
   handleSessionList as sharedSessionList,
   handleSessionContext as sharedSessionContext,
   handleSessionTimeout as sharedSessionTimeout,
@@ -2070,12 +2074,8 @@ export function handleMessage(raw: unknown, ctxOverride?: ConnectionContext): vo
     case 'directory_listing': {
       const cb = getCallback('directoryListing');
       if (cb) {
-        cb({
-          path: typeof msg.path === 'string' ? msg.path : null,
-          parentPath: typeof msg.parentPath === 'string' ? msg.parentPath : null,
-          entries: Array.isArray(msg.entries) ? msg.entries as DirectoryEntry[] : [],
-          error: typeof msg.error === 'string' ? msg.error : null,
-        });
+        const payload = sharedDirectoryListing(msg);
+        cb({ ...payload, entries: payload.entries as DirectoryEntry[] });
       }
       break;
     }
@@ -2083,12 +2083,8 @@ export function handleMessage(raw: unknown, ctxOverride?: ConnectionContext): vo
     case 'file_listing': {
       const fileBrowserCb = getCallback('fileBrowser');
       if (fileBrowserCb) {
-        fileBrowserCb({
-          path: typeof msg.path === 'string' ? msg.path : null,
-          parentPath: typeof msg.parentPath === 'string' ? msg.parentPath : null,
-          entries: Array.isArray(msg.entries) ? msg.entries as FileEntry[] : [],
-          error: typeof msg.error === 'string' ? msg.error : null,
-        });
+        const payload = sharedFileListing(msg);
+        fileBrowserCb({ ...payload, entries: payload.entries as FileEntry[] });
       }
       break;
     }
@@ -2096,14 +2092,7 @@ export function handleMessage(raw: unknown, ctxOverride?: ConnectionContext): vo
     case 'file_content': {
       const fileContentCb = getCallback('fileContent');
       if (fileContentCb) {
-        fileContentCb({
-          path: typeof msg.path === 'string' ? msg.path : null,
-          content: typeof msg.content === 'string' ? msg.content : null,
-          language: typeof msg.language === 'string' ? msg.language : null,
-          size: typeof msg.size === 'number' ? msg.size : null,
-          truncated: msg.truncated === true,
-          error: typeof msg.error === 'string' ? msg.error : null,
-        });
+        fileContentCb(sharedFileContent(msg));
       }
       break;
     }
@@ -2111,10 +2100,7 @@ export function handleMessage(raw: unknown, ctxOverride?: ConnectionContext): vo
     case 'write_file_result': {
       const fileWriteCb = getCallback('fileWrite');
       if (fileWriteCb) {
-        fileWriteCb({
-          path: typeof msg.path === 'string' ? msg.path : null,
-          error: typeof msg.error === 'string' ? msg.error : null,
-        });
+        fileWriteCb(sharedWriteFileResult(msg));
       }
       break;
     }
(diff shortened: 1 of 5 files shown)


blamechris:
## Code Review: PR #3129

### Summary

Migrates four file-operation message handlers (`directory_listing`, `file_listing`, `file_content`, `write_file_result`) from inline parsing in `packages/{app,dashboard}/src/store/message-handler.ts` into shared extractors in `@chroxy/store-core/handlers`. Continues the #2661 unification work.

The shared extractors normalize incoming WS payloads (string-typed strings, array-typed arrays, strict `truncated === true`) and return platform-agnostic shapes; concrete entry types (`DirectoryEntry` / `FileEntry`) stay downstream and are applied via cast at the call site. Behavior-preserving on both clients.

### Strengths

- **Behaviour-preservation is verified.** Tests explicitly assert the inline guards' quirks: empty-string passes through, `truncated: 'true'`/`1`/`{}` resolve to `false`, zero size passes through, non-string fields default to `null`. This is the right level of rigour for a refactor whose stated goal is no behaviour change.
- **Pattern alignment.** The new handlers match the surrounding code (`handleSessionList`, `handlePermissionRulesUpdated`) — same `Record<string, unknown>` input, same JSDoc note about deferred per-element validation, same explicit `as unknown[]` cast inside the `isArray` branch.
- **Comment header is excellent.** The block at line 1490-1500 of `index.ts` explains the design split (payload normalisation in core, callback dispatch at call site, concrete entry types downstream) and pre-empts the obvious "why isn't entry shape validated?" question.
- **Forward-compatible.** `handleWriteFileResult` is extracted now even though the dashboard doesn't yet use it — saves a follow-up migration when dashboard gains write support.
- **Test count and structure.** 26 tests is generous coverage; per-handler `describe` blocks and explicit edge-case naming (`returns truncated=false for truthy non-boolean values`) make failures easy to triage.

### Issues Found

#### Critical (Must Fix)

None.

| `packages/store-core/src/handlers/index.ts` | 1503-1513, 1486-1495 | `DirectoryListingPayload` and `FileListingPayload` are structurally identical. Consider deduping with a shared `EntriesPayload` interface (or a single `handleEntriesListing` factory). | Pure structural duplication — same four fields, identical normalisation logic. The only divergence is the JSDoc comment explaining their downstream split. Not blocking — explicit named types are arguably clearer at the call site. |

#### Nitpicks (Optional)

- `handleDirectoryListing` and `handleFileListing` have byte-identical bodies. If you keep the two distinct types (per the suggestion above), a one-line comment cross-referencing the sister function would make the duplication intentional rather than accidental.
- The `as unknown[]` cast inside `Array.isArray(...) ? (msg.entries as unknown[]) : []` is technically redundant — `Array.isArray` narrows to `unknown[]` already. Kept for stylistic consistency with the rest of the file (e.g. `handlePermissionRulesUpdated` uses the same form), so leave as-is.

| Dedup `DirectoryListingPayload` / `FileListingPayload` into a shared shape | https://github.com/blamechris/chroxy/issues/3131 | Pure cosmetic dedup. The named types document intent at downstream call sites; collapsing them is reversible later if a third entries-style payload appears. |

### Architecture Notes

This continues the established #2661 pattern: the `handleX(msg) -> Payload` shape stays simple and pure, callbacks remain at the call site, downstream concrete entry types stay where they're consumed. Keeping `entries: unknown[]` in core is the right call — pulling `DirectoryEntry`/`FileEntry` into store-core would force the dashboard's UI-shaped types into a low-level shared package.

The thin spread-and-cast at call sites (`cb({ ...payload, entries: payload.entries as DirectoryEntry[] })`) is acceptable for now but worth tracking: once 3-4 more callback-style handlers migrate this way, a `forwardEntries(payload, callback)` helper might be warranted to avoid the manual spread/cast on every call site.

### Verdict

Behaviour-preserving, well-tested, follows existing patterns. The single suggestion is cosmetic and tracked as a follow-up.

blamechris:
## Review Comments Addressed

**Total:** 0 inline comments to process.

Copilot completed its review (state: COMMENTED) and reviewed all 5 changed files but generated no inline comments. The agent-review pass posted one optional follow-up suggestion (cosmetic dedup of identical payload interfaces), tracked at #3131.

Issue #3163, opened 2026-05-03
app: stop button missing after returning to a busy session — even with the v0.6.13 defensive fallback


blamechris:
## Summary

Two related app symptoms surfaced while dogfooding v0.6.14:

1. **Stop button missing on tab return.** Send a prompt, navigate away (session/tab switch), come back while Claude is still working — the input bar shows the send button instead of the stop button.
2. **Assistant text concatenated into tool_use bubbles.** Same shape as #3071 — but the v0.6.13 defensive fallback in \`packages/app/src/store/message-handler.ts:1348-1374\` is supposed to prevent exactly this. So the symptom is recurring through a path the fallback doesn't cover.

These were observed on the same session: the user sent a prompt that involved several Bash tool calls and a final assistant response. The mobile app showed the tool bubbles but never displayed the response text. The response was visible neither on the mobile app nor the Tauri desktop dashboard (a separate dashboard fix is in #3161).

## Repro

1. Open the app, start a session with a Claude provider.
2. Send a prompt that produces multiple sequential tool calls before the final response (e.g. \"What's in this repo? Run a few greps.\").
3. While Claude is working, switch session tabs (or background and foreground the app).
4. Return to the original session — observe:
   - Input bar shows send (not stop), even though Claude is still streaming.
   - When the response finally completes, the assistant text is missing OR appears partially fused into the last tool_use bubble.

### Stop button bug
- Selectors look correct (\`selectStreamingMessageId\` reads \`activeSession(s)?.streamingMessageId\` per-session).
- \`switchSession\` at \`packages/app/src/store/connection.ts:1227\` only sets \`activeSessionId\` + \`sessionNotifications\` — does NOT clear streaming state. So a return-switch should restore.
- Possible causes:
  - Persistence rehydration overwriting \`streamingMessageId\` on remount?
  - \`useShallow\` re-subscription dropping the live update during the switch?
  - Lifecycle hook in SessionScreen clearing streaming state on focus/blur?

### Recurring text concatenation
- The defensive fallback at lines 1348-1374 only triggers when the existing message at \`deltaId\` is non-response AND no remap is registered.
- Possible variants the fallback misses:
  - Cross-session deltas (\`capturedSessionId\` ≠ \`activeSessionId\`, session not in \`sessionStates\`).
  - Deltas arriving AFTER \`stream_end\` cleared the remap (\`_ctx.deltaIdRemaps.delete(out.messageId)\` at the stream_end call site).
[...]

Issue #2611, opened 2026-03-21
bug(desktop): chat tab only shows TodoWrite — missing assistant text response


blamechris:
## Description

In the Tauri desktop app, when an assistant response contains both tool calls (e.g. TodoWrite) and text output, the chat tab sometimes only renders the tool call — the actual text response is missing. The output tab shows everything correctly.

**This is intermittent** — the same type of response (tool calls + text) renders correctly sometimes and fails other times, making reproduction difficult.

## Steps to Reproduce

1. Ask the agent to perform a task that involves tool calls (e.g. TodoWrite, Bash) + a text summary
2. Observe the chat tab in the desktop dashboard
3. Intermittently, only the tool call is shown — the assistant's text response is not rendered

## Expected Behavior

The chat tab should always show the assistant's text response, not just the tool call.

## Actual Behavior

Chat tab intermittently shows only the tool call (e.g. "TodoWrite") — the text portion of the response is absent. Output tab renders everything correctly. The mobile app also rendered the same response correctly.

## Investigation Notes

- Intermittent — identical response shapes sometimes render correctly, sometimes don't
- May be a race condition in message processing or rendering
- Could be related to the `stream_start` ID collision pattern (see memory on `_deltaIdRemaps`) — if the text block arrives before/after certain tool events, the ID mapping may fail
- Could also be a timing issue with how the dashboard processes batched WebSocket messages
- The output tab always works, suggesting the data arrives correctly but the chat-tab filter/renderer drops it under certain conditions
- **Hard to verify fixes** due to intermittent nature — may need added logging or a stress test to reproduce reliably

PR #3167, opened 2026-05-03
fix(app,dashboard): never apply stream deltas to non-response messages

blamechris:
## Summary

Adds a type guard to `flushPendingDeltas` in both clients (dashboard + app) so a stream delta can only land on a message of type `'response'`. Belt-and-suspenders against any future server regression or race that lets an ID collision slip past the existing defensive remap in `handleStreamDelta`.

## Why

After PR #3165 (server emits unique `tool_start` ids) and PR #3161 (dashboard defensive remap parity with the app), the actual root cause is fixed and ID collisions don't reach `flushPendingDeltas` in normal flow. This adds a final safety net that's strictly better than the prior behaviour: if a colliding delta ever does reach the flush, it will not poison a `tool_use` bubble's `content` with response text.

## Trade-off

The dashboard's existing `#2611` orphan-create safety net is preserved: when no message exists at the delta id, a response is still created. The new behaviour only differs when a non-response message (e.g. `tool_use`) exists at the colliding id:

- **Before:** delta concatenated onto the tool_use's `content` — visible corruption.
- **After:** delta routes to orphan path which creates a response at the same id; ChatView dedup makes that response invisible. Data loss, but **not** corruption — strictly better than poisoning a tool_use bubble.

In practice this scenario is unreachable now that #3165 is in. The guard is purely defense against future server regressions.

## Test plan

- [x] Dashboard suite — 1293/1293 pass, including a new regression test that simulates the race (delta queued before the colliding tool_use is added to state) and asserts the tool_use bubble stays pristine through flush
- [x] App suite (Jest) — 111/111 in `packages/app/src/__tests__/store/message-handler.test.ts`, including a parallel regression test
- [x] Dashboard `tsc --noEmit` clean
- [x] App `tsc --noEmit` clean

## Related

- Server fix: #3165 (now merged) — eliminates the underlying collision
- Defensive remap port: #3161 (merged) — dashboard parity with the app's `handleStreamDelta` collision detection
- Issue #3163 — separate stop-button regression on session switch (unrelated)

### DIFF
--- a/packages/app/src/__tests__/store/message-handler.test.ts
+++ b/packages/app/src/__tests__/store/message-handler.test.ts
@@ -1749,6 +1749,13 @@ describe('stream_delta handler', () => {
     const toolUse = ss.messages.find((m) => m.id === 'msg-race' && m.type === 'tool_use');
     // tool_use bubble must remain pristine — no delta concatenation
     expect(toolUse?.content).toBe('ls');
+    // App's flushPendingDeltas has no orphan-create safety net (the dashboard's
+    // #2611 fix is dashboard-only), so the colliding delta is silently dropped.
+    // No response message is created at the colliding id. This is more
+    // aggressive data loss than the dashboard but still strictly better than
+    // corrupting the tool_use bubble.
+    const responseAtSameId = ss.messages.find((m) => m.id === 'msg-race' && m.type === 'response');
+    expect(responseAtSameId).toBeUndefined();
   });
 });
 
--- a/packages/app/src/__tests__/store/message-handler.test.ts
+++ b/packages/app/src/__tests__/store/message-handler.test.ts
@@ -1713,6 +1713,43 @@ describe('stream_delta handler', () => {
     const toolUseMsg = ss.messages.find((m) => m.id === 'msg-1');
     expect(toolUseMsg?.content).toBe('ls');
   });
+
+  // Belt-and-suspenders: even if a stream_delta sneaks past the defensive
+  // remap in handleStreamDelta (e.g. the colliding tool_use is added to state
+  // AFTER the delta is queued in pendingDeltas), flushPendingDeltas itself
+  // must never apply delta text onto a non-response message.
+  it('flushPendingDeltas type-filter prevents tool_use corruption when collision slips past defensive remap', () => {
+    const store = createMockStore({
+      activeSessionId: 's1',
+      sessions: [{ sessionId: 's1', name: 'S1' } as any],
+      sessionStates: { s1: { ...createEmptySessionState(), messages: [] } },
+    });
+    setStore(store as any);
+    _testMessageHandler.setContext(createMockContext() as any);
+
+    // Step 1: dispatch delta when no message exists at this id — defensive
+    // remap can't catch the collision since the tool_use isn't there yet.
+    _testMessageHandler.handle({ type: 'stream_delta', messageId: 'msg-race', sessionId: 's1', delta: 'must not leak' });
+
+    // Step 2: race condition — tool_use is added AFTER the delta is queued
+    // but BEFORE the 100ms batcher flushes.
+    store.setState((s: any) => ({
+      sessionStates: {
+        ...s.sessionStates,
+        s1: {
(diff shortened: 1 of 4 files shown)


blamechris:
## Code Review: PR #3167

### Summary
Adds a `m.type === 'response'` type guard to `flushPendingDeltas` in both dashboard and app, plus parallel regression tests. This is the third defensive layer on top of merged PRs #3165 (server-side unique tool_start ids) and #3161 (dashboard `handleStreamDelta` defensive remap). The change is small, CI is green, and the design choice to drop colliding deltas (vs corrupt a tool_use bubble) is sound.

### Strengths
- Single change applied symmetrically to both branches (sessionStates + flat) in both clients — internally consistent.
- Trade-off is verified: dashboard `ChatView.tsx` dedups by id keeping FIRST occurrence, so the orphan-create path pushes a response at the colliding id that is then invisible — no corruption.
- Type guard is correctly scoped: `response` is the only `ChatMessage.type` that legitimately accepts delta concatenation. `thinking` is filtered at stream_start; everything else is mutated via dedicated handlers.
- Comment explicitly frames this as defense, not a current bug fix — honest about being belt-and-suspenders.

### Issues Found

#### Critical (Must Fix)
None.

| # | Issue | Status |
|---|-------|--------|
| 1 | Tests assert ONLY the negative (`tool_use.content === 'ls'`); they don't lock in what happens to the delta. Without positive assertions, the suite would still pass if the whole delta pipeline regressed to a no-op. | **FIX** — `9e26c16b27` adds positive assertions: dashboard test asserts the orphan response was pushed at the colliding id with the delta content; app test asserts no response message was created (silent drop, since the app has no orphan-create) and adds a comment explaining the divergence. |
| 2 | App `flushPendingDeltas` has no orphan-create safety net (#2611 was dashboard-only). After this PR, a colliding delta in the app is fully dropped, slightly worse degradation than the dashboard. | **FOLLOW-UP** — [#3168](https://github.com/blamechris/chroxy/issues/3168) (port #2611 to the app for symmetry). |
| 3 | Comment "Defense against future server regressions or races" — given PR #3165 just removed the underlying collision, the "races" language overstates: the test scenario requires the server to send out-of-order ids, which is a regression rather than a natural race. | **FIX** — `9e26c16b27` tightens the comment to "future server regressions that reintroduce id collisions" in both files. |

#### Nitpicks (Optional)
- The `if (d && m.type === 'response')` pattern is duplicated 4 times across two files. Could be extracted as a one-liner helper, but the duplication matches the surrounding style — not worth changing.

### Architecture Notes
This is the third defensive layer on a single failure mode (id collision causing delta-onto-tool_use corruption): server fix (#3165 unique ids), client remap (#3161 `handleStreamDelta` lazy-creates suffixed response on collision), and now this guard in `flushPendingDeltas`. The failure mode is high-impact (silent visible corruption of tool bubbles), the guard is one boolean check per message in a hot path that's already iterating, and the dashboard's orphan-create path means delta data is preserved (just invisible). Defense-in-depth is reasonable here: a single character locks down the invariant that delta content concatenates only onto response messages, regardless of any future ordering bug.

### Verdict
Approve

blamechris:
## Review Comments Addressed

Copilot's inline review produced 0 formal comments but the review body contained two **low-confidence-suppressed** observations both pointing at the same architectural concern: my type-guard fix routed colliding deltas to the existing orphan-create path, which pushed a response with the SAME id as the existing tool_use — introducing duplicate ids in the messages array. The agent-review missed this; the concern is real and addressed.

| # | Concern | Outcome | Reference |
| 1 | (Copilot, suppressed, sessionStates branch) Orphan-create produces duplicate ids when delta collides with non-response message | FIX | `3e16066314` |
| 2 | (Copilot, suppressed, flat-messages branch) Same duplicate-id risk | FIX | `3e16066314` |

**Total:** 2 concerns addressed
- Fixed: 2 (both in `3e16066314` — orphan path now suffixes to `${msgId}-response` on collision and registers a remap, mirroring `handleStreamDelta`'s defensive remap pattern)
- False positives: 0
- Follow-up issues created: 0

PR #3174, opened 2026-05-03
fix(app): adopt dashboard's #2611 orphan-create safety net for stream_delta (#3168)

blamechris:
Closes #3168

## Summary

The dashboard's `flushPendingDeltas` (added in #2611) creates a response message for orphaned deltas — protecting against dropped `stream_start` frames. The app handler had no such fallback: a delta with no matching message was silently dropped. After #3167's type-guard, a colliding delta in the app was fully lost with no trace, while the dashboard preserved the data as an orphan response.

This PR brings the app to parity.

## Changes

- Port orphan-create logic into **both** branches of app's `flushPendingDeltas` — the `sessionStates` branch and the active-session fallback. Each loops over deltas not matched by the type-filtered map and either appends to an existing target message or creates a new \`response\` at the wire id (suffixed with \`-response\` if a non-response message already occupies that id).
- Register the suffixed remap in \`_ctx.deltaIdRemaps\` on collision so subsequent deltas route correctly.
- Update the existing #3167 regression test to assert the orphan is now created (matches dashboard behaviour).
- Add new test mirroring dashboard's #2611 test: dispatch two deltas with no preceding \`stream_start\`, assert a \`response\` message is created at the wire id with concatenated content.

## Test plan
- [x] \`packages/app\` jest: 116 passed (1 new)
- [x] Type check: clean
- [x] Both new orphan-create branches exercised by tests (collision branch via #3167-update, no-collision branch via new #3168 test)

### DIFF
--- a/packages/app/src/__tests__/store/message-handler.test.ts
+++ b/packages/app/src/__tests__/store/message-handler.test.ts
@@ -1749,13 +1749,41 @@ describe('stream_delta handler', () => {
     const toolUse = ss.messages.find((m) => m.id === 'msg-race' && m.type === 'tool_use');
     // tool_use bubble must remain pristine — no delta concatenation
     expect(toolUse?.content).toBe('ls');
-    // App's flushPendingDeltas has no orphan-create safety net (the dashboard's
-    // #2611 fix is dashboard-only), so the colliding delta is silently dropped.
-    // No response message is created at the colliding id. This is more
-    // aggressive data loss than the dashboard but still strictly better than
-    // corrupting the tool_use bubble.
-    const responseAtSameId = ss.messages.find((m) => m.id === 'msg-race' && m.type === 'response');
-    expect(responseAtSameId).toBeUndefined();
+    // Orphan-create (#2611, ported in #3168) suffixes the response id when
+    // there's a non-response collision, so the messages array does not contain
+    // duplicate ids. Matches dashboard behaviour.
+    const orphan = ss.messages.find((m) => m.id === 'msg-race-response');
+    expect(orphan?.type).toBe('response');
+    expect(orphan?.content).toBe('must not leak');
+    // No two messages share an id.
+    const ids = ss.messages.map((m) => m.id);
+    expect(new Set(ids).size).toBe(ids.length);
+  });
+
+  // #3168: port of dashboard's #2611 orphan-create safety net. When a
+  // stream_start frame is dropped (e.g. session not yet in store at the time
+  // it arrived), the subsequent stream_delta has no message to attach to.
+  // Pre-#3168 the app silently dropped these orphan deltas; now it lazy-creates
+  // a response message at the wire id so data is preserved.
+  it('lazy-creates orphan response when stream_delta arrives with no matching message (#3168)', () => {
+    const store = createMockStore({
+      activeSessionId: 's1',
+      sessions: [{ sessionId: 's1', name: 'S1' } as any],
+      sessionStates: { s1: { ...createEmptySessionState(), messages: [] } },
+    });
+    setStore(store as any);
+    _testMessageHandler.setContext(createMockContext() as any);
+
+    // Skip stream_start — simulate the dropped/raced case
+    _testMessageHandler.handle({ type: 'stream_delta', messageId: 'msg-orphan', sessionId: 's1', delta: 'After tool ' });
+    _testMessageHandler.handle({ type: 'stream_delta', messageId: 'msg-orphan', sessionId: 's1', delta: 'response' });
+
+    jest.runAllTimers();
+
+    const ss = store.getState().sessionStates.s1;
+    const orphan = ss.messages.find((m) => m.id === 'msg-orphan');
+    expect(orphan?.type).toBe('response');
+    expect(orphan?.content).toBe('After tool response');
   });
 });
 
--- a/packages/app/src/store/message-handler.ts
+++ b/packages/app/src/store/message-handler.ts
@@ -548,16 +548,42 @@ function flushPendingDeltas(): void {
   for (const [sessionId, deltas] of bySession) {
     if (sessionId && newSessionStates[sessionId]) {
       const sessionState = newSessionStates[sessionId];
+      const matched = new Set<string>();
       // Type guard: never apply deltas to non-response messages, even if id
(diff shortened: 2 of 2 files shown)


blamechris:
## Review of PR #3174 — Port dashboard's #2611 orphan-create safety net

### Summary

Targeted backport of the dashboard's `flushPendingDeltas` orphan-create safety net (#2611) into the app. The change preserves orphaned `stream_delta` payloads that have no matching `response` message at flush time — either because `stream_start` was dropped/raced, or because a non-`response` message (e.g. `tool_use`) is occupying the same id. Without this, post-#3167 the app silently drops the delta. With this PR, the delta is preserved as a fresh `response` (suffixed with `-response` on collision and registered in `_ctx.deltaIdRemaps` so subsequent deltas route correctly).

The two new app loops are byte-for-byte mirrors of the dashboard sessionStates branch, modulo the variable name (`_ctx.deltaIdRemaps` vs `_deltaIdRemaps`). The active-session fallback receives the same logic, which is sensible — it's the app's analogue of the dashboard's flat-state branch.

### Strengths

- **Exact dashboard parity** in the sessionStates branch (lines 548-574 vs dashboard 416-451). The `matched` Set, the `colliding` check, the `-response` suffix, the `_ctx.deltaIdRemaps.set(msgId, targetId)` registration, and the in-place mutation pattern (`finalMessages = updatedMessages` aliasing the freshly `.map()`-produced array) all match.
- **Both code paths covered**: the sessionStates branch and the active-session fallback both get the orphan-create logic, so a delta with `sessionId: null` and an active session set is also protected.
- **Cleanup hooks unchanged and adequate**: `_ctx.deltaIdRemaps` is cleared per-message on `stream_end` (line 1402) and globally on `result` (line 1478) and via `clearPermissionSplits()` (line 452). No new entries are added that bypass these clears, so the new orphan-create paths inherit existing memory-safety guarantees.
- **Tests pivot cleanly**: the existing #3167 regression test now asserts the orphan IS created (matching dashboard), and the new #3168 test mirrors the dashboard's #2611 test for the no-collision happy path. Both new assertion paths cover the previously-silent data-loss cases.
- **Type narrowness preserved**: `type: 'response' as const` keeps the literal type narrow before the `as ChatMessage` cast, and the literal already supplies `id`, `type`, `content`, `timestamp` — all required fields per `packages/store-core/src/types.ts:24`. The cast is safe and idiomatic, again matching dashboard.

### Critical (Must Fix)

None.

### Suggestions (Should Consider)

| # | Location | Suggestion |
|---|----------|------------|
| 1 | `packages/app/src/store/message-handler.ts:548-614` | The orphan-create block is now duplicated verbatim across the sessionStates branch and the active-session fallback branch. Consider extracting a small `applyOrphanDeltas(messages, deltas, matched)` helper that returns the mutated array and registers remaps. The dashboard has the same duplication, so a refactor (in both) would also reduce drift risk on the next backport. Out of scope for this PR — see deferred item. |

### Nitpicks

- The comment block at lines 558-566 (`// Safety net: create response messages for orphaned deltas...`) is identical in both branches — a single shared comment + helper would make future audits easier. (Same observation as Suggestion #1.)
- `finalMessages = updatedMessages` is an alias rather than a copy. This is fine because `.map()` returned a fresh array, but a one-line comment noting "intentionally aliased — `updatedMessages` was just produced by `.map()` and is safe to mutate in place" would help future readers who might worry about state-shape immutability conventions.

### Deferred Items

| Item | Issue |
|------|-------|
| Extract shared `applyOrphanDeltas` helper across app + dashboard `flushPendingDeltas` to reduce duplication and drift | https://github.com/blamechris/chroxy/issues/3176 |

### Architecture Notes — dashboard parity

| Concern | Dashboard | App (this PR) | Match |
|---------|-----------|---------------|-------|
| `matched` Set gates type-filtered map | Yes (line 416) | Yes (line 551) | ✓ |
| `colliding` check uses `m.type !== 'response'` | Yes (line 437) | Yes (line 562) | ✓ |
| Suffix is `${msgId}-response` | Yes (line 438) | Yes (line 563) | ✓ |
| Append vs create branching | Yes | Yes | ✓ |
| Remap registration on collision | `_deltaIdRemaps.set(msgId, targetId)` | `_ctx.deltaIdRemaps.set(msgId, targetId)` | ✓ |
| Cleanup on `stream_end` per message | Yes (line 983) | Yes (line 1402) | ✓ |
| Cleanup on `result` (global) | Yes (line 1743) | Yes (line 1478) | ✓ |
| Fallback branch — different state shape | Flat `messages` via functional setState | `sessionStates[activeId].messages` | Justified divergence |

The fallback-branch divergence is structural (dashboard has both flat and per-session state; app has only per-session) and the porting is faithful within each shape.

### Edge cases verified

- **Multiple unique-id orphans in one flush**: each iteration of the `for (const [msgId, delta] of deltas)` loop is independent. Each unmatched `msgId` produces its own orphan or appends to its own existing target. ✓
- **Delta with no `sessionId` field**: routes through the active-session fallback branch (line 588), which now also has the orphan-create logic. ✓
- **Reconnect/replay scenarios**: if the server replays history and a `stream_delta` lands before the corresponding `stream_start` is registered (or `stream_start` is a no-op because the message already exists), the orphan-create here is the final guard. The defensive remap in `handleStreamDelta` (lines 1351-1377) handles the synchronous case; this PR closes the asynchronous-flush case. ✓
- **`streamingMessageId` on orphan create**: neither the dashboard nor this PR sets `streamingMessageId` when lazy-creating the orphan response. That's consistent — `flushPendingDeltas` is downstream of any UI streaming-state tracking, and `stream_end` will clear streamingMessageId regardless. No new bug here.

### Verdict

**Approve.**

Faithful port with matching tests and zero behavioural divergence in the contested paths. The duplication noted in Suggestions is real but pre-existing in the dashboard and should be addressed in a dedicated refactor PR rather than expanded here.

blamechris:
## Review Comments Addressed

No inline review comments. Copilot's top-level review (id 4215844454) noted "no comments" on the changed files. The agent-review (Chroxy Inspector) posted a top-level review with verdict **Approve** and one deferred suggestion (refactor: extract shared helper) tracked in #3176.

**Total:** 0 inline comments to address
- Fixed: 0
- False positives: 0
- Follow-up issues created: 0 (one created during agent-review: #3176)

Issue #3176, opened 2026-05-03
Refactor: extract shared applyOrphanDeltas helper across app + dashboard flushPendingDeltas


blamechris:
After PR #3174 ported the dashboard's #2611 orphan-create safety net into the app, the orphan-create loop is duplicated three times across the codebase:

- `packages/dashboard/src/store/message-handler.ts:434-447` (sessionStates branch)
- `packages/dashboard/src/store/message-handler.ts:467-482` (flat-state fallback)
- `packages/app/src/store/message-handler.ts:558-574` (sessionStates branch)
- `packages/app/src/store/message-handler.ts:598-614` (active-session fallback)

All four blocks have identical core logic:

\`\`\`ts
for (const [msgId, delta] of deltas) {
  if (matched.has(msgId)) continue;
  const colliding = msgs.some((m) => m.id === msgId && m.type !== 'response');
  const targetId = colliding ? \`\${msgId}-response\` : msgId;
  const existing = msgs.find((m) => m.id === targetId);
  if (existing) {
    const idx = msgs.indexOf(existing);
    msgs[idx] = { ...existing, content: existing.content + delta };
  } else {
    msgs.push({ id: targetId, type: 'response' as const, content: delta, timestamp: Date.now() } as ChatMessage);
  }
  if (colliding) remapsRef.set(msgId, targetId);
}
\`\`\`

### Proposal

Extract into a shared helper. Two options:

1. **Local file helper** in each \`message-handler.ts\` — simple, no cross-package dependency.
2. **Move to \`@chroxy/store-core\`** — already houses shared handlers and would make divergence between app/dashboard impossible by construction.

Option 2 is preferred for the same reason #2611's safety net was ported to the app: any future regression in one consumer should be caught at the shared layer.

### Acceptance criteria

- A single \`applyOrphanDeltas(messages, deltas, matched, remapsRef)\` (or returning the mutation result) with one definition.
- All four call sites use it.
- Existing tests (\`#2611\`, \`#3167\`, \`#3168\`) still pass.
- No behavioural change.

### Why deferred

Out of scope for #3174, which is a faithful port. A refactor that touches both the app and the dashboard handler should be its own PR with its own review focus.

Filed from review of #3174.

blamechris:
Addressed in PR #3180 (commit `d4e465351`) — included in the 16-issue from-review batch cleanup. GitHub's auto-close only matched the first issue after the `Closes` keyword; closing manually with cross-reference.

PR #3263, opened 2026-05-04
test(store-core): direct unit tests for applyOrphanDeltas helper (#3182)

blamechris:
## Summary

- 15 direct unit tests for \`applyOrphanDeltas\` (extracted to store-core in #3176, previously exercised only through higher-level message-handler tests).
- Coverage: all three branches (no-collision, collision with non-response, existing-response concat), the matched-set short-circuit, and the mutate-in-place contract callers depend on.

Closes #3182

## Test Plan

- [x] All new tests pass — \`orphan-deltas.test.ts\` (15 tests)
- [x] Existing tests unbroken — full store-core suite passes (680 tests)
- [x] No source changes — pure test addition

### DIFF
--- a/packages/store-core/src/orphan-deltas.test.ts
+++ b/packages/store-core/src/orphan-deltas.test.ts
@@ -1,0 +1,265 @@
+/**
+ * Direct unit tests for applyOrphanDeltas (#3182).
+ *
+ * The helper has three load-bearing branches:
+ *   1. No collision — append a fresh `response` message at the message id.
+ *   2. Collision (existing non-response at the id, e.g. `tool_use`) — use
+ *      a suffixed id `${msgId}-response` and record the remap so future
+ *      `stream_delta` events route to the suffixed message.
+ *   3. Existing matching response at the (possibly suffixed) id — concatenate
+ *      delta onto its content via index assignment.
+ *
+ * The helper also mutates `messages` in place (push / index assignment) — that
+ * contract matters because the dashboard/app callers pass the result of a
+ * `messages.map(...)` and expect the orphans appended to the same reference.
+ *
+ * Higher-level message-handler tests already cover the integration paths;
+ * these direct tests lock the contract so a future refactor of any single
+ * branch can't silently regress before the integration layer notices.
+ */
+import { describe, it, expect } from 'vitest'
+import type { ChatMessage } from './types'
+import { applyOrphanDeltas } from './orphan-deltas'
+
+function makeMessage(overrides: Partial<ChatMessage> & { id: string }): ChatMessage {
+  return {
+    type: 'response',
+    content: '',
+    timestamp: 1,
+    ...overrides,
+  }
+}
+
+describe('applyOrphanDeltas', () => {
+  describe('branch 1: no collision', () => {
+    it('appends a fresh response message when the id is unused', () => {
+      const messages: ChatMessage[] = []
+      const deltas = new Map([['m-1', 'hello']])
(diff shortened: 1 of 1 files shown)


blamechris:
## Code Review: PR #3263

### Summary

Direct unit tests for `applyOrphanDeltas` — pure test addition with no source changes. Locks the three-branch contract that's currently only exercised indirectly through `dashboard/store/message-handler.test.ts` and the equivalent app handler tests.

### Strengths

- **Clean branch enumeration.** The describe-block hierarchy mirrors the helper's three structural branches (no collision, collision with non-response, existing matching response). Easy to map a future regression to the test that should fail.
- **Mutate-in-place contract is locked.** The helper's callers (dashboard sessionStates + flat, app sessionStates + active-session — four sites total) all depend on receiving the same array reference back after `applyOrphanDeltas` mutates it. Three tests assert `messages === before` — exactly what the four call sites need.
- **Short-circuit semantics are explicit.** "skips collision-remap when matched fires" is the subtle invariant — without it, a delta that's already been applied to an existing `response` would still pollute `remapsRef` and cause a future delta to route to a non-existent suffixed id. The test exercises this directly.
- **Spread-vs-mutation regression test.** `'preserves other ChatMessage fields when concatenating'` would catch a refactor that changes `messages[idx] = { ...existing, content: ... }` to `existing.content +=` — the latter would leak the previous-message reference back into the array.
- **Tests are self-contained.** `makeMessage` factory keeps each test concise and the partial-overrides spread keeps test setup reading top-down.

### Issues Found

#### Critical (Must Fix)

_None._

| `packages/store-core/src/orphan-deltas.test.ts` | 137-152 | The "concatenates onto the suffixed response when collision was previously registered" test seeds `remapsRef` with the prior remap, but `applyOrphanDeltas` doesn't actually consult `remapsRef` on input — it always re-derives the suffix from the current `messages` state. The remap is output-only. Worth a one-line clarifying comment that the seeded remap is documenting the prior call's effect, not feeding into this call's behaviour. | Avoid future readers thinking the helper reads `remapsRef`. |
| `packages/store-core/src/orphan-deltas.test.ts` | 31 | `makeMessage` defaults `timestamp: 1`. A few tests assert specific timestamps; the helper itself uses `Date.now()` for new orphans, so any test that compares timestamps to a literal could flake under fake timers. The tests don't currently do this, but a comment on the factory ("timestamp is a placeholder; helper sets Date.now() on new orphans") would help future authors. | Defensive — no current bug. |

#### Nitpicks (Optional)

- The "treats other non-response types (thinking) as collision too" test could enumerate every non-response type in the union (`user_input`, `tool_use`, `thinking`, `prompt`, `error`, `system`) for true exhaustiveness, but the current spot-check is sufficient.
- The empty-inputs describe block is a single test that arguably belongs alongside the "no collision" branch — but separating it makes the file structure clearer.

_None._

### Architecture Notes

The helper is the canonical implementation across four call sites and lives at the store-core layer (shared between dashboard and app). Direct tests at this layer catch regressions before they reach the platform-specific message-handler tests, which currently provide the same coverage indirectly. This is the right pattern — same as `replay-dedup.test.ts` for `isReplayDuplicate`.

### Verdict

Pure test addition, clean branch coverage, locks the mutate-in-place contract that's load-bearing for all four call sites.

blamechris:
## Review Comments Addressed

| — | No inline comments — Copilot summary only | — | — |

**Total:** 0 comments addressed. Copilot reviewed and posted a summary acknowledging the coverage; no inline findings to fix.

PR #3265, opened 2026-05-04
refactor(store-core): aggregate validateGitElements drop log (#3184)

blamechris:
## Summary

- Replace per-element \`console.debug\` in \`validateGitElements\` with a single aggregated line per payload: \`[handlerName] dropped N/M malformed elements\`.
- 1000-file pathological case now emits 1 log line instead of 1000.

Closes #3184

## Test Plan

- [x] All new tests pass — 50-malformed-entries case verifies single log; clean payload + empty array verify no log
- [x] Existing tests unbroken — full store-core suite passes (668 tests)
- [x] Operator signal preserved — handler name + drop scope still in the single line

### DIFF
--- a/packages/store-core/src/handlers/handlers.test.ts
+++ b/packages/store-core/src/handlers/handlers.test.ts
@@ -2560,6 +2560,51 @@ describe('handleDiffResult', () => {
     debugSpy.mockRestore()
   })
 
+  // #3184: aggregate the per-element drop logs into ONE bounded log per
+  // payload. A pathological case (1000-file diff where every entry is
+  // malformed because of a server-side regression) previously emitted 1000
+  // console.debug lines per `diff_result` event; the new contract emits a
+  // single line per call with the drop count.
+  it('emits a single aggregated debug log per payload, not one per dropped element (#3184)', () => {
+    const debugSpy = vi.spyOn(console, 'debug').mockImplementation(() => {})
+    // Construct 50 malformed entries — without the aggregation fix this
+    // would emit 50 console.debug calls.
+    const files = Array.from({ length: 50 }, (_, i) => ({
+      path: `bad${i}.txt`, // missing status
+      additions: 0,
+      deletions: 0,
+      hunks: [],
+    }))
+    handleDiffResult({ files })
+
+    expect(debugSpy).toHaveBeenCalledTimes(1)
+    // The single line must carry the count and total so an operator can see
+    // the failure scope at a glance.
+    const message = debugSpy.mock.calls[0]?.[0] as string
+    expect(message).toMatch(/handleDiffResult\.files/)
+    expect(message).toMatch(/50.*50|50\/50/)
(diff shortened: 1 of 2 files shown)


blamechris:
## Code Review: PR #3265

### Summary

`validateGitElements` aggregates its drop log to one line per payload. A 1000-file malformed diff goes from 1000 console.debug lines to 1 — same operator signal (handler + count) at bounded cost.

### Strengths

- **Bounded log, preserved signal.** `[handlerName] dropped N/M malformed elements` keeps the two pieces an operator actually needs (which handler, how bad). The dropped element's index, removed in this refactor, was rarely actionable on its own.
- **Tests cover the three cases that matter.** Aggregation test (50 → 1), no-log-on-clean, no-log-on-empty. The aggregation test asserts both the count and the handler name in the single line so a future regression splitting the message can't pass.
- **No leak surface change.** The element value is still not logged, matching the original "keep the debug log narrow" intent.
- **Comment hygiene.** The new doc explains the why (Metro/Vite/browser noise from server-side regressions) and references the issue. A future reader who finds this and wants to add per-element logging can see the prior decision.

### Issues Found

#### Critical (Must Fix)

_None._

| `packages/store-core/src/handlers/index.ts` | 1900 | The aggregated log dropped the rejected indices. Consider whether knowing _which_ indices failed has any debugging value (e.g. "are the bad entries clustered or scattered?"). Probably not — operators chasing a regression have access to the full payload via the dev tools network panel. Defer indefinitely. | The refactor is the right call; just noting the tradeoff. |
| `packages/store-core/src/handlers/index.ts` | 1894 | A separate counter via `out.length` and `arr.length` (i.e., `dropped = arr.length - out.length`) would shave the explicit `dropped++` increment. Stylistic only. | Defer indefinitely. |

#### Nitpicks (Optional)

- The aggregation test uses `Array.from({ length: 50 }, ...)` — could be 10 to keep the test description literal-friendly. 50 is fine and signals "many" without ceremony.
- The regex `/50.*50|50\/50/` could be tightened to just `/50\/50/` since the message format is deterministic. Defensive flexibility now, premature loosening later — current form is fine.

The issue body mentioned an optional follow-up: when `dropped === total && total > 0`, attach a synthetic error to the payload so the UI distinguishes "server regression dropped everything" from "clean repo". That's intentionally out of scope here — it changes the consumer contract. If wanted, that's a separate issue (probably worth opening for the next sweep).

### Architecture Notes

`validateGitElements` is the choke point for git-result fail-soft validation across `diff_result` / `git_status_result` / `git_branches_result`. Aggregating at this layer is the right granularity — every caller benefits without any per-call-site change.

### Verdict

Bounded log, three regression tests, clean comment trail. No issues to fix.

blamechris:
## Review Comments Addressed

| — | No inline comments — Copilot summary only | — | — |

**Total:** 0 comments addressed.

Issue #3279, opened 2026-05-04
skills-loader: priority-aware two-pass tier budget (_collectCandidates) (#3275)


blamechris:
## Context

Core implementation of #3275. **Depends on the comparator helper sub-task (`_compareByPriorityThenName`) AND the bounded-read helper sub-task (`_readFrontmatterOnly`)** — both must merge before this can start. See parent #3275 for the full design rationale.

## Problem

`loadActiveSkills` in `packages/server/src/skills-loader.js` (lines 241-672) currently sorts entries alphabetically and stops reading once the per-tier byte budget is exceeded (lines 308-328 setup, 467-481 budget check). When the budget is tight, a low-priority skill that sorts FIRST fills the budget and crowds out a high-priority skill that sorts LATER. The post-merge `_enforceTotalBudget` (skills-budget.js:62) only sees what was loaded, so priority can't help skipped skills.

## Change

When `tierBudget !== null`, replace the alphabetical-cutoff loop with a two-pass approach. Keep the existing single-pass alphabetical loop intact for the `tierBudget === null` back-compat path.

### Pass 1: `_collectCandidates` (new private helper)

For every entry (after readdir + alphabetical sort for tiebreak determinism), run the FULL TOCTOU-safe validation cluster — extension/disabled-suffix → `statSync` → `openSync` → `fstatSync` → `realpathSync` → `allowedRoots` containment → `dev+ino` re-check (current lines 333-460). Then either:

- **parseCache hit** (matching `mtimeMs+size` on `realPath`): use `cached.frontmatter` directly. Skip the partial read. Set `cachedFrontmatter: cached` on the descriptor so pass 2 can short-circuit further.
- **Cache miss**: call `_readFrontmatterOnly(fd, fstat.size, { maxBytes: 4096 })` and extract `priority` via `_priorityOf({ metadata: frontmatter })`.

`closeSync(fd)` in a `finally` before each iteration ends. Drop the partial buffer immediately. Keep ONLY a small descriptor:

```js
{ entry, fullPath, realPath, fstat: { size, mtimeMs, dev, ino }, priority, cachedFrontmatter }
```

### Pass 2: priority-sorted full read

Sort candidates with the shared `_compareByPriorityThenName`. Walk the sorted list with a running `tierTotalBytes`. For each candidate:

- If `tierTotalBytes + fstat.size > tierBudget` → `continue` (NOT `break` — preserves the existing "smaller-later-fits" semantic).
[...]

PR #3291, opened 2026-05-05
fix(server): align pass-1 sort tiebreak with _enforceTotalBudget (#3287)

blamechris:
Closes #3287

## Summary

- Pre-computes the skill stem (`entry.slice(0, -(ext.length + 1))`) as `name` on the candidate descriptor in `_collectCandidates`
- Updates the pass-1 sort adapter to use `candidate.name` directly instead of recomputing the stem inline via `lastIndexOf`
- Removes the two redundant `const name = entry.slice(...)` derivations in the pass-2 loop (cached path and cache-miss path)
- Adds a regression test: `abc.md` and `abc-extra.md` at equal priority with a single-slot budget — verifies `abc.md` wins (stem `"abc"` < `"abc-extra"`), not `abc-extra.md` (which wins when full filenames are compared because `'-'` 0x2D < `'.'` 0x2E)

## Test plan

- [ ] New test `'pass-1 tiebreak and _enforceTotalBudget agree for prefix-pair filenames'` in `describe('sort tiebreak consistency (#3287)')` passes
- [ ] Confirmed the test fails when the sort adapter is reverted to use `a.entry` (full filename) instead of `a.name` (stem)
- [ ] All existing `priority-aware per-tier budget (#3279)` tests still pass
- [ ] Full test suite passes (pre-existing failures: TunnelManager, WebTaskManager — both require external tools)

### DIFF
--- a/packages/server/src/skills-budget.js
+++ b/packages/server/src/skills-budget.js
@@ -73,7 +73,7 @@ export function _compareByPriorityThenName(a, b) {
 /**
  * Apply the global skills budget (#3202). Skills are sorted by priority
  * descending (higher priority kept first), with alphabetical name as the
- * tiebreaker — same direction as the existing top-level sort. We then walk
+ * tiebreaker — same order enforced by `_compareByPriorityThenName`. We then walk
  * the list, accumulating bytes until we'd exceed the cap; the first skill
  * that wouldn't fit (and every later one) is dropped.
  *
--- a/packages/server/src/skills-loader.js
+++ b/packages/server/src/skills-loader.js
@@ -344,26 +344,17 @@ export function loadActiveSkills(dir, opts = {}) {
 
     // Sort by priority desc, name asc for deterministic tiebreaking.
     // _compareByPriorityThenName expects { name, metadata: { priority } };
-    // we construct an adapter object from the candidate descriptor — the
-    // comparator only reads `name` and `metadata.priority`, so wrapping
-    // the descriptor's raw priority value is sufficient and avoids having
-    // to thread a full frontmatter object through _collectCandidates.
-    candidates.sort((a, b) => {
-      // Strip the extension from the candidate's filename so the
-      // alphabetical tiebreak uses the same extension-free name that every
-      // downstream code path (and _compareByPriorityThenName's callers on
-      // the single-pass path) operates on. Without this, "a-b.md" would
-      // sort before "a.md" because '-' (0x2D) < '.' (0x2E), diverging from
-      // how skills are sorted everywhere else.
-      const extA = a.entry.slice(a.entry.lastIndexOf('.') + 1).toLowerCase()
-      const extB = b.entry.slice(b.entry.lastIndexOf('.') + 1).toLowerCase()
-      const nameA = a.entry.slice(0, -(extA.length + 1))
-      const nameB = b.entry.slice(0, -(extB.length + 1))
-      return _compareByPriorityThenName(
-        { name: nameA, metadata: { priority: a.priority } },
-        { name: nameB, metadata: { priority: b.priority } },
-      )
-    })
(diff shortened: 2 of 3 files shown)


blamechris:
## Code Review: PR #3291

### Summary

Tight, focused fix for the pass-1 / `_enforceTotalBudget` tiebreak inconsistency (#3287). Pre-computes the stem on the candidate descriptor in `_collectCandidates`, threads it through the pass-1 sort adapter and the pass-2 loop destructuring, and removes the two now-redundant `const name = entry.slice(...)` derivations. The regression test is load-bearing and the diff is surgically scoped.

---

### Strengths

- **Root-cause fix**: Moving the stem computation to the descriptor's construction site means there is exactly one place where `"abc.md"` → `"abc"` can go wrong. Both sort sites now read the same field from the same source.
- **JSDoc updated**: `_collectCandidates`'s return-type comment now includes `name` and explains the #3287 invariant. The `_enforceTotalBudget` block comment was also tightened to reference `_compareByPriorityThenName` by name rather than describing the sort implicitly — good for auditability.
- **No scope creep**: Diff touches exactly `skills-loader.js`, `skills-budget.js` (comment only), and the test file. Nothing else changed.
- **CI green**: All 12 checks pass.

---

### Issues Found

#### Critical (Must Fix)

_None._

| `skills-loader.js` | `_collectCandidates` push | The `entry` field is still included in the descriptor (`entry, name, fullPath, …`) but `entry` is no longer consumed by pass-2 (the destructuring at the loop head dropped it). Keeping it is harmless, but it's a silent dead field on every descriptor. A follow-up could drop `entry` from the returned shape once no callers need it, to keep the descriptor minimal. | The `entry` field was the source of the original bug; leaving it on the object creates future temptation to re-introduce the extension-ful comparison. |

#### Nitpicks (Optional)

- The sort call could be written as a one-liner with the arrow removed, e.g. `candidates.sort(_compareByPriorityThenName.bind(null, …))` — but the current explicit adapter (`(a, b) => _compareByPriorityThenName({name: a.name, …}, {name: b.name, …})`) is clearer about the shape mismatch between the raw descriptor and the `{name, metadata}` interface. No change needed.

---

### Test Verification

The new test is load-bearing. Without the fix:

- Pass-1 sort would use `a.entry` / `b.entry` (full filenames).
- `"abc-extra.md"` sorts before `"abc.md"` because `'-'` (0x2D) < `'.'` (0x2E) at index 3.
- `abc-extra` becomes `skills[0]`; assertion `skills[0].name === 'abc'` fails.

With the fix:

- Pass-1 sort uses `a.name` / `b.name` (stems `"abc-extra"` vs `"abc"`).
- `"abc"` sorts before `"abc-extra"` alphabetically.
- `abc` becomes `skills[0]`; assertion passes.

Budget math is correct: each file is ~1022 bytes (22-byte frontmatter + 1000-byte body); `maxTotalBytes: 1100` fits exactly one.

---

| Drop `entry` from the candidate descriptor return shape | Safe to defer — `entry` is inert in pass-2 now and has no correctness impact. Removing a public-ish struct field deserves its own small PR with a search for any test helpers or logging that might still reference it. |

---

### Verdict

- [x] Approve

blamechris:
## check-pr result

| PR | Comments | Changes | Issues |
|----|----------|---------|--------|
| #3291 | 0 Copilot inline comments — nothing to process | none | Deferred: #3293 (drop dead `entry` field from candidate descriptor). Created: none. Closed: none. |

**Copilot verdict:** COMMENTED (no inline suggestions) — reviewed 3/3 files, generated no comments.

**Agent-review verdict:** Approve. One suggestion deferred to #3293 (`entry` field now dead on candidate descriptor — drop it). No action needed on this PR.

**CI:** All 12 checks pass. PR is mergeable.

PR #3294, opened 2026-05-05
test(server): document single-file constraint on Guard 2 trigger (#3292)

blamechris:
Closes #3292


blamechris (commented):
**Verdict: Approve**

Comment-only diff confirmed — 6 lines added inside the existing Guard 2 comment block in `skills-loader-toctou.test.js`, no executable code changed.

Comment accuracy verified against the mock logic: `fstatCallCount` is a flat call counter incremented on every `fstatSync` invocation. With one skill file, call #1 is pass-1 and call #2 is pass-2, which is exactly where the fabricated ino is injected. A second file would advance call #2 to pass-1 of that file, leaving the actual pass-2 Guard 2 check at call #3 without the fabricated ino — the guard would silently stop firing while the test continued to pass via a different path. The new comment captures this precisely.

Copilot review (COMMENTED): summary only, no inline concerns raised. 0 inline comments to process.

PR #3295, opened 2026-05-05
refactor(server): drop dead entry field from _collectCandidates descriptor (#3293)

blamechris:
Closes #3293

### DIFF
--- a/packages/server/src/skills-loader.js
+++ b/packages/server/src/skills-loader.js
@@ -972,7 +972,7 @@ export function loadActiveSkills(dir, opts = {}) {
  * entry so pass 1 never holds more than one fd open at a time.
  *
  * Returns an array of lightweight candidate descriptors:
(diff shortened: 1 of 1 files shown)


blamechris:
Diff is exactly what it says: `entry,` removed from `candidates.push({…})` and JSDoc updated to match. No other files touched.

Correctness verified:
- Pass-2 loop destructure (`name, fullPath, label, realPath, fstat, cachedFrontmatter`) never referenced `entry` — removal has no runtime effect.
- Repo-wide grep for `candidate.entry` / `c.entry` / descriptor `.entry` accesses: zero hits in skills-loader or any consumer.
- JSDoc shape now matches the actual object.
- 167/167 skills-loader tests pass on the branch.
- All CI checks green.

Copilot summary review had no inline comments or actionable findings.

LGTM — safe to squash merge.

Issue #3298, opened 2026-05-05
dashboard: pending-review section + skill_trust_grant UI (#3206 PR C)


blamechris:
## Context

Sub-task C of #3206 (first-activation prompt for skills/community/<author>/ namespace). **Depends on PR B** (server-side WS protocol + trust store).

## Scope

Dashboard UI for the community-skill trust prompt. After this PR, when the server emits `skill_trust_request`, the dashboard's SkillsPanel renders a "Pending review" section with a `Trust 'alice'` button per pending skill. Clicking sends `skill_trust_grant`; on `skill_trust_granted` the skill moves out of pending into the appropriate auto/manual section.

### Component (`packages/dashboard/src/components/SkillsPanel.tsx`)

- [ ] New "Pending review" section above the existing Always-on / Manual sections. Renders only when there are pending community skills. Each row shows: skill name, description, `from: <author>` badge, and a `Trust '<author>'` button.
- [ ] Click `Trust '<author>'` button → invokes `onGrantTrust(skillName, author)` prop.
- [ ] Section is hidden when `auth_ok.capabilities.skillTrustGrant !== true` (older servers don't support runtime grant).

### Store / message handlers (`packages/dashboard/src/store/`)

- [ ] Add `trustState?: 'pending' | 'trusted'`, `communityAuthor?: string` fields to `SessionSkillInfo` type.
- [ ] Add `pendingCommunitySkills?: Array<{name: string, author: string}>` field to `SessionState`.
- [ ] `handleSkillTrustRequest(state, msg)` — adds `{name, author}` to `pendingCommunitySkills`. Idempotent (don't double-add).
- [ ] `handleSkillTrustGranted(state, msg)` — removes the matching entry from `pendingCommunitySkills`. The subsequent `skills_list` rebroadcast (sent by the server post-grant) updates the rest of the state.
- [ ] Update `handleSkillsList` if needed to surface `trustState` / `communityAuthor` from server payload.

### App wiring (`packages/dashboard/src/App.tsx` or wherever `<SkillsPanel>` is mounted)

- [ ] Pass `pendingCommunitySkills` from session state into `<SkillsPanel>`.
- [ ] Pass `onGrantTrust` callback that sends `{ type: 'skill_trust_grant', skillName, author, scope: 'author' }` over WS.

### Tests

- [ ] `SkillsPanel.test.tsx`:
  - Pending review section renders only when `pendingCommunitySkills` has entries
[...]

PR #3351, opened 2026-05-05
test(dashboard): cross-author collision for skill_trust_granted (#3309)

blamechris:
Closes #3309

## Summary

- Adds one test to the `skill_trust_granted (#3298)` describe block in `message-handler.test.ts`
- The test populates `pendingCommunitySkills` with two entries sharing the same `name` (`skill-x`) but different authors (`alice`, `bob`)
- Fires `skill_trust_granted` for `alice/skill-x` and asserts only bob's entry remains, confirming the dual-field filter (`name && author`) works correctly for cross-author collisions

## Test plan

- [x] `npm test` — 1404 tests pass (1403 existing + 1 new)
- [x] `npm run typecheck` — no errors

### DIFF
--- a/packages/dashboard/src/store/message-handler.test.ts
+++ b/packages/dashboard/src/store/message-handler.test.ts
@@ -1539,6 +1539,24 @@ describe('dashboard message-handler dispatch', () => {
         // List should be unchanged (no valid match)
         expect(state.pendingCommunitySkills).toEqual([{ name: 'skill-a', author: 'alice' }])
       })
+
+      it('does not remove a same-name entry from a different author', () => {
+        const empty = createEmptySessionState()
+        store = createMockStore(baseState({
+          activeSessionId: 's1',
+          sessionStates: {
+            s1: { ...empty, pendingCommunitySkills: [
+              { name: 'skill-x', author: 'alice' },
+              { name: 'skill-x', author: 'bob' },
+            ] },
+          },
+        }))
+        setStore(store)
+        handleMessage({ type: 'skill_trust_granted', skillName: 'skill-x', author: 'alice', sessionId: 's1' }, ctx() as any)
+
+        const state = (store.getState() as any).sessionStates.s1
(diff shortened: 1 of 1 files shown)


blamechris:
## Code Review: PR #3351

### Summary

Pure test addition — one new `it` block inside the existing `skill_trust_granted (#3298)` describe in `message-handler.test.ts`. Closes #3309, which called out the missing cross-author collision coverage for `handleSkillTrustGranted`.

### Strengths

- **Test directly matches the issue spec.** #3309 prescribed the exact scenario (two `pendingCommunitySkills` entries with the same `name`, different `author`, grant one, verify the other survives), and the new test implements it verbatim — including the suggested variable names and assertion shape.
- **Correct assertion.** The `toEqual([{ name: 'skill-x', author: 'bob' }])` assertion is tight: it verifies both that alice's entry was removed *and* that bob's entry survived with its full shape intact. No loose length check.
- **Follows established test patterns.** `createEmptySessionState()` + `createMockStore(baseState({...}))` + `setStore()` + `handleMessage()` is the pattern every other test in the block uses. No drift.
- **Self-contained.** The store is re-created inside the `it` body — no reliance on the outer `beforeEach` store — consistent with the other tests in this describe block that need custom initial state.
- **Implementation coverage confirmed.** The filter in `handleSkillTrustGranted` is `!(p.name === skillName && p.author === author)` — a conjunction. This test is the only one that isolates the `author` half of that conjunction; without it, a regression to a name-only filter would go undetected.

### Issues Found

#### Critical (Must Fix)

None.

None.

#### Nitpicks (Optional)

- `s1` as both the session key and the `activeSessionId` is fine for a unit test, but a comment like `// two skills share the same name, different authors` would make the intent scannable without reading the data. Very minor — take or leave.

None. The PR is surgical and complete.

### Architecture Notes

No architecture impact. The test exercises only the store's read path via `getState()` after a dispatched message — no WebSocket, no server, no mobile concerns.

### Verdict

The test exactly covers what #3309 asked for. CI is green on all dashboard/server/app checks (Desktop Rust Tests still running but unrelated to this change). No issues found.

PR #3352, opened 2026-05-05
chore(server): pin @anthropic-ai/claude-code in sidecar Dockerfile (#3330)

blamechris:
Closes #3330

## What

Pins `@anthropic-ai/claude-code` in the sidecar `Dockerfile` to an exact version via a Docker build ARG, making image builds reproducible.

## Changes

- **`packages/server/sidecar/Dockerfile`** — adds `ARG CLAUDE_CODE_VERSION=2.1.128` and uses `@anthropic-ai/claude-code@${CLAUDE_CODE_VERSION}` in the global install step
- **`packages/server/sidecar/PROTOCOL.md`** — new "claude-code Version Pinning" section documenting the ARG default, the build-time override, and the manual bump cadence

## Pinning strategy: build ARG with exact default

The `ARG` approach was chosen over a hard-coded version string because:
- It keeps the default readable at the top of the Dockerfile alongside the intent comment
- CI / release pipelines can override without editing the file: `docker build --build-arg CLAUDE_CODE_VERSION=x.y.z .`
- The default `2.1.128` matches the current `latest` tag on npm (verified at PR time)

**Bump cadence:** manual PR to update the ARG default, reviewed before landing. No automated renovation — intentional so breaking changes surface at review time.

## Test plan

- [ ] Dockerfile builds without errors: `docker build packages/server/sidecar/`
- [ ] Override works: `docker build --build-arg CLAUDE_CODE_VERSION=2.1.0 packages/server/sidecar/`
- [ ] CI passes (no Node.js source changes, no test impact)


blamechris:
## Code Review: PR #3352

### Summary

Pins \`@anthropic-ai/claude-code\` in the sidecar Dockerfile via \`ARG CLAUDE_CODE_VERSION=2.1.128\`, with a build-time override mechanism. Adds a documentation section to \`PROTOCOL.md\` explaining the pin and bump cadence. Tiny, focused scope: 2 files, 26 lines added.

### Strengths

- ARG default is correct: \`2.1.128\` is the current npm \`latest\` at PR time (confirmed via \`npm show @anthropic-ai/claude-code version\`).
- The override pattern (\`docker build --build-arg CLAUDE_CODE_VERSION=x.y.z .\`) is idiomatic Docker and will work exactly as documented.
- Comment in the Dockerfile explicitly notes the "keep in sync" intent with the rest of the server package — good signal for future maintainers.
- \`npm ci\` is already used for the sidecar's own deps, so the reproducibility story is consistent throughout the image.
- Intentional no-automation decision is documented rather than left as an unexplained omission.

### Issues Found

#### Critical (Must Fix)

None.

| \`packages/server/package.json\` | \`dependencies\` | \`@anthropic-ai/claude-code\` is not a declared dependency of the server package — it is installed globally inside the container. The Dockerfile comment says "keep it in sync with the version used by the rest of the server package", but there is no corresponding version anchor in \`packages/server/package.json\`. Consider adding a \`devDependencies\` entry (or a comment constant) so the version lives in one canonical place and the Dockerfile ARG default can reference it. | Right now the comment is aspirational; the sync is manual with no machine-readable tie. |

#### Nitpicks (Optional)

- PROTOCOL.md section header capitalises only the first word: `## claude-code Version Pinning`. The surrounding section headers use title case (`## Ping / Pong Keepalive`, `## Error Cases`). Recommend `## claude-code Version Pinning` → `## Version Pinning` or match the existing casing style consistently.
- The two-space double-indent in PROTOCOL.md code block (\`docker build --build-arg ...\`) is consistent with the rest of the file — fine as-is.

| Add Renovate/Dependabot rule to flag \`CLAUDE_CODE_VERSION\` drift automatically | [#3354](https://github.com/blamechris/chroxy/issues/3354) | Intentionally out of scope for this pin-only PR; requires Renovate \`regexManagers\` config that doesn't exist yet — separate change |

### Architecture Notes

The sidecar is a standalone package (\`chroxy-pod-agent\`) that does not share the monorepo's \`node_modules\`; \`@anthropic-ai/claude-code\` is installed globally in the image rather than as a workspace dependency. That architecture makes a single canonical version location tricky, but the Dockerfile ARG default is the next-best anchor. The suggestion above (a devDependency or comment constant in \`packages/server/package.json\`) would close the gap without changing the build model.

PROTOCOL.md is the right place for this documentation — it already covers the full operational contract for the sidecar, and version management is part of that contract. A sidecar-specific README would be redundant given the scope and detail already in PROTOCOL.md.

### Verdict

**Approve** — the pin is correct, the override mechanism works, and the documentation is clear. The "keep in sync" comment is slightly aspirational without a machine-readable anchor, but that is a follow-up concern tracked in #3354, not a blocker.

Issue #3206, opened 2026-05-03
server: first-activation prompt for skills/community/<author>/ namespace


blamechris:
## Parent epic
Part of #2959 — skills trust model.

## Scope
Skills under `skills/community/<author>/` are treated as untrusted by default. First time the user tries to activate, the dashboard prompts: "This skill came from <author>. Review and confirm to enable."

## Acceptance criteria
- [ ] Skill loader marks community-namespace skills as `trustState: 'pending'`
- [ ] Pending skills are filtered out of default-active set even if `activation: auto`
- [ ] WS message `skill_trust_request` sent to the dashboard on first activation attempt
- [ ] User confirms → `trustState: 'trusted'` recorded; subsequent activations bypass the prompt
- [ ] Tests: prompt fires first time, doesn't fire after trust granted

---
Parent epic: #2959

blamechris:
## Decomposed into 3 sub-PRs

| PR | Issue | Scope | Dependencies |
|----|-------|-------|--------------|
| A | #3296 | Loader: community-namespace detection + `trustState: 'pending'` tagging + filter + `communityTrustChecker` callback seam + `community/<author>/` walk | — |
| B | #3297 | Trust store schema migration + WS protocol (`skill_trust_request`/`grant`/`granted`) + handler + BaseSession wiring + capability advertisement | #3296 |
| C | #3298 | Dashboard: "Pending review" section + `Trust '<author>'` button + `skill_trust_request`/`granted` handlers in store | #3297 |

**Reuse decision:** new flow distinct from `skill_trust_accept` (different semantics — first-activation namespace blessing vs hash-mismatch acceptance), but shares the on-disk `skills-trust.json` file with additive migration. Author-grain trust (`Trust 'alice'` blesses all of alice's skills); per-skill grain stored too for future revocation.

**Migration risk:** legacy trust files have path-keyed records at root; new shape nests them under `skills` and adds `communityTrust` at root. `_load()` heuristically detects legacy and migrates on first flush. Existing fail-open guarantee covers corrupted files.

**CLI-only limitation:** community skills in dashboard-less sessions stay pending forever; manual workaround is editing `skills-trust.json` directly. A future `chroxy skills trust <author>` subcommand could be added — out of scope here.

blamechris:
All 3 sub-PRs merged:

| PR | Issue | Commit | What |
|----|-------|--------|------|
[...]

PR #3353, opened 2026-05-05
test(server): community-gate two-pass path coverage (#3300)

### LINKED ISSUE #3300: test(server): community-gate tests for two-pass path (maxTotalBytes)
## Context — Identified during review of PR #3299.

All 16 new community-namespace gate tests added in PR #3299 exercise only the single-pass code path (no `maxTotalBytes` opt). The two-pass cache-fast-path (lines 510–536 in `skills-loader.js`) and the two-pass cache-miss-path (lines 721–746) each contain distinct community-gate blocks that are structurally identical to the single-pass version but textually separate.

A regression in one copy would not be caught by the current test suite.

## Original Review Comment

> All 16 new tests exercise only the single-pass code path (no `maxTotalBytes` opt). The two-pass cache-fast-path and two-pass cache-miss-path each contain distinct community-gate blocks (lines 510–536 and 721–746) that are structurally identical but textually separate. A regression in one copy would not be caught by the existing test suite. Adding a few community-gate integration tests with `maxTotalBytes` set (to force the two-pass path) would verify all three implementations remain consistent.

## What to add

Add integration tests in the `community-namespace gate (#3206 PR A)` describe block that pass `maxTotalBytes` to force the two-pass `_collectCandidates` path. Minimum coverage:
- pending community skill excluded from default-active set (two-pass path)
- `trustStore.inspect()` not called for pending community skill (two-pass path)
- trusted community skill loads with `trustState: 'trusted'` (two-pass path, cache-miss)
- trusted community skill loads with `trustState: 'trusted'` (two-pass path, cache-fast-path via `parseCache` Map pre-populated)

Ref: PR #3299, part of #3206.

blamechris:
Closes #3300

## Summary

- Adds 4 integration tests to the `community-namespace gate (#3206 PR A)` describe block in `skills-loader.test.js`
- All 4 tests pass `maxTotalBytes: 1024 * 1024` to force the two-pass `_collectCandidates` code path
- Covers both the cache-fast-path (lines 510–536) and the cache-miss-path (lines 721–746) community-gate blocks, which were previously untested

## Tests added

1. **Pending skill excluded from default-active set** — two-pass path
2. **`trustStore.inspect()` not called for pending skills** — two-pass path
3. **Trusted skill loads with `trustState: 'trusted'`** — two-pass cache-miss path (no `parseCache` supplied)
4. **Trusted skill loads with `trustState: 'trusted'`** — two-pass cache-fast-path (pre-populated `parseCache` Map)

## Test plan

- All 4 new tests pass: `node --test --test-name-pattern "two-pass path" packages/server/tests/skills-loader.test.js`
- All pre-existing failures confirmed pre-existing on base branch (unrelated to this change)


blamechris:
## Code Review: PR #3353

### Summary

Adds 4 integration tests to the `community-namespace gate (#3206 PR A)` describe block covering the two-pass `_collectCandidates` path (activated by `maxTotalBytes`). Tests mirror existing single-pass community-gate tests and directly target the two structurally-separate gate copies at lines 510–536 (cache-fast-path) and 721–746 (cache-miss-path).

### Strengths

- `maxTotalBytes: 1024 * 1024` is the correct activation signal — `tierBudget` is set to a non-null value only when `Number.isFinite(opts.maxTotalBytes) && opts.maxTotalBytes > 0` (line 361), so these tests are guaranteed to enter the two-pass branch at line 449 and never fall through to the single-pass path at line 797.
- Tests 3 and 4 assert all five observable fields (`name`, `active`, `trustState`, `communityAuthor`, `inspectCallCount`), which is specific enough to catch the most likely regression forms.
- The cache warm-up strategy in test 4 is correct: a first call with `parseCache: warmCache` populates the map via the cache-miss path's `parseCache.set()` at line 699; the second call with the same `Map` triggers `cacheHit` in `_collectCandidates` (line 1355) and the `cachedFrontmatter` branch in pass 2 (line 488). The path is actually exercised.
- Test isolation is sound: `communityDir` is a fresh `mkdtempSync` per test via `beforeEach`; no shared cache state leaks between tests.
- Style conforms to project conventions: no semicolons, single quotes, plain JS.

### Issues Found

#### Critical (Must Fix)

None.

| `skills-loader.test.js` | test 2 (inspect-not-called) | Also assert the returned pending skill shape when `includeInactive:true` | Test 2 exercises the `includeInactive:true` pending branch (lines 526–533 / 737–744) but only checks `inspectCallCount`. A regression that silently drops the pending entry instead of pushing it would pass undetected. Adding `assert.equal(skills.length, 1); assert.equal(skills[0].trustState, 'pending')` closes the gap. |

#### Nitpicks (Optional)

- Test 1 only asserts `skills.length === 0`. Functionally equivalent to the single-pass mirror test so consistent, but adding `assert.equal(skills.length, 0, '...')` is already there (message is present). No change needed.
- The comment block correctly references lines 510–536 and 721–746 — these are accurate against the current source.

| Verify `onCommunityTrustPending` callback fires in two-pass path (both sub-paths); assert `includeInactive:true` pending shape; add fail-open (no `communityTrustChecker`) two-pass test | [#3360](https://github.com/blamechris/chroxy/issues/3360) | These are additive coverage gaps, not correctness bugs in the existing tests. The PR closes the primary regression risk for the gate logic. The three uncovered branches are low-risk (callback firing, fail-open fallback) or exercised-but-unasserted — suitable for a follow-up. |

### Architecture Notes

Tests are correctly scoped within the existing `describe('community-namespace gate')` block and use the same `communityDir` fixture pattern as surrounding tests. No new test infrastructure or exports needed. The `_collectCandidates` function is tested only indirectly through `loadActiveSkills` (the only public entry point), which is the right level for integration tests.

### Verdict

- [x] **Approve** — The 4 tests correctly exercise the two-pass code path, assertions are specific enough to catch the primary regression cases, and the cache-fast-path warm-up strategy is correct. One follow-up issue (#3360) tracks three smaller coverage gaps that are out of scope for this PR.

PR #3357, opened 2026-05-05
feat(server): DockerBackend execInEnvironment honors env + cwd opts (#3312)

### LINKED ISSUE #3312: feat(server): implement execInEnvironment opts.env and opts.cwd in DockerBackend
## Context — Identified during review of PR #3311.

The Backend interface (`types.js`) specifies `execInEnvironment(containerId, { cmd, env?, cwd?, timeout? })` with `env` and `cwd` as supported opts. The DockerBackend implementation destructures only `{ cmd, timeout }` and silently ignores `env` and `cwd`.

Since `execInEnvironment` is a forward-looking method (not yet called by EnvironmentManager), the incomplete implementation is not a runtime bug today. But the interface is a contract K8sBackend and callers will rely on. Having opts in the signature that are silently dropped is misleading.

## Fix

In `packages/server/src/environments/backends/docker.js`, update `execInEnvironment` to:
- Pass `--env KEY=VAL` flags for each entry in `opts.env` (if provided)
- Pass `--workdir DIR` flag for `opts.cwd` (if provided)

## Notes

- `execInEnvironment` is used by K8s WS bridge sidecar pattern (#3192) — this should land before or alongside that work.
- Add a test case in `docker.test.js` covering both opts.

Part of #2450

blamechris:
Closes #3312

## Summary

- `execInEnvironment` previously destructured only `{ cmd, timeout }` and silently dropped `opts.env` and `opts.cwd`, violating the Backend interface contract defined in `types.js`
- `opts.cwd` is now forwarded as `--workdir <cwd>` to `docker exec`
- `opts.env` entries are now forwarded as `--env KEY=VAL` flags to `docker exec`
- When neither opt is provided the invocation shape is identical to before (no regression)

## Allowlist policy

`execInEnvironment` does **not** apply an allowlist filter on `opts.env`. This is intentional and different from `streamCliInEnvironment`:

- `streamCliInEnvironment` runs the long-lived Claude Code CLI and receives the host's full `process.env` as its `env` argument — an allowlist (`FORWARDED_ENV_KEYS`) prevents accidental secret leakage
- `execInEnvironment` receives an explicit, caller-constructed env object; `process.env` is never consulted. The caller is responsible for passing only what the command needs. This is documented in the JSDoc.

## Tests

6 new cases added to `tests/environments/backends/docker.test.js`:

1. Baseline: no opts → `['exec', containerId, 'bash', '-c', cmd]` exactly
2. `opts.cwd` only → `--workdir` flag present, container ID positioned after flag
3. `opts.env` only → `--env KEY=VAL` per entry
4. Both opts together → flags appear before container ID, `bash -c cmd` at end
5. Non-zero exit → rejects with stderr message
6. `process.env` isolation → unlisted host vars never appear in exec args

All 42 tests in the docker backend suite pass. The 11 pre-existing failures in the full suite (tunnel integration, WebTaskManager, encryption) are unrelated and present on `main`.


blamechris:
## Code Review: PR #3357

### Summary

Fixes a silent contract violation: `execInEnvironment` accepted `opts.env` and `opts.cwd` in its JSDoc but destructured only `{ cmd, timeout }`, silently dropping both fields. The fix adds the missing destructuring and wires them to `docker exec --workdir` and `--env KEY=VAL` flags. Six new unit tests cover the happy paths plus a process.env isolation regression.

---

### Strengths

- `execFile` (array args, no shell) is used throughout — `--workdir` and `--env KEY=VAL` are passed as separate argv elements, so spaces, quotes, and shell metacharacters in values are handled correctly without any quoting logic needed.
- The no-allowlist design is sound. `EnvironmentManager` never calls `execInEnvironment` today (confirmed: 0 hits in environment-manager.js), and the method receives a caller-constructed env object rather than `process.env`. The JSDoc makes the caller-responsibility contract explicit.
- `types.js` interface JSDoc was updated alongside the implementation — opts.env and opts.cwd now appear with correct `@param` tags.
- The process.env isolation test (test 6) is particularly valuable: it pollutes `process.env` with a sentinel variable, calls `execInEnvironment`, and asserts the sentinel never reaches the docker args.
- The "GREETING=hello world" test implicitly verifies space safety — the value travels through execFile's argv array without shell-splitting.

---

### Issues Found

#### Critical (Must Fix)

None.

---

| `docker.js` | `env` loop | Add `String(val)` coercion: `` `${key}=${String(val)}` `` | Template literal coercion already works for numbers/booleans (e.g. `{ PORT: 3000 }` → `PORT=3000`), but the JSDoc types opts.env as `Object` with no value type constraint. Explicit `String(val)` signals the intent clearly and future-proofs against a caller passing `null`/`undefined` (which would silently produce `KEY=null` or `KEY=undefined` and be hard to debug). |
| `types.js` | `execInEnvironment` JSDoc | Narrow `@param {Object} [opts.env]` to `@param {Object.<string,string>} [opts.env]` | Matches actual usage (string key→value pairs) and surfaces the no-null constraint to IDE tooling. |
| `docker.test.js` | New test suite | Add a test for `opts.cwd` containing a space (e.g. `'/work space/src'`) | Spaces in workdir paths are handled correctly by execFile's argv array, but having a test documents and protects this behavior explicitly. Without it, a future refactor that adds naive string interpolation would go undetected. |

---

#### Nitpicks

- The `if (env)` guard passes for an empty object `{}`, which produces zero `--env` flags — this is correct behavior, but a brief comment like `// empty env object → no flags` would make the intent obvious to a future reader.
- The inline JSDoc on `docker.js` is more detailed than the interface JSDoc in `types.js` (the latter doesn't mention the no-allowlist policy). Mentioning it briefly in the `types.js` `@param` description (e.g., "no allowlist applied — caller is responsible for passing only required vars") would help readers understand the design contract without needing to trace to the implementation file.

---

| Add `String(val)` coercion + JSDoc type narrowing for opts.env | https://github.com/blamechris/chroxy/issues/3358 | Defensive, not a correctness fix — no current caller passes non-string values. Low risk, worth tracking. |
| Add space-in-cwd test | https://github.com/blamechris/chroxy/issues/3359 | Gap in test coverage, not a bug. Safe to add in a follow-up. |

---

### Architecture Notes

`execInEnvironment` sits below `EnvironmentManager` in the stack — it's a low-level shellout helper, not an orchestration method. The no-allowlist decision is correct at this layer: EnvironmentManager constructs explicit env objects for specific operations (install steps, one-off commands) and is not a conduit for session environment passthrough. The design stays consistent with the existing pattern where `streamCliInEnvironment` owns the allowlist concern for the long-lived Claude Code process.

---

### Verdict

**Approve** — implementation is correct, safe, and well-tested. The `String(val)` coercion and type-narrowing suggestions are worth following up but do not block merge.

blamechris:
**Follow-up issue links (see main review above):**

- String(val) coercion + JSDoc narrowing: https://github.com/blamechris/chroxy/issues/3361
- Space-in-cwd test: https://github.com/blamechris/chroxy/issues/3362

PR #3358, opened 2026-05-05
docs(server): renameEnvironment no-op contract for non-Docker backends (#3313)

blamechris:
Closes #3313

## Summary

- Expands the `renameEnvironment` JSDoc in `types.js` to explicitly document that this is a Docker-only concept. Non-Docker backends that do not use mutable container names (e.g. K8sBackend) MUST implement it as a no-op — the restore flow calls it unconditionally, so a rejection aborts the restore.
- Implements the no-op in `K8sBackend`: replaces `Promise.reject(notImplemented(...))` with `Promise.resolve()` and adds an inline comment explaining why.
- Removes `renameEnvironment` from `NOT_IMPLEMENTED_REASON` (it is no longer deferred).
- Updates the K8sBackend Phase-1 stubs test: removes `renameEnvironment` from the rejects-with-NotImplementedError table and adds a dedicated `doesNotReject` assertion.

## Test plan

- [ ] `K8sBackend Phase-1 stubs` suite passes (7 tests: 6 stubs still reject, 1 new no-op assertion resolves)
- [ ] No other test failures introduced


blamechris:
## Code Review: PR #3358

### Summary

This PR documents and enforces the `renameEnvironment` no-op contract for non-Docker backends. The core changes are three coupled edits: (1) expanded JSDoc in `types.js`, (2) K8sBackend implementation changed from `Promise.reject(notImplemented(...))` to `Promise.resolve()`, and (3) the K8sBackend Phase-1 stubs test updated to assert `doesNotReject` instead of `rejects`. There is also an unrelated test in `packages/dashboard/src/store/message-handler.test.ts` that was already committed on the `test/3309-cross-author-collision-test` branch (PR #3351) and appears in this PR's diff only because that branch was used as the base here.

---

### Strengths

- **Scope expansion is defensible.** The issue title says "docs" but the real problem was a divergence between documented intent and live code. `environment-manager.js:260` calls `renameEnvironment` unconditionally with no catch — a `NotImplementedError` rejection from K8sBackend would have silently aborted the entire restore flow. Aligning the code to match the documented "no-op for non-Docker backends" intent is the correct fix, not a scope creep.
- **No callers rely on the throw.** `environment-manager.js` calls `renameEnvironment` without a try/catch and without branching on success vs. failure — it already swallowed errors at the `restore()` level. `backend-seam.test.js` mocks the method to return `Promise.resolve()`, which is consistent. No real caller expected a `NotImplementedError`; anyone who did would already have seen the restore abort. Changing to `Promise.resolve()` is safe.
- **`doesNotReject` assertion is correct.** The new test verifies the actual production behavior rather than a stubbed behavior that was already known to be wrong.
- **Comment quality is good.** The inline comment in `k8s.js` and the `NOT_IMPLEMENTED_REASON` note explain the architectural reasoning concisely. The extended JSDoc in `types.js` gives implementors of future backends unambiguous guidance.
- **`_newName` parameter kept (not removed).** K8s could theoretically tag pods in future — keeping the signature aligned with the Backend interface is correct.

---

### Issues Found

#### Critical (Must Fix)

None.

| `types.js` JSDoc | `@param` description | The phrase "or ignored (K8s / no-op backends)" is accurate but slightly misleading — parameters are not "ignored" in the JS sense; they just aren't used. Consider "unused on no-op backends" for precision. | Minor clarity nit, but JSDoc is the contract spec here. |
| `k8s.test.js` | New `doesNotReject` test | The test asserts only that the promise resolves — it does not assert it resolves with `undefined` (i.e. `Promise.resolve()` with no value). If a future implementor accidentally returns a value, the test won't catch it. Consider `assert.strictEqual(await backend.renameEnvironment('pod-id', 'new-name'), undefined)` instead. | `Promise<void>` contract per the Backend interface. |

#### Nitpicks

- The `NOT_IMPLEMENTED_REASON` comment in `k8s.js` line 32-33 now reads "Keep this in sync with the Backend interface (types.js)" — since `renameEnvironment` was intentionally removed from this map, it may be worth appending "methods listed here are stubs; no-op methods are handled separately" to prevent future confusion.
- The PR title says `docs(server)` but the change also modifies runtime behavior (`k8s.js`). A more accurate type would be `fix(server)` or `fix(server,docs)`. Cosmetic; does not affect correctness.

---

### Unrelated Diff: `message-handler.test.ts`

The diff includes a change to `packages/dashboard/src/store/message-handler.test.ts` (the "cross-author collision" test for `skill_trust_granted`). This is commit `82c598328` from the `test/3309-cross-author-collision-test` branch, which is already tracked separately in PR #3351. It appears in this PR's diff because PR #3358's branch was cut from that branch's tip rather than from `origin/main` directly.

**Action needed:** Before merging, rebase `docs/3313-rename-environment-no-op-contract` onto `origin/main` (or wait for #3351 to merge first). As-is, squash-merging this PR will include the `message-handler.test.ts` change in its squash commit, which will conflict with or duplicate #3351.

This is a **blocker** only if #3351 has not yet been merged when this PR lands.

---

### Architecture Notes

`environment-manager.js` is the only production caller of `renameEnvironment`. The call site at line 260 has no error-handling wrapper — it lets a rejection propagate up and abort `restore()`. The backend-seam test confirms this is intended: the test mock already returns `Promise.resolve()`, so the seam contract was always "resolves or no-op." The K8sBackend behavior change closes the gap between the contract and the implementation.

---

### Verdict

**Approve with one action item:** resolve the `message-handler.test.ts` overlap with PR #3351 before merging (rebase onto `origin/main` after #3351 lands, or merge #3351 first). The three substantive changes — JSDoc, K8s implementation, and K8s test — are correct and complete.

The scope expansion from "docs only" to "docs + behavior fix" is the right call. A JSDoc that says "K8s MUST be a no-op" while the code throws `NotImplementedError` would be worse than the issue title implies — it would be a silent restore abort on the production K8s path.

No follow-up issues created — the two suggestions above are minor and do not warrant tracked issues at this stage.

PR #3363, opened 2026-05-05
fix(server): K8sBackend agentToken survives server restart via Secret lookup (#3339)

### LINKED ISSUE #3339: fix(server): persist K8s agentToken across server restart
## Context

Identified during review of PR #3331 (K8sBackend streaming exec, #3320), Copilot review comment 3190679220.

After PR #3331's fixes, K8sBackend stores per-Pod agent tokens in an internal `_agentTokens` Map registered by `createEnvironment()` and consulted by `streamCliInEnvironment()`. This makes the in-process flow work without callers having to thread the token through opts.

However the token is held only in memory. After a chroxy server restart:

- The Pod is still running in the cluster (its Secret has the token).
- But the K8sBackend's `_agentTokens` map is empty.
- The next `streamCliInEnvironment()` call throws `agentToken is required`.

## Required behavior

On `EnvironmentManager.reconnect()` for K8s-backed environments, the token must be repopulated. Two acceptable patterns:

1. Persist the token in the environment record (encrypt at rest) and re-register on reconnect.
2. Read the K8s Secret directly via `readNamespacedSecret({ name: chroxy-token-<envId> })` and re-register on reconnect.

(2) is preferred — keeps the token canonically in the cluster, no encryption-at-rest concerns on the chroxy side.

## Acceptance Criteria

- [ ] EnvironmentManager.reconnect() repopulates K8sBackend._agentTokens for K8s-backed environments
- [ ] Test: after reconnect(), streamCliInEnvironment() succeeds without an explicit opts.agentToken

## Original Review Comment

> This method now returns backend-specific state that the current EnvironmentManager never persists: it still stores only containerId/containerCliPath, getContainerInfo() exposes only those fields, and teardown still calls destroyEnvironment(env.containerId) with no secretName. In practice that means K8s environments created via the shared manager cannot supply the required agentToken for later execs and will leak the per-pod Secret on destroy.

PR #3331 has addressed:
- Per-Pod Secret cleanup on destroy (auto-derived in K8sBackend; no caller plumbing needed)
- agentToken auto-registration on create (no caller plumbing needed)
- Stable single-process flow

What remains: token survival across server restart.

blamechris:
Closes #3339

## Summary

After a chroxy server restart the Pod is still running in the cluster, but `K8sBackend._agentTokens` is an empty in-memory Map. The next `streamCliInEnvironment()` call threw `agentToken is required`, making all K8s sessions unreachable until the environment was re-created.

## Approach (option 2 from the issue — Secret as source of truth)

The K8s Secret (`chroxy-token-<envId>`) was already the canonical storage for the token. The `_agentTokens` Map is just a cache. This PR makes the cache miss path transparent:

1. **`K8sBackend._readAgentToken(podName, ns)`** — reads `readNamespacedSecret`, decodes the base64 value, caches it in `_agentTokens`, returns `null` on 404.

2. **`streamCliInEnvironment`** — token resolution now follows `opts.agentToken || _agentTokens.get(podName) || _readAgentToken(podName, ns)`. The entire chain is wrapped in `Promise.resolve()` so the synchronous fast-path (cache hit) has identical latency to before; only the Secret fetch adds a round-trip (once per pod per process lifetime after it's cached).

3. **`K8sBackend.reconnectAgentToken(podName)`** — explicit warm-up hook for EnvironmentManager.reconnect(). Returns `true`/`false` based on whether the Secret exists.

4. **`EnvironmentManager.reconnect()`** — calls `reconnectAgentToken` (when the backend supports it) for each environment after restoring status, so the token cache is pre-populated before any session reconnects.

## Restart recovery story

1. Server restarts; `_agentTokens` is empty.
2. `EnvironmentManager.reconnect()` iterates persisted envs, calls `getEnvironmentStatus()` (Pod is Running), then calls `K8sBackend.reconnectAgentToken('chroxy-env-<id>')`.
3. `reconnectAgentToken` reads `chroxy-token-<id>`, decodes the token, inserts into `_agentTokens`.
4. Subsequent `streamCliInEnvironment()` call hits the cache — zero extra latency.

If `reconnectAgentToken` is not called (e.g. a very fast reconnect before `reconnect()` runs), the lazy-fetch in `streamCliInEnvironment` handles it transparently on the first call.

## Tradeoffs

- The change from synchronous throw to async exit(-1) on the missing-token path is a minor semantic change. Three existing tests that expected `assert.throws` were updated to check for `exit(-1)` instead. No external API surface changed.
- A non-404 API error (e.g. Forbidden) during `_readAgentToken` propagates as a rejected promise, causing exit(-1) and a `warn` log. This is intentional — it matches the existing `streamCliInEnvironment` dial-failure behavior.

## Tests

- `K8sBackend._readAgentToken()` — 5 cases: cache-and-return, 404 returns null, non-404 throws, empty Secret field, single API call per fetch
- `K8sBackend.reconnectAgentToken()` — 3 cases: Secret found (true), 404 (false), non-404 error propagates
- `K8sBackend agentToken lazy-load after server restart (#3339)` — 3 cases: lazy fetch succeeds, Secret-not-found → exit(-1), post-reconnect cache hit
- `EnvironmentManager.reconnect() — reconnectAgentToken delegation (#3339)` — 3 cases: called for each env, error absorbed, skipped for backends without the method (Docker)

All 90 K8s tests and 78 environment-manager tests pass.


blamechris:
## Code Review: PR #3363

### Summary

Fixes the K8s session reconnect bug (#3339): after a server restart the in-memory `_agentTokens` Map is empty, so the next `streamCliInEnvironment` call threw synchronously. This PR makes the K8s Secret the canonical source of truth and the in-memory Map a cache, with a lazy-fetch fallback and an explicit warm-up hook for `EnvironmentManager.reconnect()`.

### Strengths

- Clean layering: `opts.agentToken → _agentTokens cache → _readAgentToken` is easy to follow and handles all three callers (test, happy-path, restart).
- 404 is correctly **not** cached — if a pod is recreated with a new token, the next fetch will return the new value rather than a stale null.
- No token value in any log line — only pod name and secret name are logged.
- `EnvironmentManager.reconnect()` absorbs `reconnectAgentToken` errors with a `warn` instead of letting a single bad token stop all other environments from reconnecting.
- Test coverage is thorough for the new units: 5 `_readAgentToken` cases, 3 `reconnectAgentToken` cases, 3 lazy-load scenarios, 3 delegation cases.
- Three updated tests correctly reflect the async → exit(-1) behavioural change.
- base64 roundtrip is correct: `randomBytes(32).toString('base64url')` → stored as `stringData` → K8s API re-encodes to `.data` as padded base64 → `Buffer.from(raw, 'base64').toString('utf8')` recovers the original base64url string (Node's base64 decoder accepts base64url alphabet).

### Issues Found

| `k8s.js` | `_readAgentToken` | Method always calls `readNamespacedSecret`, never checks `_agentTokens` first. Two concurrent `streamCliInEnvironment` calls on an empty cache both reach `_readAgentToken` before either populates the map, resulting in two API reads for the same Secret. | Dedup at `_readAgentToken` level (check map first) or coalesce with an in-flight promise. Benign today (idempotent write), but wastes an API call. Tracked in #3371. |
| `types.js` | Backend interface | `reconnectAgentToken` is not documented as an optional method. `EnvironmentManager.reconnect()` duck-types it with `typeof … === 'function'`. | Documents the contract for future backend implementers and makes the capability discoverable. Tracked in #3372. |

#### Nitpicks

- `reconnectAgentToken` is called even when `env.status` ended up as `'stopped'` or `'error'` after `getEnvironmentStatus`. This is correct for K8s (Secret may outlive the Pod), but a comment explaining the intent would help future readers.
- `_readAgentToken` comment says "caching is used by streamCliInEnvironment (|| tokenOrPromise chain)" — accurate, but a reader scanning `_readAgentToken` alone could reasonably expect the method itself to be cache-aware. Minor.

| Deduplicate concurrent Secret fetches in `_readAgentToken` | [#3371](https://github.com/blamechris/chroxy/issues/3371) | Benign (idempotent write, low frequency). Correctness is unaffected; optimisation only. |
| Document `reconnectAgentToken` as optional in `types.js` | [#3372](https://github.com/blamechris/chroxy/issues/3372) | Pure documentation. No behaviour change needed. |

### Architecture Notes

The approach (Secret = canonical store, Map = cache) is the right call. The alternative (re-writing the token to persistent storage on the chroxy side) would duplicate K8s state and complicate rotation. The lazy-fetch path is confined to `_readAgentToken` and does not affect any other call sites.

The duck-typing check (`typeof this._backend.reconnectAgentToken === 'function'`) is pragmatic given the Backend interface is plain JS. It's consistent with how the codebase handles other backend-specific extensions (e.g. `createComposeEnvironment`).

### Verdict

**Approve — ready to merge.**

Both deferred items are non-blocking: the concurrent-fetch issue is benign (idempotent write, negligible frequency in production), and the `types.js` gap is documentation only. The core fix is correct, well-tested, and the base64 decode path handles the base64url ↔ base64 roundtrip correctly.

PR #3365, opened 2026-05-05
feat(dashboard): richer PendingCommunitySkill — description/source/path (#3310)

blamechris:
Closes #3310

## Summary

- Extends `PendingCommunitySkill` in `types.ts` with optional `description?: string` and `path?: string` fields
- Updates `handleSkillTrustRequest` in `message-handler.ts` to capture those fields from the wire payload — the server (`event-normalizer.js`) was already emitting them via `ServerSkillTrustRequestSchema`, the handler just wasn't capturing them
- Surfaces `description` and `path` in the SkillsPanel "Pending review" row so operators can make an informed trust decision before granting first-activation trust to a community skill author
- Empty strings on the wire are normalized to `undefined` (treated as absent)

## No server-side changes needed

The data is already on the wire. `ServerSkillTrustRequestSchema` (in `@chroxy/protocol`) already declares `source`, `description`, and `path` as required string fields. `event-normalizer.js` already populates all three. This PR is purely a dashboard-side fix to capture what was always being sent.

Note: `source` from the wire is the disk path (same as `path`), not the `'global' | 'repo'` source type on `SessionSkillInfo`. The issue requested `description` and `path`; both are now captured.

## Test plan

- [x] 3 new `message-handler.test.ts` tests: captures description + path when present, omits when absent, omits when empty strings
- [x] 4 new `SkillsPanel.test.tsx` tests: renders description element, renders path element, omits description element when absent, omits path element when absent
- [x] All existing tests pass (1411/1411)
- [x] TypeScript typecheck passes


blamechris:
## Code Review: PR #3365

### Summary

Extends `PendingCommunitySkill` with optional `description?` and `path?`, wires them through `handleSkillTrustRequest`, and renders them conditionally in the SkillsPanel pending-review row. Pure dashboard change — the data has been on the wire since `ServerSkillTrustRequestSchema` was defined. 7 new tests (4 component + 3 handler) plus one bonus handler test for cross-author collision.

---

### `source` omission — verdict: defensible

`ServerSkillTrustRequestSchema` (line 351) declares `source: z.string()`. In `event-normalizer.js` (line 169) that field is populated as `data?.source || 'global'`, which is the on-disk source type string passed through from the skill loader — not a stable semantic enum. The `'global' | 'repo'` discriminator lives on `SessionSkillInfo.source` (types.ts line 250) and is populated through a completely separate `skills_list` path. Treating the wire's `source` string as equivalent to `path` would be wrong; treating it as the `SessionSkillInfo` enum would also be wrong. Dropping it and surfacing only `description` + `path` is the correct and conservative call. The issue's own suggested-fix section only asked for those two fields.

---

### Strengths

- Defensive typing in `message-handler.ts` (`typeof msg.description === 'string' && msg.description`) is consistent with every other field extraction in that file.
- `entry.description = description` / `entry.path = path` conditional assignment (not spreading `{ description: undefined }`) keeps the serialised Zustand state clean — existing snapshots that only have `{name, author}` deserialise without `undefined` keys, which is correct for optional fields.
- `title={path}` tooltip on the path span is a good UX touch for long paths.
- Backward-compat comment in `types.ts` is accurate and useful.
- Bonus `skill_trust_granted` cross-author test is a genuine gap fill.
- `@testing-library/react` was already a dashboard devDependency — no new dep introduced.

---

### Issues Found

| `SkillsPanel.tsx` | path `<span>` | `.skill-meta` carries `margin-left: 22px` (aligns under checkbox in always-on rows). Pending rows have no checkbox so the indent is visually wrong. | See #3368 |
| `SkillsPanel.tsx` | path `<span>` | `.skill-meta` has no `overflow: hidden` / `text-overflow: ellipsis`. A long absolute path will overflow the panel; the `title` tooltip helps but the visual overflow is unhandled. | See #3368 |
| `SkillsPanel.test.tsx` | component tests | No test with both `description` and `path` set simultaneously. All four tests set one or the other. A combined test would catch layout regressions where the two elements interact. | See #3369 |

#### Nitpicks

- `{(pendingCommunitySkills!).map(...)` — the non-null assertion on line 204 is pre-existing, not introduced here. Fine to leave.
- The description `<span>` sits outside `.skill-row` (a sibling block), while the path `<span>` also sits outside. Both render as block-level via `.skill-meta` or effectively block via the parent layout. Consistent — no issue.

---

| Path span CSS: fix left-margin indent + add overflow truncation | [#3368](https://github.com/blamechris/chroxy/issues/3368) | Pure visual polish; does not affect correctness or test coverage |
| Add combined description+path SkillsPanel test; clarify empty-string JSX guard | [#3369](https://github.com/blamechris/chroxy/issues/3369) | Belt-and-suspenders; existing coverage is sufficient for the stated scope |

---

### Architecture Notes

Clean layering: schema → normalizer → handler → type → component, with no server-side changes. The optional-field approach is the right serialisation contract for a dashboard store that may hold pre-upgrade snapshots. The `skill-meta` class reuse is the one rough edge (see #3368) but is an existing CSS limitation, not a new problem introduced here.

### Verdict

**Approve** — the implementation is correct, the `source` omission is well-reasoned, backward compat is handled, and the test coverage is proportionate to the change. Two minor polish items tracked in #3368 and #3369.

PR #3366, opened 2026-05-05
fix(server): case-normalize 'community' segment in skills walk (#3301)

blamechris:
Closes #3301

## Changes

- **Walk loop** (`loadActiveSkills`): normalize the directory entry via `entry.toLowerCase()` before comparing to `'community'` when `_PATH_COMPARE_CASE_INSENSITIVE` is true (macOS HFS+/APFS and Windows NTFS). Uses the original casing to build the `communityPath` so `statSync`/`readdirSync` hit the real path on disk.
- **`_isCommunityNamespace`**: normalize `segments[0]` with `_PATH_COMPARE_CASE_INSENSITIVE ? segments[0].toLowerCase() : segments[0]` before the equality check, keeping walk and detection in sync.

Both sites now follow the same `_PATH_COMPARE_CASE_INSENSITIVE` convention already used elsewhere in the file for path-containment checks.

## Tests

Three new tests, all guarded by `{ skip: !isCaseInsensitivePlatform }` (run on darwin/win32, skipped on Linux CI):

1. `_isCommunityNamespace` — `Community/alice/x.md` (capital C) → `isCommunity: true`
2. `_isCommunityNamespace` — `COMMUNITY/alice/x.md` (all-caps) → `isCommunity: true`
3. Walk integration — creates a real `Community/alice/caps.md` tree on a temp dir, asserts skill is discovered with `trustState: 'trusted'` and `communityAuthor: 'alice'`

All 188 existing skills-loader tests continue to pass.

### DIFF
--- a/packages/server/src/skills-loader.js
+++ b/packages/server/src/skills-loader.js
@@ -386,10 +386,11 @@ export function loadActiveSkills(dir, opts = {}) {
   // itself and each author-level subdir to mirror the top-level entry filter.
   for (const entry of entries) {
     if (typeof entry !== 'string') continue
-    if (entry !== 'community') continue
-    // `entry` is the literal string 'community'. Verify it is actually a
-    // directory (not a file named 'community.md' would never reach here, but
-    // a file named 'community' might).
+    const entryCmp = _PATH_COMPARE_CASE_INSENSITIVE ? entry.toLowerCase() : entry
+    if (entryCmp !== 'community') continue
+    // `entry` is the (case-normalized) string 'community'. Verify it is
+    // actually a directory (not a file named 'community.md' would never reach
+    // here, but a file named 'community' might).
     const communityPath = join(dir, entry)
     let communityDirSt
     try {
@@ -1676,7 +1677,8 @@ export function _isCommunityNamespace(realPath, dirReal) {
   if (!rel || rel.startsWith('..')) return { isCommunity: false, author: null }
   const segments = rel.split(sep)
   if (segments.length < 3) return { isCommunity: false, author: null }
-  if (segments[0] !== 'community') return { isCommunity: false, author: null }
+  const seg0 = _PATH_COMPARE_CASE_INSENSITIVE ? segments[0].toLowerCase() : segments[0]
+  if (seg0 !== 'community') return { isCommunity: false, author: null }
   const author = segments[1]
   if (!author || author === '.' || author === '..' || author.startsWith('.')) {
     return { isCommunity: false, author: null }
--- a/packages/server/tests/skills-loader.test.js
+++ b/packages/server/tests/skills-loader.test.js
@@ -2570,6 +2570,31 @@ describe('skills-loader', () => {
         assert.equal(_isCommunityNamespace(null, '/skills').isCommunity, false)
         assert.equal(_isCommunityNamespace('/skills/community/alice/x.md', null).isCommunity, false)
       })
+
+      // #3301: case-normalisation on HFS+/APFS/NTFS
+      const isCaseInsensitivePlatformCI = process.platform === 'darwin' || process.platform === 'win32'
+
+      it(
+        'detects Community/ (capital C) as community namespace on case-insensitive platforms',
+        { skip: !isCaseInsensitivePlatformCI },
+        () => {
+          const dirReal = '/skills'
+          const result = _isCommunityNamespace('/skills/Community/alice/x.md', dirReal)
+          assert.equal(result.isCommunity, true)
+          assert.equal(result.author, 'alice')
+        },
+      )
+
+      it(
+        'detects COMMUNITY/ (all-caps) as community namespace on case-insensitive platforms',
+        { skip: !isCaseInsensitivePlatformCI },
+        () => {
+          const dirReal = '/skills'
+          const result = _isCommunityNamespace('/skills/COMMUNITY/alice/x.md', dirReal)
+          assert.equal(result.isCommunity, true)
+          assert.equal(result.author, 'alice')
+        },
+      )
     })
 
     // ── Loader integration tests ──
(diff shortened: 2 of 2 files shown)


blamechris:
## Code Review: PR #3366

### Summary

Fixes a user-visible gap on macOS/Windows: directories named `Community/` or `COMMUNITY/` under the skills root were silently skipped because the walk loop used a strict string equality check. The fix applies `_PATH_COMPARE_CASE_INSENSITIVE` (already in use for path containment checks) at both the walk gate and inside `_isCommunityNamespace`, keeping the two sites in sync. Disk operations always use the original casing, which is correct. Three platform-guarded integration tests verify the fix end-to-end.

---

### Strengths

- **Surgical change** — exactly two lines changed in production code; the fix is local and easy to audit.
- **Consistent with existing conventions** — re-uses `_PATH_COMPARE_CASE_INSENSITIVE` instead of inventing a new constant, keeping the approach uniform with `_pathContains` and `_resolveRoots`.
- **Original casing preserved for disk ops** — `const communityPath = join(dir, entry)` still uses the raw `entry`, not `entryCmp`. `statSync`, `readdirSync`, and `realpath` all receive the real on-disk name. No regression risk.
- **`_isCommunityNamespace` consistency** — normalizes `segments[0]` with the same constant, so the walk loop and the namespace classifier cannot diverge.
- **Walk `break` is intact** — after finding the first matching `community`-like entry the loop breaks. On a case-sensitive FS where both `community/` and `Community/` exist as distinct directories, only the first directory-entry-order winner is processed (same as before for lowercase). Acceptable for v1.
- **Two-pass path unaffected** — `_collectCandidates` consumes the `entries` array that the walk loop has already populated with pre-resolved `{ entry, fullPath }` objects; it does not re-scan for `'community'` itself, so no second fix site is needed there.
- **Good test coverage** — the three new tests are real filesystem tests (not just string mocks), covering both the unit-level `_isCommunityNamespace` and the full `loadActiveSkills` integration path.

---

### Issues Found

#### Critical (Must Fix)

None.

| `skills-loader.js` | `_isCommunityNamespace` JSDoc | Update the Rules bullet "First segment … must be **exactly** `'community'`" to reflect that the match is case-insensitive on darwin/win32 | The doc now says the opposite of the code on those platforms; future readers will be confused |
| `skills-loader.js` | `_PATH_COMPARE_CASE_INSENSITIVE` constant | Add a one-line comment flagging it as a platform-default **heuristic** (not a filesystem probe) | Prevents a future maintainer from treating it as a filesystem fact — see deferred issues below |

#### Nitpicks

- The walk comment still reads `// entry is the (case-normalized) string 'community'` — minor: `entry` is the *original* casing; `entryCmp` is the normalized form. Could read `// entryCmp is 'community'; entry retains original casing for disk ops` for precision.
- `isCaseInsensitivePlatformCI` is a more specific name than needed for a local `const`; `isCaseInsensitivePlatform` (as used in the existing `$HOME boundary` test) would be consistent.

---

### Deferred Items

| Document that `_PATH_COMPARE_CASE_INSENSITIVE` / `_isCaseInsensitiveFs()` are OS heuristics with known non-conforming configs (case-sensitive APFS, ZFS case-insensitive on Linux); update `_isCommunityNamespace` JSDoc to match | [#3376](https://github.com/blamechris/chroxy/issues/3376) | Pure documentation/comment update; no user-visible behavior change; out of scope for this bug fix |
| Add platform-gated tests asserting that `Community/` is NOT a community namespace on Linux (`isCommunity: false`), and document the trust-transfer gap when migrating `skills-trust.json` from macOS to Linux | [#3378](https://github.com/blamechris/chroxy/issues/3378) | New test scope; Linux CI environment needed to run non-skipped; deferred to avoid scope creep |
| Add user-facing documentation noting that `community/` must be lowercase for cross-platform portability, and that `skills-trust.json` path keys are platform-cased (lowercased on mac/win, verbatim on linux) | [#3379](https://github.com/blamechris/chroxy/issues/3379) | Documentation-only; the behavior is working as designed, the gap is a known quirk worth documenting separately |

---

### Focused Analysis of Raised Questions

**Is `_PATH_COMPARE_CASE_INSENSITIVE` the right constant?**  
It's the right *existing* constant to use — applying a new one would be inconsistent. The name is an accurate description of the intent ("compare paths case-insensitively on this platform"). The underlying limitation is that it is an OS heuristic, not a filesystem probe. That limitation pre-dates this PR and affects every other callsite equally. Using it here is the correct choice for consistency; the documentation gap is tracked in #3376.

**Better approach: runtime filesystem probe?**  
A runtime probe (create temp file, stat with altered case) would be more accurate but adds startup I/O and complexity for an edge case that affects a tiny fraction of users (case-sensitive APFS, ZFS case-insensitive on Linux). Not worth the trade-off in v1. Worth reconsidering if user reports come in; #3376 captures this as a future option.

**Always normalize to lowercase regardless of OS?**  
This would fix the Linux insensitive-mount case but would break the explicit Linux case-sensitive contract: on a standard ext4 system `community/alice` and `Community/alice` are legitimately different directories. Silently treating them as the same would be surprising and harder to reason about than the current OS heuristic.

**Potential surprise on Linux: two different trust namespaces.**  
Confirmed: on Linux, `community/alice` and `Community/alice` produce different `author` strings from `_isCommunityNamespace`. The trust store `byAuthor` lookup is a plain object key comparison — `byAuthor['alice']` vs `byAuthor['alice']` (same if both directories have the same author segment) but if a user names their directory `Community` on Linux, `_isCommunityNamespace` returns `isCommunity: false` because `seg0` is `'Community'` not `'community'`. The skill is completely excluded from the community namespace, not trusted as a separate one. This is documented as a known quirk in #3379.

**Trust transfer: macOS → Linux.**  
The trust store `byPath` keys are lowercased on macOS (`_normalizePathKey` → lowercase) but verbatim on Linux. A ledger written on macOS with key `/home/user/.chroxy/skills/community/alice/foo.md` (lowercased) will not match the verbatim path on Linux if the user's path contains uppercase segments. The `byAuthor` index (`alice`) is not case-normalized and is portable as long as the author directory name is the same string — so author-level grants survive the migration. Path-level grants do not. Tracked in #3379.

**`_isCommunityNamespace` casing consistency.**  
`seg0` is normalized before the equality check; `segments[1]` (the author) is returned verbatim. This is intentional and correct: the author name comes from the filesystem and is passed to the trust checker and stored as `communityAuthor` — it should preserve the real on-disk casing so `readdirSync` and trust lookups can use it directly. No inconsistency here.

---

### Architecture Notes

The change is entirely within the skills-loader subsystem. It does not touch WS protocol, session management, or the trust store grant flow. The two fixed sites (`loadActiveSkills` walk loop and `_isCommunityNamespace`) are the only places where `'community'` appears as a string comparison, so the fix is complete.

---

### Verdict

**Approve — ready to merge.**  
The fix is correct, minimal, and consistent with existing conventions. The two in-code documentation gaps (JSDoc + constant comment) are trivial and tracked in follow-up issues. No behavior regressions introduced.

PR #3367, opened 2026-05-05
feat(server): K8sBackend native imagePullPolicy option (#3343)

blamechris:
Closes #3343

## Summary

- Adds a first-class `imagePullPolicy` option to `K8sBackend` — accepted at construction time (applies to all environments) and as a per-call override on `createEnvironment` (takes precedence over the constructor value)
- When neither is set the field is omitted from the Pod spec, leaving Kubernetes to apply its own default behaviour (`Always` for `:latest` tags, `IfNotPresent` otherwise)
- Removes the `_api.createNamespacedPod` monkey-patch from the kind-based integration test, replacing it with the native `imagePullPolicy: 'IfNotPresent'` constructor option

## Changes

**`packages/server/src/environments/backends/k8s.js`**
- Constructor destructures `imagePullPolicy`, stores as `this._imagePullPolicy`
- `createEnvironment` destructures `opts.imagePullPolicy` (aliased as `callImagePullPolicy`)
- Resolves effective policy: per-call opt > constructor opt > omit
- Builds `containerSpec` object and conditionally sets `containerSpec.imagePullPolicy` before assembling the Pod manifest

**`packages/server/tests/environments/backends/k8s.test.js`**
- 4 new unit tests: default (absent), constructor-set, per-call-set, per-call-overrides-constructor

**`packages/server/tests/integration/k8s-sidecar-roundtrip.test.js`**
- Replaces the 12-line `_api.createNamespacedPod` monkey-patch with `imagePullPolicy: 'IfNotPresent'` on the `K8sBackend` constructor

## Test plan

- [ ] `node --test packages/server/tests/environments/backends/k8s.test.js` — 83 pass, 0 fail
- [ ] All 4 new `imagePullPolicy` unit tests pass
- [ ] Integration test (`RUN_K8S_INTEGRATION=1`) no longer monkey-patches the API


blamechris:
## Code Review: PR #3367

### Summary

Adds a first-class `imagePullPolicy` option to `K8sBackend` — accepted at construction time (applies to all Pods) and as a per-call override on `createEnvironment` (takes precedence). Omitting it entirely leaves the field absent from the Pod manifest, deferring to Kubernetes' own defaulting behaviour (`Always` for `:latest`, `IfNotPresent` otherwise). Replaces the 12-line `_api.createNamespacedPod` monkey-patch in the integration test with the new native option. 4 new unit tests cover all four combinations (absent, ctor-set, call-set, call-overrides-ctor).

---

### Strengths

- **Correct backwards-compatible default.** Omitting `imagePullPolicy` from the Pod spec entirely (rather than setting a hard default) is the right call. Users who never set the option get unchanged K8s behaviour — no surprise `Always` forcing a pull for non-`:latest` images, and no surprise `IfNotPresent` silently masking an outdated image for `:latest` tags.
- **Integration test monkey-patch removal is clean.** The old patch mutated `backend._api.createNamespacedPod` directly — fragile against client-library changes and invisible to the reader. Replacing it with `imagePullPolicy: 'IfNotPresent'` on the constructor is both more readable and exercises the real production code path.
- **JSDoc union type `'Always'|'IfNotPresent'|'Never'` documents valid values** at both the constructor and `createEnvironment` levels — a clear contract for IDEs and reviewers.
- **`containerSpec` extraction from the diff.** The PR's diff shows a pre-existing inline object being extracted to `containerSpec` before the `imagePullPolicy` conditional is applied. The final shape (as seen in `k8s.js` line 318–346) is a clean build-up pattern consistent with the existing `resources` and `volumeMounts` conditionals.
- **Per-call override respects constructor value correctly.** `callImagePullPolicy || this._imagePullPolicy` is the right resolution order; the falsy-check is appropriate because the valid values are all truthy strings (no valid value is an empty string or `0`).
- **4 unit tests are well-targeted.** Absent / ctor-set / call-set / call-overrides-ctor — all four meaningful states are exercised, including the `hasOwnProperty` assertion for the absent case (not just `=== undefined`, which could hide `enumerable: false` scenarios).
- **Dashboard test addition (`message-handler.test.ts`) is a clean addition** of a missing edge-case for `skill_trust_granted` with two same-name skills from different authors.

---

### Issues Found

#### Critical (Must Fix)

None.

| `k8s.js` constructor | ~88 | Validate `imagePullPolicy` against `['Always', 'IfNotPresent', 'Never']` and throw immediately on unrecognised value | A typo like `'IfNotpresent'` passes silently through to the K8s API which returns an opaque 422. A one-liner guard gives an actionable error message at construction time. Tracked in #3375. |
| `types.js` | — | Add `opts.imagePullPolicy` to `createEnvironment` JSDoc with a K8s-specific note ("honoured only by K8sBackend") | `EnvironmentManager` passes opts straight through; callers who configure the manager should be able to discover this option from the interface definition, not just by reading `k8s.js`. Tracked in #3377. |

#### Nitpicks (Optional)

- The constructor comment added at line ~86 of the diff (`* @param {'Always'|'IfNotPresent'|'Never'} [opts.imagePullPolicy]`) is placed _before_ the `@param {'portforward'|'clusterip'}` line in the source but _after_ `sidecarImage` in the existing alphabetical run — consistent with the source order, no action needed.

---

| Runtime validation of `imagePullPolicy` value against K8s enum | [#3375](https://github.com/blamechris/chroxy/issues/3375) | Purely additive guard — doesn't affect the correctness of this PR's primary goal (the option works correctly for valid values). Reasonable to keep scope small. |
| Add `opts.imagePullPolicy` to `createEnvironment` opts JSDoc in `types.js` | [#3377](https://github.com/blamechris/chroxy/issues/3377) | Documentation-only change, no behaviour impact. Better handled in a docs-focused pass alongside any other Backend interface doc gaps. |

---

### Architecture Notes

This change fits cleanly into the existing `K8sBackend` pattern: constructor-level defaults (like `sidecarImage`, `connectMode`) that can be overridden per-call (like `namespace`, `image`). The implementation mirrors how `namespace` and `image` overrides already work. No other components are affected — `EnvironmentManager` passes opts through verbatim, so this opt is immediately usable by callers without manager-level changes.

The dashboard `message-handler.test.ts` change appears in this PR because the branch was cut after the recent `skill_trust_granted` work was merged; it adds a missing edge-case test unrelated to K8s. It is a net positive addition.

---

### Verdict

- [x] **Approve** — Ready to merge. The primary goal is fully achieved: the monkey-patch is gone, the native option works correctly with proper backwards-compat default, and the unit tests are thorough. The two deferred items are purely additive improvements tracked in #3375 and #3377.

PR #3370, opened 2026-05-05
feat(server): K8sBackend createEnvironment workspace mount + resource limits

blamechris:
## Summary

- Mount `opts.cwd` as a `hostPath` volume at `/workspace` inside the Pod (satisfies the workspace visibility requirement from #3316)
- Translate `opts.memoryLimit` / `opts.cpuLimit` into `resources.limits` and `resources.requests` on the sidecar container
- Map `opts.mounts` (Docker-style `host:container[:ro]` strings) into additional `hostPath` volumes + `volumeMounts`
- Expand `opts.forwardPorts` (bare numbers or `hostPort:containerPort` strings) into extra `containerPort` entries alongside the built-in `AGENT_PORT`

## Workspace mount strategy

Uses `hostPath` with `type: DirectoryOrCreate`. Works automatically for single-node clusters (kind, minikube, Docker Desktop) where the Pod's node shares the host filesystem. For multi-node clusters operators must either ensure the path exists on the scheduled node or provision a PVC and pass it via `opts.mounts` (leaving `opts.cwd` unset). This is documented in the JSDoc.

The kind-based integration test in `tests/integration/k8s-sidecar-roundtrip.test.js` exercises this naturally since kind mounts the host filesystem into the single control-plane node.

## Memory quantity normalisation

Docker accepts `"2g"` / `"512m"` while Kubernetes expects `"2Gi"` / `"512Mi"`. `_normaliseMemoryQuantity()` converts Docker-style single-letter suffixes to K8s binary-SI equivalents. Values already using K8s suffixes are passed through unchanged.

## Test plan

- [ ] 17 new unit tests across 4 `describe` blocks:
  - `K8sBackend.createEnvironment() — workspace mount (#3316)` (2 tests)
  - `K8sBackend.createEnvironment() — resource limits (#3316)` (6 tests)
  - `K8sBackend.createEnvironment() — additional mounts (#3316)` (4 tests)
  - `K8sBackend.createEnvironment() — forwardPorts (#3316)` (5 tests)
- [ ] All 96 K8s unit tests pass (0 failures)
- [ ] `eslint src/environments/backends/k8s.js` — clean

Closes #3316

### DIFF
--- a/packages/server/src/environments/backends/k8s.js
+++ b/packages/server/src/environments/backends/k8s.js
@@ -157,9 +157,32 @@ export class K8sBackend {
    * Returns as soon as `createNamespacedPod` resolves — the Pod has been accepted
    * by the API server but may not yet be scheduled or Running.
    *
+   * Workspace mount strategy — hostPath:
+   *   `opts.cwd` is mounted into the Pod as a `hostPath` volume at `/workspace`.
+   *   This is the simplest strategy for local clusters (kind, minikube, Docker
+   *   Desktop) where the Node running the Pod shares the host filesystem.  For
+   *   production clusters where Pods run on remote nodes the host path will not
+   *   exist; operators should provision a PVC pre-populated with the workspace and
+   *   pass it via `opts.mounts` instead.  Cluster-side requirement: the K8s node
+   *   must be able to read the path provided in opts.cwd from its local filesystem.
+   *
    * @param {Object} opts - See Backend interface in types.js
    * @param {string}   opts.envId          - Unique environment ID
+   * @param {string}   [opts.cwd]          - Host path to mount as /workspace inside the Pod
    * @param {string}   [opts.image]        - Overrides the constructor sidecarImage
+   * @param {string}   [opts.memoryLimit]  - K8s memory quantity string (e.g. "2Gi").
+   *   Applied to both `resources.limits.memory` and `resources.requests.memory`.
+   *   Accepts Docker-style suffixes ("g"/"m") and standard K8s suffixes ("Gi"/"Mi").
+   * @param {string}   [opts.cpuLimit]     - K8s CPU quantity string (e.g. "2" or "500m").
+   *   Applied to both `resources.limits.cpu` and `resources.requests.cpu`.
+   *   A plain integer or float (e.g. "2", "0.5") is valid K8s CPU quantity syntax.
+   * @param {number[]|string[]} [opts.forwardPorts] - Extra ports to expose from the container
+   *   (in addition to the built-in AGENT_PORT).  Each value may be a bare port number
+   *   or a "hostPort:containerPort" string; only the containerPort is used in the Pod spec.
+   * @param {string[]} [opts.mounts]       - Additional volume mounts in Docker-style
+   *   "hostPath:containerPath[:ro]" format.  Each entry is translated into a
+   *   `hostPath` volume + corresponding `volumeMount`.  The volume name is derived
+   *   from the entry index ("extra-vol-0", "extra-vol-1", …).
    * @param {Object}   [opts.containerEnv] - Extra environment variables
    * @param {string}   [opts.namespace]    - Overrides the constructor default namespace
    * @param {'Always'|'IfNotPresent'|'Never'} [opts.imagePullPolicy] - Per-call override for the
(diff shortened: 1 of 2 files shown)


blamechris:
## Code Review: PR #3370

### Summary

Closes #3316 — adds four features to `K8sBackend.createEnvironment` in a single coherent pass: workspace mount via `hostPath`, resource limits, additional mounts, and port forwarding. 230 LoC of implementation plus 17 new tests across 4 `describe` blocks. All 96 K8s unit tests pass; CI is green.

The refactored Pod-assembly (steps 4–8) is a clear improvement over the previous inline literal. JSDoc is thorough. The `_parseMountString` / `_parseContainerPort` / `_normaliseMemoryQuantity` helpers are well-named and have focused responsibilities.

---

### Strengths

- Clean step-by-step Pod assembly (volumes → ports → resources → containerSpec → podSpec). Much easier to follow than the previous monolithic literal.
- All new opts are optional and default-absent, so existing callers get the identical Pod spec as before — backward compat preserved.
- Deduplication of AGENT_PORT in `forwardPorts` is handled and tested.
- Graceful degradation for unparseable mounts (`log.warn` + `continue`) rather than throwing — correct for best-effort host config.
- Tests cover happy path, zero-opt baseline, `:ro` flag, combined `cwd` + `mounts`, and the `hostPort:containerPort` string format. Coverage is solid for a new feature set.
- The `_normaliseMemoryQuantity` regex correctly handles case-insensitive Docker suffixes and passes K8s-style values through unchanged.
- Code style is consistent with the rest of the file (no semicolons, single quotes, ES modules).

---

### Issues Found

#### Critical (Must Fix)

None. The implementation is correct for its stated scope.

---

| `k8s.js` | `_normaliseMemoryQuantity` JSDoc | Document the SI vs binary-SI semantic difference: `2g` (Docker, SI) = 2×10⁹ bytes; `2Gi` (K8s, binary) = 2×2³⁰ bytes — ~7.4% overshoot. The behaviour errs generous and is safe, but future maintainers should understand it's an intentional approximation. Tracked: #3384 | Without this note, a maintainer reading the code may not realise that `2g → 2Gi` is not a lossless conversion |
| `k8s.js` | `createEnvironment` JSDoc | Add an explicit security note: `hostPath` volumes require the cluster to permit them (PSA `privileged` or equivalent); they are not suitable for multi-tenant clusters. Tracked: #3387 | The current comment documents the topology requirement but not the security implication |
| `k8s.js` | `_parseContainerPort` | Add `port <= 65535` upper bound check; return `null` and `log.warn` for out-of-range values (same pattern as unparseable mounts). K8s rejects out-of-range containerPorts at the API level, but the error message will be opaque. Tracked: #3386 | Minor — K8s will catch it, but a local guard gives a clearer operator error |

---

#### Nitpicks

- **Volume index gap**: when a `mounts` entry fails to parse and is skipped, the volume names become non-sequential (`extra-vol-1` instead of `extra-vol-0`). This is harmless — K8s volume names don't need to be sequential — but if sequential naming is important, use a separate counter variable incremented only on success.
- **`_parseMountString` with Windows paths**: `"C:\\Users\\foo:/workspace"` misparses as `hostPath: "C"`. K8s pods run on Linux so this is unlikely in practice, but a comment or guard would prevent silent misconfiguration. Tracked: #3388.
- **`opts.cwd` empty-string guard**: `if (cwd)` correctly skips an empty string, which is consistent with how `EnvironmentManager` validates cwd before calling (throws if blank). No action needed, but it's worth noting that the implicit falsy guard works because callers guarantee cwd is either a non-empty string or absent.

---

| Document SI vs binary-SI memory quantity semantic difference | [#3384](https://github.com/blamechris/chroxy/issues/3384) | Safe to defer — behavior errs generous, callers can use explicit `"2Gi"` today |
| Add PVC-based workspace strategy for multi-node clusters | [#3385](https://github.com/blamechris/chroxy/issues/3385) | Scope: current PR targets single-node local clusters as stated; PVC is follow-on work |
| Validate containerPort upper bound (≤ 65535) | [#3386](https://github.com/blamechris/chroxy/issues/3386) | K8s catches it, minor defensive hygiene |
| Document hostPath privilege-escalation risk for multi-tenant clusters | [#3387](https://github.com/blamechris/chroxy/issues/3387) | Safe to defer — target audience is single-node local clusters; security note adds clarity |
| Guard against Windows-style paths in `_parseMountString` | [#3388](https://github.com/blamechris/chroxy/issues/3388) | K8s nodes are Linux — low practical risk, documentation or a comment suffices |

---

### Architecture Notes

The PR correctly narrows K8sBackend's workspace strategy to `hostPath` with explicit documentation of its topology constraint. The escape hatch (`opts.mounts` for PVC-backed volumes) is a reasonable interim answer, though note that `opts.mounts` today only supports `hostPath` entries — a PVC-backed mount cannot currently be expressed this way. Issue #3385 tracks adding a first-class PVC workspace strategy.

The `resources.limits == resources.requests` pattern (setting both to the same value) is intentional and conservative: it creates a Guaranteed QoS class Pod, which avoids OOM-kill under node pressure. This is the right choice for an agent workload.

The `containerPort` entries are informational in a Pod spec — they do not open ports, they document intent. For actual port forwarding from outside the cluster, a Service or `kubectl port-forward` is required. The JSDoc for `forwardPorts` already says "only the containerPort is used in the Pod spec", which is accurate.

---

### Verdict

**Approve** — the implementation is correct, well-documented for its stated scope, and fully tested. No blocking issues. Five follow-up issues created for deferred improvements.

PR #3373, opened 2026-05-05
feat(server): chroxy-pod-agent — sessions TTL + size cap (#3349)

blamechris:
Closes #3349

## Summary

- **Idle-resume TTL** (`CHROXY_AGENT_RESUME_TIMEOUT_MS`, default 60 s): when a WS disconnects with an active session, a per-session timer is armed. If the client does not resume within the window, the child is killed (SIGTERM → SIGKILL after 5 s) and the session is evicted. Resume before TTL cancels the timer. `agent.close()` cancels all idle timers. A late `resume` for an evicted session receives `session_lost(unknown_session)` — same as any unknown session.

- **Hard session cap** (`CHROXY_AGENT_MAX_SESSIONS`, default 8): enforced on every `spawn`. When the cap is reached the oldest idle session (by `lastActiveAt`) is evicted first; falls back to the globally oldest session if all are active. Prevents linear accumulation of orphan sessions from a client that reconnects repeatedly without resuming.

- **`lastActiveAt` tracking**: refreshed in `_emitSessionFrame` and on `resume` to power oldest-first eviction. Initialised to `Date.now()` at spawn time.

- **Deterministic timer injection**: `setTimeoutFn`/`clearTimeoutFn` constructor opts. All new tests use a `makeFakeClock()` helper — no real-clock waits.

- **PROTOCOL.md**: documents the Idle Resume Window and Hard Session Cap contracts.

## Defaults

| Env var | Default | Description |
|---------|---------|-------------|
| `CHROXY_AGENT_RESUME_TIMEOUT_MS` | `60000` (60 s) | Idle window before session eviction |
| `CHROXY_AGENT_MAX_SESSIONS` | `8` | Max concurrent sessions before LRU eviction |

## Test plan

- [x] `idle timer fires after TTL and kills the child + drops session`
- [x] `resume before TTL cancels the idle timer`
- [x] `agent.close() cancels idle timers and kills children`
- [x] `session_lost(unknown_session) is delivered to a late resume after TTL eviction`
- [x] `oldest idle session is evicted when cap is reached`
- [x] All 27 pre-existing sidecar tests still pass (34 total, 0 failures)
- [x] No regression in full server suite (74 pre-existing failures, unchanged)

### DIFF
--- a/packages/server/sidecar/PROTOCOL.md
+++ b/packages/server/sidecar/PROTOCOL.md
@@ -334,6 +334,37 @@ If the agent has no record of the `sessionId` (e.g. the agent process
 restarted), it sends `{ type: 'session_lost', sessionId, reason: 'unknown_session' }`.
 The child process is gone in this case; the session is unrecoverable.
 
+### Idle Resume Window
+
+When the WS connection drops with an active session, the agent starts an
+**idle-resume timer** (default 60 s, configurable via
+`CHROXY_AGENT_RESUME_TIMEOUT_MS`).  If the client sends a valid `resume` frame
+before the timer fires, the timer is cancelled and normal live-forwarding
+resumes.  If the timer fires first, the agent:
+
+1. Sends `SIGTERM` to the child (with a 5 s `SIGKILL` escalation).
+2. Deletes the session from `_sessions`.
+
+A subsequent `resume` for the evicted session receives
+`session_lost(unknown_session)` — the same response as for a session that was
+never created.  The eviction reason is **not** exposed to the client; consumers
+should treat both cases identically (unrecoverable, open a fresh session with
+`spawn`).
+
+The resume window is a best-effort safety net against **short** network blips.
+It is not intended as durable session persistence.
+
+### Hard Session Cap
+
+The agent limits concurrent sessions to `CHROXY_AGENT_MAX_SESSIONS` (default
+8).  When a new `spawn` would exceed the cap, the agent evicts the idle session
+(no active WS) with the oldest `lastActiveAt` timestamp before inserting the
+new one.  If all sessions are active, the globally oldest session is evicted
+(and its live WS is closed with code `1001`).
+
+This cap is defense-in-depth against a buggy K8sBackend that reconnects
+repeatedly without resuming.
+
 ### PID 1 Reaping Limitation
 
 This resume mechanism survives **network blips** (transient WS disconnects)
--- a/packages/server/sidecar/agent.js
+++ b/packages/server/sidecar/agent.js
@@ -52,6 +52,26 @@ export const DEFAULT_MAX_LINE_BYTES = Number.isFinite(_PARSED_MAX_LINE_BYTES) &&
   ? _PARSED_MAX_LINE_BYTES
   : FALLBACK_MAX_LINE_BYTES
 
+// Idle resume window: how long a disconnected session waits for the client to
+// reconnect before the child is killed and the session is evicted. Prevents
+// orphaned child processes accumulating in long-lived pods.
+// Configurable via CHROXY_AGENT_RESUME_TIMEOUT_MS (default 60 s).
+const FALLBACK_RESUME_TIMEOUT_MS = 60_000
+const _PARSED_RESUME_TIMEOUT = parseInt(process.env.CHROXY_AGENT_RESUME_TIMEOUT_MS ?? `${FALLBACK_RESUME_TIMEOUT_MS}`, 10)
+const DEFAULT_RESUME_TIMEOUT_MS = Number.isFinite(_PARSED_RESUME_TIMEOUT) && _PARSED_RESUME_TIMEOUT > 0
+  ? _PARSED_RESUME_TIMEOUT
+  : FALLBACK_RESUME_TIMEOUT_MS
+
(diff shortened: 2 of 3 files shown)


blamechris:
## Code Review: PR #3373

### Summary

Adds two safety mechanisms to `chroxy-pod-agent` to cap the `_sessions` Map and prevent in-pod memory leaks: a per-session idle-resume TTL (default 60 s) that evicts disconnected sessions if not resumed, and a hard session cap (default 8) that evicts the oldest idle — or globally oldest — session on new spawns. Timer injection (`setTimeoutFn`/`clearTimeoutFn`) keeps tests deterministic. PROTOCOL.md documents both contracts. 5 new tests across 2 describe blocks; all 34 prior tests continue to pass; all CI checks green.

---

### Strengths

- **Timer injection is idiomatic.** The `setTimeoutFn`/`clearTimeoutFn` seam is the right pattern — fake-clock tests run in microseconds with no flaky real-time waits.
- **`unref()` on idle timer.** Prevents the timer from keeping the event loop alive on a near-idle pod. Correct and easy to miss.
- **Idempotent `_startIdleTimer`.** Guard on `session.idleTimer !== null` prevents double-arming on repeated close events. Matches the existing `_killChild` idempotency pattern.
- **Idle TTL vs. K8sBackend reconnect math.** With delays `[250, 500, 1000, 2000, 4000, 8000]` and 5 retries, worst-case backoff totals ~7.75 s of timer delay plus dial time. Even accounting for 2 s portforward setup per attempt, the 60 s TTL leaves ~42 s of headroom — comfortable for mobile cellular handoff and slow Kubernetes API paths.
- **`session_lost(unknown_session)` for late resume.** Returning the same response for TTL-evicted and agent-restarted sessions keeps K8sBackend's recovery path uniform. Documented clearly in PROTOCOL.md.
- **`close()` cancels all idle timers.** Verified in `agent.close()`: `_cancelIdleTimer` is called for every session before `_killChild`. No timer leak on graceful shutdown.
- **PROTOCOL.md "Idle Resume Window" and "Hard Session Cap" sections** are clear and actionable for K8sBackend authors. The connection-stays-open note for `unknown_session` is preserved so implementors know they can spawn fresh without reconnecting.
- **Dashboard test unrelated to sidecar cleanly scoped** — the extra `skill_trust_granted` test in `message-handler.test.ts` is a legitimate companion fix, not scope creep.

---

### Issues Found

#### Critical (Must Fix)

None.

---

| `agent.js` | `_evictSession` | Send a `session_lost` (or `error`) frame before `close(1001)` when evicting an **active** session | K8sBackend maps code `1001` to `finishWithExit(-1)` — the same code used for pre-session connection failures. A live session killed by the cap is silently swallowed as a connection error. Tracked as #3390. |
| `agent.js` | `_handleSpawn` | Move `_enforceSessionCap()` to after `_spawnFn()` succeeds | If `_spawnFn` throws synchronously (ENOENT, EACCES), the function returns early with no session inserted — but the cap already evicted an existing session. Rare in practice but the invariant is worth preserving. Tracked as #3392. |

#### Nitpicks (Optional)

- `_killChild` uses the real `setTimeout` (not `this._setTimeoutFn`) for its SIGKILL grace timer. This is intentional — the grace timer is a process-safety net, not part of the session eviction logic — but a comment noting the deliberate inconsistency would help future readers.
- PROTOCOL.md Error Cases table does not list the new "session evicted by cap" close(1001) path. After #3390 adds a preceding frame this table should be updated.

---

| Active-session eviction sends silent close(1001); K8sBackend sees exit(-1) instead of exit(-2) | [#3390](https://github.com/blamechris/chroxy/issues/3390) | Behavioral change to K8sBackend exit codes; needs its own PR with contract discussion |
| `_enforceSessionCap()` called before spawn; evicts a session even when spawn will fail | [#3392](https://github.com/blamechris/chroxy/issues/3392) | Low-frequency edge case; fix requires restructuring `_handleSpawn` ordering |
| Missing test: cap eviction fallback when all sessions are active | [#3395](https://github.com/blamechris/chroxy/issues/3395) | Untested path for the most risky branch (evicting a live session); no test gap blocks merge but the coverage hole warrants follow-up |

---

### Architecture Notes

The single-client constraint (`_activeWs`) and the idle timer interact cleanly: when a disconnect fires `_cleanupConnection`, `_activeWs` is set to null and the timer is armed simultaneously. A reconnecting client sets a new `_activeWs` without hitting the "another client is already connected" guard because `_activeWs` is already null. The timer is cancelled in `_handleResume` before attaching the new WS to the session — no race between `_activeWs` admission and timer cancellation that could lead to double-eviction.

The `_enforceSessionCap` → `_evictSession` path for idle sessions also correctly calls `_cancelIdleTimer` before eviction so idle timers never fire on already-deleted sessions.

---

### Verdict

**Approve — Ready to merge.**

The core mechanisms are correct, well-tested for the primary paths, and the defaults are appropriate for single-user pod contexts. The two deferred suggestions (#3390, #3392) are genuine improvements but neither is a blocker: the silent-close issue requires K8sBackend authors to know that active-session cap eviction is possible (documented in PROTOCOL.md), and the spawn-ordering issue is a rare edge case. The missing all-active-sessions test (#3395) is a coverage gap but the fallback logic is straightforward.

PR #3374, opened 2026-05-05
test(server): de-flake K8sBackend reconnect tests with injected timer (#3350)

blamechris:
Closes #3350

## Summary

- Adds `_setTimeout` / `_clearTimeout` injection seam to `SidecarProcess` (constructor opts, defaulting to globals) and threads it through `K8sBackend` constructor → `streamCliInEnvironment` → `SidecarProcess`
- All three call-sites that touched `setTimeout`/`clearTimeout` on `proc._retryTimer` (`kill()`, `finishWithExit`, `scheduleReconnect`) now go through the injected functions
- Adds a hand-rolled `createFakeClock()` + `tickAndFlush()` pair at module scope in the test file
- The five reconnect-loop tests and the double-schedule guard test are refactored to use the fake clock — `waitTicks()` / `setTimeout(r, 30)` replaced by `clock.tick()` + one `setImmediate` yield

## Test plan

- [ ] `node --test packages/server/tests/environments/backends/k8s.test.js` → 79 pass, 0 fail
- [ ] Reconnect loop describe block runs in ~12 ms (was ~9.6 s per the issue)
- [ ] No other test files touched; pre-existing unrelated failures unchanged


blamechris:
## Code Review: PR #3374

### Summary

De-flakes the K8sBackend reconnect-loop tests by adding a `_setTimeout`/`_clearTimeout` injection seam to `K8sBackend` constructor and threading it through to `SidecarProcess`. All 6 production call-sites (`kill()`, `finishWithExit`, `scheduleReconnect`) now route through the injected functions. Test-side: a hand-rolled `createFakeClock()` + `tickAndFlush()` replaces 30ms `setTimeout` waits in 6 reconnect tests. Zero changes to non-K8s code.

---

### Strengths

- **Minimal blast radius.** Only `k8s.js` and `k8s.test.js` touched. The seam is invisible to production callers that don't pass `_setTimeout`/`_clearTimeout`.
- **All three call-sites covered.** `kill()`, `finishWithExit`, and `scheduleReconnect` all go through the injected functions — no gap left for a real timer to sneak past the fake clock.
- **`tickAndFlush` comment is accurate.** The docblock correctly explains that `tick()` alone isn't enough and a `setImmediate` yield is needed for the async `_redial()` promise chain. That's the right mental model.
- **`_reconnectDelays: [0]` is a good choice.** Passing `[0]` documents intent ("value is irrelevant") without the silent semantic shift of omitting the field.
- **Idempotency guard verified.** The double-schedule test (`error + close` on the same socket) correctly relies on the `if (proc._retryTimer) return` guard and passes with the fake clock — this was the original flake source.
- **Clean style.** No semicolons, single quotes, no TypeScript — consistent with codebase conventions.

---

### Issues Found

#### Critical (Must Fix)

None.

| `k8s.js` | line 145–146 | `setTimeout` / `clearTimeout` bare references bind to the module scope at class-definition time. In Node.js globals `globalThis.setTimeout` and `setTimeout` are the same object, so this is safe in practice — but `globalThis.setTimeout` makes the intent explicit and is immune to any future local shadowing. | Low risk now, clearer intent. |
| `k8s.test.js` | `tickAndFlush` | A single `setImmediate` yield is only guaranteed to flush one microtask checkpoint. In the "chained reconnect" path (`scheduleReconnect` → dial fails → `.catch` → `scheduleReconnect` again), a failed dial queues another timer immediately. The current tests avoid this because `_dialWs` always resolves successfully — but if a test ever exercises a failing dial mid-retry loop, one `setImmediate` won't be enough. Consider documenting this constraint in the `tickAndFlush` docblock, or making `tickAndFlush` loop until `clock.pending.size === 0`. | Documents a latent assumption before someone hits it. |
| `k8s.js` | constructor naming | The public constructor opts use `_setTimeout`/`_clearTimeout` (leading underscore, matching other injection seams like `_dialWs`, `_portForward`, `_net`). The `SidecarProcess` opts use `setTimeoutImpl`/`clearTimeoutImpl` (no underscore). The asymmetry is fine since `SidecarProcess` is a private class, but the docblock for `SidecarProcess` could note "called from `streamCliInEnvironment` — not a public API" to prevent confusion. | Cosmetic, no action required. |

#### Nitpicks

- `tickAndFlush` docblock says "yield one event-loop tick" — strictly it yields one macro-task boundary via `setImmediate`, which is slightly different from a microtask tick. The distinction only matters if someone asks "why not `await Promise.resolve()` here?" — worth one word: "macro-task boundary via setImmediate".
- The comment on line 1310 (`// delay value is irrelevant — fake clock fires immediately`) is accurate and helpful. No change needed.

---

### Focused Analysis of Requested Points

**Default fallthrough binding** — `this._setTimeout = setTimeoutImpl || setTimeout` on line 145 (and 808) refers to the bare `setTimeout` identifier in a module-scope ES module file. Node.js resolves this to `globalThis.setTimeout` at parse time; there is no `this`-binding issue because it is stored as a property value, not called as a method. Correct.

**`tickAndFlush` correctness** — `clock.tick()` drains `pending` in FIFO order then `pending.clear()` before iterating, so callbacks scheduled *during* tick don't run in the same tick (they'd be queued after `pending.clear()` returns... wait — actually `pending.clear()` is called *before* the loop, so newly-queued callbacks from inside a running callback would be added to `pending` after the snapshot was taken and would NOT run in this tick). That means a chained `scheduleReconnect` (failed dial → catch → scheduleReconnect) would leave a new timer in `pending` but not fire it. The single `setImmediate` yield after `tick()` would not drain that second timer. As noted above, existing tests don't exercise this path because dials always succeed, so it's a latent issue rather than a bug today.

**Zero-delay timer semantics** — `_reconnectDelays: [0]` calls `proc._setTimeout(fn, 0)`. `fakeSetTimeout` ignores the delay argument entirely and queues `fn` for the next `tick()`. Real `setTimeout(fn, 0)` is also async (fires after the current call stack). The fake matches the intended semantics: the callback does not run synchronously. Correct.

**Test isolation** — `createFakeClock()` returns a fresh `Map` and counter each call. Each test that calls `makeBackendWithDials()` or creates its own `clock` gets an independent instance. No shared mutable state between describe blocks. Isolation is solid.

**`session_lost` test (line 1412)** — this test calls `makeBackendWithDials(1)` (which injects the fake clock) but never calls `tickAndFlush`. That's correct: `session_lost` triggers `finishWithExit(-2)` synchronously via a WS message, not via a timer. No fake-clock interaction needed.

---

| `tickAndFlush` should document (or guard against) the chained-reconnect assumption | #3375 | No failing test today; addressing it means either doc-only change or a loop that could affect test readability. Out of scope for the de-flake PR. |

---

### Architecture Notes

The injection seam follows the established `K8sBackend` pattern exactly (`_dialWs`, `_portForward`, `_net`, `_coreV1Api`). Using `_setTimeout` (underscore prefix) at the public-constructor level and `setTimeoutImpl` (no underscore) at the `SidecarProcess` private-class level is consistent with the existing convention for the two scopes. Threading through `streamCliInEnvironment` is the right approach — it keeps `SidecarProcess` free of K8sBackend coupling while still receiving the seam.

---

### Verdict

**Approve — ready to merge.**

One follow-up issue created for the `tickAndFlush` chained-timer assumption: #3375.

blamechris:
*(Correction to review above: the follow-up issue was created as #3383, not #3375. The Deferred Items table link should read: https://github.com/blamechris/chroxy/issues/3383)*

PR #3382, opened 2026-05-05
feat(server): chroxy-pod-agent — per-spawn sentinel stderr line (#3344)

blamechris:
## Summary

- `_handleSpawn` in `sidecar/agent.js` now emits a sentinel `stderr` WS frame immediately after `session_started`, before any child output is forwarded
- Sentinel format: `[chroxy-pod-agent] spawn cmd=<cmd> args=<json-args> sessionId=<uuid>\n` — carries `seq=1`, the first sequenced output frame
- `PROTOCOL.md` documents the sentinel in the `#### stderr` section with a worked example
- `agent.test.js` adds 4 new dedicated sentinel tests and updates 5 existing tests that assumed the first data frame was child output
- `k8s-sidecar-roundtrip.test.js` collects `proc.stderr` and asserts the sentinel is present after `streamCliInEnvironment`

## Sentinel format

```
[chroxy-pod-agent] spawn cmd=claude args=["--input-format","stream-json"] sessionId=550e8400-e29b-41d4-a716-446655440000
```

As a WS frame:
```json
{ "type": "stderr", "data": "[chroxy-pod-agent] spawn cmd=claude args=...\n", "seq": 1 }
```

## Integration test update

The k8s integration test (`streamCliInEnvironment delivers events over the WS bridge`) now collects `proc.stderr` and asserts `stderrBuf.includes('[chroxy-pod-agent] spawn cmd=node')`. Requires `RUN_K8S_INTEGRATION=1` and `kind` on PATH to execute.

## Test plan

- [x] `node --test packages/server/tests/sidecar/agent.test.js` — 33 pass, 0 fail (4 new sentinel tests included)
- [x] Full server test suite — pre-existing baseline failures only (74 vs 75 on main); no regressions introduced
- [ ] K8s integration: `RUN_K8S_INTEGRATION=1 node --test packages/server/tests/integration/k8s-sidecar-roundtrip.test.js` (requires kind + Docker)

Closes #3344


blamechris:
## Code Review: PR #3382

### Summary

Adds a per-spawn sentinel `stderr` WS frame (`[chroxy-pod-agent] spawn cmd=X args=[...] sessionId=Y`) emitted synchronously after `session_started` as the first sequenced output frame (seq=1). Purpose: let integration tests prove the sidecar code path specifically handled a spawn, rather than a hypothetical shorter bypass. Also updates PROTOCOL.md, adds 4 dedicated sentinel tests, and adjusts 5 existing tests whose frame-offset assumptions changed.

### Strengths

- Placement is correct: sentinel goes through `_emitSessionFrame`, so it enters the ring buffer at seq=1, gets replayed on resume, and can be gap-detected by the overflow guard — consistent treatment with all other output frames.
- The `sessionId` in the sentinel matches `session_started.sessionId` (verified by the third new test). Clients or debuggers that want to correlate the log line get a free cross-reference with no extra message.
- `seq=1` guarantee is explicit and test-covered. A resume with `lastSeq=0` will always replay the sentinel; nothing is lost on reconnect.
- The 5 updated existing tests are correct. Each uses `find()` to locate the frame of interest instead of positional indexing, which is the right approach now that offset=0 is always the sentinel.
- Ring-buffer math in the successful-resume test is correctly adjusted (bufferSize-1 data events so total frames=3 fit without eviction, `replayedCount=3` assertion is accurate).
- PROTOCOL.md prose is clear and includes the filter hint for clients that don't want to surface the sentinel to users.
- Integration test assertion (`stderrBuf.includes('[chroxy-pod-agent] spawn cmd=node')`) is appropriately scoped: it confirms both the sentinel format and the correct cmd value, without being so tight that a whitespace tweak would break it.

---

### Issues Found

#### Critical (Must Fix)

None.

---

| `sidecar/PROTOCOL.md` | Spawn sequence diagram (~line 49) | Add the sentinel arrow between `session_started` and the first child output line | The diagram is the first thing readers see; it currently shows `session_started → event → stderr → exit`, omitting the sentinel. The text body (added in this PR) is correct, but the diagram is stale. Tracked: #3391 |
| `sidecar/PROTOCOL.md` + `k8s.js` | `stderr` section / `_wireWsToProc` | Document (or filter) the fact that `args` appear verbatim in the sentinel line, which K8sBackend forwards onto `proc.stderr` unfiltered | Currently safe (only format flags), but `execInEnvironment`'s returned `{ stderr }` will contain the sentinel including any future args. If a caller ever passes credentials as flags the args will leak into the caller's stderr buffer. Tracked: #3393 |
| `agent.test.js` | Overflow test comment (~line 955) | Update "oldest **2** should be evicted" → "sentinel + oldest **2** events evicted" | The sentinel now occupies seq=1; 5 frames pushed into a buffer of 3 evicts sentinel+event0+event1, not event0+event1. Test still passes, but the comment is off-by-one. Tracked: #3394 |

#### Nitpicks (Optional)

- The `k8s.js` `_wireWsToProc` handler at the `stderr` case (line ~996) sets `firstFrameSeen = true` when the sentinel arrives, which is technically correct (the sentinel is a frame), but the comment above `firstFrameSeen` says "an `error` frame that arrives BEFORE any event/stderr means the spawn was rejected". The sentinel is now always the first stderr, so the "before any event/stderr" window has effectively collapsed to zero for `stderr`-type error frames. Not a bug, but worth a comment update if anyone is auditing the spawn-rejection logic.

---

| Sequence diagram missing sentinel arrow | [#3391](https://github.com/blamechris/chroxy/issues/3391) | Doc-only; no behaviour change; safe to fix independently |
| Sentinel args exposure in K8sBackend stderr stream | [#3393](https://github.com/blamechris/chroxy/issues/3393) | Low current risk (only format flags in args today); decision on filter vs. doc note is a separate judgement call |
| Overflow test comment stale after sentinel offset | [#3394](https://github.com/blamechris/chroxy/issues/3394) | Comment-only; test passes and the logic is correct |

---

### Architecture Notes

The sentinel enters the sequenced output path via `_emitSessionFrame`, making it a first-class buffered frame rather than a side-channel write. This means:

- **Resume replay**: a client resuming with `lastSeq=0` always gets the sentinel replayed (seq=1 is the oldest possible seq). This is correct — the sentinel is diagnostic, not actionable, but it doesn't hurt to replay it and it keeps `replayedCount` accurate.
- **Buffer pressure**: the sentinel permanently occupies one slot in the ring buffer. For the default buffer of 1000 this is negligible. The bufferSize=3 test is correctly updated to account for it.
- **K8sBackend transparency**: the sentinel flows through `proc.stderr` unfiltered. Today's only consumer (`execInEnvironment` in snapshot/container setup) doesn't parse stderr, so the extra line is harmless. The #3393 issue tracks the future-proofing question.

### Verdict

- [x] **Approve** — implementation is correct, tests are thorough, and the format is well-documented. The three deferred items are minor and tracked.

PR #3400, opened 2026-05-05
feat(server): K8sBackend SidecarProcess stdin wiring (#3336)

### LINKED ISSUE #3336: fix(server): K8sBackend.SidecarProcess — wire stdin to sidecar (paired with #3329)
## Context

Identified during review of PR #3331 (K8sBackend streaming exec, #3320). Pairs with #3329 (sidecar stdin support).

## Description

`SidecarProcess` exposes a `stdin` PassThrough but the comment in #3331 acknowledges:

> writes here are forwarded to the child (future)

There is no wiring at all — writes to `proc.stdin` go nowhere. This means K8sBackend cannot be used by the SDK's `claude -p --input-format stream-json` workflow, which writes user turns as NDJSON to the child's stdin. Today the K8sBackend is technically unable to drive an interactive Claude Code session in a pod.

#3329 tracks the agent-side stdin frame additions to PROTOCOL.md. Once that lands, the consumer-side wiring belongs here.

## Original Review Comment

> `SidecarProcess.stdin` is a PassThrough that's never wired to the WS. The SDK's `--input-format stream-json` writes user turns to stdin; without forwarding, K8sBackend can't drive an interactive Claude session. Pair this with #3329 (agent stdin frame) and add a `proc.stdin.on('data', ...)` that writes a `{ type: 'stdin', data }` frame.

## Acceptance Criteria

- [ ] After #3329 lands, attach `proc.stdin.on('data', chunk => ws.send(JSON.stringify({ type: 'stdin', data: chunk.toString() })))` in `_wireWsToProc`.
- [ ] On `proc.stdin.end()`, send `{ type: 'stdin_end' }` (or whatever EOF semantics #3329 settles on).
- [ ] Add a test that data written to `proc.stdin` arrives at the agent as a `stdin` frame.

Required before K8sBackend can replace DockerBackend in `SdkSession.spawnClaudeCodeProcess` for interactive sessions.

blamechris:
Closes #3336

**Depends on #3389 (#3329). Targets `feat/3329-sidecar-stdin-support` — once #3389 merges, this PR will be retargeted to `main`.**

## What this does

Wires `SidecarProcess.stdin` (the PassThrough exposed to the SDK session) to the sidecar WebSocket protocol's `stdin` / `stdin_end` frames introduced in #3329.  Without this, `K8sBackend.streamCliInEnvironment` returns a process handle whose `stdin` goes nowhere, making the `--input-format stream-json` workflow (used by `SdkSession`) silently drop all user turns.

## Changes

**`packages/server/src/environments/backends/k8s.js`**

- `SidecarProcess` constructor: adds `_stdinBuffer`, `_stdinWired`, `_stdinEnded` fields; attaches an accumulator `data` listener to `this.stdin` and an `end` listener that triggers `stdin_end` forwarding.
- `_wireStdin(ws, proc)`: new helper called by `_wireWsToProc` immediately after the `spawn` frame is sent. Replaces the accumulator listener with a live-forwarding listener, flushes any pre-dial buffered chunks, and sends `stdin_end` if the stream already ended during the dial.
- `_sendStdinEnd(ws)`: sends `{ type: 'stdin_end' }` over the WS if it is still open.
- `_wireWsToProc` / `sendFirst`: adds `stdin: 'pipe'` to every `spawn` frame; calls `_wireStdin` on initial spawn only (not on reconnect — see below).

## Design choices

**`stdin: 'pipe'` always**: The sidecar protocol defaults to `'pipe'` per #3329, but we set it explicitly for clarity and to ensure the agent opens a piped stdin channel even if future protocol versions change the default.

**stdin is NOT replayed on reconnect**: The agent's ring buffer only buffers output frames (`event`, `stderr`, `exit`). `stdin` frames are consumed by the in-pod child process in real time. Re-sending them on a reconnect/resume would cause the child to receive duplicate input. `_wireStdin` is only called on the initial `spawn` path; the `resume` path does not re-attach a stdin forwarder. This is documented in the `SidecarProcess` class-level JSDoc.

**Pre-dial buffering**: Writes to `proc.stdin` before the WS dial resolves are held in `_stdinBuffer`. This matches how `stdout`/`stderr` consumers can attach listeners immediately — stdin writers can write immediately too. The buffer is flushed atomically after the spawn frame is sent, so the agent sees stdin data only after it has set up the child's stdin pipe.

**WS-closed-mid-write**: If the WS closes while stdin is still open, the first write attempt emits `error` on `proc` and removes the forwarding listener. Subsequent writes are silently dropped. This gives the caller a chance to react (e.g. `execInEnvironment` already has a timeout/error handler).

## Tests (6 new in `SidecarProcess stdin wiring (#3336)`)

- spawn frame includes `stdin:"pipe"`
- write to `proc.stdin` → `stdin` frame on WS
- `proc.stdin.end()` → `stdin_end` frame on WS
- writes before WS opens are buffered and flushed on open
- WS closed mid-write emits `error` on proc
- reconnect WS receives only the `resume` frame, never `stdin` frames

### DIFF
--- a/packages/server/sidecar/PROTOCOL.md
+++ b/packages/server/sidecar/PROTOCOL.md
@@ -35,7 +35,7 @@ probes cannot carry headers).
 
 ## Handshake Sequences
 
-### New Session
+### New Session (non-interactive `-p` mode)
 
 ```
 K8sBackend                       chroxy-pod-agent
@@ -46,10 +46,31 @@ K8sBackend                       chroxy-pod-agent
      │                                  │   reject → 401 (socket end, no WS frame)
      │                                  │   accept → WS connection open
      │                                  │
-     │── { type: 'spawn', ... } ────────▶
+     │── { type: 'spawn', stdin: 'ignore', ... } ──▶
      │◀─ { type: 'session_started', sessionId }
      │◀─ { type: 'event', payload: <object|string>, seq: N }
      │◀─ { type: 'stderr', data: '...', seq: N }
+     │◀─ { type: 'exit', code: 0, seq: N }
+     │                                  │   (WS close follows within 50 ms)
+```
+
+### New Session (stream-json input format)
+
+For `--input-format stream-json` the client must open a `stdin: 'pipe'` session,
+forward NDJSON lines via `stdin` frames, then close stdin with `stdin_end` so
+the child knows the input is complete.
+
+```
+K8sBackend                       chroxy-pod-agent
+     │                                  │
+     │── WS Upgrade (Authorization: Bearer <token>) ──▶
+     │                                  │   accept → WS connection open
+     │                                  │
+     │── { type: 'spawn', stdin: 'pipe', ... } ──▶
+     │◀─ { type: 'session_started', sessionId }
+     │── { type: 'stdin', data: '{"prompt":"hello"}\n' } ──▶
+     │── { type: 'stdin_end' } ──────────▶
+     │◀─ { type: 'event', payload: <object|string>, seq: N }
      │◀─ { type: 'exit', code: 0, seq: N }
      │                                  │   (WS close follows within 50 ms)
 ```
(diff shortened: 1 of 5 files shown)


blamechris:
## Code Review: PR #3400

### Summary

Wires `SidecarProcess.stdin` (a PassThrough stream) to the sidecar WebSocket `stdin` / `stdin_end` frames. The pre-dial buffering design is sound; the reconnect non-replay choice is correct. Diff is clean — only `k8s.js` and `k8s.test.js` are touched, with zero bleed from #3389. Six new tests cover the happy paths and most edge cases.

Two items are deferred to follow-up issues (see table below). No blocking issues.

---

### Strengths

- Pre-dial buffering is clean: the accumulator→forwarder swap in `_wireStdin` is easy to follow, and the `_stdinWired` flag prevents double-handling at the handoff point.
- Reconnect skip is correctly reasoned and well-documented — both in the class-level JSDoc and in the `_wireWsToProc` inline comment.
- `_sendStdinEnd` guards on `readyState === OPEN` before sending, and is reused from both the `end` listener and the post-flush check.
- `stdin: 'pipe'` on every spawn frame is the right explicit-over-implicit call; the comment connecting it to the stream-json workflow is helpful.
- WS-error path in the live listener emits an `error` on `proc` before removing itself, giving the consumer one notification before going silent — correct behavior for a PassThrough that keeps accepting writes.
- ES modules, no semicolons, single quotes — all clean.

---

### Issues Found

#### Critical (Must Fix)

None.

| File | Lines | Suggestion | Rationale |
|------|-------|------------|-----------|
| `k8s.js` | 808–819 | `_stdinBuffer` has no cap. If the WS dial stalls and the caller writes large payloads the buffer grows without bound. `kill()` also does not clear it. | Tracked in #3401. |
| `k8s.js` | 822–828 | On reconnect `proc.stdin` remains writable but nothing forwards — silent data loss if `SdkSession` keeps writing during a blip. No runtime signal to the consumer. | Tracked in #3402. |
| `k8s.js` | 1167–1175 | Flush loop checks `ws.readyState` but not `proc.killed`. If `kill()` fires during the synchronous flush the loop continues sending frames on a socket that `kill()` just closed. The `try/catch` will swallow the resulting `ws.send` error, so it's not dangerous, but it wastes cycles and logs noise. A `&& !proc.killed` guard on the loop condition would make the intent explicit. | Minor robustness. |

#### Nitpicks

- `k8s.test.js` L1867: setting `ws.readyState = 3` manually after `triggerClose` is fine for the test, but the comment "Mark WS as closed so the live listener sees it" could clarify *why* `triggerClose` alone isn't enough (the fake WS doesn't auto-update `readyState`). Minor doc nit.
- `k8s.test.js` L1869–1870: `proc.on('error', ...)` is registered *after* `triggerClose`. In the current fake-WS implementation that's fine because the close/reconnect path runs async. Worth a one-line comment noting the error listener must be attached before the write, not before the close.

---

| Cap `_stdinBuffer` and clear it in `kill()` | [#3401](https://github.com/blamechris/chroxy/issues/3401) | Not a correctness bug in the common case; needs a size constant decision. Out of scope for this wiring PR. |
| Signal consumer when stdin is suspended after reconnect | [#3402](https://github.com/blamechris/chroxy/issues/3402) | Requires `SdkSession` changes and a design choice (event vs. close). Cross-cutting scope. |

---

### Architecture Notes

The accumulator→forwarder listener-swap pattern is a clean way to handle the pre-open window without exposing internal state to callers. The choice to keep `_wireStdin` as a module-level function (rather than a method) is consistent with the existing `_wireWsToProc` / `_sendStdinEnd` pattern in this file.

The "stdin NOT replayed on reconnect" invariant is the right default for the stream-json workflow — the in-pod child has already consumed those bytes. The risk is the silent data-loss window on the reconnect, which #3402 tracks.

Stacking is clean: the diff touches only `k8s.js` and `k8s.test.js`; the sidecar-agent side (`sidecar/agent.js`, `PROTOCOL.md`) lives entirely in #3389 and does not appear here.

---

### Verdict

**Approve** — ready to merge once #3389 lands and this is retargeted to `main`. The two deferred items (#3401, #3402) are tracked and do not block correctness of the wiring itself.

PR #3412, opened 2026-05-07
docs(server): imagePullPolicy in Backend types.js JSDoc (#3377)

### LINKED ISSUE #3377: docs(server): add imagePullPolicy to createEnvironment opts JSDoc in types.js (K8s-specific note)
## Context

Identified during review of PR #3367 (feat(server): K8sBackend native imagePullPolicy option).

## Description

PR #3367 adds \`imagePullPolicy\` JSDoc to \`K8sBackend.createEnvironment\` and to the constructor, but it does not add any mention to the Backend interface definition in \`types.js\`.

The Backend interface's \`createEnvironment\` JSDoc describes the canonical opts shape shared by all backends. Docker has no analog for \`imagePullPolicy\`, so this option is K8s-specific. The question is whether \`types.js\` should:

1. **Document it with a K8s-specific note** — add the param with a clear callout that it is only honoured by K8sBackend, and silently ignored by others. This makes the shared interface the single source of truth.
2. **Leave it out of types.js** — backend-specific opts stay documented only on the backend itself. This keeps the interface clean but means readers of types.js don't know the option exists.

Option 1 is preferred because \`EnvironmentManager\` passes the raw opts object straight through to the backend, so users who pass \`imagePullPolicy\` to the manager need to know it is a valid (if backend-specific) key.

## Original Review Comment

> Should \`imagePullPolicy\` be added to the \`createEnvironment\` opts JSDoc in \`types.js\`? The Backend interface docs are the single source of truth for EnvironmentManager callers. Docker has no analog, but a K8s-specific callout (e.g. "Honoured only by K8sBackend — Docker ignores this field") would make the option discoverable without requiring readers to open \`k8s.js\`.

## Acceptance Criteria

- [ ] \`types.js\` \`createEnvironment\` JSDoc includes \`opts.imagePullPolicy\` with a note that it is K8s-specific and silently ignored by other backends
- [ ] The wording makes clear this is optional and that omitting it defers to the K8s default behaviour

blamechris:
Closes #3377

## Summary

- Adds `opts.imagePullPolicy` to the `createEnvironment` JSDoc in `packages/server/src/environments/backends/types.js`
- Documents the three permitted values (`'Always'|'IfNotPresent'|'Never'`)
- Includes a K8s-specific callout: honoured only by K8sBackend; Docker and other backends silently ignore the field
- Documents that omitting the param defers to the K8s cluster default (no value is forced onto the Pod spec)

JSDoc-only change — no behavioural modification.

## Test plan

- [ ] Verify lint passes (excluding pre-existing dashboard-next/dist errors): `npm run lint -- --ignore-pattern 'src/dashboard-next/dist/**'`
- [ ] Verify server tests still pass: `npm test`

### DIFF
--- a/packages/server/src/environments/backends/types.js
+++ b/packages/server/src/environments/backends/types.js
@@ -54,6 +54,10 @@
  * @param {number[]|string[]} [opts.forwardPorts] - Ports to expose from the container
  * @param {string[]} [opts.mounts]        - Additional volume mounts (already validated)
  * @param {string}   [opts.postCreateCommand] - Shell command to run after setup completes
+ * @param {'Always'|'IfNotPresent'|'Never'} [opts.imagePullPolicy] - Container image pull policy.
+ *   Honoured only by K8sBackend — Docker and other backends silently ignore this field.
+ *   When absent, K8sBackend omits the field from the Pod spec and the K8s cluster default applies
+ *   (typically `IfNotPresent` for tagged images, `Always` for `latest`).
  * @returns {Promise<{ containerId: string, containerCliPath: string }>}
  *   containerId — the full container ID string returned by the runtime
  *   containerCliPath — absolute path inside the container where the CLI binary was installed


blamechris:
## Code Review: PR #3412

### Summary

JSDoc-only addition of `opts.imagePullPolicy` to the `Backend.createEnvironment` interface definition in `types.js`. Closes #3377. No behavioural change.

### Strengths

- Union type `'Always'|'IfNotPresent'|'Never'` matches the exact set K8s accepts — correct and exhaustive
- Placement after the other optional `opts.*` params and before `@returns` is consistent with the existing style
- The K8s-only callout is present and names the backend explicitly
- "silently ignore" wording is clear — callers know passing the field to Docker is safe (not an error)
- The note about cluster default (`IfNotPresent` for tagged / `Always` for `latest`) is a useful hint without being prescriptive

### Issues Found

#### Critical (Must Fix)

None.

| `types.js` | added block | Mention the per-call vs constructor-level precedence rule | `k8s.js` line 194 says "Falls back to the constructor-level option when unset." `types.js` is silent on this. Readers of the interface doc don't know a constructor default exists, so they can't predict the effective value when both are set. A one-liner like "Falls back to the K8sBackend constructor-level `imagePullPolicy` when both are set" would complete the picture. |

#### Nitpicks

- Backtick quoting `IfNotPresent` / `Always` / `latest` inside the prose sentence would be consistent with how other inline values are styled elsewhere in `types.js` (e.g. `` `docker run` ``, `` `chroxy-env-{envId}` ``), but this is minor and the current plain-text form reads fine.

None — the one suggestion above is small enough to address inline if desired; it does not warrant a separate issue.

### Architecture Notes

The addition correctly treats `types.js` as the single source of truth for `EnvironmentManager` callers, matching the stated intent of the module header ("Backends MUST NOT hold environment state of their own … The manager stores the handle"). Backend-specific opts that pass through the manager belong here.

### Verdict

- [x] **Approve** — the union type is correct, the K8s-only note is explicit, and the "silently ignore" language is unambiguous. The suggestion about documenting the constructor-level fallback is worth a one-line addition but is not a blocker.

PR #3414, opened 2026-05-07
fix(server): docker execInEnvironment env value coercion (#3361)

blamechris:
Closes #3361

## Changes

- `docker.js` `execInEnvironment`: entries with `null` or `undefined` values are now skipped instead of producing `KEY=null` / `KEY=undefined`; all surviving values are wrapped with `String(val)` to make the coercion intent explicit
- JSDoc `@param` type narrowed from `{Object}` to `{Object.<string,string>}` in both `docker.js` and `types.js`

## Tests

Four new cases added to `DockerBackend.execInEnvironment()` test suite:

| Case | Expectation |
|------|-------------|
| `null` value | entry skipped entirely |
| `undefined` value | entry skipped entirely |
| numeric value (`3000`, `0`) | coerced — `PORT=3000`, `TIMEOUT=0` (0 is not null) |
| string / empty string | passed through unchanged |

All 46 docker backend tests pass.

### DIFF
--- a/packages/server/src/environments/backends/docker.js
+++ b/packages/server/src/environments/backends/docker.js
@@ -137,10 +137,13 @@ export class DockerBackend {
 
   /**
    * @param {string} containerId
-   * @param {{ cmd: string, env?: Object, cwd?: string, timeout?: number }} opts
+   * @param {{ cmd: string, env?: Object.<string,string>, cwd?: string, timeout?: number }} opts
    * @returns {Promise<{ stdout: string, stderr: string }>}
    *
    * opts.env — all key/value pairs are forwarded as `--env KEY=VAL` flags.
+   * Entries whose value is null or undefined are silently skipped (caller
+   * mistake — passing them would produce `KEY=null` / `KEY=undefined` which
+   * is hard to debug).  All other values are coerced to String explicitly.
    * No allowlist filter is applied here: unlike streamCliInEnvironment (which
    * runs the long-lived Claude Code CLI and must never leak the full host env),
    * execInEnvironment is a general-purpose helper invoked with an explicit env
@@ -161,7 +164,8 @@ export class DockerBackend {
 
       if (env) {
         for (const [key, val] of Object.entries(env)) {
-          execArgs.push('--env', `${key}=${val}`)
+          if (val == null) continue
+          execArgs.push('--env', `${key}=${String(val)}`)
         }
       }
 
--- a/packages/server/src/environments/backends/types.js
+++ b/packages/server/src/environments/backends/types.js
@@ -131,7 +131,7 @@
  * @param {string} containerId - Container ID
  * @param {Object} opts
  * @param {string}   opts.cmd            - Shell command to run (passed as `bash -c <cmd>`)
- * @param {Object}   [opts.env]          - Extra environment variables for the exec process
+ * @param {Object.<string,string>} [opts.env] - Extra environment variables for the exec process; null/undefined values are skipped
  * @param {string}   [opts.cwd]          - Working directory inside the container
  * @param {number}   [opts.timeout]      - Timeout in ms (default 30 000)
  * @returns {Promise<{ stdout: string, stderr: string }>}
--- a/packages/server/tests/environments/backends/docker.test.js
+++ b/packages/server/tests/environments/backends/docker.test.js
@@ -806,6 +806,78 @@ describe('DockerBackend.execInEnvironment()', () => {
     )
   })
 
+  it('filters out entries whose value is null', async () => {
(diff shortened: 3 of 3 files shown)


blamechris:
## Code Review: PR #3414

### Summary

Targeted fix for `DockerBackend.execInEnvironment`: null/undefined env values are now skipped with `val == null` (covers both), surviving values are wrapped with `String(val)` to make coercion intent explicit, and JSDoc is narrowed to `Object.<string,string>` in both `docker.js` and `types.js`. Four new tests cover the added branches.

---

### Strengths

- `val == null` is exactly the right idiom — covers both `null` and `undefined` in one guard without `===` duplication
- `String(val)` intent-explicit coercion is correct: `String(0)` → `"0"`, `String(false)` → `"false"`, no surprises
- JSDoc narrowed in *both* files (`docker.js` impl + `types.js` interface) — keeps the two in sync
- The four new tests precisely target the four distinct code paths: null-skip, undefined-skip, falsy-zero passthrough, empty-string passthrough
- Test for `TIMEOUT=0` is the critical correctness proof: confirms `val == null` does **not** catch `0`, which would be a silent data loss bug
- `EMPTY=''` test is similarly important: empty string is a valid env value and must not be elided

---

### Issues Found

#### Critical (Must Fix)

None.

---

| File | Note | Rationale |
|------|------|-----------|
| `docker.js` JSDoc | The comment says *"caller mistake"* but the method still silently swallows it. Consider `log.warn(...)` on skip so a misrouted `null` surfaces in server logs without breaking the call. | Silent skip is safe but invisible in production. A single `log.warn(\`execInEnvironment: skipping null/undefined value for env key "${key}"\`)` line would make caller bugs immediately diagnosable from server logs, consistent with the existing `log.warn` usage in the same class (e.g. setup failure). This is a judgment call — the PR description justifies silence; the tradeoff is debuggability vs noise. |
| `types.js` JSDoc | `streamCliInEnvironment`'s `[opts.env]` is still typed as `{Object}` (no value type). Now that the pattern is established for `execInEnvironment`, narrowing it to `{Object.<string,string>}` would be consistent. | Low priority — `streamCliInEnvironment` uses an allowlist filter so the risk is lower, but the inconsistency in the same file is worth noting. |

---

#### Nitpicks (Optional)

- The four new tests share identical `envPairs` extraction boilerplate. A local `getEnvPairs(execCall)` helper would remove 16 duplicate lines, but the current repetition is consistent with the existing test style in this file — purely optional.

---

### Focused Review of the Key Questions

**Silent-skip vs warn-and-skip:** The silent-skip choice is defensible. `execInEnvironment` is a fire-and-forget helper; rejecting on bad input would be a breaking behavior change. However, the JSDoc comment *itself* identifies this as a "caller mistake", which argues for at least a `log.warn` — the information exists, the infrastructure exists (`log` is imported at file top), and it costs one line. This is flagged as a suggestion, not a blocker.

**`String(0)` → `"0"`:** Correct. `val == null` does not catch `0`, and `String(0)` → `"0"` produces `TIMEOUT=0` as expected. The `TIMEOUT=0` test case directly verifies this.

**JSDoc narrowing coverage:** `docker.js` `@param` and `types.js` `@param` are both updated. The Backend interface's `streamCliInEnvironment` still uses `{Object}` for its `env` param — see Suggestions above.

**4 new tests / branch coverage:** All four meaningful branches are covered: null skip, undefined skip, falsy-zero passthrough, empty-string passthrough. The tests correctly use assertion inversion (`!envPairs.some(p => p.startsWith('BAD='))`) for the skip cases rather than checking length, which would be fragile.

---

### Deferred Items

| Add `log.warn` on null/undefined skip in `execInEnvironment` | #3427 | Behavior is safe as-is; warn-on-skip is an observability improvement, not a correctness fix. Out of scope for this targeted coercion PR. |
| Narrow `streamCliInEnvironment` opts.env JSDoc to `Object.<string,string>` in `types.js` | #3428 | Consistency cleanup; streamCliInEnvironment has allowlist protection so correctness risk is lower. Separate PR keeps this one focused. |

---

### Architecture Notes

`execInEnvironment` sits between the Backend interface and the EnvironmentManager. There are currently no production callers outside the backends directory — this is wiring infrastructure. The fix is correctly scoped to the backend implementation only, with no interface surface changes. The `val == null` + `String(val)` pattern is worth mirroring in `K8sBackend.execInEnvironment` if/when it grows an env-forwarding loop (it currently delegates to `streamCliInEnvironment`).

---

### Verdict

**Approve — Ready to merge.**

The implementation is correct, concise, and consistent with existing style. Tests are well-targeted and cover the critical `0 != null` edge case explicitly. Two minor follow-up issues created for observability and consistency improvements.

PR #3416, opened 2026-05-07
ci: concurrency group for nightly-k8s-integration (#3359)

blamechris:
Closes #3359

## Summary

- Adds a top-level `concurrency` block to `nightly-k8s-integration.yml`
- `group: nightly-k8s-integration` scopes the lock to this workflow
- `cancel-in-progress: true` — a new `workflow_dispatch` (or a re-triggered schedule) cancels the in-flight run rather than queuing a second kind cluster alongside it

## Rationale for `cancel-in-progress: true`

The issue and original review comment both specify this value. The workflow runs a resource-intensive kind cluster; letting an older run continue while a newer one starts wastes double the runner minutes and risks resource contention on the shared runner pool. Cancelling the stale run is the correct behaviour — whoever triggered the newer run wants fresh results.

## Test plan

- [ ] Verify the workflow YAML is valid (no syntax errors)
- [ ] Confirm CI passes on this PR
- [ ] Optionally: trigger a `workflow_dispatch` while the nightly is mid-run and observe the older run is cancelled

### DIFF
--- a/.github/workflows/nightly-k8s-integration.yml
+++ b/.github/workflows/nightly-k8s-integration.yml
@@ -6,6 +6,12 @@ on:
     - cron: '0 6 * * *'
   workflow_dispatch:
 
+# Prevent overlapping runs: a new dispatch (or schedule) cancels an in-flight
+# run rather than spinning up a second kind cluster on the same runner pool.
+concurrency:
+  group: nightly-k8s-integration
+  cancel-in-progress: true
+
 # Read-only token — the job only reads the repo and talks to GitHub Checks.
 permissions:
   contents: read


blamechris:
## Code Review: PR #3416

### Summary

Adds a top-level `concurrency` block to `nightly-k8s-integration.yml` with `group: nightly-k8s-integration` and `cancel-in-progress: true`. Closes #3359.

Six lines of YAML, zero code changes. Scope is appropriately narrow.

### Strengths

- **Correct scope for the group name.** Using the literal workflow name as the group key means only runs of this workflow contend with each other — no accidental cross-workflow cancellations.
- **Placement is correct.** Top-level `concurrency` (not per-job) applies to the entire workflow run, which is what you want when the resource being protected is a cluster that spans the whole job.
- **Comment explains the why.** The inline comment (`# Prevent overlapping runs: …`) is clear and accurate; future readers won't have to guess the intent.
- **`cancel-in-progress: true` is the right call for this workload.** Detailed rationale below.

### `cancel-in-progress: true` — Trade-off Analysis

The alternative (`cancel-in-progress: false`, i.e., queue) would let both runs finish sequentially — fresh results from the later run, but at the cost of double (or more) runner minutes and delayed feedback for the triggering engineer.

For this workload the `true` choice is correct because:

1. **The cluster is ephemeral and PID-scoped.** `CLUSTER_NAME = chroxy-test-${process.pid}` — there is no shared persistent state between runs. Cancelling an in-flight run abandons one transient cluster; it does not corrupt any external resource.
2. **Cleanup on cancellation is safe.** When GitHub Actions cancels a job mid-run, the runner sends `SIGTERM` to the running process. The test suite registers a `SIGINT` handler that calls `deleteCluster()` unconditionally. GitHub Actions sends SIGTERM, not SIGINT, so the in-process handler won't fire — **but the kind cluster will still be cleaned up** because `helm/kind-action`'s `install_only: true` mode never creates a cluster itself, and the cluster created by the test's `before()` hook is named `chroxy-test-<pid>`. When the runner VM is recycled after cancellation, the VM (and the kind cluster running inside it) is destroyed. No leaked resource, no orphan cluster on a shared infrastructure layer.
3. **The `after()` hook and SIGINT handler are defence-in-depth, not the primary cleanup path.** Primary cleanup is runner VM teardown. Secondary cleanup is `after()` on normal exit. The SIGINT handler is a local-dev convenience. None of these are bypassed or broken by cancellation.
4. **Queuing (`false`) creates a worse failure mode here.** A queued run would start after the slower nightly finishes. If the nightly was already flaky or slow on a resource-constrained runner, the queued `workflow_dispatch` run inherits that same degraded runner state and is more likely to fail too. Cancellation + fresh start is a better signal.

**One nuance worth noting (not a blocker):** If a run is cancelled mid-`before` hook — specifically after `kind create cluster` but before the test suite starts — the SIGTERM path relies solely on runner VM teardown for cluster cleanup. This is acceptable given kind clusters are scoped to the runner VM, but if this ever moves to a persistent/shared cluster (e.g., a cloud-hosted kind-in-kind setup), the cleanup guarantee weakens. That scenario is out of scope for today's PR.

### Issues Found

#### Critical (Must Fix)

None.

None.

#### Nitpicks (Optional)

None. The diff is minimal and the comment is correct.

### Deferred Items

None. The one nuance above (persistent-cluster cleanup) is a hypothetical future concern, not a present gap. No issue warranted.

### Architecture Notes

This change touches only CI configuration — no server, app, or protocol code is affected. The change is consistent with how other workflows in the repo use concurrency groups.

### Verdict

The `cancel-in-progress: true` choice is correct. The cluster lifecycle is ephemeral and runner-scoped, cleanup on cancellation relies on VM teardown (which is guaranteed), and queuing would produce worse outcomes for this workload. No issues found.

PR #3417, opened 2026-05-07
docs(server): tickAndFlush single-tick limitation note (#3383)

blamechris:
Closes #3383

## Summary

- Added a JSDoc block to `tickAndFlush()` in `k8s.test.js` explaining the single-tick assumption
- Documents that `clock.tick()` snapshots `pending` before iterating, so a timer enqueued by a callback during that iteration is not fired by the same call
- Explains that all current tests are safe because `_dialWs` always resolves, bypassing the `.catch` branch in `scheduleReconnect`
- Provides a `tickAndFlushAll` loop skeleton (with max-iterations guard) as a reference for any future test that exercises chained reconnect timers

## Test plan

- [ ] `npm test` in `packages/server/` — k8s reconnect tests pass (0 failures in k8s suite)
- [ ] JSDoc-only change; no logic modified

### DIFF
--- a/packages/server/tests/environments/backends/k8s.test.js
+++ b/packages/server/tests/environments/backends/k8s.test.js
@@ -130,6 +130,37 @@ function createFakeClock() {
 /**
  * Fire all pending fake-clock callbacks then yield one event-loop tick so
  * any promise continuations scheduled inside those callbacks can resolve.
+ *
+ * Single-tick assumption
+ * ----------------------
+ * clock.tick() snapshots `pending`, clears the map, then iterates the
+ * snapshot.  If a callback enqueues a *new* timer during that iteration (e.g.
+ * scheduleReconnect → dial fails → .catch handler calls scheduleReconnect
+ * again), the new timer is added to `pending` *after* the snapshot was taken
+ * and is therefore NOT fired by the current tick() call.  The single
+ * setImmediate yield that follows does not drain it either.
+ *
+ * All current tests are unaffected because _dialWs always resolves
+ * successfully, so scheduleReconnect's .catch branch is never entered and no
+ * chained timer is ever queued.
+ *
+ * If a future test exercises a mid-retry dial failure (the .catch path in
+ * scheduleReconnect), it will need to call tickAndFlush() once per chained
+ * failure — or the helper should be extended with a loop that keeps ticking
+ * until pending is empty, guarded by a max-iterations cap to avoid infinite
+ * loops for genuinely self-rescheduling timers:
+ *
+ *   async function tickAndFlushAll(clock, maxIterations = 20) {
+ *     let iterations = 0
+ *     do {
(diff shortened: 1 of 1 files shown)


blamechris:
## Code Review: PR #3417

### Summary

JSDoc-only change to `tickAndFlush()` in `k8s.test.js`. Documents the single-tick assumption of the fake clock, explains why all current tests are safe, and provides a `tickAndFlushAll` reference skeleton for future tests that exercise chained reconnect failures.

No behavioral change — this is documentation only.

### Strengths

- The explanation of why `clock.tick()` snapshots `pending` before iterating is accurate and well-written. Verified against `createFakeClock()`: it does `const callbacks = [...pending.values()]; pending.clear(); for (const fn of callbacks) fn()` — the snapshot/clear pattern is exactly as described.
- The claim that current tests are safe because `_dialWs` always resolves is correct. Traced through `scheduleReconnect` → `proc._redial().catch(scheduleReconnect)` — the `.catch` path is only reachable on a dial failure, which the injected fake `_dialWs` never triggers.
- Flagging the need for `hasPending()` on `createFakeClock` is the right call — it keeps the skeleton honest rather than implying it works today.

### Issues Found

#### Critical (Must Fix)

None.

| `k8s.test.js` | skeleton | Consider checking `hasPending()` before the first `tick()` (pre-check) rather than only at the bottom of the loop | The `do...while` always fires at least one `tick()` even if `pending` is empty. A `while (clock.hasPending())` loop would be a slightly safer pattern for callers who may invoke it after `pending` is already drained. For a test helper this is low-stakes, but worth noting. |

#### Nitpicks

- The `iterations` guard fires *after* the tick on that iteration (`iterations >= maxIterations` is checked post-hoc). This means on iteration 20 the tick has already executed side effects before the error is thrown. Acceptable for a safety net, but a pre-loop cap (`if (iterations >= maxIterations) throw` at top of loop) would give a cleaner abort with no partial state.
- Minor style: the skeleton uses `async function tickAndFlushAll` — consistent with the rest of the file's style. No issue.

None. The suggestion and nitpick above are too minor to warrant tracking issues for a doc-only PR.

### Architecture Notes

The documentation correctly situates `tickAndFlush` within the fake-clock contract. The max-iterations guard in the skeleton is a good pattern — self-rescheduling timers (e.g., timers that always re-enqueue regardless of exit conditions) could produce an infinite loop without it, and 20 iterations is a reasonable cap for reconnect retry chains.

The single-`setImmediate` yield in the skeleton is sufficient: `_redial()` rejection propagates through the microtask queue (Promise chains), and microtasks all flush before `setImmediate` fires. So `clock.hasPending()` is evaluated after the `.catch(scheduleReconnect)` has had a chance to enqueue the next timer.

### Verdict

Doc is accurate, the skeleton is correct for its stated purpose, and the "must add `hasPending()`" caveat is clearly called out.

PR #3420, opened 2026-05-07
fix(server): LineLimitTransform CRLF counting (#3381)

### LINKED ISSUE #3381: LineLimitTransform: CRLF byte counted toward cap (CR not reset on \r)
## Context

Identified during review of PR #3364 (NDJSON line buffer cap).

## Description

`LineLimitTransform` only resets `_pending` on `0x0a` (LF). A `\r` byte is treated as content and increments the counter. This means:

- A CRLF line of exactly `maxBytes` characters will false-trip the guard (the CR pushes pending to `maxBytes + 1` before the LF can reset it).
- A 1 MiB-capped CRLF line gets only `maxBytes - 1 = 1_048_575` bytes of JSON content before triggering (the CR consumes one slot).

**Is this a practical problem?** Probably not — `chroxy-pod-agent` runs Claude inside Linux pods which emit LF-terminated lines. CRLF from a Windows process is hypothetical. The 1-byte discrepancy at 1 MiB is negligible.

However, if `maxLineBytes` is ever tuned to a small value for testing or resource-constrained environments, an off-by-one on the CR could cause surprising false trips.

## Original Review Comment

> `LineLimitTransform` counts `\r` as a content byte. For a CRLF line that is exactly `maxBytes` characters long, the `\r` pushes `_pending` to `maxBytes + 1` before the `\n` reset — firing `oversized_line` on a line that is actually within the content cap.

## Acceptance Criteria

- [ ] Decision documented in code comment: either (a) intentionally count transport bytes including CR (add a note), or (b) reset on `\r` as well (match readline's CRLF strip behaviour)
- [ ] If (b): update the unit test `'does not fire oversized_line when line is exactly maxBytes'` to cover the CRLF case

blamechris:
Closes #3381

## What

`LineLimitTransform` counted CR (`\r`, 0x0d) as a content byte. A CRLF line of exactly `maxBytes` content bytes would push `_pending` to `maxBytes+1` before the LF reset it, falsely triggering `oversized_line`.

## Fix

Treat CR the same as LF in the pending counter — reset `_pending` to zero rather than counting it toward the cap. This matches readline's CRLF-strip semantics: CR is a transport artefact, not content.

## Tests

Two regression tests added to the `LineLimitTransform` suite:

- `does not fire oversized_line for a CRLF line of exactly maxBytes content bytes` — verifies the false-positive is gone
- `fires oversized_line when CRLF line content exceeds maxBytes` — verifies legitimate over-cap CRLF lines still trip the guard

All 65 sidecar agent tests pass.

## Impact

Practically none today — `chroxy-pod-agent` runs inside Linux pods emitting LF-only output. Fixes a correctness issue that would matter if `maxLineBytes` is tuned to a small value or CRLF output is ever introduced.

### DIFF
--- a/packages/server/sidecar/agent.js
+++ b/packages/server/sidecar/agent.js
@@ -137,8 +137,12 @@ export class LineLimitTransform extends Transform {
 
     let offset = 0
     for (let i = 0; i < chunk.length; i++) {
-      if (chunk[i] === 0x0a /* '\n' */) {
-        // Newline resets the pending counter for the next line.
+      if (chunk[i] === 0x0a /* '\n' */ || chunk[i] === 0x0d /* '\r' */) {
+        // LF resets the line counter.  CR is treated the same so that a CRLF
+        // line of exactly maxBytes content bytes does not false-trip the guard
+        // (without this, the CR byte pushes _pending to maxBytes+1 before the
+        // LF resets it).  readline strips CR from CRLF pairs, so excluding it
+        // from the content count matches readline's semantics.
         this._pending = 0
         offset = i + 1
       } else {
--- a/packages/server/tests/sidecar/agent.test.js
+++ b/packages/server/tests/sidecar/agent.test.js
@@ -1119,6 +1119,26 @@ describe('PodAgent', () => {
       assert.equal(oversized, true)
     })
 
+    // CRLF regression tests (#3381) ----------------------------------------
+
+    it('does not fire oversized_line for a CRLF line of exactly maxBytes content bytes (#3381)', async () => {
+      // Before the fix, CR was counted as a content byte — a CRLF line of
+      // exactly maxBytes pushed _pending to maxBytes+1 before the LF could
+      // reset it, causing a false oversized_line.
+      const t = new LineLimitTransform({ maxBytes: 10 })
+      const crlfLine = 'A'.repeat(10) + '\r\n'
+      const { oversized } = await collect(t, [crlfLine])
+      assert.equal(oversized, false)
+    })
+
+    it('fires oversized_line when CRLF line content exceeds maxBytes (#3381)', async () => {
+      // 11 content bytes + CRLF must still trip the guard.
+      const t = new LineLimitTransform({ maxBytes: 10 })
+      const crlfOver = 'A'.repeat(11) + '\r\n'
+      const { oversized } = await collect(t, [crlfOver])
+      assert.equal(oversized, true)
+    })
+
     it('drops all data after the oversized_line event fires (second write is suppressed)', async () => {
       const t = new LineLimitTransform({ maxBytes: 5 })
       const received = []


blamechris:
## Code Review: PR #3420

### Summary

Fixes a false-positive `oversized_line` event in `LineLimitTransform` when processing CRLF-terminated lines at exactly `maxBytes` content bytes. Before this fix, the CR byte incremented `_pending` to `maxBytes+1` before the subsequent LF could reset it — triggering the guard incorrectly. The fix treats CR identically to LF (reset `_pending = 0`, advance `offset`). Two regression tests are added.

---

### Strengths

- **Correct semantic framing.** The PR description accurately identifies the root cause and justifies the fix by reference to readline's CRLF-strip semantics: CR is a transport artefact, not content, so excluding it from the byte count is the right model.
- **Symmetric implementation.** CR and LF now share the same branch — no separate handling, no state flag — keeping the hot loop simple.
- **Regression tests are precise.** The two new tests hit the exact boundary (`exactly maxBytes` and `maxBytes+1` with CRLF), mirroring the existing LF-only boundary tests, making the coverage symmetrical.
- **No existing tests disturbed.** The CR branch was dead code before this change; all LF-only tests are unaffected by the new condition.

---

### Issues Found

#### Critical (Must Fix)

None.

| `agent.js` | 140 | The comment says "LF resets the line counter" but the condition now covers both CR and LF. Consider "LF or CR resets…" for clarity. | Minor readability — the inline comment on the `if` branch currently undersells the dual-byte logic. |

#### Nitpicks (Optional)

- The test comment block header says `// CRLF regression tests (#3381) ---` — the trailing dashes differ in length from the other section headers in the file. Harmless, but inconsistent.

---

### Semantic Correctness — Focused Questions

**Does CR-resets match readline strip-CR semantics?**

Yes. Node's `readline.createInterface` with default options strips a trailing `\r` from every line it emits. `LineLimitTransform` sits *upstream* of readline and its job is to count bytes for the overflow guard — not to strip anything. Since readline will discard the CR from the line content, counting it toward `maxBytes` would mis-charge content budget against a transport byte. Resetting `_pending` on CR is the correct model.

**What about CR-only line endings (old Mac `\r` without `\n`)?**

The new code treats bare `\r` as a line terminator, which means CR-only line endings are handled correctly: each CR-separated segment is treated as a "line" for overflow counting purposes. The guard fires if any segment exceeds `maxBytes`. Since the CR byte still passes through unchanged (only `_pending` is reset, not the data), readline downstream receives the raw bytes and handles them according to its own rules. In practice, old Mac CR-only output is non-existent from any modern tool or container process, so this path is only a theoretical edge case — and the behavior is safe.

**Does this affect any existing test that depends on CR being content?**

No. All pre-existing `LineLimitTransform` tests use LF-only input. The CR branch was unreachable before this PR. Verified: all 63 tests pass after the change.

**Split-chunk CRLF (CR at end of chunk N, LF at start of chunk N+1)?**

Handled correctly. CR in chunk N resets `_pending = 0`; LF in chunk N+1 resets `_pending = 0` again (harmless double-reset). No false fire, no missed count.

**`offset` passthrough when oversized fires?**

Correct. `offset` is advanced to `i+1` on both CR and LF. When the oversized guard trips, `chunk.slice(0, offset)` passes all complete lines (including any trailing `\r\n`) to readline before dropping the overrun. The partial passthrough logic works identically for CRLF as it did for LF.

---

### Deferred Items

None. No follow-up issues warranted.

---

### Architecture Notes

`LineLimitTransform` is a focused, single-responsibility stream transform. This change stays well within its scope. The fix does not alter the WS protocol, session lifecycle, or readline integration.

---

### Verdict

**Approve — ready to merge.**

The fix is correct, the justification is sound, the regression tests are well-targeted, and no existing behaviour is disturbed. The CR-only and split-chunk edge cases are handled safely.

PR #3424, opened 2026-05-07
test(server): two-pass community-gate uncovered branches (#3360)

blamechris:
Closes #3360

Adds three tests to the `community-namespace gate (#3206 PR A)` describe block, covering the two-pass `_collectCandidates` code path branches identified in the issue:

**1. `onCommunityTrustPending` callback in two-pass path**
A single test drives both sub-paths by using a shared `parseCache` Map. The first call (cold cache) hits the cache-miss branch (lines 731–735); the second call (warm cache) hits the cache-fast-path branch (lines 519–524). Both assert the callback fires once with the correct `{ name, author, path }` payload.

**2. `includeInactive:true` pending shape in two-pass path**
Existing test 2 passed `includeInactive: true` but only checked `inspectCallCount === 0`. This new test asserts the full returned entry has `{ active: false, trustState: 'pending', communityAuthor: 'alice' }`, catching regressions where the entry is silently dropped or a field is omitted.

**3. Fail-open without `communityTrustChecker` in two-pass path**
Mirrors the existing single-pass fail-open test for the two-pass `_collectCandidates` path. Omitting `communityTrustChecker` should default `trustedC`/`trustedM` to `true` (lines 515–517 and 726–728); asserts the skill loads as active with `trustState: 'trusted'`.

All three tests run under `maxTotalBytes: 1024 * 1024` (1 MiB) to activate the two-pass path without triggering budget cuts.

## Test plan

- `node --test tests/skills-loader.test.js` → 195 pass, 0 fail

### DIFF
--- a/packages/server/tests/skills-loader.test.js
+++ b/packages/server/tests/skills-loader.test.js
@@ -2932,6 +2932,85 @@ describe('skills-loader', () => {
       assert.equal(inspectCallCount, 1, 'inspect() must run for trusted community skills (two-pass cache-fast-path)')
     })
 
+    it('two-pass path: onCommunityTrustPending fires once with correct payload (cache-miss + cache-fast-path)', () => {
+      // Covers both the cache-miss branch (lines 731–735) on the first call and
+      // the cache-fast-path branch (lines 519–524) on the warm-cache second call.
+      mkdirSync(join(communityDir, 'community', 'alice'), { recursive: true })
+      writeFileSync(join(communityDir, 'community', 'alice', 'pending.md'), '# Pending skill\n\nBody.\n')
+
+      const warmCache = new Map()
+
+      // First call: cache cold → cache-miss branch fires onCommunityTrustPending
+      const pendingCallsCold = []
+      loadActiveSkills(communityDir, {
+        communityTrustChecker: () => false,
+        onCommunityTrustPending: (info) => pendingCallsCold.push(info),
+        includeInactive: true,
+        maxTotalBytes: 1024 * 1024,
+        parseCache: warmCache,
+      })
+
+      assert.equal(pendingCallsCold.length, 1, 'cache-miss: callback must fire exactly once')
+      assert.equal(pendingCallsCold[0].name, 'pending')
+      assert.equal(pendingCallsCold[0].author, 'alice')
+      assert.ok(typeof pendingCallsCold[0].path === 'string', 'path must be a string')
+
+      // Second call: warmCache is now populated → cache-fast-path fires onCommunityTrustPending
+      const pendingCallsWarm = []
+      loadActiveSkills(communityDir, {
+        communityTrustChecker: () => false,
+        onCommunityTrustPending: (info) => pendingCallsWarm.push(info),
+        includeInactive: true,
+        maxTotalBytes: 1024 * 1024,
+        parseCache: warmCache,
+      })
+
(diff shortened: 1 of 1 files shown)


blamechris:
## Code Review: PR #3424

### Summary

Adds three tests to the `community-namespace gate (#3206 PR A)` describe block covering uncovered branches in the two-pass `_collectCandidates` path. All three tests target symmetry gaps between the single-pass and two-pass code paths that exist at lines 510–536 (cache-fast-path) and 722–748 (cache-miss-path) in `skills-loader.js`.

### Strengths

- **Correct branch targeting.** Test 1 drives both the cache-miss branch (lines 731–735) and the cache-fast-path branch (lines 519–524) using a single shared `warmCache` Map. The two-call pattern is the minimal design to exercise both sub-branches in one test.
- **Test 2 catches a real regression class.** The existing test only asserted `inspectCallCount === 0`; the new test asserts the full shape `{ active, trustState, communityAuthor }` of the returned entry, which would catch a silent field omission or a dropped entry.
- **Test 3 mirrors the existing single-pass fail-open test precisely.** Pattern match is clean: no `communityTrustChecker` → `trustedM/trustedC` defaults to `true` → skill surfaces as `active: true, trustState: 'trusted'`.
- **All tests pass (195/195, 0 fail).** Verified locally on the PR branch.
- **CI is fully green.** All 12 checks pass.
- **Budget value (1 MiB) is well-chosen.** Large enough that budget enforcement is never triggered, keeping the tests focused on gate logic only — matches the comment in the existing two-pass test block.
- **`beforeEach`/`afterEach` isolation.** Each test gets a fresh `mkdtempSync` directory; no shared state between tests.

### Issues Found

#### Critical (Must Fix)

None.

| File | Approx. Line | Suggestion | Rationale |
|------|-------------|------------|-----------|
| `skills-loader.test.js` | new test 1 | Assert `path` value equals the actual file path, not just `typeof === 'string'` | The existing single-pass `onCommunityTrustPending` test (`fires onCommunityTrustPending callback once per pending community skill`) asserts the exact `path` value with `assert.ok(pendingCalls[0].path.endsWith('pending.md'))`. Consistency with that precedent would make regressions where the path is wrong (e.g., pass-1 realPath vs pass-2 realPath2 diverge) detectable. |
| `skills-loader.test.js` | new test 2 | Add assertion `skills[0].active === false` is accompanied by `skills[0].trustState === 'pending'` on the same object, not just `assert.equal(skills[0].active, false)` as a precondition check | Already present — noted as a positive, no action needed. Ignore. |

#### Nitpicks

- Test 1 title says "cache-miss + cache-fast-path" but the sub-label ordering in the two comment blocks is "First call: cache cold → cache-miss" then "Second call: warm cache → cache-fast-path". That ordering matches the execution order, which is ideal — no change needed, just noting it reads well.
- Minor: `pendingCallsCold` / `pendingCallsWarm` are clear names. No concern.

No out-of-scope items identified. The one suggestion above (asserting exact `path` value) is minor enough to be in-scope if the author agrees, but is not blocking.

### Architecture Notes

The PR is purely additive test coverage with zero production code changes. It correctly targets the symmetric duplication between the single-pass loader path and the two-pass `_collectCandidates` path — a known maintenance burden flagged in the source comments. No architecture concerns.

### Verdict

- [x] **Approve** — Ready to merge. All three tests are correct, well-scoped, and green. The only suggestion (asserting exact `path` value in test 1 for parity with the single-pass equivalent) is a minor nicety, not a blocker.

PR #3425, opened 2026-05-07
fix(dashboard): SkillsPanel pending row path styling (#3368)

blamechris:
Closes #3368

## Summary

- Add `.skill-pending-path` CSS rule to `components.css` that overrides the 22 px `margin-left` inherited from `.skill-meta` — pending rows have no checkbox, so the indent was misaligned
- Add `overflow: hidden; text-overflow: ellipsis; white-space: nowrap` to the same rule so long absolute paths (e.g. `/Users/chris/very/long/nested/project/.chroxy/skills/community/alice/skill.md`) truncate with an ellipsis rather than overflowing the panel
- The `title` tooltip on the path element is already present in the component; hover accessibility is unchanged
- Two new tests assert `skill-pending-path` class is applied and the title attribute equals the full path

## Test plan

- [x] `path element carries skill-pending-path class for overflow truncation (#3368)` — new test passes
- [x] `path element has title tooltip with full path (#3368)` — new test passes
- [x] Full dashboard test suite: 96 files / 1418 tests pass
- [x] TypeScript typecheck clean

### DIFF
--- a/packages/dashboard/src/components/SkillsPanel.test.tsx
+++ b/packages/dashboard/src/components/SkillsPanel.test.tsx
@@ -539,6 +539,41 @@ describe('SkillsPanel (#3209)', () => {
       expect(screen.queryByTestId('skill-pending-path-alice/alice-skill')).toBeNull()
     })
 
+    // #3368: path span must carry .skill-pending-path so CSS can zero the
+    // margin-left (no checkbox in pending rows) and enable overflow truncation.
+    it('path element carries skill-pending-path class for overflow truncation (#3368)', () => {
+      renderPanel({
+        skills: [],
+        pendingCommunitySkills: [{
+          name: 'alice-skill',
+          author: 'alice',
+          path: '/Users/chris/very/long/nested/project/.chroxy/skills/community/alice/skill.md',
+        }],
+        onGrantTrust: vi.fn(),
+        capabilities: { skillTrustGrant: true },
+      })
+      const pathEl = screen.getByTestId('skill-pending-path-alice/alice-skill')
+      expect(pathEl.classList.contains('skill-pending-path')).toBe(true)
+    })
+
+    // #3368: title tooltip must be present so the full path is accessible on
+    // hover even when the text is truncated by overflow: hidden / text-overflow.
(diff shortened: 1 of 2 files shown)


blamechris:
## Code Review: PR #3425

### Summary

Targeted CSS-only fix for #3368. Adds `.skill-pending-path` to `components.css` to zero out the 22 px `margin-left` inherited from `.skill-meta` (which exists to align metadata under the checkbox in always-on/manual rows — inapplicable to pending rows that have no checkbox) and to apply `overflow: hidden / text-overflow: ellipsis / white-space: nowrap` so long absolute paths truncate gracefully. The class was already applied in the component (`className="skill-meta skill-pending-path"`); this PR adds only the missing CSS rule and two regression tests.

### Strengths

- Minimal, surgical diff — zero component-logic changes, zero risk of behavioural regression
- CSS specificity is clean: `.skill-pending-path` is a sibling class on the same element as `.skill-meta`, so `margin-left: 0` reliably overrides `margin-left: 22px` without needing `!important` or deep selectors
- `display: block` is already set on `.skill-meta`, so the truncation trio (`overflow: hidden; text-overflow: ellipsis; white-space: nowrap`) will work correctly — no extra layout change needed
- `title` tooltip on the path element was already present in the component; the second new test pins that contract explicitly, which is good
- Comment block in CSS clearly explains *why* the override exists and references the issue number — easy to audit later
- New tests reuse the existing `renderPanel` / `getByTestId` patterns exactly, no friction

### Issues Found

#### Critical (Must Fix)

None.

None.

#### Nitpicks (Optional)

- The two new tests duplicate the fixture setup (`renderPanel({ skills: [], pendingCommunitySkills: [{ name, author, path }], onGrantTrust, capabilities })`) identically. A shared `const longPath = …; const setup = () => renderPanel(…)` at the describe-block level would reduce copy-paste, but this is a genuinely minor style point given the tests are only 15 lines each and self-contained.

### Deferred Items

None — scope is entirely contained in this PR.

### Architecture Notes

No protocol, state, or component-logic changes. CSS-only. Fits squarely in the dashboard theming layer; no cross-package impact.

### Verdict

PR #3426, opened 2026-05-07
fix(server): K8sBackend port range validation (#3386)

blamechris:
Closes #3386

## Changes

- `_parseContainerPort`: added `port < 1 || port > 65535` guard; returns `null` for out-of-range values
- Callsite in `createEnvironment`: on `null` result, emits `log.warn` with the offending entry and skips it (same pattern as the unparseable mount warning)
- 6 new tests covering: valid port accepted, port 0 rejected, port 65536 rejected, negative port rejected, non-integer string rejected, mixed list drops invalid and keeps valid sibling

### DIFF
--- a/packages/server/src/environments/backends/k8s.js
+++ b/packages/server/src/environments/backends/k8s.js
@@ -305,7 +305,11 @@ export class K8sBackend {
       for (const entry of forwardPorts) {
         // Accept bare port number or "hostPort:containerPort" strings.
         const containerPort = _parseContainerPort(entry)
-        if (containerPort && containerPort !== AGENT_PORT) {
+        if (containerPort === null) {
+          log.warn(`createEnvironment: ignoring invalid forwardPorts entry "${entry}" (must be 1-65535)`)
+          continue
+        }
+        if (containerPort !== AGENT_PORT) {
           ports.push({ containerPort })
         }
       }
@@ -1507,7 +1511,7 @@ function _parseMountString(mountStr) {
  *   - A bare number or numeric string: "8080" → 8080
  *   - A "hostPort:containerPort" string: "9000:8080" → 8080
  *
- * Returns null for non-numeric or zero values.
+ * Returns null for non-numeric, zero, or out-of-range (> 65535) values.
  *
  * @param {string|number} entry
  * @returns {number | null}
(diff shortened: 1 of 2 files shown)


blamechris:
## Code Review: PR #3426

### Summary

Fixes #3386 by adding an upper-bound guard to `_parseContainerPort` so ports outside 1–65535 return `null`. The callsite in `createEnvironment` is refactored from a truthy check to an explicit `=== null` check (correct, since valid ports can be any integer), emitting `log.warn` on bad entries before skipping them. Six new tests cover boundaries and mixed-list behaviour.

### Strengths

- The `=== null` refactor at the callsite is the right call — the old `if (containerPort && ...)` would have silently dropped port `0` but accepted anything truthy, which was already guarded only by `> 0`. The new two-step `null` check + AGENT_PORT dedup is cleaner and explicit.
- `log.warn` pattern matches the existing unparseable-mount warning at line 288 — consistent style.
- JSDoc on `_parseContainerPort` updated to reflect the new behaviour.
- Tests follow the existing `createMockApi` fixture pattern — no new helpers needed.
- Boundary cases are thorough: `0`, `-1`, `65536`, `99999`, non-numeric string, and the mixed-list survival test.

### Issues Found

#### Critical (Must Fix)

None.

| `k8s.test.js` | New `it` descriptions | Replace "silently drops" with "drops (with warn)" or similar | The code emits `log.warn`, so "silently drops" is a mild misnomer. Low stakes, but descriptions are documentation. |
| `k8s.test.js` | Port range tests | Add a `"hostPort:65536"` colon-format case | The existing colon tests only cover valid values; an out-of-range colon string (e.g. `'9000:65536'`) is a natural extension of the new guard and is not covered. |

#### Nitpicks

- The new JSDoc still says `Returns null for non-numeric, zero, or out-of-range (> 65535) values` but the code now uses `port < 1` (which excludes `0` and negatives), so the phrase should arguably read `non-numeric, or out-of-range (< 1 or > 65535)`. Trivial but worth keeping in sync.

### Deferred Items

| Colon-format out-of-range test (`"9000:65536"`) | #3427 | Narrow gap in coverage; doesn't block this fix shipping |

### Architecture Notes

Change is entirely self-contained inside `K8sBackend`. No WS protocol impact, no cross-package changes, no state machine side effects. The `log.warn` approach matches how the mount-string parser handles bad input — consistent operational behaviour for ops teams reading logs.

### Verdict

- [x] Approve — Ready to merge (suggestions and nitpick are non-blocking)

PR #3427, opened 2026-05-07
docs(server): K8sBackend memory quantity SI-to-binary semantics (#3384)

blamechris:
Closes #3384

## Summary

- Adds a JSDoc block to `_normaliseMemoryQuantity` explaining the SI vs binary-SI approximation: Docker `g`/`m`/`k` are decimal units (10^9, 10^6, 10^3) while Kubernetes `Gi`/`Mi`/`Ki` use powers of two, resulting in ~7.4 % / ~4.9 % overshoot for the common `g`/`m` cases
- Documents that the conversion intentionally errs generous (safe) and that operators needing exact semantics can pass a K8s-native quantity string (e.g. `"2Gi"`, `"2000000000"`) which is returned unchanged
- Adds a one-line cross-reference comment at the call site in `createEnvironment`

### DIFF
--- a/packages/server/src/environments/backends/k8s.js
+++ b/packages/server/src/environments/backends/k8s.js
@@ -314,6 +314,8 @@ export class K8sBackend {
     // 6. Build resource limits/requests.
     //    Convert Docker-style suffixes (g → Gi, m → Mi) to K8s quantity strings.
     //    A plain integer/float string (e.g. "2", "0.5") is already valid K8s CPU syntax.
+    //    Note: the g→Gi / m→Mi mapping errs generous (~7 % / ~5 % over SI).
+    //    See _normaliseMemoryQuantity JSDoc for details.
     const resources = {}
     if (memoryLimit || cpuLimit) {
       const limits = {}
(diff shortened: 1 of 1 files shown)


blamechris:
## Code Review: PR #3427

### Summary

Doc-only change adding a JSDoc block to `_normaliseMemoryQuantity` and a cross-reference comment at the call site in `createEnvironment`. Explains why Docker's single-letter suffixes (`g`/`m`/`k`) map to K8s binary-SI quantities (`Gi`/`Mi`/`Ki`) with a ~7.4 % / ~4.9 % generous overshoot, and documents the intentional design decision to err on the generous side.

### Strengths

- The math is correct. `2 GiB = 2,147,483,648 B` vs `2 GB = 2,000,000,000 B` → 7.37 % overshoot; `512 MiB = 536,870,912 B` vs `512 MB = 512,000,000 B` → 4.86 % overshoot. Numbers in the JSDoc match.
- Existing function behaviour is unchanged — no risk of regression.
- The escape hatch ("supply a K8s-native quantity string") is clearly documented alongside the approximation.
- Call-site cross-reference is minimal and points readers to the right place without duplicating prose.
- Follows the existing JSDoc style in the file (mapping table, `@param`, `@returns`).

### Issues Found

#### Critical (Must Fix)

None.

None.

#### Nitpicks (Optional)

- The JSDoc uses `"g" = 10^9 bytes` inline prose (`10^9` / `10^6` / `10^3`). In rendered JSDoc viewers these render as plain text, not superscripts, which is fine — but `1e9` / `1e6` / `1e3` would be equally readable and consistent with how JS numeric literals look. Minor preference only.
- The call-site comment says `~7 % / ~5 %` (rounded). The JSDoc says `~7.4 % / ~4.9 %`. Both are correct; the call-site rounding is intentional for brevity.

### Architecture Notes

`_normaliseMemoryQuantity` is a private helper with a single call site. The JSDoc is the appropriate place for this explanation — it lives right above the implementation and will surface in any IDE hover or `jsdoc` output. No structural concerns.

### Verdict

- [x] **Approve** — Ready to merge. Doc-only, math verified, no behaviour change.

PR #3428, opened 2026-05-07
docs: community/ namespace lowercase requirement for portability (#3379)

blamechris:
Closes #3379

## Summary

- Adds a **Community namespace** section to `docs/skills.md` — the first user-facing documentation for the `community/<author>/` skill layout
- Documents the Linux case-sensitivity requirement: only `community/` (exact lowercase) is recognised on Linux; `Community/` or `COMMUNITY/` are silently treated as ordinary top-level skill directories and bypass the trust gate
- Documents the trust file migration caveat: `skills-trust.json` stores `by-path` keys lowercased on macOS/Windows and verbatim on Linux, so migrating the ledger across platforms may require re-trusting community skills (author-level grants in `by-author` are unaffected)

## What changed

`docs/skills.md` — 32 lines added, no code changes.

### DIFF
--- a/docs/skills.md
+++ b/docs/skills.md
@@ -86,6 +86,38 @@ git clone git@github.com:you/my-chroxy-skills.git ~/.chroxy/skills
 
 Repo-overlay skills (`<repo>/.chroxy/skills/`) are intended to live alongside the project — commit them to the repo so every contributor's Chroxy session inherits the same conventions automatically.
 
+## Community namespace
+
+Skills shared by third parties live under a `community/<author>/` subdirectory of any skills root:
+
+```
+~/.chroxy/skills/
+├── coding-style.md
+└── community/
+    └── alice/
+        ├── typescript-conventions.md
+        └── review-checklist.md
+```
+
+Community skills are subject to a first-activation trust prompt — Chroxy will not inject them until you explicitly grant trust for the author. This prevents a cloned or downloaded skill set from silently influencing the model's behaviour without your consent.
+
+### Linux case-sensitivity requirement
+
+**Always name the directory `community/` in lowercase.** The behaviour differs by platform:
+
+- **macOS and Windows** — filesystems are case-insensitive by default. `Community/`, `COMMUNITY/`, and `community/` all refer to the same directory and are all recognised as the community namespace.
+- **Linux** — the filesystem is case-sensitive. Only the exact name `community/` is recognised. A directory named `Community/` or `COMMUNITY/` on Linux is silently treated as an ordinary top-level skills directory and is **not** subject to the community trust gate — its skills are either loaded without a trust prompt or rejected, depending on configuration.
+
+Using lowercase `community/` everywhere is the portable convention that works on all platforms.
+
+### Trust file migration between platforms
+
+The trust ledger (`~/.chroxy/skills-trust.json`) stores path-based grant records. On macOS and Windows the stored paths are lowercased; on Linux they are stored verbatim as resolved by the filesystem.
+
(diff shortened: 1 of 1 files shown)


blamechris:
## Code Review: PR #3428

### Summary

Doc-only PR that adds a **Community namespace** section (32 lines) to `docs/skills.md`. Covers the `community/<author>/` directory layout, the Linux case-sensitivity requirement, and the cross-platform `skills-trust.json` migration caveat. Closes #3379.

### Strengths

- Precisely mirrors the implementation: `skills-loader.js` uses `_PATH_COMPARE_CASE_INSENSITIVE` (`darwin | win32`) and `_normalizePathKey` in `skills-trust.js` lowercases keys only on those platforms — the docs match the actual code behaviour exactly.
- The Linux "silently treated as ordinary top-level skills directory" consequence is the most surprising footgun; calling it out explicitly with bold text is the right call.
- Author-level grants surviving migration (only `by-path` keys are at risk) is a useful precision — most users will hold author grants, so the caveat lands with the right severity.
- Placement between "Sharing skills" and "Scope" is natural — it follows the flow from basic setup → sharing → community extensions → scope summary.
- Workaround instruction is actionable and covers both the UI and the manual-edit path.

### Issues Found

#### Critical (Must Fix)

None.

| `docs/skills.md` | Trust file migration section | Clarify that `by-path` keys in `skills-trust.json` are the *ledger lookup keys*, not the stored display paths. The current phrasing "stored paths are lowercased" could be read as "the file paths written to disk are altered", when it is actually only the JSON key used for lookup that differs. | `skills-trust.js` `_normalizePathKey` normalises the *key* — `flush()` writes `byPathOut` verbatim from `communityTrust.byPath` which is populated on load via the same normalised key. The distinction matters when a user is manually editing the JSON. |
| `docs/skills.md` | Linux case-sensitivity section | Consider adding a one-liner about *what happens* to skills loaded from `Community/` (not `community/`) on Linux: they are loaded without a trust prompt (or blocked, depending on config). The current text says they "bypass the trust gate" but does not say whether that means unconditional load or unconditional block. | `skills-loader.js` line 389–390: a directory that does not compare-equal to `'community'` is skipped entirely in the community walk, so those skills are loaded as ordinary top-level skills — they go through hash checking but not the community-trust-prompt path. A user wondering "will my skills still load if I accidentally used `Community/`?" can't tell from the doc alone. |

#### Nitpicks (Optional)

- "filesystems are case-insensitive by default" — technically APFS can be formatted case-sensitive; the code acknowledges this with the "intentionally does NOT probe the actual mount" comment. The doc's "by default" qualifier is sufficient and matches the code's own documented heuristic — no change needed, just flagging for awareness.
- The tree diagram uses only one author (`alice`) with two skill files. A two-author example would better illustrate that the trust prompt is per-author, not per-file — but this is style preference and the single-author example is simpler for first-time readers.

| Clarify that it is the ledger *key* that is normalised, not the stored path value | #3434 | Minor precision improvement; the existing phrasing is not wrong and a separate commit keeps this PR focused on closing #3379. |
| State the load outcome for skills in a wrongly-cased directory on Linux | #3435 | Additive detail; the current warning is already the critical safety net. Good for a follow-up docs pass. |

### Architecture Notes

The new section is pure documentation and introduces no runtime changes. It accurately reflects the two-file implementation split (`skills-loader.js` for discovery, `skills-trust.js` for the ledger) and the v2 storage shape (`communityTrust.by-author` / `communityTrust.by-path`). Nothing in the diff touches source, tests, CI, or protocol messages.

### Verdict

The documentation is accurate, well-scoped, and directly addresses both acceptance criteria in #3379. The two suggestions above are worth tracking but do not block merge.

PR #3429, opened 2026-05-07
test(server): docker exec --workdir spaces (#3362)

blamechris:
Closes #3362

## Summary

- Add regression test for `DockerBackend.execInEnvironment` with a `cwd` containing a space (`/work space/src`)
- Asserts the path arrives as a single unescaped argv element (`args[wIdx + 1] === '/work space/src'`)
- Asserts the path is not split (`/work` and `space/src` must not appear as separate elements)

## Why

`execFile` receives args as an array, so spaces in paths are naturally safe — no quoting or escaping needed. Without this test, a future refactor that naively joins the args array into a shell string would break spaces-in-paths silently. The test documents and protects this behavior.

## Test plan

- [x] New test passes: `node --test tests/environments/backends/docker.test.js` — 43/43
- [x] No new failures in full suite (`npm test` — 78 pre-existing failures unchanged)

### DIFF
--- a/packages/server/tests/environments/backends/docker.test.js
+++ b/packages/server/tests/environments/backends/docker.test.js
@@ -746,6 +746,27 @@ describe('DockerBackend.execInEnvironment()', () => {
     assert.ok(execCall.args.indexOf('ctr-abc') > wIdx + 1)
   })
 
+  it('passes --workdir as a single argv element when cwd contains spaces', async () => {
+    const mockExec = createMockExecFile({ results: { exec: '' } })
+    const backend = new DockerBackend({ _execFile: mockExec })
+
+    await backend.execInEnvironment('ctr-abc', {
+      cmd: 'pwd',
+      cwd: '/work space/src',
+    })
+
+    const execCall = mockExec.calls.find(c => c.args[0] === 'exec')
+    const wIdx = execCall.args.indexOf('--workdir')
+    assert.ok(wIdx >= 0, '--workdir flag must be present')
+    // The path with a space must arrive as one unescaped array element — not split
+    // or shell-quoted. This guards against any future refactor that naively joins
+    // args into a shell string before passing to docker exec.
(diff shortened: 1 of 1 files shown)


blamechris:
## Code Review: PR #3429

### Summary

Adds a single regression test to `DockerBackend.execInEnvironment()` asserting that a `cwd` value containing a space (`/work space/src`) is passed to `execFile` as one unescaped array element — not split. Closes #3362. No production code changes.

### Strengths

- **Follows established file pattern exactly.** Uses `createMockExecFile` and the `args.indexOf('--workdir')` idiom already in the describe block — no new helpers or imports needed.
- **Three complementary assertions.** `assert.equal` on the positive case, plus two `assert.ok(!includes(...))` guards for the split-path failure modes. Covers the regression from both directions.
- **Comment explains the _why_, not the _what_.** The inline note about naive array-join refactors is exactly the kind of forward-looking documentation that makes a regression test worth keeping.
- **Self-contained.** Fresh `createMockExecFile` instance; no shared state with adjacent tests.
- **Confirmed green.** Ran the suite locally against the PR branch: 43/43 pass, 0 fail.

### Issues Found

#### Critical (Must Fix)

None.

None.

#### Nitpicks (Optional)

- The PR description says "43/43" but main currently has 42 tests in this file — the count is correct *after* this test is added. Minor wording ambiguity; no action needed.

### Deferred Items

None — test-only PR with well-scoped intent.

### Architecture Notes

`execInEnvironment` uses `execFile` (array args), so spaces in paths are inherently safe at the OS level. The test documents this invariant so that if `execInEnvironment` is ever refactored to shell out via a string (e.g. `exec` with `shell: true`), the breakage is immediately visible. Fits cleanly in the existing `DockerBackend.execInEnvironment()` describe block.

### Verdict

- [x] **Approve** — Ready to merge. Correct assertions, green suite, no structural concerns.

PR #3430, opened 2026-05-07
fix(server): chroxy-pod-agent spawn-before-eviction ordering (#3392)

blamechris:
Closes #3392

## Summary

- Moves `_enforceSessionCap()` to after `_spawnFn()` succeeds in `_handleSpawn`
- Previously the cap evicted an idle session before attempting to spawn; a synchronous spawn failure (ENOENT, EACCES, ENOMEM) left the pod with one fewer session slot and no new session to fill it
- Fix is minimal: remove the early cap call, spawn first, enforce cap only once the child exists and will occupy a slot

## Changes

- `packages/server/sidecar/agent.js`: reorder the two operations — delete 4 lines, add 6 comment+call lines at the right place
- `packages/server/tests/sidecar/agent.test.js`: new test in `max-sessions cap` describe — verifies that a synchronous spawn failure at the cap limit leaves `_sessions.size` at 1 (the original session is not evicted)

## Test plan

- [ ] `node --test tests/sidecar/agent.test.js` — 64 pass, 0 fail
- [ ] New test: `failed synchronous spawn does not evict an existing session (#3392)` — passes
- [ ] Existing max-sessions cap tests still pass (oldest idle evicted, fallback to active)

### DIFF
--- a/packages/server/sidecar/agent.js
+++ b/packages/server/sidecar/agent.js
@@ -562,10 +562,6 @@ export class PodAgent {
       return
     }
 
-    // Enforce the concurrent session cap before creating a new session. Evicts
-    // the oldest idle session if needed so the Map never exceeds _maxSessions.
-    this._enforceSessionCap()
-
     // stdin option controls how the child's stdin is set up (#3329):
     //   'pipe'    — (default) writable stdin; client feeds data via stdin frames.
     //               Required for --input-format stream-json workflows.
@@ -597,6 +593,12 @@ export class PodAgent {
       this._send(ws, { type: 'error', message: `spawn failed: ${err.message}` })
       return
     }
+
+    // Spawn succeeded -- now enforce the concurrent session cap. Evicts the
+    // oldest idle session if needed so the Map never exceeds _maxSessions.
+    // Enforcing after spawn ensures a failed spawn (ENOENT, EACCES, etc.)
+    // never evicts an existing session unnecessarily (#3392).
+    this._enforceSessionCap()
 
     // Assign a sessionId and create the session object.
     const sessionId = randomUUID()
--- a/packages/server/tests/sidecar/agent.test.js
+++ b/packages/server/tests/sidecar/agent.test.js
@@ -1748,6 +1748,72 @@ describe('PodAgent', () => {
         await capAgent.close()
       }
     })
+
+    it('failed synchronous spawn does not evict an existing session (#3392)', async () => {
+      // Arrange: one idle session already in the map at the cap limit.
+      // The second spawn throws synchronously -- _enforceSessionCap must NOT
+      // have run before the throw, so the existing session must survive.
+      let spawnCallCount = 0
+      const mockChild = new EventEmitter()
+      mockChild.stdout = new PassThrough()
+      mockChild.stderr = new PassThrough()
+      mockChild.kill = () => true
+
+      const spawnFn = (_cmd, _args, _opts) => {
+        spawnCallCount += 1
+        if (spawnCallCount === 1) {
+          // First call succeeds -- establishes the existing session.
+          return mockChild
+        }
+        // Second call fails synchronously (e.g. binary not found).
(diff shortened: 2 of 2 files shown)


blamechris:
## Code Review: PR #3430

### Summary

Moves `_enforceSessionCap()` from before `_spawnFn()` to after it — specifically to the gap between a successful synchronous spawn and `_sessions.set()`. A synchronous throw from `_spawnFn` (ENOENT, EACCES, ENOMEM) now returns early without touching the session map. One new test verifies the invariant at cap=1.

### Strengths

- Fix is minimal and surgical — 4 lines deleted, 6 lines added, no new abstractions
- The new placement is the only correct position: after spawn confirms a child exists and needs a slot, before that slot is registered
- Existing cap tests (oldest-idle eviction, fallback-to-active eviction) are untouched and still pass
- New test is well-structured: establishes one idle session at the cap limit, then triggers a synchronous throw on the second spawn, and asserts both `_sessions.size` and `_sessions.has(sid1)` — covers the exact invariant from #3392
- Error frame assertion (`assert.match(errFrame.message, /spawn failed/)`) verifies the client gets a meaningful error rather than silence
- Comment on the moved call explains the `why` concisely, including the issue reference

### Issues Found

#### Critical (Must Fix)

None.

| `agent.js` | 451 | Update the `_enforceSessionCap` header comment: "Called before inserting a new session" is now slightly inaccurate | It's still called before `_sessions.set`, but the word "before inserting" implies it runs before anything spawn-related. A more precise phrasing: "Called after spawn succeeds, before registering the new session." Low stakes but the comment was already used to justify the old (buggy) ordering — worth keeping it accurate. |

#### Nitpicks (Optional)

- Test uses `setTimeout(r, 30)` twice for timing. 30 ms is tight in a loaded CI environment. The existing cap test uses `setTimeout(r, 20)` for spawn and `setTimeout(r, 30)` for disconnect settle — the new test mirrors this pattern, which is fine. No change needed, just a note.
- `resumeTimeoutMs: 60_000` comment says "long TTL so idle timer does not race" — correct and matches the existing cap test.

### Ordering Analysis

The concern stated in the PR focus: **does spawn-then-evict still work correctly when cap=N and we're at N?**

Yes. The sequence is fully synchronous within a single Node.js event loop turn:

1. `_spawnFn()` succeeds → child exists in memory but not in `_sessions`
2. `_enforceSessionCap()` runs: `_sessions.size` is still N → evicts oldest idle → `_sessions.size` = N-1
3. `_sessions.set(sessionId, session)` → `_sessions.size` = N again

There is no async gap between steps 2 and 3. `_handleSpawn` has no `await` or `.then()` calls — it runs synchronously to completion before the event loop can deliver another message. Two concurrent `_handleSpawn` calls cannot interleave. The window between eviction and insertion is safe.

**The async spawn failure path is unaffected.** When `_spawnFn` returns a child object (no throw) but the OS fires an async `error` event later, the flow is: eviction fires → session inserted into map → `child.on('error')` eventually removes it via `_sessions.delete`. This is pre-existing behavior unchanged by this PR, and is already covered by the `async spawn errors` test suite.

**Edge case: `_maxSessions = 1`, all sessions active.** `_enforceSessionCap` falls back to evicting the globally oldest session even if it has a live `activeWs`. This was pre-existing and is unchanged. The new test uses a disconnected (idle) session as the eviction target, so it doesn't exercise the live-WS-eviction fallback — but that path isn't touched by this PR.

| Update stale `_enforceSessionCap` header comment ("Called before inserting a new session" → "Called after spawn succeeds, before registering the new session") | #3448 | One-line comment fix; doesn't affect correctness and is separate from the ordering fix |

### Architecture Notes

This change tightens the invariant that "a session slot is never consumed unless a child actually exists to fill it." The fix belongs entirely in `_handleSpawn` — no callers of `_enforceSessionCap` are affected. The async spawn error handler (`child.on('error')`) remains the correct cleanup path for async failures and is intentionally left alone.

### Verdict

- [x] **Approve** — Correct fix, minimal footprint, well-tested. The ordering is sound and the new test covers the exact failure mode from #3392.

PR #3431, opened 2026-05-07
fix(server): K8sBackend dedup concurrent _readAgentToken (#3371)

blamechris:
Closes #3371

## Problem

Two concurrent `streamCliInEnvironment` calls on the same pod with an empty `_agentTokens` cache both evaluated the `||` lazy-fetch chain simultaneously and called `_readAgentToken` before either could populate the map. The result was two redundant K8s API reads for the same Secret.

## Solution

Add a `_pendingAgentTokens: Map<podName, Promise>` that stores the in-flight promise while a fetch is in progress. Concurrent callers return the same promise and share the result. The entry is cleared in a `finally` block so:
- concurrent calls coalesce onto the same fetch
- subsequent calls (reconnect path, Secret rotation) always start a fresh fetch

## Changes

- `packages/server/src/environments/backends/k8s.js`
  - Add `this._pendingAgentTokens = new Map()` in constructor
  - Convert `async _readAgentToken()` to a synchronous method that wraps an `async` IIFE — stores the promise in `_pendingAgentTokens`, returns it to concurrent callers, deletes it in `finally`

- `packages/server/tests/environments/backends/k8s.test.js`
  - Update the sequential-call test: comment corrected, now asserts `apiCalls === 2` (back-to-back calls each issue their own fetch after the first resolves)
  - Add `describe('K8sBackend._readAgentToken() concurrent deduplication (#3371)')` with three tests:
    - two concurrent calls share one API request and both get the correct token
    - `_pendingAgentTokens` is cleared after resolution
    - `_pendingAgentTokens` is cleared even when the API throws

### DIFF
--- a/packages/server/src/environments/backends/k8s.js
+++ b/packages/server/src/environments/backends/k8s.js
@@ -135,6 +135,12 @@ export class K8sBackend {
     // K8s-specific agentToken arg through the manager.
     this._agentTokens = new Map()
 
+    // In-flight _readAgentToken promises, keyed by pod name. Coalesces
+    // concurrent callers so only one K8s API request is made per pod at a time
+    // (deduplicates the race where two streamCliInEnvironment calls arrive
+    // before either has populated _agentTokens).
+    this._pendingAgentTokens = new Map()
+
     // Reconnect backoff config — injectable for deterministic unit tests.
     // Validate the delay schedule defensively: an empty array (or non-finite
     // entries) would make `delay` undefined and let `setTimeout` fire on the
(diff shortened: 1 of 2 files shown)


blamechris:
## Code Review: PR #3431

### Summary

Adds a `_pendingAgentTokens: Map<podName, Promise>` to `K8sBackend` to coalesce concurrent `_readAgentToken` calls for the same pod into a single K8s API request. The in-flight promise is stored synchronously before the first `await`, and cleared in a `finally` block so every post-settlement caller starts a fresh fetch. Three new tests cover the dedup invariant, post-settlement clear, and error-path clear.

### Strengths

- **Correct placement of `set` before any `await`**: The IIFE creates `pending`, then `_pendingAgentTokens.set(podName, pending)` is called synchronously before returning. There is no gap in which a concurrent caller could slip past the guard and issue a second API call. This is the only correct ordering in a single-threaded event loop.
- **`finally` rather than `then`/`catch`**: Using `finally` for the delete ensures cleanup on both resolution and rejection paths without duplicating the call.
- **Settle-and-clear race is safe**: Asked specifically about the window between "finally fires delete" and "a new caller arrives". In Node's single-threaded event loop, the `finally` microtask and any new synchronous call cannot interleave. A caller that arrives after the `finally` delete has completed will correctly see `has(podName) === false` and start a fresh fetch. A caller that arrives *before* the `finally` runs will see the still-live entry and coalesce. No intermediate state is observable.
- **Reconnect path preserved**: `reconnectAgentToken` calls `_readAgentToken` and awaits it. Because `_pendingAgentTokens` is cleared on settlement, a reconnect that fires after a prior fetch completes always issues a new K8s read — which is the correct behaviour for Secret rotation (new token value will be fetched and overwrite `_agentTokens`).
- **Secret rotation preserved**: `_agentTokens.set(podName, token)` runs inside the IIFE's `try` block, and `_pendingAgentTokens.delete` runs in `finally`. Any rotate-then-reconnect sequence: first reconnect writes the new token to `_agentTokens`, clears `_pendingAgentTokens`; subsequent `streamCliInEnvironment` hits `_agentTokens.get(podName)` on the fast path and uses the rotated token without another API call. Rotation semantics are unchanged from before this PR.
- **Updated sequential-call test is accurate**: The old comment said "does not short-circuit — caching is done by streamCliInEnvironment" which was misleading. The new comment correctly explains that sequential calls each issue their own fetch because `_pendingAgentTokens` is cleared between them. `apiCalls === 2` is the right assertion.
- **Test uses deferred promise**: The concurrent test correctly holds the API response open with a `new Promise` + `resolveSecret` pattern so both `p1` and `p2` are genuinely in-flight before either resolves. `assert.strictEqual(p1, p2)` is a strong identity check, not just value equality.
- **Style compliance**: No semicolons, single quotes, ES modules — all consistent with server code standards.

### Issues Found

#### Critical (Must Fix)

None.

| `k8s.js` | ~930 | Consider also checking `_pendingAgentTokens` in `streamCliInEnvironment`'s `tokenOrPromise` chain (`opts.agentToken \|\| _agentTokens.get \|\| _pendingAgentTokens.get \|\| _readAgentToken`) | The current chain calls `_readAgentToken` which *internally* coalesces, so the dedup works. This is functionally correct. A direct check in the chain would make the dedup intent visible at the call site, but the internal guard is sufficient and cleaner. Low priority. |

#### Nitpicks (Optional)

- The `describe` block title includes `(#3371)` — this matches the existing `(#3339)` pattern in the file, so it's consistent. Fine as-is.
- No missing `destroy()` cleanup needed — `_pendingAgentTokens` does not need explicit teardown since the promises themselves don't hold any external resources; they will resolve/reject naturally when the K8s API call completes.

None. The suggestion above is a cosmetic/readability point and not worth a tracked issue.

### Architecture Notes

The change is fully contained within `K8sBackend`. The two-layer token cache (`_pendingAgentTokens` for in-flight coalescing, `_agentTokens` for post-settlement caching) mirrors a standard request-deduplication pattern and does not affect any WS protocol paths, session lifecycle, or other backends. The `streamCliInEnvironment` → `tokenOrPromise` chain is unchanged at the call site.

### Verdict

PR #3433, opened 2026-05-07
test(server): cap eviction all-active fallback (#3395)

blamechris:
Closes #3395

## Summary

Adds coverage for the fallback path in `_enforceSessionCap` where every session has an active WS client (no idle sessions to prefer). Without this path tested, a bug in the fallback loop would go undetected.

**Scenario tested:**
1. `maxSessions: 2`
2. Spawn session A, disconnect WS → goes idle
3. Spawn session B, disconnect WS → goes idle
4. Patch both sessions: cancel idle timers, set `activeWs` to a close spy, pin `lastActiveAt` (A=1000, B=2000) so ordering is deterministic
5. Spawn session C → `_enforceSessionCap` finds no idle sessions → falls back to evicting globally oldest (A)

**Assertions:**
- Session count stays at cap (2)
- Session A deleted from `_sessions`, session B retained
- Session A's child receives SIGTERM
- Session A's fake WS receives `close(1001)`
- Session B's fake WS is not closed

## Test plan

- [x] `node --test tests/sidecar/agent.test.js` — 64 pass, 0 fail
- [x] Full suite shows no new failures vs baseline (76 fail with change vs 77 without)

### DIFF
--- a/packages/server/tests/sidecar/agent.test.js
+++ b/packages/server/tests/sidecar/agent.test.js
@@ -1835,6 +1835,106 @@ describe('PodAgent', () => {
         await capAgent.close()
       }
     })
+
+    it('falls back to evicting oldest active session when all sessions are active', async () => {
+      // Cap=2. Spawn A and B sequentially (each over its own WS connection so
+      // the single-connection policy is satisfied), then simulate both sessions
+      // being "active" by patching their activeWs to a mock WS with a close
+      // spy. Spawning C triggers the fallback path in _enforceSessionCap: no
+      // idle sessions → evict globally oldest by lastActiveAt → session A is
+      // evicted with ws.close(1001).
+      const children = []
+      const spawnFn = (_cmd, _args, _opts) => {
+        const mock = createMockSpawn()
+        children.push(mock.child)
+        return mock.child
+      }
+
+      const { agent: capAgent, port: capPort } = await startAgent({
+        spawnFn,
+        maxSessions: 2,
+        resumeTimeoutMs: 999_999,  // long TTL so idle timers don't race
+      })
+
+      try {
+        // --- Session A ---
+        const ws1 = connect(capPort, TOKEN)
+        await waitOpen(ws1)
+        const ws1Msgs = []
+        ws1.on('message', (d) => { try { ws1Msgs.push(JSON.parse(d.toString())) } catch {} })
+        ws1.send(JSON.stringify({ type: 'spawn', cmd: 'claude', args: [] }))
+        await new Promise((r) => setTimeout(r, 20))
+        const sidA = ws1Msgs.find((m) => m.type === 'session_started').sessionId
+        ws1.close()
+        await new Promise((r) => setTimeout(r, 30))
+
+        // --- Session B (spawned after A so its lastActiveAt is naturally newer) ---
+        const ws2 = connect(capPort, TOKEN)
+        await waitOpen(ws2)
+        const ws2Msgs = []
+        ws2.on('message', (d) => { try { ws2Msgs.push(JSON.parse(d.toString())) } catch {} })
+        ws2.send(JSON.stringify({ type: 'spawn', cmd: 'claude', args: [] }))
+        await new Promise((r) => setTimeout(r, 20))
+        const sidB = ws2Msgs.find((m) => m.type === 'session_started').sessionId
+        ws2.close()
+        await new Promise((r) => setTimeout(r, 30))
+
+        assert.equal(capAgent._sessions.size, 2, 'two sessions present before patching')
+
+        // Simulate both sessions being actively connected: cancel idle timers,
+        // then set activeWs to a mock WS so _enforceSessionCap sees no idle
+        // sessions and must fall back to the oldest-active eviction path.
+        const closeCallsA = []
+        const fakeWsA = { close(code, reason) { closeCallsA.push({ code, reason }) } }
+        const closeCallsB = []
+        const fakeWsB = { close(code, reason) { closeCallsB.push({ code, reason }) } }
+
+        const sessionA = capAgent._sessions.get(sidA)
+        const sessionB = capAgent._sessions.get(sidB)
+
+        capAgent._cancelIdleTimer(sessionA)
+        capAgent._cancelIdleTimer(sessionB)
+        sessionA.activeWs = fakeWsA
+        sessionB.activeWs = fakeWsB
+
+        // Pin timestamps: A is definitively older than B. Without this the two
+        // spawns may happen within the same millisecond on fast machines.
+        sessionA.lastActiveAt = 1000
+        sessionB.lastActiveAt = 2000
+
+        // --- Session C: cap enforced, must evict session A (oldest active) ---
+        const ws3 = connect(capPort, TOKEN)
+        await waitOpen(ws3)
+        const ws3Msgs = []
+        ws3.on('message', (d) => { try { ws3Msgs.push(JSON.parse(d.toString())) } catch {} })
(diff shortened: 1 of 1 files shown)


blamechris:
## Code Review: PR #3433

### Summary

Adds a test for the fallback path in `_enforceSessionCap` where all sessions have an active WS connection (no idle sessions to prefer). The strategy is correct: spawn A and B via real connections, then disconnect them (making them idle), patch both sessions to re-attach fake close-spy WS objects and pin `lastActiveAt` for deterministic ordering, then spawn C — which forces the fallback loop and evicts A (the oldest by timestamp).

Two unrelated additions ride along: a JSDoc update for `imagePullPolicy` in `types.js` (part of a prior commit in the branch), and `renovate.json` which is already on `main` from #3354 and will be a no-op on merge.

All 64 tests pass (verified locally on the PR branch).

---

### Strengths

- **Correct fallback coverage.** The existing cap test only exercised the idle-preference path. This test deterministically exercises the second `for` loop in `_enforceSessionCap` — a real gap that could silently regress.
- **Deterministic ordering.** Pinning `sessionA.lastActiveAt = 1000` / `sessionB.lastActiveAt = 2000` is the right call. Without it, two spawns within 1ms on a fast machine could produce the wrong eviction winner, making the test flaky.
- **Minimal fake WS.** The `fakeWsA`/`fakeWsB` objects are lean — just a `close()` spy. The production eviction path (`_evictSession` line 442) calls `close()` directly without a `readyState` guard, so a full WS mock isn't needed and the minimal shape is correct.
- **Idle-timer cancellation.** Calling `_cancelIdleTimer()` on both sessions before patching `activeWs` correctly prevents the real idle timers from firing during the test and producing spurious evictions.
- **`finally` cleanup.** `capAgent.close()` is always called, consistent with the rest of the suite.
- **`resumeTimeoutMs: 999_999`.** A long TTL prevents a race between the idle timer and the test's own eviction assertion. The existing idle test uses `60_000`; using `999_999` here is fine and avoids the same issue.

---

### Issues Found

#### Critical (Must Fix)
None.

| `agent.test.js` | 1787 | `await new Promise((r) => setTimeout(r, 20))` / `30` scattered 5× | These are timing sleeps inherited from the existing cap test. They work but are brittle on a loaded CI machine. The existing test gets away with it because the assertions are simple. For this test, the 20ms wait before reading `session_started` is the riskiest: if the message hasn't arrived yet, `.find()` returns `undefined` and `sidA`/`sidB` are `undefined`, causing a silent wrong-assertion rather than a clear failure. Consider using `waitForMessages` (already in scope) to wait for the `session_started` frame event-driven instead of time-driven. |
| `agent.test.js` | ~1845 | `ws3` is not checked for `session_started` | Session C's `session_started` message is never read from `ws3Msgs`. This is intentional (the test doesn't need it), but adding a brief wait for C's `session_started` would guard against the test passing before eviction has actually occurred. Currently the 20ms sleep covers this, but it's implicit. |

#### Nitpicks

- `resumeTimeoutMs: 999_999` vs the adjacent test's `resumeTimeoutMs: 60_000` — no functional issue, but a comment explaining why the longer value was chosen would help readers. (The existing test has no comment either.)
- Trailing comma style on `resumeTimeoutMs: 999_999,  // long TTL...` line: two spaces before the comment vs one on the adjacent test. Trivial.

---

| Event-driven `session_started` wait in cap tests (both existing and new) | #3434 | Affects the existing idle-cap test too — scope is larger than this PR. Worth tracking as a test-reliability improvement. |

---

### Architecture Notes

The test correctly reaches into `_sessions`, `_cancelIdleTimer`, and `activeWs` — all private-by-convention (`_` prefix) but intentionally accessible for test purposes. This is consistent with every other cap/eviction test in the suite. No concerns there.

The `renovate.json` in the diff is already on `main` (landed via #3354). It will produce no diff on merge — not a problem.

---

### Verdict

- [x] **Approve** — Test is correct, deterministic, and passes cleanly. The timing-sleep concern is pre-existing and tracked as a follow-up. Ready to merge.

blamechris:
## Code Review: PR #3433

### Summary

Adds a test for the fallback path in `_enforceSessionCap` where all sessions have an active WS connection (no idle sessions to prefer). The strategy is correct: spawn A and B via real connections, then disconnect them (making them idle), patch both sessions to re-attach fake close-spy WS objects and pin `lastActiveAt` for deterministic ordering, then spawn C — which forces the fallback loop and evicts A (the oldest by timestamp).

Two unrelated additions ride along: a JSDoc update for `imagePullPolicy` in `types.js` (part of a prior commit in the branch), and `renovate.json` which is already on `main` from #3354 and will be a no-op on merge.

All 64 tests pass (verified locally on the PR branch).

---

### Strengths

- **Correct fallback coverage.** The existing cap test only exercised the idle-preference path. This test deterministically exercises the second `for` loop in `_enforceSessionCap` — a real gap that could silently regress.
- **Deterministic ordering.** Pinning `sessionA.lastActiveAt = 1000` / `sessionB.lastActiveAt = 2000` is the right call. Without it, two spawns within 1ms on a fast machine could produce the wrong eviction winner, making the test flaky.
- **Minimal fake WS.** The `fakeWsA`/`fakeWsB` objects are lean — just a `close()` spy. The production eviction path (`_evictSession` line 442) calls `close()` directly without a `readyState` guard, so a full WS mock isn't needed and the minimal shape is correct.
- **Idle-timer cancellation.** Calling `_cancelIdleTimer()` on both sessions before patching `activeWs` correctly prevents the real idle timers from firing during the test and producing spurious evictions.
- **`finally` cleanup.** `capAgent.close()` is always called, consistent with the rest of the suite.
- **`resumeTimeoutMs: 999_999`.** A long TTL prevents a race between the idle timer and the test's own eviction assertion. The existing idle test uses `60_000`; using `999_999` here is fine and avoids the same issue.

---

### Issues Found

#### Critical (Must Fix)
None.

| `agent.test.js` | ~1787 | `await new Promise((r) => setTimeout(r, 20))` / `30` scattered 5× | These timing sleeps are brittle on a loaded CI machine. The 20ms wait before reading `session_started` is the riskiest: if the message hasn't arrived yet, `.find()` returns `undefined` and `sidA`/`sidB` are `undefined`, causing a silent wrong-assertion rather than a clear failure. `waitForMessages` (already in scope) would make this event-driven and reliable. |
| `agent.test.js` | ~1845 | `ws3` `session_started` not awaited | Session C's `session_started` is never read from `ws3Msgs`. Intentional, but the 20ms sleep is the implicit guard that eviction has completed. An event-driven wait would make this explicit. |

#### Nitpicks

- `resumeTimeoutMs: 999_999` vs the adjacent test's `60_000` — no functional issue, but a comment explaining the longer value would help future readers.
- Two spaces before the inline comment on `resumeTimeoutMs: 999_999,  // long TTL...` vs one space elsewhere. Trivial.

---

| Event-driven `session_started` wait in cap-eviction tests (both existing and new) | [#3440](https://github.com/blamechris/chroxy/issues/3440) | Affects the existing idle-cap test too — scope is larger than this PR. Worth tracking as a test-reliability improvement. |

---

### Architecture Notes

The test correctly reaches into `_sessions`, `_cancelIdleTimer`, and `activeWs` — all private-by-convention (`_` prefix) but intentionally accessible for test purposes. This is consistent with every other cap/eviction test in the suite. No concerns there.

The `renovate.json` in the diff is already on `main` (landed via #3354). It will produce no diff on merge — not a problem.

---

### Verdict

- [x] **Approve** — Test is correct, deterministic, and passes cleanly. The timing-sleep concern is pre-existing and tracked as a follow-up. Ready to merge.

PR #3436, opened 2026-05-07
test(server): Linux-side case-sensitive community namespace tests (#3378)

blamechris:
Closes #3378

## Summary

- Adds two `_isCommunityNamespace` unit tests gated on `process.platform === 'linux'` asserting that `Community/` and `COMMUNITY/` are **not** recognised as the community namespace (exact lowercase `community/` required on case-sensitive FS)
- Adds one `loadActiveSkills` walk integration test gated on Linux asserting a skill under `Community/alice/` is **not** discovered (complements the existing darwin/win32 test that asserts the opposite)
- Adds a code comment documenting the trust-transfer gap: a `skills-trust.json` grant recorded on macOS under `Community/alice/foo.md` is invisible on Linux because the path is never classified as a community namespace; the workaround is to rename the directory to lowercase `community/`

## Test plan

- [ ] CI (Linux runner) should execute the three new Linux-gated tests and assert the `isCommunity: false` / empty-skills-array invariants
- [ ] The two darwin-gated `_isCommunityNamespace` tests and one darwin-gated walk test continue to skip on Linux (verified: 3 skipped on macOS run)
- [ ] Full `skills-loader.test.js` suite: 195 tests, 0 failures, 3 skipped (macOS local run)

### DIFF
--- a/packages/server/tests/skills-loader.test.js
+++ b/packages/server/tests/skills-loader.test.js
@@ -2595,6 +2595,29 @@ describe('skills-loader', () => {
           assert.equal(result.author, 'alice')
         },
       )
+
+      // #3378: Linux (case-sensitive FS) must NOT treat Community/ as the community namespace
+      it(
+        'Community/ (capital C) is NOT a community namespace on Linux (case-sensitive FS)',
+        { skip: process.platform !== 'linux' },
+        () => {
+          const dirReal = '/skills'
+          const result = _isCommunityNamespace('/skills/Community/alice/x.md', dirReal)
+          assert.equal(result.isCommunity, false, 'Community/ must not match on Linux — exact case required')
+          assert.equal(result.author, null)
+        },
+      )
+
+      it(
+        'COMMUNITY/ (all-caps) is NOT a community namespace on Linux (case-sensitive FS)',
+        { skip: process.platform !== 'linux' },
+        () => {
+          const dirReal = '/skills'
+          const result = _isCommunityNamespace('/skills/COMMUNITY/alice/x.md', dirReal)
+          assert.equal(result.isCommunity, false, 'COMMUNITY/ must not match on Linux — exact case required')
+          assert.equal(result.author, null)
+        },
+      )
     })
 
     // ── Loader integration tests ──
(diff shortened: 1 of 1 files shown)


blamechris:
## Code Review: PR #3436

### Summary

Test/comment-only PR. Adds three Linux-gated (`{ skip: process.platform !== 'linux' }`) assertions to `skills-loader.test.js` addressing issue #3378:

1. Two `_isCommunityNamespace` unit tests confirming `Community/` and `COMMUNITY/` return `isCommunity: false` on case-sensitive Linux FS.
2. One `loadActiveSkills` integration walk test confirming skills under `Community/alice/` are not discovered on Linux.
3. A block comment documenting the trust-transfer gap when moving a `skills-trust.json` from macOS to Linux (no test — comment only, satisfying acceptance criterion 2 from the issue).

No source changes. No new dependencies.

---

### Strengths

- Platform guard is tight and idiomatic (`process.platform !== 'linux'`). The existing darwin/win32 tests use the complementary guard — the two sets form a clean pair.
- The walk test writes into the `communityDir` tmp fixture that already has `beforeEach`/`afterEach` teardown, so no temp-dir leakage.
- The trust-transfer comment is clear and actionable: names the exact cause (`_PATH_COMPARE_CASE_INSENSITIVE` is false on Linux → `'Community' !== 'community'`), the consequence (skill simply isn't discovered, no silent data loss), and the workaround (rename to lowercase). That's exactly the right level for a caveat comment.
- Three tests match the three acceptance criteria items in issue #3378 (two unit tests + one walk integration test), and the comment satisfies criterion 2.

---

### Issues Found

#### Critical (Must Fix)

None.

| File | Approx line | Suggestion | Rationale |
|------|-------------|------------|-----------|
| `skills-loader.test.js` | walk Linux test | Assert `skills.length === 0` with a brief note that the dir itself is created on a real temp FS — on macOS/Windows `mkdirSync('Community/alice')` would silently resolve to the `community/` inode if it already existed; no issue on the actual Linux CI runner, but worth a one-liner comment so a future reader understands why the test has no fallback | Low risk; the `mkdirSync` is fine in isolation on Linux. |

#### Nitpicks

- The comment block for the trust-transfer gap is appended after the closing `})` of the walk test but _before_ the outer `})` — visually it reads as a dangling comment not attached to any `it()`. Consider placing it immediately inside a descriptive `it.todo()` or a plain `it('trust-transfer caveat: moving skills-trust.json from macOS to Linux', { skip: true }, () => { /* ... */ })` so the scenario is surfaced in the test output as a `skipped` entry rather than silent prose. This is optional; a comment is acceptable per the issue's acceptance criteria.

---

### Deferred Items

None. All items from issue #3378 acceptance criteria are addressed within this PR scope.

---

### Architecture Notes

This PR has no runtime surface — test file only. The implementation under test (`_isCommunityNamespace` at `packages/server/src/skills-loader.js:1669`) is unchanged and correct: `_PATH_COMPARE_CASE_INSENSITIVE` is only true on darwin/win32, so the Linux case-sensitive path was already working; these tests make the contract explicit and CI-verifiable on Linux runners.

---

### Verdict

The three tests fulfill the issue's acceptance criteria exactly. No source changes means zero regression risk. The trust-transfer comment is well-placed and actionable. No blocking issues found.

PR #3437, opened 2026-05-07
docs(server): community namespace case-sensitivity heuristic (#3376)

### LINKED ISSUE #3376: doc: clarify case-sensitivity heuristic for community namespace detection
## Context

Identified during review of PR #3366.

## Description

`_PATH_COMPARE_CASE_INSENSITIVE` in `skills-loader.js` and `_isCaseInsensitiveFs()` in `skills-trust.js` both use `process.platform === 'darwin' || process.platform === 'win32'` as a proxy for filesystem case-insensitivity. The constants are named as filesystem facts, but they are OS heuristics.

Edge cases that silently misbehave today:
- **macOS case-sensitive APFS** (`diskutil apfs addVolume ... APFS (Case-sensitive)`): `_PATH_COMPARE_CASE_INSENSITIVE` is `true`, so `Community/` is treated as `community/` — but on disk they are separate inodes. The walk finds `Community/` and routes it through the community trust path, but the same directory could coexist with a separate `community/` directory.
- **Linux with case-insensitive mounts** (ZFS `casesensitivity=insensitive`, `ciopfs`, FAT/exFAT, WSL with NTFS volumes): `_PATH_COMPARE_CASE_INSENSITIVE` is `false`, so `Community/` on an insensitive Linux mount is silently ignored and never surface in the community namespace.

Additionally, the JSDoc Rules comment in `_isCommunityNamespace` still says "First segment of the relative path must be exactly `'community'`" — this is no longer accurate after PR #3366 patched the comparison but left the doc unchanged.

## What should be done

1. Add a JSDoc note to `_PATH_COMPARE_CASE_INSENSITIVE` (and `_isCaseInsensitiveFs`) documenting that this is a platform-default heuristic, not a filesystem probe. Mention the known non-conforming configurations (case-sensitive APFS, case-insensitive Linux mounts).
2. Update the `_isCommunityNamespace` JSDoc Rules bullet to reflect that on case-insensitive platforms the first segment is matched case-insensitively.
3. Consider a runtime filesystem probe (create a temp file, stat with altered case) as a follow-up if the heuristic becomes a real user pain point — probably in a separate issue given the added startup cost.

## Original Review Comment

> `_PATH_COMPARE_CASE_INSENSITIVE` implies "compare paths case-insensitively" — a filesystem fact. But it's an OS heuristic. A user on macOS with a case-sensitive APFS volume or on Linux with a case-insensitive mount will get wrong behavior silently. The JSDoc for `_isCommunityNamespace` also still says the first segment must be "exactly `'community'`", which is stale after this PR.

## Acceptance Criteria

- [ ] `_PATH_COMPARE_CASE_INSENSITIVE` has a JSDoc comment noting the heuristic limitation
- [ ] `_isCaseInsensitiveFs()` in `skills-trust.js` has a matching note (already has a partial one — needs the case-sensitive APFS gap mentioned explicitly)
- [ ] `_isCommunityNamespace` JSDoc Rules bullet updated to reflect platform-conditional matching

blamechris:
Closes #3376

## Summary

JSDoc-only clarification for the OS-level case-sensitivity heuristic used in community namespace detection and trust ledger normalisation. No behavioral change.

- **`_PATH_COMPARE_CASE_INSENSITIVE`** (`skills-loader.js`): added a `Platform-default heuristic` label and documented both non-conforming configurations that silently misbehave:
  - macOS case-sensitive APFS volume: flag is `true` but `Community/` and `community/` are distinct inodes
  - Linux with a case-insensitive mount (ZFS, ciopfs, FAT/exFAT, WSL on NTFS): flag is `false` but the filesystem folds case, so `Community/` is never matched as a community namespace

- **`_isCommunityNamespace` Rules bullet** (`skills-loader.js`): updated stale "must be exactly `'community'`" claim to accurately describe that on darwin/win32 the comparison is lowercased (so `Community/` and `COMMUNITY/` also match), while on Linux the match is exact.

- **`_isCaseInsensitiveFs()`** (`skills-trust.js`): expanded the existing partial note to explicitly call out the case-sensitive APFS gap (two files collide in the ledger to the same normalised key) and the Linux case-insensitive mount gap (skill looked up by different case is treated as untrusted). Clarified what a proper filesystem probe would require.

### DIFF
--- a/packages/server/src/skills-loader.js
+++ b/packages/server/src/skills-loader.js
@@ -1622,12 +1622,34 @@ export function loadActiveSkillsLayered({
   return _enforceTotalBudget(filtered, totalCap)
 }
 
+// Platform-default heuristic for filesystem case-sensitivity.
+//
 // macOS (HFS+/APFS) and Windows (NTFS) are case-insensitive by default. A
 // `cwd` like `/Users/Bob/proj` and a homedir like `/Users/bob` resolve to the
 // same directory but compare unequal as strings, which would defeat both the
 // $HOME boundary check and the global-skills-dir guard. Lowercase before
 // compare on those platforms to make the equality check actually correspond
 // to "same directory on disk".
+//
+// IMPORTANT — this is an OS heuristic, not a filesystem probe. Two known
+// non-conforming configurations will misbehave silently:
+//
+//   • macOS case-sensitive APFS volume (created with
+//     `diskutil apfs addVolume ... "APFS (Case-sensitive)"`): this flag is
+//     `true`, so paths are lowercased before comparison even though the
+//     filesystem treats `Community/` and `community/` as distinct inodes.
+//     In practice this means a `Community/` directory is matched as if it
+//     were `community/`, but both could legitimately coexist on disk.
+//
+//   • Linux with a case-insensitive mount (ZFS `casesensitivity=insensitive`,
+//     ciopfs, FAT/exFAT, or a WSL workspace on an NTFS volume): this flag is
+//     `false`, so path comparison is case-sensitive even though the underlying
+//     filesystem folds case. A `Community/` directory on such a mount will not
+//     be recognised as a community namespace.
+//
+// A runtime filesystem probe (create a temp file, stat with altered case)
+// would close both gaps but adds startup I/O. Track in a follow-up if the
+// heuristic becomes a real user pain point.
 const _PATH_COMPARE_CASE_INSENSITIVE =
   process.platform === 'darwin' || process.platform === 'win32'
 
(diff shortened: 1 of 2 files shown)


blamechris:
## Code Review: PR #3437

### Summary

JSDoc-only clarification for the OS-level case-sensitivity heuristic in `skills-loader.js` (`_PATH_COMPARE_CASE_INSENSITIVE` constant and `_isCommunityNamespace` Rules bullet) and `skills-trust.js` (`_isCaseInsensitiveFs()`). No behavioral change — zero diff to executable code.

### Strengths

- All three documentation sites that needed updating were updated: the constant, the community-namespace rules bullet, and the trust ledger normaliser.
- The non-conforming configurations (case-sensitive APFS, case-insensitive Linux mounts) are documented with concrete examples (`diskutil` command, ZFS/ciopfs/FAT/WSL) rather than vague hand-waves.
- The "why not probe" rationale is now explicit (startup I/O cost) rather than just asserting "intentionally not probed."
- The forward-pointer ("Track in a follow-up if it becomes a real user pain point") keeps the door open without committing to the extra complexity now.
- The `_isCommunityNamespace` Rules bullet now accurately describes the lowercasing path (`darwin / win32 by default`) instead of the stale "exactly `'community'`" claim.
- Code style is consistent throughout: bullet-point indentation, backtick quoting, no trailing whitespace.

### Issues Found

#### Critical (Must Fix)

None.

None.

#### Nitpicks (Optional)

- `skills-loader.js` line 1643 (`_PATH_COMPARE_CASE_INSENSITIVE` block comment) mentions `Community/` and `community/` as the example pair for the APFS case, while the `skills-trust.js` block uses `Foo.md` / `foo.md`. Consistent examples across both files would make a diff-reader's life marginally easier — but this is purely cosmetic and not worth a change.

### Deferred Items

None — issue #3376 fully enumerated all three doc sites, and all three were addressed.

### Architecture Notes

Pure documentation change. No WS protocol impact, no behavioral change, no mobile or server runtime impact. The described behavior (platform heuristic, known non-conforming configs, no filesystem probe) was already the implementation; this PR only makes it legible.

### Verdict

- [x] Approve — Ready to merge

PR #3438, opened 2026-05-07
fix(server): sidecar sentinel args truncation (#3393)

blamechris:
Closes #3393

## Summary

- Truncates the `args` array in the sidecar's per-spawn sentinel stderr line to the first 3 elements, eliding the rest as `...[N more]`
- The sentinel is ring-buffered and replayed to reconnecting clients; full args exposure could leak credential flags (`--api-key`, `--password`) into `execInEnvironment`'s returned `{ stderr }` and any `proc.stderr` consumer
- Documents the risk and the `env`-over-args guidance in PROTOCOL.md

## Changes

- `packages/server/sidecar/agent.js` — truncation logic in `_handleSpawn` + JSDoc comment
- `packages/server/sidecar/PROTOCOL.md` — updated sentinel format spec, added truncation examples and security note
- `packages/server/tests/sidecar/agent.test.js` — 3 new tests: args ≤ 3 (no truncation), args > 3 (elision), correct count in summary string

## Test plan

- [ ] `node --test tests/sidecar/agent.test.js` — 64 pass, 0 fail (includes 3 new truncation tests)
- [ ] Existing sentinel tests pass unmodified (none asserted on > 3 args)

### DIFF
--- a/packages/server/sidecar/PROTOCOL.md
+++ b/packages/server/sidecar/PROTOCOL.md
@@ -249,10 +249,18 @@ per-spawn sentinel line, emitted by the agent immediately after the child is
 spawned.  Its format is:
 
 ```
-[chroxy-pod-agent] spawn cmd=<cmd> args=<json-args> sessionId=<uuid>
+[chroxy-pod-agent] spawn cmd=<cmd> args=<truncated-json-args> sessionId=<uuid>
 ```
 
-Example:
+`args` is **truncated to the first 3 elements**.  If the full args array has
+more than 3 elements the remainder is replaced with a summary string
+`"...[N more]"`:
+
+```
+args=["-e","console.log(1)","...[2 more]"]
+```
+
+Example sentinel frame (≤ 3 args — shown in full):
 
 ```json
 {
@@ -261,6 +269,23 @@ Example:
   "seq": 1
 }
 ```
+
+Example sentinel frame (> 3 args — remainder elided):
+
+```json
+{
+  "type": "stderr",
+  "data": "[chroxy-pod-agent] spawn cmd=claude args=[\"--input-format\",\"stream-json\",\"--output-format\",\"...[2 more]\"] sessionId=550e8400-e29b-41d4-a716-446655440000\n",
+  "seq": 1
+}
+```
+
+**Security note:** The sentinel is buffered in the agent's ring buffer and
+replayed to reconnecting clients.  Callers **must not** pass secret material
+(e.g. `--api-key`, `--password`) as CLI flags — even truncated, the first 3
+args appear in the sentinel line and could surface in `execInEnvironment`'s
+returned `{ stderr }` string or in any consumer that reads `proc.stderr` from
+`streamCliInEnvironment`.  Pass secrets via `env` instead (see #3393).
 
 This line is intentionally distinct from real child stderr so integration tests
 can assert that the sidecar code — not a shorter path that bypasses the agent —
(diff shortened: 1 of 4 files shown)


blamechris:
## Code Review: PR #3438

### Summary

Closes #3393 — adds arg truncation to the per-spawn sentinel stderr frame so that secret CLI flags beyond position 3 are never replayed to reconnecting clients. Accompanied by a security note + two PROTOCOL.md examples, JSDoc in the implementation, and three integration tests. Also introduces `renovate.json` with a custom regex manager to track the pinned `@anthropic-ai/claude-code` version in the sidecar Dockerfile.

---

### Strengths

- **Correct approach**: truncating at the source (the sentinel emitter) rather than filtering at the consumer means the secret never enters the ring buffer in the first place. Filtering in `_wireWsToProc` would only protect the immediate caller, not any future replay path.
- **Const is well-named and local**: `SENTINEL_MAX_ARGS = 3` is scoped to the method and clearly labelled — no magic number.
- **Tests cover the boundary exactly**: `len == 3` (no-op), `len == 5` (elides 2), `len == 7` (elides 4) — boundary, typical, and arbitrary-count cases are all represented.
- **Existing `sentinel includes correct cmd, args` test unaffected**: it uses a 2-arg spawn, safely under the limit, so no pre-existing test is silently invalidated.
- **renovate.json addition is well-scoped**: `automerge: false` is the safe default; the custom regex manager correctly targets only the `CLAUDE_CODE_VERSION` ARG in the sidecar Dockerfile.

---

### Issues Found

#### Critical (Must Fix)

None.

| `sidecar/PROTOCOL.md` | brief example block | Inline example shows **2** real args + `"...[2 more]"`, but the prose says "first **3** elements". Change to `["-e","console.log('hi')","--flag","...[2 more]"]` (3 real + 1 summary) so the example matches the stated rule. | A reader learning the protocol from the brief inline example will infer a limit of 2, contradicting the text. The JSON frame example below correctly shows 3 real args. |
| `k8s.js` `_wireWsToProc` | `case 'stderr':` | The sentinel frame still flows through `proc.stderr` unfiltered. Truncation reduces the blast radius, but the sentinel line itself (with up to 3 plaintext args) still appears in `execInEnvironment`'s `{ stderr }` string. A one-line guard (`if (frame.data.startsWith('[chroxy-pod-agent] spawn ')) break`) would fully satisfy the "filter in K8sBackend" acceptance criterion from #3393. | The issue listed filtering as an alternative mitigation. This PR chose truncation + documentation, which is valid, but the first 3 args (which could still contain non-secret but noisy flags) will continue to pollute stderr captures. |

#### Nitpicks (Optional)

- The two PROTOCOL.md JSON frame examples are placed after the security note rather than before it. Reading them in the opposite order (example first, security note after) would be more natural.
- `SENTINEL_MAX_ARGS` is declared with `const` inside the method body on every call. It's essentially a module-level constant — could be moved to the top of the file alongside the other tunable constants (`FALLBACK_BUFFER_SIZE`, `KILL_GRACE_MS`, etc.) to make it configurable in the same pattern. Minor; current placement is correct.

---

| Filter sentinel line from `proc.stderr` in `_wireWsToProc` | [#3393](https://github.com/blamechris/chroxy/issues/3393) | Still open — the "filter in K8sBackend" acceptance criterion was not addressed by this PR. The security note covers the documentation criterion. Consider addressing the filter criterion in a follow-up. |
| Move `SENTINEL_MAX_ARGS` to module-level constant | — | Cosmetic; not worth a standalone issue. Call it out if a refactor of agent.js constants is ever done. |

---

### Architecture Notes

The change is entirely contained within `sidecar/agent.js` and its corresponding test file. It does not touch the WS protocol handshake, session management, ring-buffer eviction, or any server-side path outside the sidecar. `K8sBackend`'s `_wireWsToProc` is unchanged — the truncation decision is correctly made at the point of emission, not at the point of consumption.

The new `renovate.json` sits at the repo root and is out of scope for the sentinel fix, but it's a clean, self-contained addition and does not conflict with anything.

---

### Verdict

The core change is correct, well-tested, and well-documented. The inline PROTOCOL.md example inconsistency is a minor doc nit; the K8sBackend filter gap was already tracked in #3393 which remains open. Neither blocks merge.

PR #3442, opened 2026-05-07
fix(server): chroxy-pod-agent active-session eviction sends session_lost (#3390)

blamechris:
Closes #3390

## Problem

When `_enforceSessionCap` evicts an **active** session (all slots live, cap hit by a new spawn), `_evictSession` called `ws.close(1001, 'session evicted')` with no preceding frame. `K8sBackend._wireWsToProc` maps close code `1001` → `finishWithExit(-1)` — the same exit code used for pre-session connection failures — so the consumer received no actionable signal distinguishing a pod-cap eviction from a stale TCP drop.

## Fix

Before `ws.close(1001)`, `_evictSession` now sends:

```js
this._send(session.activeWs, { type: 'session_lost', sessionId, reason: 'evicted_by_cap' })
```

`K8sBackend` already handles all `session_lost` frames with `finishWithExit(-2)` (session-loss semantics), so no change to the backend is needed. The consumer now sees `exit(-2)` instead of `exit(-1)` for an active-session cap eviction.

## Changes

- `sidecar/agent.js` — `_evictSession`: send `session_lost(evicted_by_cap)` before `ws.close(1001)` when `activeWs` is present at eviction time. Frame is sent directly (not via `_emitSessionFrame`) so it is not buffered or seq-stamped as session output.
- `sidecar/PROTOCOL.md` — add `evicted_by_cap` row to `session_lost` reason table; update Hard Session Cap section to document the new frame; add row to Error Cases table.
- `tests/sidecar/agent.test.js` — new TDD test asserting `session_lost` frame is sent **before** `ws.close(1001)` and carries the correct `sessionId` and `reason: 'evicted_by_cap'`.

## Test

```
# tests 64
# pass  64
# fail  0
```

### DIFF
--- a/packages/server/sidecar/PROTOCOL.md
+++ b/packages/server/sidecar/PROTOCOL.md
@@ -333,10 +333,13 @@ Sent in response to a `resume` frame when the session is unrecoverable.
 |---------------------|------------------------------------------------------------|
 | `unknown_session`   | No record of the given `sessionId` (agent restarted, etc.) |
 | `buffer_overflow`   | `lastSeq` predates the oldest buffered frame — gap detected |
+| `evicted_by_cap`    | Session was forcibly evicted because all slots were active and the hard session cap was hit by a new spawn |
 
 For `unknown_session` the connection stays open so the client can open a fresh
 session with `spawn`.  For `buffer_overflow` the agent closes the WS with code
-`1008` because continuing on the same WS would still see a partial stream.
+`1008` because continuing on the same WS would still see a partial stream.  For
+`evicted_by_cap` the agent closes the WS with code `1001`; the session is
+unrecoverable (the child has been killed).
 
 ```json
 {
@@ -470,8 +473,13 @@ It is not intended as durable session persistence.
 The agent limits concurrent sessions to `CHROXY_AGENT_MAX_SESSIONS` (default
 8).  When a new `spawn` would exceed the cap, the agent evicts the idle session
 (no active WS) with the oldest `lastActiveAt` timestamp before inserting the
-new one.  If all sessions are active, the globally oldest session is evicted
-(and its live WS is closed with code `1001`).
+new one.  If all sessions are active, the globally oldest session is evicted:
+the agent sends `{ type: 'session_lost', sessionId, reason: 'evicted_by_cap' }`
+to the live consumer before closing the WS with code `1001`.
+
+`K8sBackend` maps `session_lost` to `exit(-2)` regardless of reason, so the
+caller can distinguish a cap eviction (session-loss semantics) from a
+pre-session connection failure (which maps to `exit(-1)`).
 
 This cap is defense-in-depth against a buggy K8sBackend that reconnects
 repeatedly without resuming.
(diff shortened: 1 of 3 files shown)


blamechris:
## Code Review: PR #3442

### Summary

Fixes a silent-signal gap in `_evictSession`: when the hard session cap evicts a session that has a live WebSocket, the consumer previously received only `ws.close(1001)` — indistinguishable from a pre-session TCP drop (also `close(1001)` with no preceding frame). The fix is a single `session_lost(reason: 'evicted_by_cap')` frame emitted directly to the live WS before the close handshake. `K8sBackend`'s existing `session_lost` handler already maps to `exit(-2)`, so no backend change is needed and the consumer can now distinguish forced eviction (`-2`) from connection failure (`-1`). PROTOCOL.md is updated with a new reason row, Hard Session Cap narrative, and Error Cases table entry. One integration test asserts frame-before-close ordering via a fake-WS callLog.

---

### Strengths

- **Minimal blast radius.** One line added to `_evictSession`; the backend requires zero changes.
- **Correct bypass of `_emitSessionFrame`.** Sending directly via `_send` keeps the eviction frame out of the session's seq-stamped ring buffer, preventing the ghost frame from surfacing if the client later attempts `resume`.
- **`_send` already has a readyState guard.** If the WS is somehow already closing, the guard silently drops the frame; the subsequent `close()` call is wrapped in its own `try/catch`. No double-fault path.
- **Test is genuinely integration-level.** It spawns a real `PodAgent` over a real HTTP port, creates two sessions, then surgically patches their `activeWs` with fake WS objects to force the all-active fallback path. The `callLog` array captures send/close in insertion order, making the ordering assertion deterministic and not reliant on timing.
- **Documentation is complete.** All three doc sites updated: `session_lost` reason table, Hard Session Cap narrative, Error Cases table.

---

### Issues Found

#### Critical (Must Fix)

_None._

| `sidecar/agent.js` | 447 | The `reason` passed to `_evictSession` is `'max_sessions'` (from the cap enforcer) but the frame uses the hard-coded string `'evicted_by_cap'`. These are two names for the same event. Consider either (a) passing `'evicted_by_cap'` as the eviction reason so the callsite and frame reason align, or (b) adding a comment noting the intentional rename so future readers don't introduce a third name. | Minor — currently harmless, but the mismatch is a small footgun if someone later adds a reason-forwarding log line. |
| `sidecar/PROTOCOL.md` | §Session Lost After Agent Restart | The two existing sequence diagrams (`unknown_session`, `buffer_overflow`) in the "Session Lost" section have no matching diagram for `evicted_by_cap`. Consistency would suggest one, though the prose in the Hard Session Cap section covers the case adequately. | Low priority — existing diagrams are reactive (triggered by a `resume`), whereas `evicted_by_cap` is proactive (pushed to the live consumer), so the asymmetry may be intentional. |

#### Nitpicks

- `fakeWsB` in the test has no-op `send()` and `close()` — it's fine, but a comment like `// fakeWsB — newer session, survives; we only care it stays attached` would make the intent immediately clear to the next reader.
- The `reason` column for `evicted_by_cap` in the `session_lost` table wraps to a long sentence. Existing rows are one-liners; a short form like `"Session forcibly evicted; hard cap hit while all slots active"` would match the table's style.

---

| Align `'max_sessions'` eviction-reason string with `'evicted_by_cap'` frame reason, or add an explicit mapping comment | [#3445](https://github.com/blamechris/chroxy/issues/3445) | One-line rename + doc change; no behaviour change. Safe to do post-merge as a standalone cleanup. |

---

### Architecture Notes

The fix lives entirely within the session-cap enforcement path inside `PodAgent`. It does not touch the WebSocket message dispatch loop, the session ring buffer, or `K8sBackend`. The direct `_send` call (bypassing `_emitSessionFrame`) is the correct choice: the eviction frame is a control signal to the live consumer, not a seq-stamped output frame, and must not be replayed to a future `resume` consumer. This matches how `session_lost(unknown_session)` is already sent directly in the `resume` handler.

---

### Verdict

The core fix is correct, minimal, and well-tested. The one follow-up issue (#3445) is a cosmetic string-alignment concern, not a correctness problem.

PR #3443, opened 2026-05-07
fix(server): K8sBackend imagePullPolicy enum validation (#3375)

### LINKED ISSUE #3375: feat(server): validate imagePullPolicy against K8s enum at constructor + createEnvironment callsite
## Context

Identified during review of PR #3367 (feat(server): K8sBackend native imagePullPolicy option).

## Description

The current implementation accepts any string for \`imagePullPolicy\` — both at constructor time and as a per-call override. Kubernetes only accepts three values: \`Always\`, \`IfNotPresent\`, and \`Never\`. An unrecognised value (e.g. \`'IfNotpresent'\` — wrong capitalisation, no space) is silently passed through to the Kubernetes API, which then rejects the Pod creation with an opaque 422 Unprocessable Entity error. The error message from the K8s API is not always clear about which field failed.

A short guard at both the constructor and `createEnvironment` call-site would catch typos immediately with a helpful message:

\`\`\`js
const VALID_PULL_POLICIES = new Set(['Always', 'IfNotPresent', 'Never'])

// In constructor
if (imagePullPolicy && !VALID_PULL_POLICIES.has(imagePullPolicy)) {
  throw new Error(
    \`K8sBackend: invalid imagePullPolicy "${imagePullPolicy}" — must be Always, IfNotPresent, or Never\`
  )
}
\`\`\`

The same check should be applied to the per-call override in \`createEnvironment\`.

## Original Review Comment

> Should the option validate against \`'Always' | 'IfNotPresent' | 'Never'\` or accept any string? K8s only accepts those three values. A typo like \`'IfNotpresent'\` would be silently forwarded to the API, which rejects the Pod with a 422 — not a great DX. A guard with a clear error message at construction and per-call would prevent this with zero cost.

## Acceptance Criteria

- [ ] Constructor throws immediately with a clear message when \`imagePullPolicy\` is set to an unrecognised value
- [ ] \`createEnvironment\` throws when the per-call override is an unrecognised value
- [ ] Unit tests cover the invalid-value error path for both callsites
- [ ] JSDoc union type \`'Always'|'IfNotPresent'|'Never'\` already documents valid values — the runtime guard makes that contract enforceable

blamechris:
Closes #3375

## What

Guard the `imagePullPolicy` option against unrecognised values at both callsites in `K8sBackend`:

- **Constructor** — throws `TypeError` immediately if an invalid string is passed
- **`createEnvironment`** — throws `TypeError` before any API calls so no Secret or Pod is created

Valid values remain `Always`, `IfNotPresent`, `Never`. `null`/`undefined` still omits the field so Kubernetes applies its own default.

## Why

A typo like `'IfNotpresent'` was silently forwarded to the K8s API, which rejected the Pod creation with an opaque 422 Unprocessable Entity. The new guard produces a clear error message at the point of misconfiguration:

```
TypeError: K8sBackend: invalid imagePullPolicy "IfNotpresent" in constructor opts — must be one of: Always, IfNotPresent, Never
```

## Tests

8 new unit tests in `k8s.test.js` under `K8sBackend imagePullPolicy validation (#3375)`:

- All three valid values accepted at constructor and per-call
- `undefined`/`null` accepted at both sites
- `'IfNotpresent'` (wrong cap) throws `TypeError` with the offending value + allowed list + context label at constructor
- `'IfNotpresent'` throws at `createEnvironment` with `createEnvironment opts` context
- Arbitrary string `'always'` throws at constructor
- No Secret or Pod API call is made when per-call validation fails

All 126 existing + new k8s tests pass.

### DIFF
--- a/packages/server/src/environments/backends/k8s.js
+++ b/packages/server/src/environments/backends/k8s.js
@@ -47,6 +47,26 @@ function notImplemented(method) {
   return err
 }
 
+/** Valid Kubernetes imagePullPolicy values */
+const VALID_PULL_POLICIES = new Set(['Always', 'IfNotPresent', 'Never'])
+
+/**
+ * Validates an imagePullPolicy value.  Throws TypeError for unrecognised strings.
+ * Passes through null/undefined (meaning: omit the field, let K8s apply its default).
+ *
+ * @param {string|null|undefined} value
+ * @param {string} context - Describes where the value came from (for the error message)
+ */
+function validateImagePullPolicy(value, context) {
+  if (value == null) return
+  if (!VALID_PULL_POLICIES.has(value)) {
+    throw new TypeError(
+      `K8sBackend: invalid imagePullPolicy "${value}" in ${context} — ` +
+      `must be one of: Always, IfNotPresent, Never`
+    )
+  }
+}
+
 /**
  * K8sBackend implements the Backend interface (see types.js) using the
  * Kubernetes API via @kubernetes/client-node.
@@ -90,6 +110,7 @@ export class K8sBackend {
     connectMode, _coreV1Api, _portForward, _dialWs, _net,
     _reconnectDelays, _maxRetries,
     _setTimeout: setTimeoutImpl, _clearTimeout: clearTimeoutImpl } = {}) {
+    validateImagePullPolicy(imagePullPolicy, 'constructor opts')
     this._namespace = namespace || 'default'
     this._sidecarImage = sidecarImage || DEFAULT_SIDECAR_IMAGE
     this._imagePullPolicy = imagePullPolicy || null
@@ -205,6 +226,7 @@ export class K8sBackend {
       memoryLimit, cpuLimit, forwardPorts, mounts,
       imagePullPolicy: callImagePullPolicy,
     } = opts
+    validateImagePullPolicy(callImagePullPolicy, 'createEnvironment opts')
     const ns = namespace || this._namespace
     const podName = `chroxy-env-${envId}`
     const secretName = `chroxy-token-${envId}`
--- a/packages/server/tests/environments/backends/k8s.test.js
+++ b/packages/server/tests/environments/backends/k8s.test.js
@@ -371,6 +371,100 @@ describe('K8sBackend.createEnvironment()', () => {
 })
 
 // ─────────────────────────────────────────────────────────────────────────────
+// K8sBackend imagePullPolicy validation (#3375)
+// ─────────────────────────────────────────────────────────────────────────────
+
+describe('K8sBackend imagePullPolicy validation (#3375)', () => {
+  it('constructor accepts valid values: Always, IfNotPresent, Never', () => {
+    const api = createMockApi()
+    for (const policy of ['Always', 'IfNotPresent', 'Never']) {
+      assert.doesNotThrow(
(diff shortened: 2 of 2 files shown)


blamechris:
## Code Review: PR #3443

### Summary

Adds a `validateImagePullPolicy` module-level helper to `K8sBackend` that guards the `imagePullPolicy` option against unrecognised strings at two callsites: the constructor and `createEnvironment`. The helper throws `TypeError` with a diagnostic message that names the offending value, the valid options, and the context (constructor vs. per-call). `null`/`undefined` short-circuits through so the field is omitted from the Pod spec and Kubernetes applies its own default. Eight new unit tests exercise the full matrix.

### Strengths

- **Synchronous, early guard** — validation fires before any async work, so no Secret is orphaned on invalid input. The "no Pod is created" test verifies this explicitly.
- **`Set` for membership test** — constant-time lookup, trivially extensible if K8s ever adds a policy value.
- **`==` null check** — correctly passes through both `null` and `undefined` without requiring two comparisons.
- **Error message quality** — includes the offending value, the context label, and the full allowed list. Immediately actionable from logs.
- **JSDoc on both the helper and the constructor/createEnvironment params** — consistent with the surrounding code.
- **Test coverage** — all three valid values, both null sentinels, both error paths, and the "no side effects on rejection" invariant. 8 tests for 8 lines of logic is appropriate.
- **Style is fully consistent** — no semicolons, single quotes, ES module, no TypeScript.

### Issues Found

#### Critical (Must Fix)

None.

| `k8s.js` | 116 | `imagePullPolicy \|\| null` silently converts the empty string `''` to `null` | An empty string is not a valid policy and would not be caught at the constructor, only at runtime via the `if (imagePullPolicy)` guard later. A pre-validated value can only ever be `'Always'`, `'IfNotPresent'`, `'Never'`, or `null`/`undefined` at this point, so `imagePullPolicy ?? null` is the correct idiom. The `||` form is a latent inconsistency — practically harmless since validation runs first, but worth tightening. |
| `k8s.test.js` | 440–447 | `createEnvironment accepts undefined imagePullPolicy` test will resolve even if `createNamespacedPod` returns `{}` without a real Pod status, which means the assertion does not actually prove the happy-path completes correctly | This is a pre-existing characteristic of the test suite, not introduced here. The validation-path tests introduced in this PR are correct. No action required in this PR, but worth noting. |

#### Nitpicks (Optional)

- Comment at line 282 says `// 4. Resolve imagePullPolicy` but the preceding workspace-mount block is also labelled `// 4. Build volumes`. Step numbering drifted. Not introduced by this PR.

### Deferred Items

| Replace `imagePullPolicy \|\| null` with `imagePullPolicy ?? null` on line 116 | #3444 | One-line clean-up, zero risk, but out of scope for a validation-focused PR. |

### Architecture Notes

The helper is correctly scoped as a module-level function (not a static method), consistent with `notImplemented` and `_parseMountString` in the same file. It adds no new state and no new events. Placement before the class declaration keeps the class body focused on behaviour. The dual-callsite pattern (constructor + `createEnvironment`) is the right design: constructor validation catches misconfigured backends at startup; per-call validation catches dynamic overrides passed by `EnvironmentManager` or tests without requiring a new backend instance.

### Verdict

Tight, focused change. Correct error type, correct null semantics, correct placement in the call flow, well-tested. The `||` vs `??` inconsistency is worth a follow-up but does not affect correctness given the validation that precedes it.

PR #3451, opened 2026-05-07
docs(server): document load outcome for wrongly-cased community dir on Linux

blamechris:
## Summary

- Clarifies the Linux case-sensitivity section of `docs/skills.md` to state explicitly what happens to skills placed inside a `Community/` or `COMMUNITY/` directory on Linux.
- Previous wording said skills "are either loaded without a trust prompt or rejected, depending on configuration", which left users unable to predict behaviour.
- Verified actual behaviour against `packages/server/src/skills-loader.js` (community walk skips entries that do not compare-equal to `'community'` on case-sensitive platforms; top-level walk drops directory entries via the `dotIdx <= 0` extension check) and the Linux-only test in `packages/server/tests/skills-loader.test.js` that asserts `skills.length === 0` for a `Community/alice/caps.md` fixture.
- New wording: skills inside a wrongly-cased directory are silently skipped (not loaded at all, no warning), and top-level files alongside the wrongly-cased directory continue to load normally.
- Adds a troubleshooting hint pointing users at the directory name when community skills go missing on Linux.

Closes #3435

## Test plan

- [x] Docs-only change — no code or tests modified
- [x] Cross-checked claim against `_isCommunityNamespace`, the community walk in `skills-loader.js` (~lines 387-440), and the existing Linux-skip test in `skills-loader.test.js`

### DIFF
--- a/docs/skills.md
+++ b/docs/skills.md
@@ -106,9 +106,9 @@ Community skills are subject to a first-activation trust prompt — Chroxy will 
 **Always name the directory `community/` in lowercase.** The behaviour differs by platform:
 
 - **macOS and Windows** — filesystems are case-insensitive by default. `Community/`, `COMMUNITY/`, and `community/` all refer to the same directory and are all recognised as the community namespace.
-- **Linux** — the filesystem is case-sensitive. Only the exact name `community/` is recognised. A directory named `Community/` or `COMMUNITY/` on Linux is silently treated as an ordinary top-level skills directory and is **not** subject to the community trust gate — its skills are either loaded without a trust prompt or rejected, depending on configuration.
+- **Linux** — the filesystem is case-sensitive. Only the exact name `community/` is recognised. A directory named `Community/` or `COMMUNITY/` on Linux is silently skipped: the loader does not recurse into it, so any skills inside (e.g. `Community/alice/foo.md`) are **not loaded at all** — they neither appear as community skills nor as ordinary top-level skills, and no warning is emitted. Top-level skill files alongside the wrongly-cased directory continue to load as normal.
 
-Using lowercase `community/` everywhere is the portable convention that works on all platforms.
+Using lowercase `community/` everywhere is the portable convention that works on all platforms. If your community skills are unexpectedly missing on a Linux machine, check the directory name first.
 
 ### Trust file migration between platforms
 


blamechris:
## Code Review: PR #3451

### Summary

Docs-only change to the Linux case-sensitivity section of `docs/skills.md`. Replaces ambiguous wording ("either loaded without a trust prompt or rejected, depending on configuration") with a concrete statement of behaviour: skills inside a wrongly-cased `Community/` or `COMMUNITY/` directory on Linux are silently skipped, no warning is emitted, and top-level files alongside the wrongly-cased directory continue to load. Adds a troubleshooting hint.

### Strengths

- **Accurate to the code.** Verified against `packages/server/src/skills-loader.js`:
  - The community walk at line 387-440 only recurses when `entryCmp === 'community'`, where `entryCmp = _PATH_COMPARE_CASE_INSENSITIVE ? entry.toLowerCase() : entry`. On Linux `_PATH_COMPARE_CASE_INSENSITIVE` is `false`, so a `Community/` entry never matches and the walk skips it without recursing.
  - The top-level walk at line 821-822 drops any entry where `dotIdx <= 0` (no extension), so a `Community/` directory entry is filtered out before any stat/readdir — directory entries cannot be loaded as ordinary top-level skills regardless.
  - Confirmed by the Linux-only test at `packages/server/tests/skills-loader.test.js:3079` which asserts `skills.length === 0` for a fixture containing `Community/alice/caps.md`. Issue #3378 is the test's tracking ID.
- **Correctly contradicts the issue's premise.** Issue #3435 incorrectly stated skills "are then discovered as ordinary top-level skills". The PR author identified this and documented the actual behaviour rather than copying the issue text — good due diligence.
- **Good UX touch.** The added troubleshooting sentence ("If your community skills are unexpectedly missing on a Linux machine, check the directory name first") is practical because the loader emits no warning, making misconfiguration silent.
- **Scope discipline.** Two-line change, no unrelated edits.

### Issues Found

#### Critical (Must Fix)

None.

None — the change is minimal, accurate, and self-contained.

#### Nitpicks (Optional)

- The example path `Community/alice/foo.md` works fine, but the section above the heading already uses `alice/typescript-conventions.md` as the canonical author/skill example. Mirroring that exact filename would be marginally more consistent. Not worth a follow-up.

None.

### Architecture Notes

The doc accurately reflects the two-stage filter (top-level extension gate + community-walk lowercase-exact match). Both gates are deliberate:

1. **Extension gate** (line 821-822) prevents the loader from treating any subdirectory at the top level as a skill source — this is what stops `Community/` from being walked.
2. **Lowercase-exact match in the community walk** (line 389-390) is the explicit Linux/case-sensitive policy added in #3378 to prevent silent trust bypass.

The combined effect is that wrongly-cased `Community/` directories are effectively invisible to the loader on Linux. Documenting this saves users from a confusing failure mode where their skills are present on disk but never injected.

The "no warning is emitted" detail is worth preserving in the doc — the loader has no logging path that fires for a directory it deliberately skipped, so users can't grep logs for evidence of the misconfiguration. The added troubleshooting hint is the right mitigation.

### Verdict

Two-line docs-only change with verified accuracy. No code, tests, or unrelated files touched. Closes #3435 cleanly.

PR #3471, opened 2026-05-07
fix(sidecar): close WS in send callback to avoid flush race

blamechris:
## Summary

When the agent emits an `error(line_too_long)` frame on oversized stdout it then closes the WS. Previously the close was scheduled on a fixed 50ms `setTimeout`, which can race with `ws.send()` flushing the frame on a busy event loop and drop the frame on the floor — the client would see only `close(1008, 'line_too_long')` with no preceding error frame.

The fix:
- Extends `_send` and `_emitSessionFrame` to accept an optional completion callback, invoked when the underlying `ws.send(json, cb)` callback fires (or synchronously on closed-socket / send-throws paths so callers always make progress).
- Moves the `ws.close(...)` in every post-error path into that callback so the close runs only after the frame has been flushed to the socket buffer.

Sites updated:
- `oversized_line` handler (line_too_long error → close 1008)
- async spawn `error` handler (exit frame → close 1000)
- child `close` handler (exit frame → close 1000)
- `_evictSession` (session_lost evicted_by_cap → close 1001)
- `_handleResume` duplicate-client (error → close 1008)
- `_handleResume` resume-with-gap (session_lost buffer_overflow → close 1008)

## Test plan

- [x] `node --test packages/server/tests/sidecar/agent.test.js` — 73/73 pass
- [x] `node --test packages/server/tests/environments/backends/k8s.test.js` — 136/136 pass
- [x] Existing #3380 regression test (no spurious exit frame after line_too_long) still passes
- [x] Existing #3390 ordering test (session_lost before close on cap-evict) still passes
- [x] New test: deterministic ordering via instrumented fake — `send → send_cb_fired → close(1008)`
- [x] New test: `_send` catch-path still invokes the close callback when `ws.send` throws synchronously

Closes #3399

### DIFF
--- a/packages/server/sidecar/agent.js
+++ b/packages/server/sidecar/agent.js
@@ -447,10 +447,14 @@ export class PodAgent {
     // without a preceding frame).  The frame is sent directly — not through
     // _emitSessionFrame — because we are mid-eviction and do not want the
     // frame buffered or seq-stamped as session output.
+    // Close in the send callback so the session_lost frame is flushed to the
+    // socket buffer before the WS close handshake (#3399).
     if (session.activeWs) {
-      this._send(session.activeWs, { type: 'session_lost', sessionId, reason: 'evicted_by_cap' })
-      try { session.activeWs.close(1001, 'session evicted') } catch {}
+      const ws = session.activeWs
       session.activeWs = null
+      this._send(ws, { type: 'session_lost', sessionId, reason: 'evicted_by_cap' }, () => {
+        try { ws.close(1001, 'session evicted') } catch {}
+      })
     }
 
     this._sessions.delete(sessionId)
@@ -658,15 +662,15 @@ export class PodAgent {
       this._emitSessionFrame(session, { type: 'error', message: `spawn failed: ${err.message}` })
       // Async spawn never produced a child — there will be no 'close' event
       // to clean up the session. Synthesize an exit and drop the session
-      // entry so the agent does not leak it indefinitely.
-      this._emitSessionFrame(session, { type: 'exit', code: -1 })
-      setTimeout(() => {
+      // entry so the agent does not leak it indefinitely.  Closing in the
+      // ws.send callback for the exit frame avoids a flush race (#3399).
+      this._emitSessionFrame(session, { type: 'exit', code: -1 }, () => {
         if (session.activeWs && session.activeWs.readyState === 1) {
           session.activeWs.close(1000, 'spawn failed')
         }
         this._cancelIdleTimer(session)
         this._sessions.delete(sessionId)
-      }, 50)
+      })
     })
 
     // stdout — read line-by-line; each NDJSON line becomes one 'event' frame.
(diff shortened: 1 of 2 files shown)


blamechris:
## Code Review: PR #3471

### Summary

Replaces the fixed `setTimeout(50)` close-after-send pattern in `packages/server/sidecar/agent.js` with a callback-driven sequence that closes the WebSocket only after `ws.send()` has flushed the preceding frame to the socket buffer. Closes #3399.

The change extends `_send` and `_emitSessionFrame` with an optional completion callback, then routes the close call through that callback at every site where the agent emitted a final frame followed by a close — six in total. The wrapper synthesises a callback for the `ws.readyState !== 1` and `ws.send` synchronous-throw paths so callers always make progress.

### Strengths

- Eliminates the original race entirely: close now strictly happens after the ws.send callback fires (or synchronously in failure paths), rather than racing a 50 ms timer against a busy event loop.
- Caught and fixed every call site, not just the one in the issue: `_evictSession`, `_handleResume` duplicate-client, `_handleResume` resume-with-gap, `child.on('error')` exit frame, `child.on('close')` exit frame, plus the original `oversized_line` path.
- Backward-compatible API: callback is optional, so existing call sites that don't need flush ordering keep working unchanged.
- `_send` catch-path still invokes the callback — without this, a synchronous send throw would leave the session pinned forever.
- `_emitSessionFrame` defers the callback via `queueMicrotask` when there is no `activeWs`, matching the async semantics of `ws.send(json, cb)` and preventing reentrancy bugs.
- `_handleResume` paths added defensive `try { ws.close(...) } catch {}` — good belt-and-braces given the close runs from a queued callback.
- `_evictSession` now nulls `session.activeWs` *before* calling `_send` (capturing the ref in a local), preventing a re-entrant `_emitSessionFrame` from double-sending on the WS we're about to close.
- Two new tests with deterministic ordering: one uses a fake WS that defers the send cb by 5ms to widen the race window, the other forces `ws.send` to throw and verifies the close still runs.
- Test fakes for the existing #3390 ordering test were updated to mirror the real `ws.send(data, cb)` signature.

### Issues Found

#### Critical (Must Fix)

None.

| `packages/server/sidecar/agent.js` | 350-353 | `_handleConnection` reject path uses raw `ws.send(msg, cb)` directly rather than going through `_send()`. | Minor inconsistency — every other reject/error path now flows through `_send` with a callback. Routing this one through `_send` too would make the pattern uniform and pick up the catch-path safety for free. Pre-existing (predates this PR), so deferring is fine — flagging as a follow-up nit. |
| `packages/server/sidecar/agent.js` | 433-461 | `_evictSession` accepts a `reason` parameter but the frame always carries `'evicted_by_cap'`. Pre-existing. | Out of scope for #3399, but worth tracking — when an idle-timeout eviction lands here the consumer sees `evicted_by_cap` which is misleading. |

#### Nitpicks (Optional)

- The new line_too_long ordering test waits `setTimeout(30ms)` after triggering. On heavily loaded CI this could theoretically be tight (5ms send-cb delay + close + assertion). 50 ms would be safer, though 30 ms is consistent with the rest of the file's polling cadence. Not blocking.
- `_emitSessionFrame` JSDoc says "fires on the next microtask" — accurate, but the `_send` JSDoc just above could mention that the cb fires async-via-ws-send when the socket is open vs sync-on-error otherwise. Minor.

| Route `_handleConnection` reject path through `_send` for consistency | (filed below) | Pre-existing behavior; refactor is a tidy-up, not part of the #3399 race fix. |

### Architecture Notes

This change is internal to `PodAgent` (the in-pod sidecar agent) and does not alter the wire protocol — clients still see the same `error(code=line_too_long)` frame followed by `close(1008, 'line_too_long')`. The fix simply guarantees the frame actually arrives. K8sBackend's reconnect/exit code mapping is unchanged.

The pattern (callback threaded through emit → send → close) is cleaner than alternatives I considered (e.g., wrapping `ws.send` in a Promise, or polling `bufferedAmount`) — it stays close to the underlying `ws` API and keeps the call sites readable.

### Verdict

Approve — ready to merge once Copilot/CI is green. Solid, focused fix with no regressions in the existing 73-test sidecar suite.

blamechris:
## Review Comments Addressed

No inline review comments to address — Copilot review did not fire on this PR within the polling window (~16 min). The agent-review summary comment above stands as the sole review record.

| — | No inline comments to process | — | — |

- Fixed: 0
- False positives: 0
- Follow-up issues created: 0 (1 created during agent-review: #3473)
- Existing issues closed: 0

CI: 12/12 green. Ready to merge.

PR #3472, opened 2026-05-07
fix(server): cap SidecarProcess pre-dial stdin buffer

### LINKED ISSUE #3401: SidecarProcess: cap pre-dial stdin buffer to prevent unbounded memory growth
## Context

Identified during review of PR #3400 (K8sBackend stdin wiring for #3336).

## Description

`SidecarProcess._stdinBuffer` accumulates stdin writes before the WS dial resolves. There is no cap on how large this buffer can grow. If the WS dial hangs (slow cluster, network partition, DNS failure) and the caller writes large payloads — e.g. a large prompt or a binary attachment — the buffer will grow without bound.

Additionally, `kill()` does not clear `_stdinBuffer`. If `kill()` is called before the dial resolves the buffer is held until `_wireStdin()` is called, which never happens in that path — so it is effectively leaked until the proc object is GC'd.

## Suggested Fix

1. Add a `MAX_STDIN_BUFFER` constant (e.g. 1 MB) and enforce it in the `data` accumulator — drop and log a warning if exceeded.
2. In `kill()`, clear `this._stdinBuffer = []` to release memory immediately on the kill path.

## Original Review Comment

> `_stdinBuffer` is unbounded. If the WS dial takes a long time (slow cluster, DNS failure) and the caller is piping large stdin payloads, memory grows unchecked. `kill()` also doesn't clear the buffer — if kill fires before the dial resolves, the buffer is held until the object is GC'd.

## Acceptance Criteria

- [ ] `_stdinBuffer` is capped at a reasonable byte limit (e.g. 1 MB); excess writes are dropped with a `log.warn`
- [ ] `kill()` clears `_stdinBuffer = []`
- [ ] Test: writing > cap bytes before WS opens drops the excess and emits a warning

blamechris:
## Summary

- Caps `SidecarProcess._stdinBuffer` at a configurable byte limit (default 1 MiB) so a fast producer cannot OOM the server while the WS dial is hung. Over-cap writes are dropped entirely (no truncation — preserves NDJSON line integrity) with a single warning per overflow streak.
- Clears `_stdinBuffer` (and resets `_stdinBufferBytes`) inside `kill()`. Previously, if `kill()` fired before the dial resolved, `_wireStdin()` never ran and the buffered bytes were held until the proc object was GC'd.
- Threads `_maxStdinBufferBytes` through the `K8sBackend` constructor so unit tests can inject a small cap (no need to write a megabyte of data per test case).

Closes #3401

## Test plan

- [x] New unit tests in `packages/server/tests/environments/backends/k8s.test.js` (`SidecarProcess pre-dial stdin buffer cap (#3401)` describe block):
  - Under-cap writes: all 3 chunks (63 B total, 100 B cap) flush correctly when WS opens
  - Over-cap drop: 50 B fits, 60 B + 80 B both dropped; exactly the original 50-byte payload flushes (not a truncation)
  - `kill()` clears the pre-dial buffer (both array and byte counter)
- [x] Existing 135 tests in the same file continue to pass (138 total)
- [x] `k8s-sidecar-roundtrip.test.js` integration test passes

### DIFF
--- a/packages/server/src/environments/backends/k8s.js
+++ b/packages/server/src/environments/backends/k8s.js
@@ -15,6 +15,16 @@ const DESTROY_TIMEOUT_MS = 30_000
 const DEFAULT_RECONNECT_DELAYS = [250, 500, 1_000, 2_000, 4_000, 8_000]
 const DEFAULT_MAX_RETRIES = 5
 
+/**
+ * Cap on bytes held in `_stdinBuffer` before the WS dial resolves (#3401).
+ * If the dial hangs (slow cluster, DNS failure) and the caller is piping
+ * large stdin payloads, an unbounded buffer would grow until the server
+ * OOMs.  Writes that would push the buffer past the cap are dropped with
+ * a warning.  1 MiB is generous for a few large prompts but bounds the
+ * worst-case memory hold at one process-worth.
+ */
+const DEFAULT_MAX_STDIN_BUFFER_BYTES = 1 * 1024 * 1024  // 1 MiB
+
 /** Default sidecar image — overridden via constructor option */
 const DEFAULT_SIDECAR_IMAGE = 'chroxy-pod-agent:latest'
 
@@ -103,12 +113,13 @@ export class K8sBackend {
    * @param {Function} [opts._dialWs]                   - Injected WS dial factory for testing: (url, token) => WebSocket
    * @param {number[]} [opts._reconnectDelays]          - Override backoff delays in ms for testing
    * @param {number}   [opts._maxRetries]               - Override max reconnect retries for testing
+   * @param {number}   [opts._maxStdinBufferBytes]      - Override SidecarProcess pre-dial stdin buffer cap (#3401)
    * @param {Function} [opts._setTimeout]               - Override setTimeout for deterministic testing
    * @param {Function} [opts._clearTimeout]             - Override clearTimeout for deterministic testing
    */
   constructor({ namespace, inCluster, kubeconfigPath, sidecarImage, imagePullPolicy,
     connectMode, _coreV1Api, _portForward, _dialWs, _net,
-    _reconnectDelays, _maxRetries,
+    _reconnectDelays, _maxRetries, _maxStdinBufferBytes,
     _setTimeout: setTimeoutImpl, _clearTimeout: clearTimeoutImpl } = {}) {
     validateImagePullPolicy(imagePullPolicy, 'constructor opts')
     this._namespace = namespace || 'default'
@@ -173,6 +184,10 @@ export class K8sBackend {
     this._maxRetries = Number.isFinite(_maxRetries) && _maxRetries >= 0
       ? _maxRetries
       : DEFAULT_MAX_RETRIES
+    // Pre-dial stdin buffer cap (#3401) — injectable for tests.
+    this._maxStdinBufferBytes = Number.isFinite(_maxStdinBufferBytes) && _maxStdinBufferBytes > 0
+      ? _maxStdinBufferBytes
+      : DEFAULT_MAX_STDIN_BUFFER_BYTES
     // Timer seam: allow tests to substitute a fake clock and avoid real-time polling.
     this._setTimeout = setTimeoutImpl || setTimeout
     this._clearTimeout = clearTimeoutImpl || clearTimeout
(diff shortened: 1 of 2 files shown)


blamechris:
## Code Review: PR #3472

### Summary

Caps `SidecarProcess._stdinBuffer` (pre-dial stdin write buffer) at 1 MiB by default, with the cap injectable via a constructor option for tests. Implements drop-and-warn semantics on overflow and clears the buffer in `kill()` so memory is released immediately when a kill races a hung WS dial. Adds three focused unit tests covering under-cap pass-through, over-cap drop with state preservation, and `kill()` cleanup. Exactly the scope laid out in #3401's acceptance criteria.

### Strengths

- **Drop policy preserves NDJSON line integrity.** Over-cap chunks are dropped *whole*, not truncated — explicit comment calls this out. The right call: a mid-line truncation would corrupt the consumer's stream-json reader and produce confusing parse errors downstream.
- **Single-warning-per-process pattern.** `_stdinBufferDropped` flag avoids log spam from a producer hammering stdin in a tight loop. The comment ('further drops will be silent') accurately describes the resulting behavior.
- **`kill()` clears both the array and the byte counter.** Specifically addresses the leak path called out in #3401: kill before dial → `_wireStdin()` never runs → buffer held until GC. Also resets in the `_wireStdin()` flush path so the counter is consistent across both release sites.
- **Cap is wired through `K8sBackend` constructor as `_maxStdinBufferBytes`.** Enables tests to inject a 100-byte cap rather than allocate megabytes per test case. Validation (`Number.isFinite && > 0`) matches the existing pattern for `_maxRetries` / `_reconnectDelays`.
- **Test design.** Three tests cover the three distinct paths cleanly: pass-through, drop, kill-clear. `makeBackendWithPendingDial` helper holds the WS in `CONNECTING` state which is exactly the production failure mode (slow cluster / DNS hang).
- **Style matches house rules.** ES modules, no semicolons, single quotes, JSDoc-typed opts.

### Issues Found

#### Critical (Must Fix)

None.

| packages/server/src/environments/backends/k8s.js | 1119–1121 | Comment refers to 'single warning per overflow streak' but the implementation is single warning per *process lifetime* (pre-dial window). `_stdinBufferDropped` is never reset, even if subsequent under-cap writes succeed. | Tighten the wording to 'logs once per pre-dial window' or explicitly reset the flag whenever a write succeeds (which would restore the streak semantic). Either is fine; the current behavior is reasonable, the comment is just slightly misleading. |
| packages/server/tests/environments/backends/k8s.test.js | n/a | Consider one extra assertion in the existing 'writes before WS opens are buffered and flushed on open' test (line 2787) to verify `proc._stdinBufferBytes === 0` after the flush. | Locks in the post-flush invariant so a future change that forgets to reset the counter would be caught. Two-line addition. |
| packages/server/src/environments/backends/k8s.js | n/a | The drop policy is 'drop and warn,' but the consumer (SdkSession) has no runtime signal that bytes were lost. This mirrors the gap #3402 documents for the post-reconnect path. | Out of scope here, but worth a follow-up issue: emit a `'stdin_dropped'` event so the consumer can react (e.g. tear down the session rather than send half a turn). Filed as #3474. |

#### Nitpicks (Optional)

- `DEFAULT_MAX_STDIN_BUFFER_BYTES = 1 * 1024 * 1024` — the leading `1 *` is harmless but could be `1024 * 1024` for less visual noise.
- The new tests pass `args: []` to `streamCliInEnvironment` while existing tests in the same file pass `['-p']`. Either works; consistency is mildly nicer.

| Emit `stdin_dropped` event so consumer can detect pre-dial buffer overflow | https://github.com/blamechris/chroxy/issues/3474 | Out of scope for the cap-and-warn fix this PR implements (which exactly matches #3401's acceptance criteria). The signaling concern mirrors #3402's post-reconnect gap and deserves its own decision/PR. |

### Architecture Notes

The `K8sBackend` → `SidecarProcess` → WS path is the only consumer of this code; Docker and local backends don't have a pre-dial buffer because their stdin is a real OS pipe. This change is contained to the K8s path and does not touch the WS protocol, the SDK session, or any cross-cutting state. The drop semantics are visible only through `proc.stdin.write()` returning `true` while bytes are silently discarded — same observable shape as a slow consumer dropping bytes, which the SDK already tolerates via its NDJSON reader. No protocol implications.

### Verdict

- [x] **Approve** — Ready to merge after CI passes. The implementation matches #3401's acceptance criteria exactly, tests are well-targeted, and the only structural concern (consumer signaling) is out of scope and tracked as #3474.

PR #3475, opened 2026-05-07
fix(sidecar): handle backpressure on child.stdin.write()

### LINKED ISSUE #3396: chroxy-pod-agent: backpressure not handled on child.stdin.write()
## Context

Identified during review of PR #3389 (stdin support for chroxy-pod-agent, #3329).

## Description

In \`_handleStdin\`, the write to child stdin is fire-and-forget:

\`\`\`js
session.child.stdin.write(data)
\`\`\`

\`stream.Writable.write()\` returns \`false\` when the internal buffer is full (highWaterMark reached) signaling that the caller should stop writing until the \`drain\` event fires. This is Node's cooperative backpressure mechanism.

The current implementation ignores the return value. If the client streams many \`stdin\` frames rapidly to a slow child process:

1. Each \`write()\` call silently buffers data in the \`stdin\` WritableStream's internal buffer.
2. The buffer grows without bound until either the child drains it or the agent process OOMs.
3. No error is returned to the client — the memory growth is invisible.

This matters because the use-case for this feature is \`--input-format stream-json\` where the client sends potentially large prompts as NDJSON. A multi-turn conversation with large tool outputs forwarded back as context could exceed hundreds of KB.

## Suggested fix

Pause the WS message delivery while the stdin stream is draining:

\`\`\`js
const ok = session.child.stdin.write(data)
if (!ok) {
  ws.pause()
  session.child.stdin.once('drain', () => ws.resume())
}
\`\`\`

Or alternatively track a \`_stdinDraining\` flag on the session and queue/drop frames.

For the v1 use-case (single prompt + stdin_end before child starts processing) the risk is low since the total payload is bounded. Backpressure matters more in multi-turn streaming scenarios which arrive in #3336.

## Acceptance Criteria

- [ ] \`child.stdin.write()\` return value is checked
- [ ] WS ingress is paused while stdin is draining, OR frames are queued with a bounded queue that returns an error if exceeded
- [ ] Test: rapid large stdin frames do not accumulate unboundedly in memory

blamechris:
## Summary

- Inspect `child.stdin.write()` return value in `_handleStdin`; pause WS message delivery when it returns `false` and resume on the next `'drain'` event from `child.stdin`.
- Add a `_stdinDraining` session flag so a long burst of writes does not stack drain listeners or call `ws.pause()` repeatedly while already paused.
- Skip `ws.resume()` if the session has been detached during draining (cleanup-during-drain race).
- Document the cooperative backpressure behaviour in `PROTOCOL.md`.

The fix prevents a fast WS client from growing the stdin WritableStream's internal buffer without bound (and OOMing the pod) when streaming `--input-format stream-json` payloads to a slow child.

Closes #3396

## Test plan

- [x] `node --test packages/server/tests/sidecar/agent.test.js` — 75/75 pass (71 existing + 4 new).
- [x] New tests cover:
  - pause + drain happy path
  - idempotent re-arming (one drain listener, one `pause` call across multiple in-flight backpressured writes)
  - no `pause` when `write()` returns `true`
  - drain after WS detach does not call `resume()` on a stale ws
- [x] Full server test suite delta: no new failures introduced (baseline 74 pre-existing flakes on `origin/main`, same set after).

### DIFF
--- a/packages/server/sidecar/PROTOCOL.md
+++ b/packages/server/sidecar/PROTOCOL.md
@@ -179,6 +179,17 @@ If the child was spawned with `stdin: 'ignore'` or `stdin: 'inherit'`,
 `child.stdin` is `null`; `stdin` frames are **silently dropped** in that case
 (no error is returned) so callers do not need to track the stdin mode.
 
+**Backpressure (#3396).**  Writes to `child.stdin` are cooperative: when the
+child's stdin buffer reaches its `highWaterMark`, `child.stdin.write()` returns
+`false`.  In that case the agent calls `ws.pause()` on the connection,
+deferring further frame delivery to the kernel socket buffer (TCP-level
+backpressure) until the child consumes its buffered input.  On the next
+`drain` event from `child.stdin` the agent calls `ws.resume()` and frame
+delivery continues.  This is transparent to the client — no frame is lost,
+delayed beyond the kernel socket buffer, or surfaced as an error — but a
+client sending faster than the child can read may observe slower TCP
+acknowledgements during the back-pressured window.
+
 #### `stdin_end`
 
 Close the write end of the child's stdin pipe (equivalent to EOF).  After this
--- a/packages/server/sidecar/agent.js
+++ b/packages/server/sidecar/agent.js
@@ -620,6 +620,9 @@ export class PodAgent {
       buffer: [],
       lastActiveAt: Date.now(),
       idleTimer: null,
+      // Cooperative backpressure flag for child.stdin.write() (#3396).
+      // True while a 'drain' listener is armed and the WS is paused.
+      _stdinDraining: false,
     }
     this._sessions.set(sessionId, session)
     ws._sessionId = sessionId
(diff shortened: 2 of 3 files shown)


blamechris:
## Code Review: PR #3475

### Summary

Adds cooperative backpressure to `_handleStdin` in the chroxy-pod-agent sidecar. When `child.stdin.write()` returns `false` the agent calls `ws.pause()` on the connection and registers a one-shot `'drain'` listener that calls `ws.resume()` — pushing further frame delivery into the kernel socket buffer (TCP-level backpressure) instead of growing the stdin WritableStream's buffer in process memory.

A `_stdinDraining` session flag guards the listener from re-arming on every write, and the drain callback skips `resume()` if the session has detached during draining. PROTOCOL.md is updated to document the behaviour. Four new unit tests cover the happy path, idempotent re-arming, no-pause-when-write-returns-true, and the stale-WS race.

### Strengths

- **Idempotency guard.** `_stdinDraining` prevents drain-listener stacking and redundant `pause()` calls during a sustained burst — verified by the second test.
- **Stale-WS race covered.** The drain callback explicitly checks `session.activeWs` and `readyState === 1` before calling `resume()`, and the dedicated test for the disconnect-during-drain case is exactly the scenario that would otherwise crash with `Cannot read properties of null`.
- **Code style matches surrounding.** No semicolons, single quotes, ES modules, EventEmitter pattern preserved.
- **Test isolation.** The fake `EventEmitter` stdin lets the suite drive `write()` return value and `'drain'` deterministically — no flakiness risk.
- **Documentation.** PROTOCOL.md update is in the right section, calls out that the behaviour is transparent to the client, and explicitly references the issue number.

### Issues Found

#### Critical (Must Fix)

None.

| `packages/server/sidecar/agent.js` | ~795 | Consider whether a wedged child (one that never drains) should eventually fail rather than pin the WS in `paused` indefinitely. | Today a child that hangs after accepting input will keep the WS paused forever. The idle-resume TTL eviction (#3349) and session cap (#3392) eventually clean up, but the WS itself does not surface a timeout to the client. Probably out of scope for this PR — see follow-up. |
| `packages/server/sidecar/agent.js` | ~778 | The comment block could note that `ws.pause()` only pauses **inbound** message delivery — outbound `_send()` calls (e.g. session frames from stdout/stderr) continue to flow normally. | Avoids future readers wondering whether output is also paused. The current "transparent to the client" wording covers it implicitly but could be more explicit. |

#### Nitpicks (Optional)

- The `try { ... } catch {}` block around the write+pause sequence currently swallows errors from both `child.stdin.write()` and `ws.pause()`. The pause-arm path is unlikely to throw, but if it ever does, `_stdinDraining` would not be set and the next write would re-attempt. This is recoverable; just worth a one-line comment in the catch.
- All four new tests use `await new Promise((r) => setTimeout(r, 20))` — consistent with the surrounding stdin-forwarding suite, so this matches house style. Worth keeping in mind for future flake hardening if these grow.

| Wedged-child timeout: detect when a session has been paused on stdin drain for an unusually long time and surface to client | https://github.com/blamechris/chroxy/issues/3476 | Out of scope for the immediate backpressure fix; existing TTL/cap protections provide an eventual backstop. Wanted to track separately so the wedged-child failure mode does not get lost. |

### Architecture Notes

The fix sits cleanly inside `_handleStdin` and reuses the existing `session` lifecycle. It does not introduce any new WS frame types or break the protocol. The `_stdinDraining` flag is stored on the session (which already persists across reconnects and is cleaned up by `_evictSession`), so a reconnecting client that resumed a paused session would land in a clean state — though strictly speaking the drain listener is on `child.stdin` and would not survive a session-level reattach if the child were swapped (it isn't, in current code).

PROTOCOL.md is the right place for the documentation. The new note follows the same "**Backpressure (#3396).**" callout pattern used by the rest of the doc.

### Verdict

The implementation is correct, well-tested, well-documented, and aligns with the style of the rest of the sidecar agent. Suggestions are non-blocking; the wedged-child concern is tracked as a follow-up.

PR #3441, opened 2026-05-07
test(dashboard): SkillsPanel pending row combined fields (#3369)

blamechris:
Closes #3369

## Summary

- Adds a combined test where both `description` and `path` are present on a pending row — both elements must render. This guards against future JSX changes that could introduce mutual-exclusion logic or CSS that hides one field when the other is present.
- Adds empty-string tests (`description: ''`, `path: ''`) to verify blank values produce no rendered element.
- Adds whitespace-only tests (`description: '   '`, `path: '   '`) that directly exercise the new `?.trim()` JSX guard.
- Changes `{description && ...}` and `{path && ...}` to `{description?.trim() && ...}` and `{path?.trim() && ...}` in the pending-row render path, with a comment explaining the relationship to the message-handler normalisation.

## Test count

+5 tests (1416 → 1421), all passing. `tsc --noEmit` clean.

PR #3483, opened 2026-05-07
fix(dashboard): apply .trim() guard to CheckpointTimeline description

blamechris:
## Summary

`packages/dashboard/src/components/CheckpointTimeline.tsx:63` rendered `<div class="cp-desc">` whenever `checkpoint.description` was truthy. If the upstream checkpoint event ever delivered a whitespace-only string, the result was a blank div with layout (margins, line-height) but no visible text — invisible padding in the timeline.

This swaps the guard to `checkpoint.description?.trim() && ...` so whitespace-only values short-circuit. The display value is unchanged — `checkpoint.description` is rendered untrimmed (mirrors the SkillsPanel guards from #3441 / #3458).

## Changes

- `CheckpointTimeline.tsx`: line 63 swapped to a `?.trim()` boolean guard, plus a comment cross-referencing #3441/#3458.
- `CheckpointTimeline.edge.test.tsx`: added a `description trim guard (#3461)` describe block with four cases — verbatim render, undefined, whitespace-only spaces, whitespace-only mixed (`\t\n  `).

## Test plan

- [x] `cd packages/dashboard && npm test` — 1433 tests pass (96 files).
- [x] `cd packages/dashboard && npx tsc --noEmit` — clean.
- [x] New tests fail before the fix (verified) and pass after.

Closes #3461


blamechris:
## Code Review: PR #3483

### Summary

Tiny, focused fix that mirrors the SkillsPanel `?.trim()` guard pattern from #3441 / #3458 onto `CheckpointTimeline.tsx:63`. If the server ever emits a whitespace-only `checkpoint.description`, the timeline previously rendered a blank `<div class="cp-desc">` with margins/line-height but no visible text. The new guard short-circuits on whitespace while preserving the authored value (display still renders `checkpoint.description`, not the trimmed result).

### Strengths

- **Display value preserved.** `?.trim()` is used as a boolean predicate only; the rendered text is `checkpoint.description` untrimmed. Matches the explicit acceptance criterion in #3461 and the pattern documented in #3458.
- **Cross-references in code.** The inline comment links #3441 / #3458 so future readers see the family of guards at a glance.
- **TDD discipline.** Tests were added before the fix and verifiably failed without the guard (`<div class="cp-desc">   </div>` reproduced for both whitespace-only cases).
- **Edge coverage.** Four cases: verbatim render, undefined, spaces-only, mixed whitespace (`\t\n  `). The mixed case catches the realistic newline/tab scenario.
- **Test placement.** Added to the existing `CheckpointTimeline.edge.test.tsx` rather than spawning a third file — keeps the edge-case coverage discoverable.
- **Optional-chaining handles undefined safely.** Even though `Checkpoint.description: string` is non-optional in `store-core/types.ts`, `?.trim()` defends against runtime drift if the WS payload ever lies about the shape (which is exactly the "what if upstream misbehaves" failure mode this issue is about).

### Issues Found

#### Critical (Must Fix)

None.

None — the change is one line of production code with proportionate test coverage.

#### Nitpicks (Optional)

- `checkpoint.name` on line 56-57 has the same shape (`{checkpoint.name || fallback}`) and would render whitespace as the literal name (not the unnamed-fallback). Out of scope for #3461 (different field, different ticket), but worth a follow-up issue if we want consistent whitespace hygiene across the row. Filing below.

| Mirror trim-guard onto `checkpoint.name` fallback | (filed below) | Different field, different ticket — #3461 specifically scopes `description`. |

### Architecture Notes

- The guard lives in the dashboard render path only; the Checkpoint shape is defined in `@chroxy/store-core` (`types.ts:261`). No protocol or store-shape changes required — purely a render-time defense.
- The mobile app's `CheckpointView` (`packages/app/src/components/CheckpointView.tsx`) is a separate render path and may want the same treatment, but that's out of scope here.

### Verdict

Surgical fix, faithful to the issue's acceptance criteria, with TDD-verified test coverage. No regressions in the 1433-test dashboard suite. `tsc --noEmit` clean.

blamechris:
Filed follow-up nitpick as #3484 (whitespace-only `checkpoint.name` fallback).

Issue #3222, opened 2026-05-03
feat(server): skills-loader global budget — enforce per-tier before merge to bound peak memory


blamechris:
Spotted in PR #3220 review.

\`loadActiveSkillsLayered\` in \`packages/server/src/skills-loader.js\` loads global + repo skills fully into memory, merges, THEN enforces \`maxTotalSkillBytes\` via \`_enforceTotalBudget\`. Per-skill cap (32KB default) is enforced at read time, but the global cap (256KB default) is post-merge.

## Why it's not urgent

In the common case — a few global skills + a few repo skills — peak memory is well under 1MB. No real-world report of pressure.

## Why it's worth tracking

A user dropping a directory of 200 valid 32KB skills under \`~/.chroxy/skills/\` would transiently allocate 6.4MB just to read them all and then drop most via the budget. On embedded / edge deployments (low-RAM mini PCs, devcontainers with tight cgroup limits) this is the kind of thing that surprises people.

## Suggested approach

Two options, in order of preference:

1. **Two-pass per tier**: stat first, sort by priority, then read only as many files as fit the per-tier budget (allocating budget proportionally between tiers, repo wins ties). Keeps the existing post-merge pruning as a final safety net.
2. **Streaming budget check**: track running byte total in \`loadActiveSkills\` and stop reading once the sum exceeds a per-call cap (which the layered loader divvies up). Less code than (1), more aggressive.

Either way, the public API stays the same — only internals change.

## Out of scope for this issue

Don't add a new option for per-tier budget split until we have a concrete operator request. Default behaviour should be transparent.

PR #3485, opened 2026-05-07
fix(server): sort skills-loader community walk for deterministic order

blamechris:
## Summary

The community walk under `<root>/community/<author>/` in `skills-loader.js` (lines ~402-434) appended `readdirSync` results to the `entries` array in raw filesystem order. Although the top-level `entries` array is sorted alphabetically before the walk, the community entries pushed in later bypassed that sort.

Filesystem `readdir` order is platform-dependent and can shift over time on the same machine. When a tier byte budget cuts off skills mid-walk and skill names tie at the priority/name tiebreaker, the admitted set could differ between runs on the same host.

## Fix

Sort both `authorEntries` (community-level subdir listing) and `authorFiles` (per-author file listing) before iterating so the walk processes community entries in the same alphabetical order as the top-level scan.

## Test plan

- [x] New test file `packages/server/tests/skills-loader-community-walk-order.test.js` mocks `readdirSync` (via `mock.module('fs')`) to return community author dirs and files in reverse-alphabetical order, then asserts the budget-cutoff result is alphabetically deterministic.
  - Confirmed FAILING before the fix (2/2 sub-tests).
  - Confirmed PASSING after the fix (2/2 sub-tests).
- [x] Full skills test suite passes (`skills-loader.test.js`, `skills-loader-toctou.test.js`, `skills-loader-community-walk-order.test.js`, `skills-trust.test.js`, `skills-integration.test.js`): 267 pass, 0 fail, 4 skipped.
- [x] Full server test suite: 75 failures on this branch vs 78 failures on `main` baseline — all unrelated to this change (pre-existing flakes).

Closes #3302


blamechris:
## Code Review: PR #3485

### Summary

Targeted fix for the community-skills walk in `packages/server/src/skills-loader.js` so that the per-tier byte budget produces deterministic results across runs. The two `readdirSync` calls inside the walk (`<root>/community/` and `<root>/community/<author>/`) now return sorted lists, mirroring the alphabetical sort already applied to the top-level `entries` array a few lines above.

This is a small, well-scoped fix to a real determinism bug. Without it, two runs on the same machine (after an FS reorder) could admit different community skills under tight memory budgets, surprising operators of #3222's per-tier guardrail.

### Strengths

- **Causally specific test design.** `mock.module('fs', { namedExports })` is used in the same pattern as the existing `skills-loader-toctou.test.js`, so the new test file fits the repo's idioms exactly. Only `readdirSync` is overridden — every other fs op delegates to the real module, keeping the tmpdir fixture realistic.
- **Test names exploit name-collision tiebreaks.** Both sub-tests use identical skill stems (`skill.md` across two authors; `one.md`/`two.md` across three authors) so the priority/name sort tiebreaker collapses to insertion order — making the test fail before the fix and pass after, with no accidental coverage from the post-merge `_enforceTotalBudget` re-sort.
- **Comments cite the issue (#3302) and explain why.** Both inline comments in `skills-loader.js` and the test file's module-level comment explain the bug semantics, not just the mechanics.
- **Skip guard for `mock.module`.** The test correctly emits a single skip if `--experimental-test-module-mocks` is missing, matching the existing pattern in `skills-loader-toctou.test.js`.
- **Existing skills test suite still green** (267/267, 4 skipped) — and full server suite shows 75 failures vs 78 on `main` baseline (3 fewer, all unrelated/flaky).

### Issues Found

#### Critical (Must Fix)

None.

| `packages/server/src/skills-loader.js` | 410, 433 | Consider matching the comparator used by the top-level sort (line 371): `.sort((a, b) => (a < b ? -1 : a > b ? 1 : 0))`. | The default `Array.prototype.sort()` and the explicit comparator produce identical results for ASCII strings (both are UTF-16 code-unit lexicographic), but using the same explicit comparator everywhere would make it visually obvious that all three sites apply the same ordering. Functional behaviour is identical — purely a consistency/readability nit. |

#### Nitpicks (Optional)

- The new test file is 142 lines, all dedicated to two sub-tests. It could live alongside the existing `skills-loader.test.js` `community-namespace gate` describe block to reduce file proliferation. However, that file already mixes mocked-fs and real-fs tests, and the `mock.module('fs')` install is module-level (must run before the import of `skills-loader.js`) — keeping the `mock.module`-using tests in their own file matches the precedent set by `skills-loader-toctou.test.js`. Leaving as-is is the right call.

None — the only suggestion is a stylistic nit not worth a tracked issue.

### Architecture Notes

The fix is at the right layer. `loadActiveSkills` is the single entry point for both single-pass and two-pass (priority-aware, tier-budgeted) paths, and the community walk runs before either path branches. Sorting at the walk site means both paths inherit deterministic input.

The fix also remains correct under the post-#3279 two-pass priority path: `_collectCandidates` builds descriptors in entry order, then `candidates.sort(_compareByPriorityThenName)` re-sorts by priority desc, name asc. When skill stems are unique across authors, the re-sort hides the bug; when stems collide (which is realistic — many authors will publish a `skill.md` or `index.md`), the priority tiebreak falls back to insertion order, exposing the bug. The fix correctly removes that exposure.

Memory impact: `.sort()` mutates in place, no extra allocation beyond what `readdirSync` already returns. Zero performance cost on the hot path.

### Verdict

The fix is correct, minimal, well-tested, and matches existing project conventions. The only finding is a stylistic suggestion not blocking merge.

blamechris:
## Review Comments Addressed

Copilot review completed and generated no inline comments ("Copilot reviewed 2 out of 2 changed files in this pull request and generated no comments"). Agent-review verdict was Approve with no critical findings.

- Fixed: 0
- False positives: 0
- Follow-up issues created: 0

PR #3487, opened 2026-05-07
feat(server): aggregate warn on EnvironmentManager.reconnect() failure

### LINKED ISSUE #3464: server-cli: log aggregate warn when environmentManager.reconnect() returns false
## Context

Identified during review of PR #3462.

## Description

PR #3462 made `EnvironmentManager.reconnect()` return `Promise<boolean>` so callers can detect when at least one environment was marked unreachable (e.g. its K8s Secret was GC'd, so `reconnectAgentToken` returned `false`).

The single existing caller in `packages/server/src/server-cli.js` discards the return value:

```js
await environmentManager.reconnect()
log.info(`EnvironmentManager ready (${environmentManager.list().length} environment(s))`)
```

When one or more environments are unreachable on startup, the only signal an operator currently sees is the per-env `warn` log buried in the reconnect loop ("Environment X credential source is gone — marking unreachable"). A single aggregate warn at the boot path would make this much more visible in startup logs and on tray-app dashboards.

## Original Review Comment

> `server-cli.js:243` — caller still discards the return value. Optional follow-up: log a single aggregate warn when `await environmentManager.reconnect()` returns `false`.

## Acceptance Criteria

- [ ] When `environmentManager.reconnect()` returns `false`, log a `warn` at the boot path summarizing the count of unreachable environments (e.g. via `environmentManager.list().filter(e => e.status === 'error').length`).
- [ ] The existing healthy-startup `info` log remains for the all-healthy case (avoid double-logging).
- [ ] No regression in the existing test suite — this is a logging-only change.

blamechris:
## Summary

PR #3462 made `EnvironmentManager.reconnect()` return `Promise<boolean>`, but the single caller in `server-cli.js` was discarding the result. When one or more environments are unreachable on startup, the only signal was a buried per-env `warn` log — invisible in startup logs and the tray-app dashboard.

This PR extracts a small `logEnvironmentManagerReconnectResult` helper that:

- always emits the existing `info` summary so the healthy-startup line is unchanged
- when `reconnect()` resolves `false`, additionally emits a single aggregate `warn` summarising the count of unreachable environments derived from `environmentManager.list().filter(e => e.status === 'error').length`

Closes #3464

## Acceptance criteria

- [x] When `environmentManager.reconnect()` returns `false`, log a `warn` summarising the count of unreachable environments
- [x] Existing healthy-startup `info` log remains for the all-healthy case (no double-logging)
- [x] No regression in existing tests (the 6 unrelated failures in `tunnel.integration`, `web-task-manager`, `ws-server*`, `models_updated broadcasting`, and `provider capability gates` reproduce on clean `main`)

## Test plan

- [x] `cd packages/server && npm test -- tests/server-cli-environment-reconnect.test.js` — 4 new tests pass
- [x] `cd packages/server && npm test -- tests/server-cli-banner.test.js tests/environment-manager.test.js` — 86 related tests still pass
- [x] `npm run lint` — no new warnings


blamechris:
## Code Review: PR #3487

### Summary

Wires up the `Promise<boolean>` return value from `EnvironmentManager.reconnect()` (added in #3462) at the only caller — the boot path in `server-cli.js`. Extracts a small `logEnvironmentManagerReconnectResult` helper that always emits the existing healthy `info` summary and additionally emits a single aggregate `warn` when one or more environments came back unreachable. Closes #3464.

### Strengths

- **Pure helper, dependency-injected logger** — matches the existing `buildServerBanner` testability pattern. The 4-test suite covers all four meaningful branches (healthy populated, healthy empty, unhealthy with mixed statuses, allHealthy authoritative over stale env.status) without booting `startCliServer()`.
- **No double-logging.** The `info` line is unconditional and the `warn` is gated on `!allHealthy`, so the all-healthy path is byte-identical to today.
- **JSDoc is excellent** — names the issue numbers, explains the contract from #3462, and warns future readers that the helper is exported purely for tests.
- **Trusts the manager's contract** rather than re-deriving health from `list()`. The 4th test explicitly pins this: a stale `status: 'error'` with `allHealthy: true` does not produce a warn. Good defensive choice.
- **Per-env warnings still fire** in `EnvironmentManager.reconnect()` above the aggregate, so the new line correctly says "see per-environment logs above for details."

### Issues Found

#### Critical (Must Fix)

None.

| `src/server-cli.js` | 121 (helper body) | Add a short inline comment noting that the unreachable-count derivation only stays accurate while every code path that sets `allHealthy = false` in `EnvironmentManager.reconnect()` also sets `env.status = 'error'`. | Today both paths in `environment-manager.js` (missing `containerId`, `reconnectAgentToken === false`) maintain that invariant, but a future contributor flipping `allHealthy` from a different branch (e.g. a metrics/quotas check) without setting per-env status would silently produce `0 environment(s) unreachable` warns. A two-line comment makes the invariant explicit. Related to #3478 which already tracks one gap (reconnectAgentToken throw). |

#### Nitpicks (Optional)

- `tests/server-cli-environment-reconnect.test.js` rolls a local `captureLogger` instead of using `createSpy()` from `tests/test-helpers.js`. The local version is fine and arguably more readable here, but `createSpy()` is the project convention for newer tests (see `codex-session.test.js`, `gemini-session.test.js`). Not worth changing.
- The boot-path call discards the returned boolean (`await logEnvironmentManagerReconnectResult(environmentManager, log)`). That's intentional — startup shouldn't fail on a single unreachable env — but worth confirming there's no upstream consumer (push/dashboard) that should know about it. Looks like the per-env `status: 'error'` already surfaces through `environment-manager.js`'s `environments_reconnected` event, so this is fine.

| Document the `allHealthy <-> status='error'` invariant in `EnvironmentManager.reconnect()` and the new helper | [#3492](https://github.com/blamechris/chroxy/issues/3492) | Two-line comment in two files; nice-to-have hardening that doesn't belong on the issue's hotfix path. |

### Architecture Notes

This change correctly fits the boot-path layering: `startCliServer()` orchestrates components and emits user-visible status; `EnvironmentManager` owns per-env detail. The helper preserves that boundary by reading from `list()` (public API) rather than reaching into `_environments`. The exported-pure-helper pattern (`buildTunnelWarmingStatus`, `buildServerBanner`, now `logEnvironmentManagerReconnectResult`) is now well-established in this file and stays consistent.

The aggregate warn is also visible to log listeners (`addLogListener` in `logger.js`), which means the desktop tray dashboard and any future monitoring hooks will see it without further plumbing — exactly the surfacing the issue called for.

### Verdict

Small, well-tested, low-risk fix that closes the original observability gap with no protocol or data-flow changes.

PR #3496, opened 2026-05-07
fix(server): use --no-trunc in DockerBackend container listing

blamechris:
## Summary

`DockerBackend._listChroxyContainers()` used `docker ps -q`, which returns 12-character truncated container IDs. `DockerBackend.createEnvironment()` persists the full 64-character ID returned by `docker run`, and the `reconcile()` loop does an exact string comparison against stored IDs.

The mismatch meant every known container appeared to be an orphan on every reconcile cycle and was destroyed.

Add ``--no-trunc`` to the ``ps`` invocation so both sides of the comparison use the same full-length form.

## Changes

- ``packages/server/src/environments/backends/docker.js`` — pass ``--no-trunc`` to ``docker ps`` in ``_listChroxyContainers()`` and document the rationale inline.
- ``packages/server/tests/environments/backends/docker.test.js`` — assert ``listEnvironments()`` invokes docker ps with ``--no-trunc`` (regression guard at the unit level).
- ``packages/server/tests/environment-manager.test.js`` — higher-level reconcile() test where the mock simulates real Docker behavior (truncates to 12 chars unless ``--no-trunc`` is present), confirming a known full-length container ID is NOT destroyed.

## Test plan

- [x] New unit test fails before the fix and passes after
- [x] New reconcile test fails before the fix (known container would be destroyed) and passes after
- [x] All existing docker / environment-manager / backend-seam tests still pass (148/148)
- [ ] Verify on a live Docker host that ``reconcile()`` no longer eats known containers after restart

Closes #3314


blamechris:
## Code Review: PR #3496

### Summary

Adds `--no-trunc` to `DockerBackend._listChroxyContainers()` so `docker ps -q` returns the full 64-character container IDs that match the form persisted by `createEnvironment()`. Without this, `reconcile()`'s exact-string comparison treats every known container as an orphan and destroys it — a critical bug for any user with persistent environments who restarts the daemon.

The fix is one line plus an inline comment, complemented by two regression tests:
- A unit test asserting `--no-trunc` is in the args.
- An integration-style test that simulates real Docker truncation behavior and exercises `reconcile()` end-to-end through `EnvironmentManager`.

### Strengths

- **Minimal, surgical fix.** A single argument change at the right layer (the backend shellout helper).
- **Inline comment explains the why and links the issue.** Future maintainers won't be tempted to revert it.
- **Two-level test coverage.** The unit test guards against accidental flag removal; the higher-level reconcile test simulates real Docker truncation behavior (returning 12-char IDs unless `--no-trunc` is present), making it a true regression test that actually fails before the fix.
- **TDD discipline.** Tests fail first, fix applied, tests pass — verified in commit history.
- **Preserves existing test conventions.** Uses the same mock execFile shape and test seeding pattern as the surrounding suite.

### Issues Found

#### Critical (Must Fix)
None.

None blocking. One observation:

| `packages/server/tests/environment-manager.test.js` | 1968 | Consider also asserting that the `inspect` step (or other reconcile path) is *not* called for the known container | Strengthens the negative assertion: today the test relies on `removedContainers.length === 0`, which is correct but doesn't exclude the broader category of "did anything dangerous happen to my container". Low priority — current assertion is sufficient. |

#### Nitpicks (Optional)
- `truncatedId` is computed but never asserted directly (only used to drive the mock branch). Minor — keeping it explicit aids readability.

None. The fix is self-contained and complete.

### Architecture Notes

This sits in the ALL-Docker-shellout-owns-everything seam (`DockerBackend`), which is the right place. The fix is independent of `EnvironmentManager` (the consumer) and `K8sBackend` (which has `listEnvironments` deferred to Phase 2 anyway). The `_inspectContainer` and `_renameContainer` helpers also accept full or truncated IDs because Docker resolves both, but `reconcile()` is the only path where exact-string comparison matters — so this is the only place the fix is needed.

The `WARN`-level reconcile log already prefixes container IDs with `.slice(0, 12)` for readability, which means logs remain unchanged after this fix. Good.

### Verdict

- [x] **Approve - Ready to merge**
- [ ] Request Changes
- [ ] Comment

Solid bug fix with strong test coverage. No blockers.

PR #3522, opened 2026-05-07
fix(server): tighten reconnectAgentToken non-true return check

blamechris:
## Summary

Closes #3495.

Identified during review of #3491 (closes #3478). The `EnvironmentManager.reconnect()` check for the result of `reconnectAgentToken()` used a strict `=== false`, which silently treats `undefined` / `null` / any other falsy non-`false` value as success. No current backend (DockerBackend, K8sBackend) returns anything other than `true` / `false`, so this is **purely defensive contract tightening** — not a current bug.

## Decision (per #3495 acceptance criteria)

Adopt **strict `=== true` for success** as the contract. Anything else — `false`, `undefined`, `null`, or any other non-`true` value from a misbehaving / future backend — is treated as the same unreachable signal:

- `env.status = 'error'`
- `allHealthy = false`
- operator-visible warn log

This is consistent with the existing #3478 decision that a thrown `reconnectAgentToken` is the same unreachable signal as returning `false`.

## Changes

- `packages/server/src/environment-manager.js`: flip `ok === false` → `ok !== true`. Inline comment and JSDoc enumerate the new "any non-`true` return is unreachable" rule and cross-reference #3495.
- `packages/server/tests/environment-manager.test.js`: 2 new tests asserting `reconnect()` returns `false` and marks the env `error` for both `undefined` and `null` returns from `reconnectAgentToken`.

## Test plan

- [x] New tests fail before the fix, pass after the fix (TDD verified)
- [x] All 86 environment-manager tests pass (was 84, +2 new)
- [x] No new failures vs. main in full server test suite — pre-existing flakes (BaseSession import, `@anthropic-ai/claude-agent-sdk` hoist, integration tests) confirmed identical on `origin/main`

### DIFF
--- a/packages/server/src/environment-manager.js
+++ b/packages/server/src/environment-manager.js
@@ -407,7 +407,9 @@ export class EnvironmentManager extends EventEmitter {
    *   - it has no `containerId`
    *   - `getEnvironmentStatus` reports the container is stopped
    *   - `getEnvironmentStatus` throws (container/handle not found)
-   *   - `reconnectAgentToken` returns `false` (credential source GC'd)
+   *   - `reconnectAgentToken` returns any non-`true` value, e.g. `false`
+   *     (credential source GC'd) or `undefined` / `null` from a misbehaving
+   *     backend (#3495)
    *   - `reconnectAgentToken` throws (transient error treated as same signal)
    */
   async reconnect() {
@@ -452,10 +454,15 @@ export class EnvironmentManager extends EventEmitter {
       // A thrown error is treated as the same unreachable signal as returning
       // false (#3478): the operator cannot rely on the credential, so the
       // environment must be marked error and reconnect() must report failure.
+      //
+      // Defensive (#3495): treat any non-`true` return (including `undefined` /
+      // `null` from a misbehaving or future backend) as the same unreachable
+      // signal. The contract is "strict `=== true` for success" — a missing
+      // credential is just as unusable as `false`.
       if (typeof this._backend.reconnectAgentToken === 'function') {
         try {
           const ok = await this._backend.reconnectAgentToken(env.containerId)
-          if (ok === false) {
+          if (ok !== true) {
             env.status = 'error'
             allHealthy = false
             log.warn(`Environment "${env.name}" (id: ${env.id}) credential source is gone — marking unreachable`)
--- a/packages/server/tests/environment-manager.test.js
+++ b/packages/server/tests/environment-manager.test.js
@@ -786,6 +786,83 @@ describe('EnvironmentManager.reconnect()', () => {
       'reconnect() must return false when getEnvironmentStatus throws')
     assert.equal(manager.get('env-status-throw').status, 'error')
   })
+
+  // ──────────────────────────────────────────────────────────────────────────
+  // Defensive hardening for non-true backend returns (#3495)
+  //
+  // The Backend protocol JSDoc only documents `true` / `false` returns from
+  // reconnectAgentToken. A misbehaving or future backend that returns
+  // `undefined` / `null` / any other non-true value should be treated as
+  // failure — a missing credential is just as unusable as `false`.
+  // ──────────────────────────────────────────────────────────────────────────
+
+  it('marks environment unreachable when reconnectAgentToken returns undefined (#3495)', async () => {
+    const seedData = {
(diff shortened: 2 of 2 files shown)


blamechris:
## Code Review: PR #3522

### Summary

Tightens the `EnvironmentManager.reconnect()` boolean check for `reconnectAgentToken` from `ok === false` to `ok !== true`, so that any non-`true` return (`false` / `undefined` / `null` / other) flips `allHealthy = false` and marks the env `error`. JSDoc and inline comments are updated to enumerate the new "strict `=== true` for success" contract. Two new tests cover `undefined` and `null` returns.

### Strengths

- **Decision matches issue acceptance criterion.** The PR body explicitly picks the "strict `=== true` for success" contract from #3495's two options and documents it inline. Operators reading the JSDoc (or the inline comment at the catch block) no longer have to guess — both now enumerate `false` / `undefined` / `null` as equivalent unreachable signals.
- **Consistent with #3478 throw-handling.** The new "any non-`true` is unreachable" rule is the natural symmetric extension of "throw is unreachable" — same status flip, same warn log, same `allHealthy` consequence. Callers see one signal regardless of how the backend misbehaved.
- **TDD discipline.** The two new tests (`undefined` and `null`) were verified to fail before the fix and pass after. They cover the symmetric pair (no return + explicit null) and reuse the exact seed-data shape of the existing `#3478` tests, keeping the test file consistent.
- **JSDoc cross-references the issue.** `(#3495)` annotation on the new bullet point makes the historical rationale discoverable from the source — operators chasing an unfamiliar contract can `gh issue view 3495` to see why.
- **Inline comment at the call site.** The new "Defensive (#3495)" block at lines 457–460 of the diff makes it obvious WHY a misbehaving backend is treated as failure rather than success, which is the surprising direction.
- **No protocol surface changes.** Purely internal contract tightening. Single existing call site (`server-cli.js:243`) discards the return value today, so no caller-visible regression. Forward-looking: a future supervisor restart loop or `/api/health` endpoint that gates on `reconnect()` will get the correct signal.
- **Per-env serialization preserved.** No new races introduced. `reconnect()` doesn't take per-env locks (loops over snapshot), and the change is a single boolean flip — no state ordering subtleties.

### Issues Found

#### Critical (Must Fix)

None.

None within scope. The change is minimal and the intent is clear.

#### Nitpicks (Optional)

- The new test comment `/* returns undefined */` on the `async reconnectAgentToken()` body is helpful but slightly unusual for this file (no other tests use that pattern). Consistency-wise either is fine — the explicit comment is more discoverable for future readers, so leave as-is.
- Inline comment at lines 457–460 partially overlaps with the JSDoc bullet at line 410. Defensive duplication is fine for surprising semantics — the JSDoc tells the contract reader what the function returns; the inline comment tells the implementation reader why. Different audiences, different perspectives.

None. All acceptance criteria from #3495 are satisfied in this PR.

### Architecture Notes

This is the third in a chain of defensive tightenings to `EnvironmentManager.reconnect()`:

1. **#3339** — split out `reconnectAgentToken` as a per-backend reconnection hook
2. **#3478** (PR #3491) — flip `allHealthy = false` for `reconnectAgentToken` throws and `getEnvironmentStatus` failures
3. **#3495** (this PR) — extend "non-success" coverage to `undefined` / `null` returns

Each step has narrowed the contract from "best-effort" to "strict `=== true` for success", which is the right direction for a method whose return value will be used by future health-gating code (supervisor restarts, `/api/health` degraded signal). The Backend protocol JSDoc only documents `true` / `false` returns, so this PR's new check rejects accidental violations of that contract from misbehaving / future backends without changing behavior for compliant ones.

The decision to treat `undefined` / `null` as the same signal as `false` (rather than as a programming-error throw) is the right call. A backend that forgets to return a value or returns `null` instead of `false` is _likely_ buggy, but operators care about "is the environment usable?" not "is the backend strictly conformant?" — and from the operator's perspective both answers are "no".

### Verdict

- [x] **Approve - Ready to merge**
- [ ] Request Changes
- [ ] Comment

Minimal, well-scoped, defensive contract tightening with TDD-verified test coverage. The decision documented in the PR body satisfies the #3495 acceptance criteria, and the JSDoc + inline comments propagate the new contract to all future readers. No protocol changes, no behavioral regression for compliant backends.

Issue #3307, opened 2026-05-05
feat(server): return INVALID_AUTHOR when skill exists but author namespace mismatches in skill_trust_grant


blamechris:
## Problem

Flagged by Copilot on PR #3303 (comment `3186164529`).

In `handleSkillTrustGrant` (`settings-handlers.js`), when the loop finds a skill under `community/alice/` but the request claims `author: 'bob'`, it skips with `continue`. If no other root resolves the skill for `bob`, `resolvedPath` remains `null` and `SKILL_NOT_FOUND` is returned.

Clients (and the acceptance criteria) expect `INVALID_AUTHOR` to be returned when the skill exists but the claimed author doesn't match the community namespace — making it easier to distinguish "skill missing" from "wrong author".

## Fix

Track during the loop whether a namespace mismatch was detected (skill found on disk but `actualAuthor !== msg.author`). After the loop, if `resolvedPath === null` but a mismatch was detected, return `INVALID_AUTHOR` instead of `SKILL_NOT_FOUND`.

## Files

- `packages/server/src/handlers/settings-handlers.js` — `handleSkillTrustGrant`, resolution loop (~line 804)
- Tests: `packages/server/tests/handlers/settings-handlers.test.js`

PR #3523, opened 2026-05-07
test(server): guard symlinkSync usage for cross-platform portability

blamechris:
## Summary
- Wrap the `symlinkSync` call in `packages/server/tests/handlers/settings-handlers.test.js` (added in #3307 work) in a try/catch with an early return.
- On platforms where symlinks aren't permitted (Windows without dev mode, restricted/sandboxed CI) the test now skips silently instead of failing the suite.
- Pattern matches the existing symlink-defense tests in `skills-loader.test.js`, `file-ref-attachments.test.js`, etc.

## Test plan
- [x] `node --test packages/server/tests/handlers/settings-handlers.test.js` passes on macOS (Node 22) — 72/72 tests pass, including the `INVALID_AUTHOR when skill exists under a different community author (#3307)` case.
- [x] `rmSync` cleanup remains in the outer `finally` block, so the early return from the inner catch still cleans the temp dir.

Closes #3517

### DIFF
--- a/packages/server/tests/handlers/settings-handlers.test.js
+++ b/packages/server/tests/handlers/settings-handlers.test.js
@@ -1148,10 +1148,17 @@ describe('settings-handlers', () => {
         // handler resolves it via realpathSync, sees actualAuthor='alice', and
         // detects the namespace mismatch against the claimed author 'bob'.
         mkdirSync(join(skillsDir, 'community', 'bob'), { recursive: true })
-        symlinkSync(
-          join(skillsDir, 'community', 'alice', 'foo.md'),
-          join(skillsDir, 'community', 'bob', 'foo.md'),
-        )
+        // Match the existing "symlink defense" tests: skip silently on
+        // platforms (Windows / restricted CI) where symlinkSync isn't
+        // permitted, rather than failing the suite.
+        try {
+          symlinkSync(
+            join(skillsDir, 'community', 'alice', 'foo.md'),
+            join(skillsDir, 'community', 'bob', 'foo.md'),
+          )
+        } catch {
+          return
+        }
         const trustStore = makeCommunityTrustStore()
         const sessions = new Map()
         const session = createMockSession()


blamechris:
## Code Review: PR #3523

### Summary
Tiny test-only change (+11/-4) that wraps the `symlinkSync` call inside the `INVALID_AUTHOR when skill exists under a different community author (#3307)` test in a `try/catch` with an early return, so the suite skips silently on platforms where symlink creation is denied (Windows without dev mode, restricted/sandboxed CI). Pattern matches the existing symlink-defense tests in `skills-loader.test.js`, `file-ref-attachments.test.js`, and `ws-file-ops-error-paths.test.js`.

### Strengths
- Pattern is identical to the closest precedent (`skills-loader.test.js:1976`) — same comment text, same `try { symlinkSync(...) } catch { return }` shape. Easy to grep for and reason about.
- Early `return` happens inside the outer `try { ... } finally { rmSync(...) }`, so the temp `skillsDir` is still cleaned up on the skip path. Verified by re-reading lines 1142–1181.
- Comment explains both the "what" (skip) and the "why" (Windows / restricted CI) in two lines — good signal density.
- Diff is genuinely test-only, no production code touched, no new imports, no Node 22 surface area.

### Issues Found

#### Critical (Must Fix)
None.

None for this PR. The matching tests immediately above and below (`SKILL_NOT_FOUND when no skill exists for any author (#3307)` and `resolves community skills stored with .markdown extension`) do not call `symlinkSync` and therefore do not need the same guard. Confirmed by grep — only one `symlinkSync` call exists in this file.

#### Nitpicks (Optional)
- A `t.skip()` (Node test runner skip) would surface "skipped" in test output instead of silently passing. The repo convention is the silent early-return pattern, so matching it is correct here — only flagging in case the team wants to standardise on visible skips later (would be a cross-cutting cleanup, not this PR's job).

None — no out-of-scope suggestions worth tracking.

### Architecture Notes
Pure test-portability fix. No protocol, runtime, or behavioural change. Closes #3517 cleanly with one of the smallest possible diffs that satisfies the acceptance criteria.

### Verdict

PR #3526, opened 2026-05-07
chore(server): include env.id in token-refresh failure warning

blamechris:
## Summary

In `EnvironmentManager.reconnect()`, the `catch` block on `reconnectAgentToken` previously logged:

```
Environment "<name>" token refresh failed: <msg>
```

The sibling `ok === false` branch a few lines above already includes `(id: <env.id>)`. This PR adds the same `(id: <env.id>)` token to the `catch` warning so operators can correlate token-refresh failures to a specific persisted environment from the per-env list.

New format:

```
Environment "<name>" (id: <env.id>) token refresh failed: <msg>
```

One-line change in `packages/server/src/environment-manager.js`. No test changes — existing tests don't assert on the exact warn text. Verified `tests/environment-manager.test.js` (84 tests) and `tests/server-cli-environment-reconnect.test.js` (4 tests) all pass under Node 22.

Closes #3512

## Test plan

- [x] `node --test packages/server/tests/environment-manager.test.js` — 84 pass / 0 fail
- [x] `node --test packages/server/tests/server-cli-environment-reconnect.test.js` — 4 pass / 0 fail

### DIFF
--- a/packages/server/src/environment-manager.js
+++ b/packages/server/src/environment-manager.js
@@ -463,7 +463,7 @@ export class EnvironmentManager extends EventEmitter {
         } catch (err) {
           env.status = 'error'
           allHealthy = false
-          log.warn(`Environment "${env.name}" token refresh failed: ${err.message}`)
+          log.warn(`Environment "${env.name}" (id: ${env.id}) token refresh failed: ${err.message}`)
         }
       }
       // Clear stale session references — sessions don't survive server restart


blamechris:
## Code Review: PR #3526

### Summary

One-line surgical fix in `EnvironmentManager.reconnect()`. The `catch` block on `reconnectAgentToken(...)` is brought into format-parity with its sibling `if (ok === false)` branch by including `(id: ${env.id})` in the warn message. Closes #3512.

### Strengths

- **Symmetry restored**: the `ok === false` branch (line 461) and the `catch` branch (line 466) now share the same `Environment "<name>" (id: <id>)` prefix, so logs grep cleanly across both unreachable signals.
- **No test churn**: existing tests assert on env status / `reconnect()` return value, not the exact warn text — touching them was correctly avoided.
- **Style compliant**: ES module, no semicolon, single quotes, template literal — matches the surrounding file.
- **Atomic diff**: 1 line changed, 1 file, no incidental edits. Easy revert if ever needed.

### Issues Found

#### Critical (Must Fix)

None.

| `packages/server/src/environment-manager.js` | 435, 440 | The two earlier `log.warn` lines on the same `reconnect()` path (`container is stopped`, `container inspect failed`) still omit `env.id`. Bringing them to the same `(id: ${env.id})` shape would complete the parity story #3512 is chasing. | Out of scope for this PR — tracked as #3528 so the operator-correlation story is consistent across all four unreachable branches in `reconnect()`, not just two. |

#### Nitpicks (Optional)

- None. Commit message is clean (no attribution), follows the `chore(server):` convention, and references the closing issue. Branch name `feat/3512-...` would arguably be `chore/3512-...` to match the type, but this is bikeshed-level.

| Extend `(id: ${env.id})` to the `container is stopped` / `container inspect failed` warnings in the same `reconnect()` path | #3528 (https://github.com/blamechris/chroxy/issues/3528) | Out of scope for #3512 (which specifically called out the token-refresh failure path). Worth tracking so the operator-correlation story is uniform across all four unreachable branches. |

### Architecture Notes

This is a pure log-message change — no behaviour, no protocol surface, no state, no public API. The `reconnect()` method is invoked once at server boot from `server-cli.js#logEnvironmentManagerReconnectResult`, which already aggregates per-env unreachable counts; per-env detail in the warn is the only way an operator can map an aggregate "1 environment unreachable" line back to a concrete env. Including `env.id` (rather than just the name) is the right field — names can be duplicated by mistake while ids are unique by construction (`env-` + 8 random bytes hex).

### Verdict

- [x] **Approve** — Ready to merge.

PR #3529, opened 2026-05-07
fix(server): suppress sidecar close handler after terminal error closes WS

blamechris:
## Summary
- Add `_terminalErrorSent` session flag in the sidecar agent. Set it inside `_handleStdinDrainStalled` (and inside the existing `oversized_line` handler) before kill/close.
- Make the child `'close'` handler return early when the flag is set, mirroring the existing `_oversized` guard. Without this, after a `stdin_drain_stalled` recovery (which closes WS with 1011), the natural child `'close'` event could race and emit `exit` + close 1000 — clients would observe the wrong terminal frame.
- Regression test installs an instrumented fake `activeWs` that defers its `send` callback by a macrotask, fires the child `'close'` before the error-frame send callback runs, and asserts the client only sees `error(stdin_drain_stalled)` + `close(1011)` — never `exit` + `close(1000)`. Verified the test fails without the fix.

## Test plan
- [x] `node --test packages/server/tests/sidecar/agent.test.js` (Node 22) — 96/96 pass
- [x] Verified regression test fails when the close-handler guard is reverted
- [x] `node --test packages/server/tests/integration/k8s-sidecar-roundtrip.test.js` passes

Closes #3513

### DIFF
--- a/packages/server/sidecar/agent.js
+++ b/packages/server/sidecar/agent.js
@@ -528,6 +528,13 @@ export class PodAgent {
       `[chroxy-pod-agent] stdin drain stalled for session ${sessionId} (${this._stdinDrainTimeoutMs}ms) — killing wedged child`,
     )
 
+    // Mark the session as having emitted a terminal error frame BEFORE we
+    // kill the child. The subsequent child 'close' event must not re-emit a
+    // generic exit/close 1000 — the client should observe the explicit
+    // stdin_drain_stalled error + close 1011 we send below. Mirrors the
+    // _oversized guard used by the line-too-long path (#3513).
+    session._terminalErrorSent = true
+
     // Clear the draining flag so any late 'drain' event (e.g. once the child
     // is killed) does not try to resume an already-closing WS.
     session._stdinDraining = false
@@ -830,6 +837,12 @@ export class PodAgent {
       // if no 'drain' event arrives within _stdinDrainTimeoutMs and forces a
       // wedged-child recovery (error frame + kill + close 1011).
       _stdinDrainTimer: null,
+      // Terminal-error guard (#3513). Set true by any error path that emits a
+      // terminal error frame and closes the WS itself (e.g. stdin_drain_stalled,
+      // line_too_long via the _oversized flag below). The child 'close' handler
+      // checks this flag and returns early so it cannot race with the terminal
+      // error path and override the close code with a generic exit/1000.
+      _terminalErrorSent: false,
     }
     this._sessions.set(sessionId, session)
     ws._sessionId = sessionId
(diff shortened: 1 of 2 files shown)


blamechris:
## Code Review: PR #3529

### Summary
Adds a session-level `_terminalErrorSent` flag to the sidecar agent and gates the child `'close'` handler on it, so that after a terminal-error path (`_handleStdinDrainStalled`, `oversized_line`) has already killed the child and closed the WS, the natural child `'close'` event cannot race and emit a spurious `exit` + close `1000`. Pure internal hardening — no protocol changes.

### Strengths
- Tight, minimal fix that mirrors the existing `_oversized` guard (consistent with `#3380`)
- New flag is generalized: useful for any future terminal-error path, not just stdin drain stall
- Excellent regression test — uses an instrumented fake `activeWs` that defers `send` callbacks by a macrotask to deterministically reproduce the race (same shape as the `#3399` line_too_long test)
- Confirmed the test fails without the close-handler guard (TDD)
- Inline comments link the change to `#3513`, `#3380`, and explain the invariant
- Implementation now matches the documented behavior in `packages/server/sidecar/PROTOCOL.md` ("No `exit` frame follows" for both `line_too_long` and `stdin_drain_stalled`)

### Issues Found

#### Critical (Must Fix)
None.

None.

#### Nitpicks (Optional)
- The `oversized_line` handler now sets BOTH `_oversized = true` AND `_terminalErrorSent = true`, with the close-handler checking both. The dual-flag is intentional (`_oversized` is preserved for backward compatibility, `_terminalErrorSent` is the unified guard) — comment explains this clearly. A future cleanup could collapse to a single flag, but mixing now would risk subtle behavioural drift in unrelated paths that read `_oversized`. Leave as-is.
- Test declares `KILL_GRACE_MS = 25` and passes it through `killGraceMs`. It's only used to keep the fake clock budget tight; not strictly necessary but matches the surrounding test style.

None — no out-of-scope suggestions.

### Architecture Notes
The fix sits cleanly within the `PodAgent` session lifecycle. Two terminal-error paths can race with the natural child `'close'` event:
1. `_handleStdinDrainStalled` (closes 1011) — newly guarded by `_terminalErrorSent`
2. `lineGuard.once('oversized_line')` (closes 1008) — already guarded by `_oversized`, now also sets `_terminalErrorSent`

A third terminal path, `child.on('error')` (async spawn failure), is not affected because no `'close'` event will fire when the child never started — verified by reading the surrounding spawn flow. No further race surface.

This change makes the sidecar agent's WS close codes deterministic from the K8sBackend's perspective, which simplifies downstream handling of `1011` vs `1000` mappings.

### Verdict
- [ ] Request Changes
- [ ] Comment

CI: pending. All 96 sidecar tests + the k8s-sidecar-roundtrip integration test pass locally on Node 22.

PR #3533, opened 2026-05-07
fix(server): clear stale sessions on no-containerId reconnect

### LINKED ISSUE #3494: EnvironmentManager.reconnect(): no-containerId branch skips session cleanup
## Context

Identified during review of PR #3491 (closes #3478).

## Description

In `packages/server/src/environment-manager.js` `reconnect()`, the early-return path for environments that have no `containerId` is:

```js
if (!env.containerId) {
  env.status = 'error'
  allHealthy = false
  continue
}
```

This `continue` skips both the `reconnectAgentToken` block (correctly — there is nothing to reconnect) and the trailing `env.sessions = []` cleanup at the end of the loop body (incorrectly — stale session refs for an env with no container are even more useless than for one with a container).

After PR #3491, the JSDoc explicitly enumerates "no containerId" as an unreachability cause, which makes this asymmetry slightly more visible.

## Original Review Comment

> The `if (!env.containerId)` short-circuit `continue`s past the `reconnectAgentToken` block AND past `env.sessions = []`. Stale sessions on no-containerId envs will not be cleared.

## Acceptance Criteria

- [ ] Move `env.sessions = []` ahead of the `continue`, OR restructure the loop so cleanup runs in a `finally`-style trailer regardless of branch
- [ ] Add a regression test that seeds an env with no `containerId` and stale `sessions: ['stale-1']`, asserts `sessions === []` after `reconnect()`

blamechris:
## Summary

`EnvironmentManager.reconnect()` had a `continue` in the no-`containerId` branch that skipped past the trailing `env.sessions = []` cleanup at the bottom of the loop body. After a server restart, an environment with no container kept its stale session refs — which is exactly backwards: a no-container env is even less reachable than a stopped one, so its in-memory session list is even more useless.

Move the cleanup to the top of the loop body so it runs unconditionally for every persisted environment. In-memory session state never survives a server restart, so the cleanup has no value-dependent guard.

Identified during review of #3491 (closes #3478).

## Changes

- `packages/server/src/environment-manager.js` — hoist `env.sessions = []` to the top of the reconnect loop body so it runs before the no-`containerId` `continue`.
- `packages/server/tests/environment-manager.test.js` — regression test that seeds an env with `containerId: null` and `sessions: ['stale-1', 'stale-2']`, then asserts:
  - `reconnect()` returns `false`
  - `env.status === 'error'`
  - `env.sessions === []`
  - the backend is never consulted (mock throws if either `getEnvironmentStatus` or `reconnectAgentToken` is called)

## Test plan

- [x] New regression test passes
- [x] Existing reconnect tests still pass (85/85 in `environment-manager.test.js`)
- [x] Environment handler tests still pass

Closes #3494

### DIFF
--- a/packages/server/src/environment-manager.js
+++ b/packages/server/src/environment-manager.js
@@ -421,6 +421,13 @@ export class EnvironmentManager extends EventEmitter {
     let allHealthy = true
 
     for (const env of this._environments.values()) {
+      // Clear stale session references unconditionally — in-memory session
+      // state never survives a server restart, regardless of whether the
+      // environment's container is reachable, stopped, or has no containerId
+      // at all. (#3494: this used to live at the bottom of the loop body and
+      // was skipped by the no-containerId `continue` below.)
+      env.sessions = []
+
       if (!env.containerId) {
         env.status = 'error'
         allHealthy = false
@@ -473,8 +480,6 @@ export class EnvironmentManager extends EventEmitter {
           log.warn(`Environment "${env.name}" (id: ${env.id}) token refresh failed: ${err.message}`)
         }
       }
-      // Clear stale session references — sessions don't survive server restart
-      env.sessions = []
     }
 
     this._persist()
--- a/packages/server/tests/environment-manager.test.js
+++ b/packages/server/tests/environment-manager.test.js
@@ -863,6 +863,48 @@ describe('EnvironmentManager.reconnect()', () => {
     assert.equal(manager.get('env-cred-null').status, 'error',
       'env should be marked error when token refresh returns null')
   })
+
+  // Regression for #3494: previously the no-containerId branch `continue`d past
+  // the `env.sessions = []` cleanup, leaving stale session refs on an env that
+  // is even less reachable than one whose container is merely stopped.
+  it('clears stale sessions on environments with no containerId (#3494)', async () => {
+    const seedData = {
+      version: 1,
+      environments: [{
+        id: 'env-no-container',
+        name: 'no-container-env',
+        cwd: '/tmp',
+        image: 'node:22-slim',
+        containerId: null,
+        containerUser: 'chroxy',
+        containerCliPath: '/usr/local/cli.js',
+        status: 'running',
+        sessions: ['stale-1', 'stale-2'],
+        createdAt: '2026-03-17T00:00:00Z',
+        memoryLimit: '2g',
+        cpuLimit: '2',
+      }],
+    }
+    writeFileSync(statePath, JSON.stringify(seedData))
+
+    // Backend should never be consulted for a no-containerId env — assert by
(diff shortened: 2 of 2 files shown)


blamechris:
## Code Review: PR #3533

### Summary

Targeted bug fix for the no-`containerId` branch in `EnvironmentManager.reconnect()`. The `continue` at the end of that branch was skipping the trailing `env.sessions = []` cleanup, so a persisted environment with no container survived a server restart with its stale session list intact. The fix hoists the cleanup to the top of the loop body so it runs unconditionally for every iteration, regardless of which branch the rest of the body takes.

### Strengths

- **Minimal, surgical diff.** Two-line move + comment, with no behavior change for the other unreachable causes (stopped, throws, token-gone, token-throws). Each of those already passed through the trailing cleanup.
- **Cleanup hoist is the right shape.** Sessions are persisted to disk (`addSession`/`removeSession` call `_persist()`), so after `_restore()` repopulates `_environments`, every env carries stale session refs. The cleanup belongs at the top of the loop because there is no value-dependent guard — in-memory session state never survives a restart, which the new comment correctly states.
- **Test is well-designed.** Seeds the exact failure mode (`containerId: null` + `sessions: ['stale-1', 'stale-2']`), asserts all three observable post-conditions (return value, status, sessions), AND uses a backend that throws on either method to verify the no-containerId branch short-circuits before any backend call. The throwing-backend pattern is a nice positive assertion that the early `continue` is preserved and the fix didn't accidentally fall through.
- **Code style.** ES modules, single quotes, no semicolons, comment cites the issue number for future spelunkers. Matches the rest of the file.

### Issues Found

#### Critical (Must Fix)

None.

None worth blocking on. The fix is the minimal correct change for the issue as filed.

#### Nitpicks (Optional)

- Comment is slightly long (5 lines) for a 1-line action. Could be trimmed to 2-3 lines, but the historical context ("this used to live at the bottom...") is genuinely useful for anyone bisecting reconnect() behavior, so keeping it is defensible.

### Architecture Notes

`reconnect()` is the boundary between persisted state on disk and live in-memory state on a fresh server. Anything that does not survive a restart should be cleared as part of `_restore() → reconnect()`, regardless of the env's reachability outcome. Hoisting the cleanup makes that contract explicit at the loop entry rather than burying it as a trailer that any future `continue` could skip — same lesson as the original bug. Future maintenance: if more "doesn't survive a restart" fields show up, they belong next to `env.sessions = []` at the top.

### Verdict

The fix is correct, the regression test exercises the exact reported failure mode plus a positive assertion that the no-containerId short-circuit is preserved, and the diff is the minimum needed. No follow-up issues warranted.

blamechris:
## Review Comments Addressed

Copilot reviewed both changed files and generated **zero inline comments**. No fixes, false positives, or follow-up issues required.

**Total:** 0 comments to address — nothing pending.

PR #3537, opened 2026-05-07
feat(server): cumulative dropped-bytes counter + louder log severity for stdin_dropped

blamechris:
Closes #3506.

## Decision

**Both — error-level on key events + cumulative running counter.** The issue offered three options (louder severity, cumulative counter, or both). Picked option 3 because the two signals serve different audiences:

- **Loud severity** → first-time triage. An operator scanning errors needs the failure to *exist* in the right log bucket.
- **Cumulative total** → ongoing triage. Once the operator has the error in hand, they need to know "how bad is this?" — 60 bytes is annoying, 10 MiB is a serious data-loss event.

## What changes

`SdkSession._attachSidecarProcessListeners` now:

1. Tracks `_stdinDroppedBytesTotal` and `_stdinDroppedCount` for the lifetime of the session.
2. Emits every drop log line with a `cumulative=<N> bytes over <M> drops` suffix.
3. Escalates to `log.error` on:
   - the **first drop** (one-shot loud signal),
   - **every 10th drop** (`STDIN_DROPPED_ESCALATION_EVERY_N`) — so a flood of small drops still surfaces,
   - the **first time cumulative bytes cross 10 MiB** (`STDIN_DROPPED_BYTES_ERROR_THRESHOLD`) — one-shot per session.
4. All other drops continue to log at `warn`, preserving the existing per-event audit trail and not flooding logs during a hot loop.

Both thresholds are `export`ed constants so tests reference them symbolically and operators tuning them have a single source of truth.

### Threshold rationale

| Constant | Value | Why |
|---|---|---|
| `STDIN_DROPPED_BYTES_ERROR_THRESHOLD` | 10 MiB | Equal to ten full pre-dial-cap chunks. Below this, lost input is annoying; at/above this, it's almost certainly a corrupt long turn or large attachment that demands an error-level signal. |
| `STDIN_DROPPED_ESCALATION_EVERY_N` | 10 | Catches the "many small drops" case (e.g. unknown-bytes payloads that don't move the cumulative byte counter). Re-escalates at a low enough cadence to survive log filtering without spamming. |

### Edge cases

- Unknown / missing `bytes` payloads still increment `_stdinDroppedCount` but do **not** poison `_stdinDroppedBytesTotal` — the byte counter only accumulates known positive values.
- Threshold-crossing escalation is one-shot per session (`_stdinDroppedThresholdLogged`) so a flood after the crossing doesn't re-spam errors.
- The Nth-drop escalation runs in addition to the byte threshold so unknown-byte floods still raise loud signals.

## Tests

Six new unit tests in `packages/server/tests/sdk-session.test.js`:

- `accumulates dropped bytes across events` — counter sums correctly.
- `treats unknown/missing byte counts as zero` — count advances, byte total doesn't.
- `logs an error on the first stdin_dropped` — replaces the legacy "logs a warning" assertion, with cumulative-string assertion.
- `logs subsequent drops at warn level` — only the first escalates; rest are warn.
- `escalates to error every N drops` — verifies the count cadence path.
- `escalates to error when cumulative bytes cross the threshold` — verifies the byte threshold path.
- `does not re-escalate after the threshold once crossed` — one-shot semantics.

Existing tests adjusted to accept either log level on the first drop (since the level intentionally changed) and to read `STDIN_DROPPED_*` constants symbolically rather than hard-coding magic numbers.

## Test plan

- [x] `node --test packages/server/tests/sdk-session.test.js` — all 87 SDK session tests pass.
- [x] `node --test packages/server/tests/sdk-session.test.js sdk-session.test.js k8s.test.js session-manager.test.js ws-server.test.js` — 453 / 453 pass.
- [x] Full server test suite — only pre-existing unrelated failures (cloudflared integration tests requiring cloudflared, web-task-manager CLI capability detection, encryption integration flake — none touched by this change).

### DIFF
--- a/packages/server/src/sdk-session.js
+++ b/packages/server/src/sdk-session.js
@@ -651,23 +651,30 @@ export class SdkSession extends BaseSession {
   }
 
   /**
-   * Attach default warn-log listeners for SidecarProcess stdin failure
-   * signals (#3402, #3474).
+   * Attach default log listeners for SidecarProcess stdin failure
+   * signals (#3402, #3474, #3506).
    *
    * SidecarProcess (used by container/k8s spawnClaudeCodeProcess paths)
    * emits two stdin failure events the SDK itself does not surface:
    *
    *   - `stdin_disabled`  — fired when forwarding becomes unrecoverable
    *     (post-reconnect or live WS close mid-write). One-shot.
+   *     Logged at warn level.
    *   - `stdin_dropped`   — fired for every chunk that exceeds the 1 MiB
    *     pre-dial cap. Payload: `{ bytes, reason: 'pre-dial-cap' }`.
+   *     #3506: every drop logs with a cumulative running total
+   *     (`_stdinDroppedBytesTotal` + `_stdinDroppedCount`); the level
+   *     escalates to error on the first drop, every Nth drop
+   *     (`STDIN_DROPPED_ESCALATION_EVERY_N`), and when the cumulative
+   *     byte total crosses `STDIN_DROPPED_BYTES_ERROR_THRESHOLD`. All
+   *     other drops log at warn.
    *
    * Both signal silent data loss from the consumer's perspective: the
    * underlying PassThrough still accepts writes, so without an explicit
    * listener the user sees a hung turn instead of an error.  This helper
-   * provides the "warn log at minimum" guarantee — subclasses or future
-   * K8s-aware paths may override to escalate (e.g. emit error, abort the
-   * turn).
+   * provides the "log at minimum" guarantee — subclasses or future
+   * K8s-aware paths may override to escalate further (e.g. emit a
+   * session error event, abort the turn).
    *
    * Idempotent on two axes:
    *   1. Safe on non-SidecarProcess procs — Node ChildProcess (Docker path)
@@ -675,7 +682,7 @@ export class SdkSession extends BaseSession {
    *   2. Safe on repeat calls — the proc is stamped with a Symbol marker
    *      after the first wiring; subsequent calls short-circuit so a
    *      resume/reconnect caller cannot accumulate duplicate listeners
-   *      (and therefore duplicate warn logs) on the same proc.
+   *      (and therefore duplicate logs) on the same proc.
    *
    * @param {EventEmitter|null|undefined} proc — the spawned process from
    *   `spawnClaudeCodeProcess`.  May be a Node ChildProcess (Docker path)
--- a/packages/server/src/sdk-session.js
+++ b/packages/server/src/sdk-session.js
@@ -50,6 +50,19 @@ const DEFAULT_MAX_TOOL_INPUT_LENGTH = 262144
 // Symbol-keyed so it cannot collide with consumer code or test stubs.
 const SIDECAR_LISTENERS_ATTACHED = Symbol('sdk-session.sidecarListenersAttached')
 
+// Cumulative byte threshold for escalating stdin_dropped to error (#3506).
+// Defaults to 10 MiB — equivalent to ten full pre-dial-cap chunks.  Once the
+// running total of dropped bytes meets or exceeds this, a single error log
+// is emitted (one-shot per crossing) so operators triaging "why did my
+// prompt vanish?" get a loud signal even on a flood of small drops.
+export const STDIN_DROPPED_BYTES_ERROR_THRESHOLD = 10 * 1024 * 1024
+
+// Drop-count cadence for re-escalating stdin_dropped to error (#3506).
+// In addition to the byte threshold, every Nth drop event is logged at error
+// level so a stream of zero-byte / unknown-size drops still raises a loud
+// signal.  Set to 10 to balance signal-to-noise.
+export const STDIN_DROPPED_ESCALATION_EVERY_N = 10
+
 export class SdkSession extends BaseSession {
   /**
    * Human-readable label shown in the startup banner and anywhere else the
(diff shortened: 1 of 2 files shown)


blamechris:
## Code Review: PR #3537

### Summary

Implements #3506: every `stdin_dropped` event from `SidecarProcess` now accumulates a cumulative byte counter (`_stdinDroppedBytesTotal`) and a drop count (`_stdinDroppedCount`) on the `SdkSession`. The default log listener emits a `cumulative=<N> bytes over <M> drops` suffix on every line, and escalates from `warn` to `error` on the first drop, every 10th drop, and the first time cumulative bytes cross 10 MiB. All other drops continue to log at `warn` to preserve the per-event audit trail.

### Strengths

- **Both severity + counter — the right call.** The issue offered three options; option 3 was the least lossy choice. First-drop-as-error is the correct one-shot loud signal; the cumulative suffix on every line is the running diagnostic.
- **Constants are exported.** `STDIN_DROPPED_BYTES_ERROR_THRESHOLD` and `STDIN_DROPPED_ESCALATION_EVERY_N` are exported so tests reference them by name (no magic numbers in assertions) and operators tuning them have a single source of truth.
- **One-shot semantics on the threshold crossing.** `_stdinDroppedThresholdLogged` prevents a flood of small drops after the 10 MiB mark from spamming additional error lines — important on a tight loop.
- **Defensive byte parsing.** Unknown / missing `bytes` payloads still increment `_stdinDroppedCount` but do *not* poison `_stdinDroppedBytesTotal`. Negative or non-finite values are also rejected (`knownBytes && rawBytes > 0`).
- **Count-based cadence catches unknown-bytes floods.** Even when every drop has unknown bytes (so the byte threshold never crosses), every Nth drop re-escalates. Belt + suspenders.
- **Test coverage is thorough.** Six new unit tests cover counter accumulation, unknown-byte handling, first-drop escalation, every-Nth escalation, byte-threshold escalation, and one-shot semantics. Existing tests adjusted to read constants symbolically.
- **Doc commit (`60f1f46ae`) refreshes the JSDoc** to describe the new escalation policy — the helper no longer provides "warn log at minimum"; the documentation now matches the behavior.

### Issues Found

#### Critical (Must Fix)

None.

| `packages/server/src/sdk-session.js` | 685 | `_stdinDroppedBytesTotal` is incremented on `_stdinDroppedCount` advance regardless of whether the SidecarProcess listener wiring is present (it always is for SdkSession). Worth a unit test assertion that the counters survive a `_clearMessageState()` / mid-session interrupt — they're persistent across turns by design but the contract isn't asserted anywhere. | Tracked as deferred (see below). |
| `packages/server/src/sdk-session.js` | 727 | The cumulative byte total in the log is raw bytes. For a 10 MiB+ event the log line will read `cumulative=10485760 bytes` — readable but harder to scan than `cumulative=10.0 MiB`. A small `_formatBytes()` helper (KiB/MiB/GiB) would make threshold-cross logs easier to parse at a glance. | Cosmetic; the raw count is more grep-friendly for scripts, so this is a tradeoff. Tracked as deferred. |
| `packages/server/src/sdk-session.js` | n/a | The dropped-bytes counter is per-session in memory only — it doesn't reset on a `claude --resume` reconnect, which is the right behavior, but is also not surfaced anywhere in the dashboard / WS protocol. An operator who isn't tailing logs has no visibility. | Deferred — out of scope for this PR. Tracked as follow-up. |

#### Nitpicks (Optional)

- `dropCount > 1 && dropCount % STDIN_DROPPED_ESCALATION_EVERY_N === 0` (line 715) — the `dropCount > 1` guard is technically redundant with `isFirstDrop` already being captured separately, but it's cheaper than relying on short-circuit evaluation order in `escalate`. No change requested; the explicit guard is clearer.
- The new `cumulative=...` suffix on every drop log makes lines longer (~50 chars). Acceptable for `error` lines but it does add noise to `warn`. Could be omitted from sub-threshold warns and only added on escalations — but consistency is worth more than terse lines for grepping. Keep as-is.

| Persist drop counters across `_clearMessageState()` and assert with a unit test | https://github.com/blamechris/chroxy/issues/3542 | They already do (the counters are constructor-init), but no test pins the contract. Out of scope for this PR which delivers the per-event escalation. |
| Format cumulative bytes as KiB/MiB in log lines | https://github.com/blamechris/chroxy/issues/3543 | Cosmetic. Raw counts are better for scripted log parsing; if we add a `formatBytes()` helper it should be done in a small isolated PR. |
| Surface cumulative drop totals on the dashboard / WS protocol | https://github.com/blamechris/chroxy/issues/3544 | Operator visibility for ops who aren't tailing logs. Needs a `session_event` envelope addition + dashboard surface — large change, separate PR. |

### Architecture Notes

- This change is local to `SdkSession`; `CliSession`, `CodexSession`, and `GeminiSession` are unaffected because only the SDK path runs through `SidecarProcess` (the only emitter of `stdin_dropped`). No cross-cutting protocol or wire-format changes.
- The counters are private (`_stdinDroppedBytesTotal`, `_stdinDroppedCount`, `_stdinDroppedThresholdLogged`) — accessed only by the listener. Exposing them via a getter for diagnostics could come later (see deferred #3544), but isn't required by #3506.
- The escalation thresholds are exported constants. If we ever decide to make them per-deployment configurable (env var or config file), the export points are already in place — drop-in replacement.

### Verdict

- [x] **Approve** — Ready to merge.

Tests pass (87/87 in `sdk-session.test.js`, 453/453 across the affected suite group). The four pre-existing test failures in the full suite (`tunnel.integration`, `web-task-manager`, `encryption integration end-to-end`, plus session-manager flake) are environmental / unrelated and present on `main`. PR description documents the threshold rationale and the "both severity + counter" decision per the issue's acceptance criteria.

PR #3554, opened 2026-05-07
docs(server): align PROTOCOL.md line_too_long close code with implementation

blamechris:
## Summary

- Fix inconsistency in `packages/server/sidecar/PROTOCOL.md`: the **NDJSON Line Length Limit** section step 3 said the WS closes with code `1000`, but `agent.js:946` actually closes with `1008` (matching the Error Cases table earlier in the doc).
- Follow-up to PR #3520 which addressed the table but not the prose section.

## Test plan

- [x] Verified `agent.js:946` uses `session.activeWs.close(1008, 'line_too_long')`
- [x] Verified Error Cases table (line 598) already documents `close 1008`
- [x] Updated prose at line 620 to match

Closes #3548

### DIFF
--- a/packages/server/sidecar/PROTOCOL.md
+++ b/packages/server/sidecar/PROTOCOL.md
@@ -617,7 +617,7 @@ if a line exceeds the cap before a newline arrives it:
 
 1. Emits `{ type: 'error', code: 'line_too_long', message: '...', seq: N }`.
 2. Sends `SIGTERM` to the child (with a `SIGKILL` escalation after 5 s).
-3. Closes the WS with code `1000` within 50 ms.
+3. Closes the WS with code `1008` within 50 ms.
 
 **Default cap:** 1 MiB (`1 048 576` bytes) — far above any normal SDK event.
 


blamechris:
## Code Review: PR #3554

### Summary

Single-line documentation fix in `packages/server/sidecar/PROTOCOL.md`. The **NDJSON Line Length Limit** section (step 3) was the last remaining holdout from PR #3520, which had aligned the Error Cases table but not this prose paragraph. After this change, all four references to the `line_too_long` close code (table row 598, prose line 399, prose line 620, and code at `agent.js:946`) consistently say `1008`.

### Strengths

- Surgical: exactly one character changed in one line, no collateral edits.
- Correct semantics: `1008` (Policy Violation) is the appropriate WS close code for an oversized-line guard kill — `1000` (Normal Closure) would have wrongly implied a clean shutdown.
- Verified against implementation (`agent.js:946`) and against the existing Error Cases table — the doc is now self-consistent.

### Issues Found

#### Critical (Must Fix)

None.

None.

#### Nitpicks (Optional)

None.

None.

### Architecture Notes

Pure documentation change. No protocol behavior, code paths, tests, or runtime artifacts are affected. The fix removes a contradiction between the same document's table and prose, which matters for sidecar consumers reading PROTOCOL.md to handle close codes correctly (1008 implies a server-side policy violation that the client should surface, not a normal shutdown).

### Verdict

PR #3555, opened 2026-05-07
test(server): extract waitFor helper for sidecar-agent polling

blamechris:
## Summary

- Replace three predicate-polling loops in `packages/server/tests/sidecar/agent.test.js` with the existing `waitFor()` helper from `tests/test-helpers.js`
- Each migrated loop polled a session-state predicate every 10ms up to 50 iterations; the helper expresses the same intent more clearly and gives a useful timeout error message instead of silently falling through to the assert

## Sites migrated

Under the `drain timer cancelled on every kill path (#3514)` describe block:
- idle TTL eviction — wait for `session.activeWs === null` after WS close
- session-cap eviction — wait for `firstSession.activeWs === null`
- session-cap eviction — wait for `agent._sessions.has(firstSessionId)` to flip false

## Test plan

- [x] All 101 sidecar tests pass under Node 22 (`node --test packages/server/tests/sidecar/agent.test.js`)
- [x] The two specifically migrated tests (`idle TTL eviction cancels the drain timer before _killChild`, `session-cap eviction cancels the drain timer before _killChild`) both pass

Closes #3547

### DIFF
--- a/packages/server/tests/sidecar/agent.test.js
+++ b/packages/server/tests/sidecar/agent.test.js
@@ -4,6 +4,7 @@ import { EventEmitter, once } from 'node:events'
 import { PassThrough } from 'node:stream'
 import WebSocket from 'ws'
 import { PodAgent, LineLimitTransform, DEFAULT_MAX_LINE_BYTES, DEFAULT_STDIN_DRAIN_TIMEOUT_MS } from '../../sidecar/agent.js'
+import { waitFor } from '../test-helpers.js'
 
 const TOKEN = 'test-secret-token-abc123'
 const WRONG_TOKEN = 'wrong-token'
@@ -2396,9 +2397,10 @@ describe('PodAgent', () => {
         const clientClosed = once(ws, 'close')
         ws.close()
         await clientClosed
-        for (let i = 0; i < 50 && session.activeWs !== null; i++) {
-          await new Promise((r) => setTimeout(r, 10))
-        }
+        await waitFor(() => session.activeWs === null, {
+          timeoutMs: 500,
+          label: 'session.activeWs cleared after disconnect',
+        })
         assert.equal(session.activeWs, null, 'activeWs must be cleared after disconnect')
 
         // Fire the idle TTL timer.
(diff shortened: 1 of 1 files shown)


blamechris:
## Code Review: PR #3555

### Summary

Replaces three predicate-polling loops in `packages/server/tests/sidecar/agent.test.js` with the existing `waitFor()` helper from `tests/test-helpers.js`. Pure test-only refactor, +13 / -9, scoped to one file. Implements #3547.

### Strengths

- **Reuses the existing helper** instead of redefining one. `waitFor()` already lives in `tests/test-helpers.js` (lines 21-31) with `{ timeoutMs, intervalMs, label }` semantics, JSDoc, and a thrown timeout error. The author correctly chose the established API over the issue's literal suggestion (`{ timeout, interval }`).
- **Preserves the effective timeout budget.** Each loop was bounded at `50 * 10ms = 500ms`; each `waitFor` call passes `timeoutMs: 500` — no behavioural change to test runtime.
- **Useful `label` per call site.** Each `waitFor` carries a descriptive label (`session.activeWs cleared after disconnect`, `first session evicted by session-cap`, etc.) so a future timeout produces an actionable error rather than a silent fall-through into the assert message.
- **Migrates exactly the three sites called out in the issue.** I grep'd the rest of the server tests for `for (let i = 0;.*&&.*;.*i\+\+)` predicate-polling shape — there are no other instances in `packages/server/tests/`. The audit criterion in the issue is satisfied.
- **Style compliant.** ES modules, no semicolons, single quotes, matches the surrounding code.
- **Tests pass.** Confirmed all 101 sidecar tests green under Node 22, including the two specifically migrated tests (`idle TTL eviction cancels the drain timer before _killChild`, `session-cap eviction cancels the drain timer before _killChild`).

### Issues Found

#### Critical (Must Fix)
None.

None. The change is small, surgical, and exactly matches the issue's acceptance criteria.

#### Nitpicks (Optional)
- Lines 2404, 2478, and 2502-2504 retain `assert.equal(...)` / `assert.ok(...)` checks immediately after `waitFor` has already proven the predicate holds. With the old `for`-loop these asserts were the actual safety net (the loop didn't throw). With `waitFor` they're now technically redundant in the happy path, since a timeout would already throw with the descriptive `label`. They're harmless and arguably serve as documentation/defense-in-depth, so this is purely a style call — fine either way. Not blocking.

### Architecture Notes

This change has no production impact — `tests/sidecar/agent.test.js` is a test file. It does, however, gently nudge the codebase toward using `waitFor()` consistently for predicate polling, which is a positive direction: the helper already exists, is documented, and produces better failure messages than ad-hoc loops. Future sidecar work that needs to wait on async server-side state should follow this pattern.

### Verdict

Clean, minimal, exactly scoped to the issue. No follow-ups needed.

blamechris:
## Review Comments Addressed

| — | — | — | — |

**Total:** 0 inline comments to address.

Copilot review (state: `COMMENTED`) returned a descriptive overview only — no inline findings, no requested changes, no concerns. Agent-review pass returned `Approve` with 1 non-blocking nitpick (redundant `assert.equal()` after `waitFor`, kept as defense-in-depth). Nothing to fix, dismiss, or defer.

PR is ready for merge.

Issue #3493, opened 2026-05-07
k8s: audit `namespace || this._namespace` sites for the same nullish-coalescing rationale as #3459


blamechris:
## Context

Identified during review of PR #3489 (which closes #3459).

PR #3489 switched the K8sBackend constructor defaults from `||` to `??`. The same falsy-stomping pattern appears in five `createEnvironment`/`destroyEnvironment`/`streamCliInEnvironment`/etc. call sites that resolve a per-call `namespace` against the constructor-stored `this._namespace`.

## Description

The K8sBackend repeats this resolution at five sites:

```js
// packages/server/src/environments/backends/k8s.js
line 274:  const ns = namespace || this._namespace             // createEnvironment
line 521:  const ns = opts.namespace || this._namespace        // ...
line 582:  const ns = opts.namespace || this._namespace        // ...
line 718:  const ns = namespace || this._namespace             // ...
line 832:  const ns = opts.namespace || this._namespace        // ...
```

For #3459 the line 338 `callImagePullPolicy || this._imagePullPolicy` was left as `||` because `validateImagePullPolicy()` runs first and rejects empty strings — by the time control reaches line 338 the value is either nullish or a valid enum, all of which behave identically under `||` vs `??`.

The five `namespace` sites have **no equivalent validation gate**. A caller passing `namespace: ''` today would silently fall through to `this._namespace` rather than surface as an obvious error.

Two ways to resolve:

1. **Switch to `??`** at all five sites for the same reason as the constructor changes in #3459.
2. **Extract a `_resolveNamespace(opts)` helper** that centralises the resolution and applies validation, eliminating the repetition entirely (DRY win, single point of policy).

Option 2 is cleaner if a `validateNamespace()` helper is introduced alongside (RFC 1123 label format, length cap of 63 chars, lowercase, etc.). Option 1 is a one-line-each minimal change. Either is acceptable; pick based on appetite.

## Original Review Comment

> \`const ns = namespace || this._namespace\` follows the same pattern as line 338 — per-call override falls back to constructor default. Worth a brief audit pass to confirm this is consistent with the line-338 reasoning (i.e., the per-call \`namespace\` is either nullish or a real namespace string; empty string here would currently be stomped by \`this._namespace\`). [The same audit applies to lines 521, 582, 718, 832] which all use \`opts.namespace || this._namespace\` — five identical falsy-stomping sites that would benefit from a shared resolver helper or a consistent \`??\` switch.

## Acceptance Criteria

[...]

PR #3556, opened 2026-05-07
test(server): speed up destroyEnvironment slow-path test

blamechris:
## Summary
- The `destroyEnvironment preserves explicit empty-string namespace` test in `packages/server/tests/environments/backends/k8s.test.js` (~line 640) took the slow path: `deleteNamespacedPod` succeeded, then `destroyEnvironment` slept a real `POLL_INTERVAL_MS` (1s) before the first `readNamespacedPod` poll. That added a deterministic ~1s to every suite run.
- Switch the mock so `deleteNamespacedPod` itself throws 404. The idempotent fast-path exits immediately without entering the poll loop. The namespace assertion remains valid because `createMockApi` records `calls.delete` before invoking the user override.

## Approach
Issue #3551 mentioned two options:
1. Inject a custom `pollIntervalMs` constructor option.
2. Use `t.mock.timers` to advance time deterministically.

A third option — the one suggested in the issue body — is even simpler: **change which mock throws 404 so the function never enters the poll loop**. This is a single-line test edit, requires no production-code change (no new constructor option, no test-only seam), and preserves the regression coverage (#3493) that the test was added for.

The targeted assertion (`api.calls.delete[0].namespace === ''`) still validates the nullish-coalescing fix because the namespace is captured by `createMockApi`'s wrapper before the override throws.

## Measurements (local Node 22)
- Before: `real 12.02s` (171 tests)
- After:  `real 9.78s` (171 tests)
- Targeted test alone: `~1000ms` → `~3ms`

## Test plan
- [x] `node --test packages/server/tests/environments/backends/k8s.test.js` — all 171 tests pass
- [x] Targeted test still asserts the empty-string namespace contract from #3493
- [x] No production-code change

Closes #3551

### DIFF
--- a/packages/server/tests/environments/backends/k8s.test.js
+++ b/packages/server/tests/environments/backends/k8s.test.js
@@ -638,8 +638,12 @@ describe('K8sBackend per-call namespace resolution uses nullish coalescing (#349
   })
 
   it('destroyEnvironment preserves explicit empty-string namespace (not replaced by default)', async () => {
+    // Make `deleteNamespacedPod` itself 404 so destroyEnvironment exits via the
+    // idempotent fast-path without entering the 1s POLL_INTERVAL_MS poll loop
+    // (#3551). The mock still records the delete call before invoking the
+    // override, so the namespace assertion below remains valid.
     const api = createMockApi({
-      readPod: async () => { throw make404Error() },
+      deletePod: async () => { throw make404Error() },
     })
     const backend = new K8sBackend({ namespace: 'default', _coreV1Api: api })
 


blamechris:
## Code Review: PR #3556

### Summary
Test-only change. Replaces the `readPod: throw 404` mock override with `deletePod: throw 404` in the `destroyEnvironment preserves explicit empty-string namespace` test so the function exits via the idempotent fast-path instead of sleeping `POLL_INTERVAL_MS` (1s real time) before the first poll. Saves ~1-2s per suite run.

### Strengths
- **Smallest viable fix.** Issue #3551 listed two approaches (constructor `pollIntervalMs` option, `t.mock.timers`). The author chose the third option mentioned in the issue body — flip which mock throws 404 — which is a one-line override swap with zero production-code surface area. Right call: no new test-only seams to maintain, no API expansion.
- **Coverage preserved.** The assertion target (`api.calls.delete[0].namespace === ''`) is unchanged. `createMockApi` records the delete-call arguments before invoking the user override, so the namespace flow-through is still validated even though the override now throws. The #3493 regression guard remains intact.
- **Clear inline comment.** The added comment explains *why* `deletePod` throws (avoid 1s poll), references #3551, and notes the recording-vs-throw ordering invariant of the mock — future maintainers will not need to spelunk the helper to understand the test setup.
- **No production change.** Avoids the temptation to add a `pollIntervalMs` ctor option or a `_sleep` seam just for this test. Keeps `K8sBackend` constructor surface minimal.
- **Style compliance.** No semicolons, single quotes, ES modules — matches `packages/server/` conventions.

### Issues Found

#### Critical (Must Fix)
None.

None. The change is minimal, targeted, and correct.

#### Nitpicks (Optional)
None.

### Architecture Notes
This is a pure test-suite-speed improvement. No protocol or runtime behaviour changes. The `destroyEnvironment` poll loop and `POLL_INTERVAL_MS` semantics in `packages/server/src/environments/backends/k8s.js` remain unchanged — there is still a separate slow-path test (`'calls deleteNamespacedPod then polls until 404'` at line 1244) that exercises the actual polling behaviour, so reducing this test to fast-path-only does not lose coverage.

### Verdict

Single-line override swap that preserves the test's assertion target while skipping the 1s real-time sleep. No follow-ups needed.

PR #3558, opened 2026-05-07
test(server): strengthen reconnect() invariant guard test

### LINKED ISSUE #3545: environment-manager: strengthen invariant guard test to pair allHealthy=false with nearest env.status assignment
## Context

Identified during review of PR #3541 (which closes #3492).

## Description

The current invariant guard test in \`tests/environment-manager.test.js\` (\`EnvironmentManager.reconnect() invariant guard (#3492)\`) catches drift at the *count* level: it asserts that the number of \`allHealthy = false\` flips equals \`statusMatches.length - 1\` (subtracting the success-path \`'running'\` assignment). It also asserts every literal status in the function body is either \`'running'\` or in \`UNREACHABLE_STATUSES\`.

This catches the most likely contributor bug — adding a new \`allHealthy = false\` branch without any \`env.status\` assignment.

It does **not** catch a narrower contributor bug: adding a new branch that *both* flips \`allHealthy = false\` *and* assigns \`env.status = 'running'\` (or any reachable status). Such code is internally inconsistent — the env is reported healthy yet the boolean says otherwise — but the count assertion passes.

## Original Review Comment

> Strengthen invariant guard: also assert no \`allHealthy = false\` co-located with \`env.status = 'running'\`. The current test would pass if a contributor added a new branch that flips \`allHealthy = false\` AND assigns \`'running'\` (an internally-inconsistent contributor bug — unhealthy but reported running). A line-by-line scan that pairs each \`allHealthy = false\` with the *nearest preceding* \`env.status\` assignment and asserts the literal is in \`UNREACHABLE_STATUSES\` would catch this.

## Acceptance Criteria

- [ ] Replace the count-level assertion with a per-branch pairing: walk the body line-by-line, track the most-recent \`env.status = '<literal>'\` assignment, and on each \`allHealthy = false\` line assert the tracked literal is in \`UNREACHABLE_STATUSES\`
- [ ] Existing assertions (count >= 1, every literal is 'running' or in UNREACHABLE_STATUSES) remain
- [ ] Failure messages clearly indicate which branch broke the pairing
- [ ] Test deliberately mutated locally to verify it catches: (a) a new branch with no status assignment, (b) a new branch with \`status = 'running'\`, (c) a new unreachable status not in \`UNREACHABLE_STATUSES\`

## Deferral Rationale

Current guard catches the documented invariant from #3492 (count-level drift). Pairing logic adds significant test complexity (parsing nesting, tracking assignment ordering across branches) for a narrower bug class. Worth tracking but not worth blocking PR #3541.

blamechris:
Closes #3545.

## Summary

Original guard (#3492) inspected counts only — it asserted that the number of `allHealthy = false` flips equals (status assignments - 1) and that every literal status was `'running'` or in `UNREACHABLE_STATUSES`. The count check misses a narrower contributor bug: a new branch that flips `allHealthy = false` AND assigns `env.status = 'running'` (or any other reachable status) — internally inconsistent, yet the count balances.

This PR strengthens the guard by walking the function body line-by-line:

- Track the most-recent `env.status = '<literal>'` assignment.
- On each `allHealthy = false` line, assert the tracked literal exists and is in `UNREACHABLE_STATUSES`.
- Consume the tracker after each pairing so a sibling branch cannot reuse a stale assignment.
- Failure messages quote the exact line numbers of both the flip and the offending status assignment.

Existing top-level assertions (count >= 1 flip, every literal is `'running'` or in `UNREACHABLE_STATUSES`) are preserved as a sanity gate.

## Verification

Test deliberately mutated against `environment-manager.js` to confirm coverage of every acceptance-criteria case:

| Mutation | Caught? | Failure message |
|---|---|---|
| (a) new branch flipping `allHealthy = false` with no status assignment | yes | `expected N+1 env.status assignments ... found N. A new allHealthy=false branch was added without a co-located env.status assignment` |
| (b) new branch flipping `allHealthy = false` with `env.status = 'running'` | yes | `[#3545] reconnect() line 519: 'allHealthy = false' is paired with 'env.status = 'running'' (line 518), which is NOT in UNREACHABLE_STATUSES` |
| (c) new unreachable status not in `UNREACHABLE_STATUSES` (e.g. `'quota_exceeded'`) | yes | `env.status = 'quota_exceeded' is not 'running' and not in UNREACHABLE_STATUSES` |

## Test plan

- [x] `node --test packages/server/tests/environment-manager.test.js` — 89 tests pass
- [x] Manually mutated source to verify each acceptance-criteria failure mode triggers a clear, line-numbered assertion error
- [x] Test-only change — no source edits to `environment-manager.js`

### DIFF
--- a/packages/server/tests/environment-manager.test.js
+++ b/packages/server/tests/environment-manager.test.js
@@ -2513,7 +2513,8 @@ describe('EnvironmentManager.reconnect() — reconnectAgentToken delegation (#33
 })
 
 /**
- * Issue #3492 — invariant guard for the reconnect() unreachable-count helper.
+ * Issue #3492 / #3545 — invariant guard for the reconnect() unreachable-count
+ * helper.
  *
  * The boot-path aggregate-warn helper in
  * `server-cli.js#logEnvironmentManagerReconnectResult` derives the unreachable
@@ -2524,20 +2525,26 @@ describe('EnvironmentManager.reconnect() — reconnectAgentToken delegation (#33
  * `allHealthy = false` branch (quotas, partial-restore, metrics, …) without
  * a co-located status assignment would silently undercount.
  *
- * This test inspects the source of `reconnect()` and asserts that the count
- * of `allHealthy = false` lines matches the count of preceding/adjacent
- * `env.status = '<unreachable status>'` assignments. It also asserts the
- * exported `UNREACHABLE_STATUSES` set contains every literal status string
- * actually written by `reconnect()` on its `allHealthy=false` paths.
+ * Original guard (#3492) inspected counts only — it asserted that the number
+ * of `allHealthy = false` flips equals (status assignments − 1) and that every
+ * literal status was 'running' or in UNREACHABLE_STATUSES. The count check
+ * misses a narrower contributor bug (#3545): a new branch that flips
+ * `allHealthy = false` AND assigns `env.status = 'running'` (or any other
+ * reachable status) — internally inconsistent, yet the count balances.
+ *
+ * Strengthened guard (#3545) walks the body line-by-line and pairs each
+ * `allHealthy = false` with the most-recent preceding `env.status` literal,
+ * asserting the literal is in UNREACHABLE_STATUSES. Failure messages quote the
+ * exact pairing so the broken branch is obvious.
  */
-describe('EnvironmentManager.reconnect() invariant guard (#3492)', () => {
+describe('EnvironmentManager.reconnect() invariant guard (#3492, #3545)', () => {
   it('UNREACHABLE_STATUSES contains expected statuses', () => {
     assert.ok(UNREACHABLE_STATUSES.has('error'), '\'error\' must be unreachable')
     assert.ok(UNREACHABLE_STATUSES.has('stopped'), '\'stopped\' must be unreachable')
     assert.ok(!UNREACHABLE_STATUSES.has('running'), '\'running\' must NOT be unreachable')
   })
 
-  it('every allHealthy=false branch in reconnect() co-locates env.status assignment to an UNREACHABLE_STATUSES value', () => {
+  it('every allHealthy=false branch in reconnect() pairs with the nearest preceding env.status assignment to an UNREACHABLE_STATUSES value', () => {
     const srcPath = resolve(__dirname, '../src/environment-manager.js')
     const src = readFileSync(srcPath, 'utf-8')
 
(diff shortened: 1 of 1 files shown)


blamechris:
## Code Review: PR #3558

### Summary

Test-only change that strengthens the `reconnect()` invariant guard added in #3492. Replaces a count-level check with a per-branch line-pairing walk that pairs each `allHealthy = false` flip with the most-recent preceding `env.status = '<literal>'` assignment and asserts the literal is in `UNREACHABLE_STATUSES`. Closes #3545.

### Strengths

- **Surgical scope** — single test file, no source modifications, no new dependencies
- **Mutation verification documented** — commit message and PR body enumerate all three acceptance-criteria cases (a, b, c) and confirm each fails the strengthened test with a quoted error message
- **Diagnostic line numbers** — `bodyOffsetLine = src.slice(0, startIdx).split('\n').length` lets failure messages quote the exact line of both the flip and the offending status assignment, which is a real ergonomic upgrade for the contributor who trips it
- **Belt-and-braces preservation** — the original count-level checks (`flipMatches.length > 0`, `statusMatches.length === flipMatches.length + 1`, every literal in `'running' ∪ UNREACHABLE_STATUSES`) are kept as a top-level sanity gate, so a regression in either pass surfaces independently
- **Tracker consume semantics correct** — `lastStatus = null` after each pairing prevents a flip in a sibling branch from silently reusing a stale assignment from above. Verified against the live source: 4 flips, each tightly co-located with its `env.status` on the previous line, all pair correctly
- **Code style conforms** — ES modules, no semicolons, single quotes, no new imports

### Issues Found

#### Critical (Must Fix)

None.

|---|---|---|---|
| `tests/environment-manager.test.js` | 2576 | The line-comment stripper `line.replace(/\/\/.*$/, '')` would mangle a string containing `//` (e.g., `'http://...'`) inside `reconnect()`. The current source has no such string, but a future contributor could introduce one and silently lose half a line from the regex pass. A state-tracking parser is overkill, but a guard like asserting the body has no string-literal `//` would make the failure mode loud. | Test-self-correctness against future drift; non-blocking |

#### Nitpicks (Optional)

- The final `pairingsChecked === flipMatches.length` assertion is essentially unreachable — if the line walk diverges from the top-level regex count, an earlier assertion has already fired or the source extraction is wildly broken. The comment ("belt-and-braces") is honest about its purpose; harmless to keep.
- Section divider comments use U+2500 box-drawing characters. The file already mixes Unicode (`−` minus sign in the existing docstring) so this is consistent with surrounding style.

|---|---|---|

None — no follow-up issues warranted. The single suggestion above is a marginal hardening against a hypothetical future regression in source content (string literals containing `//`), not a current bug. If such a string is ever introduced, the test would silently start over-stripping; but the count-level pass would still catch the most-likely contributor mistake (missing status assignment), so the worst case is degraded diagnostic precision rather than a false negative on the actual invariant. Not worth a tracking issue.

### Architecture Notes

This is a test-side guard for an architectural invariant documented in the JSDoc on `reconnect()` (lines 411-426 of `environment-manager.js`): every code path that flips `allHealthy = false` must co-locate an `env.status` assignment to a value in `UNREACHABLE_STATUSES`, because the boot-path aggregate-warn helper in `server-cli.js#logEnvironmentManagerReconnectResult` derives the unreachable count via `list().filter(e => UNREACHABLE_STATUSES.has(e.status)).length`. The guard is the contributor-facing enforcement of that invariant; strengthening it from count-level to per-branch pairing closes the narrowest internally-consistent contributor bug class (flip + reachable status). Aligns cleanly with the existing pattern of self-inspecting source-AST tests in this codebase.

### Verdict

- [ ] Request Changes
- [ ] Comment

PR #3563, opened 2026-05-07
test(server): pin stdin_dropped counters as session-lifetime (#3542)

blamechris:
## Summary

Pins the session-lifetime contract for the cumulative stdin_dropped counters introduced in #3506.

The flags `_stdinDroppedBytesTotal`, `_stdinDroppedCount`, and `_stdinDroppedThresholdLogged` are constructor-initialised on `SdkSession` and are intentionally NOT touched by `_clearMessageState()` — they persist across turns / `sendMessage()` cycles so the loud-signal escalation (first drop = error, every Nth drop = error, threshold-cross = error) stays one-shot per session.

The current behavior is correct, but no test pinned the contract. A future refactor of `_clearMessageState()` could silently reset the counters and the loud-signal behavior would silently regress: every turn would re-fire a "first drop" error, and the cumulative-bytes threshold one-shot would re-arm on every turn.

## Changes

Adds four tests in `sdk-session.test.js` under the `_attachSidecarProcessListeners` describe:

- `preserves stdin_dropped counters across _clearMessageState()` — emits two drops, calls `_clearMessageState()`, asserts both `_stdinDroppedBytesTotal` and `_stdinDroppedCount` are unchanged.
- `preserves _stdinDroppedThresholdLogged across _clearMessageState()` — crosses the byte threshold, then asserts the latched flag survives a clear.
- `preserves stdin_dropped counters across interrupt()` — exercises the public `interrupt()` contract; counters and threshold flag are unchanged.
- `does not re-escalate a fresh drop after _clearMessageState() to error` — fires the one-shot first-drop error in turn 1, calls `_clearMessageState()`, then asserts a turn-2 drop logs at `warn` only (no fresh error).

## Test plan

- [x] `node --test packages/server/tests/sdk-session.test.js` — 92/92 pass (4 new + 88 existing)

Closes #3542

### DIFF
--- a/packages/server/tests/sdk-session.test.js
+++ b/packages/server/tests/sdk-session.test.js
@@ -1391,5 +1391,126 @@ describe('SdkSession', () => {
       assert.equal(proc.listenerCount('stdin_disabled'), 1,
         'only one stdin_disabled listener should be attached')
     })
+
+    // -- #3542 — session-lifetime counter contract --
+    //
+    // The cumulative `_stdinDroppedBytesTotal`, `_stdinDroppedCount`, and
+    // `_stdinDroppedThresholdLogged` flags are session-lifetime by design
+    // (see #3506).  They MUST survive a `_clearMessageState()` / mid-session
+    // interrupt — otherwise the loud-signal escalation would re-fire on
+    // every turn (each turn would treat its first drop as "first drop ever")
+    // and the threshold-cross one-shot would silently regress.  These tests
+    // pin the contract so a future refactor of `_clearMessageState()` can't
+    // accidentally reset the counters.
+    it('preserves stdin_dropped counters across _clearMessageState() (#3542)', async () => {
+      const { EventEmitter } = await import('events')
+      const proc = new EventEmitter()
+      session._attachSidecarProcessListeners(proc)
+
+      proc.emit('stdin_dropped', { bytes: 100, reason: 'pre-dial-cap' })
+      proc.emit('stdin_dropped', { bytes: 250, reason: 'pre-dial-cap' })
+
+      const bytesBefore = session._stdinDroppedBytesTotal
+      const countBefore = session._stdinDroppedCount
+      assert.equal(bytesBefore, 350)
+      assert.equal(countBefore, 2)
+
+      session._clearMessageState()
+
+      assert.equal(session._stdinDroppedBytesTotal, bytesBefore,
+        '_stdinDroppedBytesTotal must survive _clearMessageState()')
+      assert.equal(session._stdinDroppedCount, countBefore,
+        '_stdinDroppedCount must survive _clearMessageState()')
+    })
+
+    it('preserves _stdinDroppedThresholdLogged across _clearMessageState() (#3542)', async () => {
+      const { EventEmitter } = await import('events')
+      const { STDIN_DROPPED_BYTES_ERROR_THRESHOLD } = await import('../src/sdk-session.js')
+
+      const proc = new EventEmitter()
+      session._attachSidecarProcessListeners(proc)
+
+      proc.emit('stdin_dropped', { bytes: 1, reason: 'pre-dial-cap' })
+      proc.emit('stdin_dropped', {
+        bytes: STDIN_DROPPED_BYTES_ERROR_THRESHOLD,
+        reason: 'pre-dial-cap',
+      })
+      assert.equal(session._stdinDroppedThresholdLogged, true,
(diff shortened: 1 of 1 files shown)


blamechris:
## Code Review: PR #3563

### Summary

Adds four tests to `sdk-session.test.js` that pin the session-lifetime contract for the cumulative `stdin_dropped` counters introduced in #3506. The PR is test-only (+121/-0, 1 file) and addresses both acceptance criteria from #3542.

### Strengths

- **Targets the exact regression vector.** The narrative comment block (lines 1395-1403) clearly explains what would silently regress if `_clearMessageState()` ever resets the counters — every turn would re-fire a "first drop" error and the threshold-cross one-shot would re-arm. This is the kind of context a future refactor author needs.
- **Uses the real exported constant.** `STDIN_DROPPED_BYTES_ERROR_THRESHOLD` is imported from `sdk-session.js` rather than hardcoding `10 * 1024 * 1024`. If the threshold ever changes, the test follows.
- **Independent tests.** Each test gets a fresh `session` (from the existing `beforeEach`) and a fresh `EventEmitter` proc, so order-of-execution does not affect outcomes.
- **Test #4 captures only post-clear logs.** Correctly attaches the log listener AFTER the turn-1 drop and clear, so the pre-clear error log doesn't pollute the assertion. This is the right way to test "post-clear behavior" without inheriting setup noise.
- **Style matches existing tests.** ES modules, no semicolons, single quotes, dynamic imports of `events` / `logger.js` matching the surrounding pattern.

### Issues Found

#### Critical (Must Fix)

None.

| `sdk-session.test.js` | 1450-1473 | The `interrupt()` test calls `session.interrupt()` on a session with `_query === null`, so the method early-returns at `if (!this._query) return` without touching anything. The test does verify the public contract (interrupt is safe to call mid-session), but it does not exercise the more meaningful scenario: counters surviving an interrupt that DOES tear down an active query. Consider stubbing `session._query` with a minimal object that has an `interrupt()` method (or noting in the comment that the test is a smoke test for the early-return path). | Without an active query, the test cannot regress if a future refactor moves counter-resets into the active-query branch of `interrupt()`. The comment ("no active query so this is a no-op") is honest but slightly understates the limitation. |
| `sdk-session.test.js` | 1395-1518 | The four tests live inside the `_attachSidecarProcessListeners` describe block, but semantically they're testing the `_clearMessageState` / `interrupt` contract. A nested `describe('counter persistence across _clearMessageState (#3542)', ...)` block would make the intent clearer in test output. | Cosmetic; the existing block already houses all stdin_dropped tests, so the grouping is defensible. |

#### Nitpicks (Optional)

- Test 1 (line 1418-1419) asserts hardcoded `350` and `2` and ALSO compares to `bytesBefore`/`countBefore` after the clear. The hardcoded asserts are belt-and-suspenders against the existing `accumulates dropped bytes across events` test changing semantics, which is fine — but redundant.
- Test 2 (line 1430-1448) mixes a precondition assertion (`_stdinDroppedThresholdLogged === true` BEFORE the clear) with the main assertion. Reads cleanly as-is, but a comment label like `// precondition` would scan faster.

| Strengthen `interrupt()` test to exercise the active-query branch (stub `_query` with mock or run a real query) | [#3565](https://github.com/blamechris/chroxy/issues/3565) | Out of scope for #3542 acceptance criteria — the issue explicitly accepts an interrupt-as-public-contract smoke test. The active-query branch is worth pinning separately. |

### Architecture Notes

This PR is a pure test addition with no production-code changes. It correctly characterizes the existing `SdkSession` behavior:

- Counters are constructor-initialized on lines 262-264.
- `_clearMessageState()` (line 1036-1044) only touches `_permissions`, `_permissionPauseCount`, `_resultTimeoutPaused`, `_resetResultTimeout` — leaving the stdin_dropped counters untouched.
- `interrupt()` (line 939-948) early-returns on `!this._query`, otherwise just calls `this._query.interrupt()` — it never touches counters in either branch.

The test file gains 121 lines (1396 → 1517 total), keeping the 22 sub-suites stable. Other 88 tests in the file remain green.

### Test Plan

- [x] `node --test --test-name-pattern="3542" packages/server/tests/sdk-session.test.js` — 4/4 pass
- [x] `node --test packages/server/tests/sdk-session.test.js` — 92/92 pass

### Verdict

**Approve** — Ready to merge after CI green. Test-only change, all four tests are correct and pin the documented contract. The suggestion about the `interrupt()` test is logged as a follow-up (#3565) for out-of-scope strengthening.

PR #3569, opened 2026-05-07
refactor(server): extract _resolveNamespace helper for K8sBackend

blamechris:
## Summary

Extracts a `_resolveNamespace(callNamespace)` instance helper on `K8sBackend` and replaces the five per-call resolution sites (`createEnvironment`, `destroyEnvironment`, `getEnvironmentStatus`, `streamCliInEnvironment`, `reconnectAgentToken`) with calls into it.

Behaviour is unchanged — the helper wraps the same `??` fallback to the constructor-stored namespace, preserving the empty-string contract from #3493 and the `null`/`undefined` → default fallback.

## Scope

Deliberately tight: just the helper extraction. The companion `validateNamespace()` (RFC 1123) idea from #3534 is **not** in this PR — it is a layered change that can land on top of this seam without re-touching the five sites.

## Test plan

- [x] Existing K8sBackend test suite (`tests/environments/backends/k8s.test.js`) continues to pass — 173 tests including the #3493 empty-string preservation cases that exercise all five sites
- [x] Two new direct tests for `_resolveNamespace()`:
  - returns the per-call override (including `''`) verbatim
  - falls back to the constructor namespace for `undefined` / `null` / no-arg
- [x] No production behaviour change — the diff at each call site is mechanical (`X ?? this._namespace` → `this._resolveNamespace(X)`)

Closes #3534

### DIFF
--- a/packages/server/src/environments/backends/k8s.js
+++ b/packages/server/src/environments/backends/k8s.js
@@ -193,6 +193,21 @@ export class K8sBackend {
     this._clearTimeout = clearTimeoutImpl || clearTimeout
   }
 
+  /**
+   * Resolve the effective namespace for a per-call invocation.
+   *
+   * Returns the per-call override when provided (including the empty string —
+   * see #3493), otherwise falls back to the constructor-stored namespace.
+   * Uses `??` so an explicit `''` is preserved verbatim and surfaces to the
+   * K8s API rather than being silently rewritten to the default.
+   *
+   * @param {string|undefined|null} callNamespace - Per-call namespace override
+   * @returns {string} Resolved namespace
+   */
+  _resolveNamespace(callNamespace) {
+    return callNamespace ?? this._namespace
+  }
+
   // ─────────────────────────────────────────────────────────────────────────
   // createEnvironment — create a sidecar Pod + per-Pod Secret
   // ─────────────────────────────────────────────────────────────────────────
@@ -271,7 +286,7 @@ export class K8sBackend {
       imagePullPolicy: callImagePullPolicy,
     } = opts
     validateImagePullPolicy(callImagePullPolicy, 'createEnvironment opts')
-    const ns = namespace ?? this._namespace
+    const ns = this._resolveNamespace(namespace)
     const podName = `chroxy-env-${envId}`
     const secretName = `chroxy-token-${envId}`
     // K8sBackend ALWAYS runs the chroxy-pod-agent sidecar — the sidecar is
@@ -518,7 +533,7 @@ export class K8sBackend {
    * @returns {Promise<void>}
    */
   async destroyEnvironment(podName, opts = {}) {
-    const ns = opts.namespace ?? this._namespace
+    const ns = this._resolveNamespace(opts.namespace)
     const secretName = opts.secretName || _deriveSecretName(podName)
 
     // Always drop the cached token first so a partial failure can't leave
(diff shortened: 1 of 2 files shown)


blamechris:
## Code Review: PR #3569

### Summary

Mechanical extraction of the five-site `(opts.)namespace ?? this._namespace` pattern into a single `_resolveNamespace(callNamespace)` instance helper on `K8sBackend`. Behaviour is preserved exactly — same `??` semantics, same empty-string contract from #3493, same `null`/`undefined` fallback. Adds two direct unit tests pinning the helper's contract.

### Strengths

- **Truly mechanical diff at the call sites.** Each replacement is `X ?? this._namespace` → `this._resolveNamespace(X)` with no other changes to the surrounding code. Easy to audit, low review risk.
- **Helper signature is uniformly applicable.** Taking a raw `callNamespace` (rather than `opts`) works cleanly across all five sites — including the two (`createEnvironment`, `streamCliInEnvironment`) that already destructure `namespace` out of opts for other reasons. An `(opts)` signature would have forced one of those call sites into an awkward shape.
- **Naming follows convention.** `_resolveNamespace` matches the underscore-prefix pattern used throughout `K8sBackend` for non-public surface (`_namespace`, `_sidecarImage`, `_dialWs`, `_readAgentToken`).
- **Tests are well-targeted.** Two focused unit tests on the helper itself, with explicit assertions on the empty-string contract and the `undefined`/`null`/no-arg fallback. The existing #3493 block continues to exercise behaviour at all five call sites — so the seam is pinned both directly and via the call sites.
- **Tight scope discipline.** The companion `validateNamespace()` (RFC 1123) idea from #3534 is correctly deferred. Landing the helper as a no-op rename today means the validator can land later without re-touching the five call sites — the issue's intended layered-improvement progression.
- **JSDoc is complete and references #3493** so the empty-string contract isn't mysterious to future readers.

### Issues Found

#### Critical (Must Fix)

None.

None.

#### Nitpicks (Optional)

None.

### Architecture Notes

The helper sits at the right seam: it is the boundary between caller intent (per-call override) and instance default (constructor-stored namespace). Having all five sites funnel through it is exactly what makes a layered RFC 1123 validator a one-line addition later — drop a `validateNamespace(ns, context)` call inside `_resolveNamespace`, or wrap the result, and every site gains validation atomically. That is the entire point of the extraction.

The instance-method shape (rather than a module-level function) is correct because the helper's whole job is to consult `this._namespace`. A static helper would just push the `this._namespace` argument back to every caller, defeating the consolidation.

No protocol surface change. No public API change. Existing #3493 behavioural coverage untouched.

### Verdict

Clean, minimal, well-tested DRY refactor. The acceptance criteria from #3534 that are in this PR's scope (extract helper, replace five sites, add direct tests) are all met. The deferred items (`validateNamespace` + RFC 1123 cases + class JSDoc update) are correctly handled by leaving them in #3534 as the layered follow-up — that issue can stay open after this PR merges, since this PR is a strict subset.

blamechris:

| `validateNamespace(ns, context)` RFC 1123 validator wired through `_resolveNamespace`, plus updated empty-string tests and class-JSDoc contract documentation | #3571 | Explicitly out of PR #3569's scope — the PR description committed to a tight refactor-only change. Tracked separately so the deferred half of #3534 isn't lost when #3534 auto-closes on this PR's merge. |

The validator becomes a one-line addition once the seam exists, which is exactly the layering #3534 envisioned.

Issue #3208, opened 2026-05-03
protocol: WS broadcast events `evaluator_rewrite` / `evaluator_clarify`


blamechris:
## Parent epic
Part of #3068 — pre-prompt evaluator hook (auto-intermediation).

## Scope
The existing `evaluate_draft_result` is request/response (manual flow). The auto-flow needs broadcast events that arrive WITHOUT a triggering client request, so the dashboard knows to render the rewrite-explanation or clarify-question UI.

## Wire shape
```
{
  type: 'evaluator_rewrite',
  sessionId: string,
  originalDraft: string,
  rewritten: string,
  reasoning: string,
  evaluatorIterationId: string  // monotonic per session for dashboard dedup
}

{
  type: 'evaluator_clarify',
  sessionId: string,
  originalDraft: string,
  clarification: string,
  reasoning: string,
  evaluatorIterationId: string,
  evaluatorIteration: number  // 1, 2, 3 — clamped to maxIterations
}
```

## Acceptance criteria
- [ ] Add Zod schemas to `@chroxy/protocol/src/schemas/server.ts`
- [ ] Document in `ws-server.js` header
- [ ] Server emits when auto-evaluation hook lands on rewrite/clarify verdicts
- [ ] Tests: schema validation

---
Parent epic: #3068

Issue #3186, opened 2026-05-03
server: auto-evaluation hook on user_input — gate before forwarding to session


blamechris:
## Parent epic
Part of #3068 — pre-prompt evaluator hook (auto-intermediation).

## Scope
Wire the existing `evaluateDraft` (already in `packages/server/src/prompt-evaluator.js`) into the `user_input` WS handler so that when the session's `promptEvaluator` toggle (other sub-task) is `true`, the message is evaluated before being forwarded.

## Behaviour
- `verdict: 'forward'` → original message goes to the session unchanged.
- `verdict: 'rewrite'` → rewritten version is sent to the session; broadcast `evaluator_rewrite` event so dashboard can show "your message was rewritten".
- `verdict: 'clarify'` → DON'T forward. Broadcast `evaluator_clarify`. Wait for user follow-up; re-evaluate (max iterations = 3).
- API error → log + forward original (fail-open) so the user isn't blocked by evaluator outages.

## Acceptance criteria
- [ ] `user_input` handler reads `session.config.promptEvaluator` and short-circuits past evaluation when false
- [ ] When true, calls `evaluateDraft` with the user's message + last N (e.g., 5) session messages as context
- [ ] Verdict-based routing per the table above
- [ ] Max iteration guard for `clarify` loop
- [ ] Fail-open on evaluator API error (log + forward original)
- [ ] Tests: each verdict path + fail-open + max-iteration cap

---
Parent epic: #3068

PR #3655, opened 2026-05-08
test(protocol): pin empty-string and extra-field policies for evaluator broadcast schemas

### LINKED ISSUE #3627: protocol: add empty-string and extra-field policy tests for evaluator broadcast schemas
## Context

PR #3625 lands `ServerEvaluatorRewriteSchema` and `ServerEvaluatorClarifySchema` (parent #3208, epic #3068). The 11 schema tests cover well-formed payloads, wrong type literals, missing required fields, and iteration boundaries.

## Suggested additional coverage

Two contract decisions are currently implicit and untested:

1. **Empty-string policy** — `originalDraft: ''`, `rewritten: ''`, `clarification: ''`, `reasoning: ''` all currently parse as valid (Zod's default `z.string()` allows empty). This is probably intentional (auto-evaluator could legitimately produce empty reasoning during a model timeout fallback), but pinning the choice with a test prevents accidental regression if someone tightens to `z.string().min(1)`.

2. **Extra-field policy** — these schemas are strict-by-default (no `.passthrough()` or `.strict()`). Zod's default is to STRIP unknown fields, not reject. A test asserting `safeParse({ ...valid, extraField: 'foo' }).success === true` (or `.error` if `.strict()`) pins the choice and protects against accidental tightening that would break older servers when newer fields land.

## Suggested next step

Add two tests per schema (4 tests total):
- `accepts empty originalDraft / rewritten / clarification / reasoning`
- `strips unknown fields for forward compat` (mirrors the pattern at the end of `ServerSessionListEntrySchema (#3573)` describe block)

## References
- PR #3625 (deferred from review)
- Issue #3208

blamechris:
## Summary
Pins two implicit contract decisions on `ServerEvaluatorRewriteSchema` and `ServerEvaluatorClarifySchema` (#3208) that were flagged in the PR #3625 review:

1. **Empty-string policy** — `z.string()` accepts empty by default. The auto-evaluator (#3186) may produce empty reasoning under timeout fallback, so this is intentional.
2. **Extra-field policy** — schemas are unscoped, so Zod strips unknown fields. Forward compat: a newer server may emit fields older clients don't recognize.

4 new tests (2 per schema). Total protocol test count: 111 (was 107).

Closes #3627

### DIFF
--- a/packages/protocol/tests/schemas.test.js
+++ b/packages/protocol/tests/schemas.test.js
@@ -892,6 +892,38 @@ describe('@chroxy/protocol schemas', () => {
       })
       assert.ok(!result.success, 'sessionId must be a string')
     })
+
+    // #3627: pin the empty-string and extra-field policies so a future
+    // tightening (z.string().min(1) or .strict()) is a deliberate decision
+    // with a failing test, not a silent regression that breaks older
+    // servers when newer fields land.
+    it('accepts empty originalDraft / rewritten / reasoning (empty-string policy)', async () => {
+      const { ServerEvaluatorRewriteSchema } = await import('../src/schemas/server.ts')
+      const result = ServerEvaluatorRewriteSchema.safeParse({
+        type: 'evaluator_rewrite',
+        sessionId: 'sess-1',
+        originalDraft: '',
+        rewritten: '',
+        reasoning: '',
+        evaluatorIterationId: 'iter-1',
+      })
+      assert.ok(result.success, 'empty strings must validate — auto-evaluator may produce empty reasoning under timeout fallback')
+    })
+
+    it('strips unknown fields for forward compat (Zod default behavior)', async () => {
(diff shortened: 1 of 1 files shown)


blamechris (commented):
LGTM. Tests correctly pin both policies and would fail-fast on a future tightening: the empty-string assertions (`assert.ok(result.success, ...)`) would break if anyone added `.min(1)` to the four string fields, and the extra-field tests would break under either `.strict()` (success flips to false) or `.passthrough()` (`result.data.someFutureField` would equal the object, not `undefined`). Style matches the surrounding `ServerSessionListEntrySchema` pattern — dynamic `await import('../src/schemas/server.ts')`, issue-number comments, `safeParse` + `assert.ok`/`assert.equal`. Verified locally: 111/111 protocol tests pass, no code changes outside `packages/protocol/tests/schemas.test.js`.

blamechris:
## Review Comments Addressed

Copilot review completed with 0 inline comments. No fixes required.

PR #3662, opened 2026-05-08
test(dashboard): assert rewrite banner survives real localStorage round-trip

### LINKED ISSUE #3647: dashboard: assert reconnect-replay rehydrates EvaluatorRewriteBanner from localStorage (real path, not handler)
PR #3643's \`message-handler-evaluator.test.ts\` includes a test titled "a rewrite system message replayed via session messages still renders only once" — but the test calls \`handleEvaluatorRewrite\` twice with the same event. That exercises the **handler-level** dedup, not the actual reconnect-replay path.

In production, reconnect replay does NOT go through the handler. Messages are loaded from \`localStorage\` via \`loadAllSessionMessages\` directly into \`sessionStates[id].messages\` during \`loadSavedConnection\`. So the dedup that matters on reconnect is "system message round-trips through JSON.stringify/parse with \`evaluator\` metadata intact and renders the banner".

## Suggested fix

Add a focused test that:
1. Calls \`handleEvaluatorRewrite\` to inject the system message.
2. Reads the persisted state via \`loadSessionMessages(sessionId)\` (after debounce flush).
3. Asserts the loaded \`ChatMessage[]\` contains an entry with \`type: 'system'\` and \`evaluator.kind === 'rewrite'\`.
4. Renders \`<App />\` with that hydrated state and asserts \`EvaluatorRewriteBanner\` appears once.

## Why deferred

The acceptance-criterion claim "visual state survives reconnect (replayed via session_messages)" is true — but the existing tests don't actually prove it.

Refs PR #3643, parent epic #3068.

blamechris:
## Summary

The existing "replay safety" test in `message-handler-evaluator.test.ts` calls `handleEvaluatorRewrite` twice with the same `evaluatorIterationId` within a single store lifecycle. That exercises handler-level dedup, not the actual reconnect-replay path.

In production, reconnect replay does NOT go through the handler — `loadSavedConnection` reads cached system messages straight out of `localStorage` via `loadAllSessionMessages` and pushes them into `sessionStates[id].messages` directly. The dedup that matters on reconnect is "the `evaluator` metadata round-trips cleanly through `JSON.stringify`/`parse`, and a follow-up live `evaluator_rewrite` for the same iteration id does NOT double-insert the banner".

## What this adds

A new describe block at the end of `message-handler-evaluator.test.ts`:

- Fires `evaluator_rewrite` so the handler pushes the banner.
- Persists via the **real** `persistSessionMessages` (uses `vi.importActual` to bypass the file-level `vi.mock('./persistence')`).
- Flushes the debounce, then recreates the store to simulate reconnect.
- Rehydrates via the **real** `loadSessionMessages` — the same call `loadSavedConnection` makes on reconnect.
- Deep-equals the rehydrated `evaluator` block against the live one to pin the full shape.
- Re-fires the live `evaluator_rewrite` post-reconnect and asserts the banner is not double-inserted.

The deep-equal on `evaluator` is the regression guard: if anyone adds a new field to the banner that doesn't survive `JSON.stringify` (e.g. a `Function`, `undefined`, or `Date` that gets coerced to a string), this test will fail.

## Test plan

- [x] `npm test --workspace=@chroxy/dashboard -- --run src/store/message-handler-evaluator.test.ts` — 16/16 pass (was 15)
- [x] Other store test suites still pass (317/317 across 11 files; the 1 unrelated failure is a pre-existing `@testing-library/react` resolution issue on `commands.test.ts`)
- [x] No new typecheck errors in the touched file

Closes #3647

### DIFF
--- a/packages/dashboard/src/store/message-handler-evaluator.test.ts
+++ b/packages/dashboard/src/store/message-handler-evaluator.test.ts
@@ -386,6 +386,126 @@ describe('dashboard message-handler — auto-evaluator (#3188)', () => {
     })
   })
 
+  // #3647 — The dedupe test above only exercises handler-level dedup within a
+  // single store lifecycle. In production, reconnect replay does NOT re-fire
+  // through `handleMessage` — `loadSavedConnection` reads the cached system
+  // message straight out of `localStorage` via `loadAllSessionMessages` and
+  // pushes it into `sessionStates[id].messages` directly. So the dedup that
+  // matters on reconnect is "the `evaluator` metadata round-trips cleanly
+  // through JSON.stringify/parse, and a follow-up live `evaluator_rewrite`
+  // for the same iteration id does NOT double-insert the banner".
+  //
+  // This test pins that contract using the REAL persistence helpers
+  // (`persistSessionMessages` + `loadSessionMessages`), not a JSON.stringify
+  // simulation. The persistence module is mocked at the top of this file for
+  // the rest of the suite, so we pull in the real implementation via
+  // `vi.importActual` here. The goal: catch a regression where someone adds a
+  // new field to the rewrite banner that doesn't survive JSON serialization
+  // (e.g. a Function, undefined, Date that gets coerced to string).
+  describe('localStorage round-trip: rewrite banner survives a real persist + rehydrate cycle (#3647)', () => {
+    it('round-trips through localStorage and dedupes a follow-up live event', async () => {
+      // Pull the real persistence helpers, bypassing the file-level vi.mock.
+      const realPersistence = await vi.importActual<typeof import('./persistence')>('./persistence')
+      const {
+        persistSessionMessages,
+        loadSessionMessages,
+        flushPendingWrites,
+        _resetForTesting,
+      } = realPersistence
+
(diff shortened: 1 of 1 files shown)


blamechris:
## Technical Review

**Verdict: LGTM**

Sharp test — addresses a real coverage gap.

The previous "replay safety" test was running through the live handler twice within the same store lifecycle. That exercises handler-level dedup but NOT the production reconnect path, which (per `loadSavedConnection`) reads cached system messages straight out of `localStorage` via `loadAllSessionMessages` and pushes them into `sessionStates[id].messages` directly.

The new test uses `vi.importActual` to bypass the file-level `vi.mock('./persistence')` and exercises `persistSessionMessages` + `flushPendingWrites` + `loadSessionMessages` for a real round-trip. The deep-equal on the `evaluator` block is the regression pin: any future field that doesn't survive `JSON.stringify` (Function, undefined, Date) breaks this test.

Bonus: also asserts the post-reconnect dedup contract (live `evaluator_rewrite` for the same iteration id doesn't double-insert against a rehydrated banner). That's the failure mode that would actually surface in production.

CI green, only adds tests — no production code changes.

PR #3711, opened 2026-05-10
fix(dashboard): move Skills to icon button in header-right + bump 0.7.9

blamechris:
## Summary

The Skills text button kept overlapping with neighbouring controls in header-center because text-button widths fight with the model dropdown's flexible column. Moved it into header-right as an icon-only button (💾 floppy glyph per user suggestion) where it sits alongside copy-transcript and gear with stable spacing — header-right zones don't compete for the flexible 1fr column.

## Changes

- **\`App.tsx\`** — Skills button now uses \`.header-icon-btn\` instead of \`.header-text-btn\`, lives in header-right BEFORE the copy button. Visible glyph is 💾 (\`&#128190;\` / U+1F4BE). \`aria-label\` and \`title\` remain "Skills" for screen readers and tooltips.
- **\`components.css\`** — removed the now-unused \`.header-text-btn\` rule + its hover/focus rules (no element references it). Consolidated the duplicate \`flex-shrink: 0\` rule into the main \`.header-icon-btn\` block.
- **\`HeaderOverflow.test.tsx\`** — simplified the no-shrink assertion to match the single \`.header-icon-btn\` rule.

## Why icon, not text

Text labels in header-center compete with the model dropdown's \`1fr\` column for space. Icons in header-right have a stable width budget — they're predictable and don't push other elements around. The \`title\` attribute restores the discoverability that the text label provided.

## Why floppy 💾

User-requested icon. It's not semantically perfect for "skills" (a toolbox 🧰 or hammer/pick ⚒ would fit better thematically), but it's recognisable as a glyph and easy to swap later if a better icon is preferred.

## Test plan

- [x] 1570/1570 dashboard tests pass
- [x] HeaderOverflow.test.tsx updated to assert the simplified .header-icon-btn rule
- [ ] Manual: rebuild Tauri 0.7.9, verify Skills 💾 button sits beside copy/gear in header-right with no overlap; clicking opens the SkillsPanel as before

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "chroxy",
-  "version": "0.7.8",
+  "version": "0.7.9",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "chroxy",
-      "version": "0.7.8",
+      "version": "0.7.9",
       "license": "MIT",
       "workspaces": [
         "packages/*"
@@ -16856,7 +16856,7 @@
     },
     "packages/app": {
       "name": "@chroxy/app",
-      "version": "0.7.8",
+      "version": "0.7.9",
       "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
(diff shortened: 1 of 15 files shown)


blamechris:
Superseded by #3713 — same change rebased on main (which advanced past 0.7.9 to 0.7.10 via #3712), now bumping to 0.7.11.

PR #3721, opened 2026-05-10
fix(dashboard): per-kind widths for header selects + bump 0.7.14 (#3720)

blamechris:
## Summary

- Fixes #3720 — the permission and thinking selects in the dashboard header inherited the model select's `min-width: 180px / max-width: 240px` floor and rendered ~2–3× wider than their content needed, visually crowding the icon row and cost/tokens display.
- Adds `data-kind="model|permission|thinking"` to each `<select>` in `ChatSettingsDropdown.tsx` so the CSS can size each one to match its longest label.

## What changed

- `ChatSettingsDropdown.tsx`: added `data-kind` to all three selects.
- `components.css`: removed the shared `min-width/max-width` from `.header-center select`; added per-kind rules:
  - model: 180–240px (unchanged)
  - permission: 110–160px (fits "Accept Edits" + chevron)
  - thinking: 80–110px (fits "Default")
- Same per-kind split inside the `@media (max-width: 600px)` block so narrow viewports keep proportional floors.
- `HeaderOverflow.test.tsx`: updated the affected two tests to validate the per-kind selectors and floors.
- Version bump: 0.7.13 → 0.7.14 across all packages.

## Test plan

- [x] Dashboard typecheck — pass
- [x] Dashboard tests — 1572/1572 pass
- [ ] Manual: install 0.7.14 Tauri, verify the permission dropdown renders narrow ("Approve" ≈ 110px wide, not 180px+)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "chroxy",
-  "version": "0.7.13",
+  "version": "0.7.14",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "chroxy",
-      "version": "0.7.13",
+      "version": "0.7.14",
       "license": "MIT",
       "workspaces": [
         "packages/*"
@@ -16856,7 +16856,7 @@
     },
     "packages/app": {
       "name": "@chroxy/app",
-      "version": "0.7.13",
+      "version": "0.7.14",
       "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
@@ -16909,7 +16909,7 @@
     },
     "packages/dashboard": {
       "name": "@chroxy/dashboard",
-      "version": "0.7.13",
+      "version": "0.7.14",
       "dependencies": {
         "@chroxy/protocol": "*",
         "@chroxy/store-core": "*",
@@ -17019,7 +17019,7 @@
     },
     "packages/desktop": {
       "name": "@chroxy/desktop",
-      "version": "0.7.13"
+      "version": "0.7.14"
     },
     "packages/protocol": {
       "name": "@chroxy/protocol",
@@ -17034,7 +17034,7 @@
     },
     "packages/server": {
       "name": "@chroxy/server",
-      "version": "0.7.13",
+      "version": "0.7.14",
       "license": "MIT",
       "dependencies": {
         "@anthropic-ai/claude-agent-sdk": "^0.2.0",
@@ -17093,7 +17093,7 @@
     },
     "packages/store-core": {
       "name": "@chroxy/store-core",
-      "version": "0.7.13",
+      "version": "0.7.14",
       "dependencies": {
         "@chroxy/protocol": "*",
         "tweetnacl": "^1.0.3",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "chroxy",
-  "version": "0.7.13",
+  "version": "0.7.14",
   "private": true,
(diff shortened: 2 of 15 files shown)


blamechris:
## Code Review: PR #3721

### Summary
Tight, well-scoped CSS fix. The shared `min-width: 180px / max-width: 240px` floor on `.header-center select` was sized for the model dropdown's longest label and was visibly wrong on the permission/thinking selects. Splitting via `data-kind="model|permission|thinking"` is the right primitive — small surface area, no regression risk, and the responsive `@media` block was updated in the same shape. Tests cover both the bare rule and the per-kind variants. CI is green across all 12 checks.

### Strengths
- Per-kind data attributes keep CSS specificity clean and intent-readable in DevTools (you can see exactly which select picked up which floor).
- Same per-kind split inside `@media (max-width: 600px)` — narrow viewports keep proportional ratios instead of all collapsing to one floor.
- New `HeaderOverflow.test.tsx` case explicitly anchors each kind's expected min/max-width, so future drift on any single kind fails loudly.
- The bare `.header-center select` block retains the truncation properties (`overflow/white-space/text-overflow`) and is asserted by a separate test — refactor stayed focused.
- No backwards-compat breakage: a `grep -rn "180px|240px"` across `packages/dashboard/src` and `packages/app/src` finds no other code relying on the old shared floor.
- Permission floor is wide enough for the longest label `"Auto (bypass)"` (13 chars + chevron + padding ≈ 110–120px) — fits within the 160px ceiling.
- `select[data-kind=...]` is plain CSS3 attribute-selector, fully supported in Tauri's WKWebView.

### Issues Found

#### Critical (Must Fix)

_None._

_None._

#### Nitpicks (Optional)
- `HeaderOverflow.test.tsx:79` lists permission labels as `"Approve"/"Accept Edits"`, but `packages/server/src/handler-utils.js` actually defines the longest as `"Auto (bypass)"` (13 chars vs 12 for "Accept Edits"). Floors still fit, but the comment slightly underplays the worst-case label and was the actual motivation for choosing 110px not 100px. Worth a one-line tweak next time you touch this file.
- `components.css:130` — bare `.header-center select` still carries `flex-shrink: 1`. Since `.header-center` is a grid item (and the existing `.header-left` test explicitly notes `flex-shrink` is a no-op on grid items), the inner select being a flex child of a grid-item-flex-container is fine, but the property is functionally redundant given `min-width: 0` on the parent already lets the column shrink. Pre-existing — don't churn for it.
- The `@media` regex `/@media \(max-width: 600px\)\s*\{[\s\S]*?\n\}/` works because the only column-0 `}` inside the block is its own closing brace (all inner one-liners and the multi-line `.header-center select` close with indented `  }`). It's correct, but brittle if anyone adds a nested rule whose closing brace lands at column 0. A more defensive regex would balance braces (e.g., parse the block by counting), but that's overkill for a CSS fixture test.

_None._

### Architecture Notes
The header layout invariants are preserved: 3-column CSS grid (`auto | minmax(0, 1fr) | auto`), `overflow: hidden` on `.header-center`, native `<select>` rendering via OS compositor for dropdown menus. No flex/grid model changes — only the per-element width floors changed. The change is purely cosmetic and cannot affect dropdown menu rendering, focus behaviour, or icon-row layout in `.header-right`.

`data-kind` is a new convention in the dashboard (no existing `data-role`/`data-variant` usage), so it doesn't conflict with anything. If more components want per-variant CSS later, this naming will scale.

### Verdict
- [x] Approve
- [ ] Request Changes
- [ ] Comment

PR #3724, opened 2026-05-10
perf(server): cache _hasClaudeOAuthCreds probe with 5s TTL (#3678)

blamechris:
## Summary

Wraps the `_hasClaudeOAuthCreds()` on-disk probe (3 `existsSync` + optional `readFileSync` + `JSON.parse`) in a module-scoped 5s TTL cache so `listProviders()` stops re-doing the disk work on every dashboard `list_providers` request and on every `auth_ok` in `ws-history.js`.

- Renames the existing probe to `_probeClaudeOAuthCreds()` (private)
- New `_hasClaudeOAuthCreds()` keys the cache on `\`${CHROXY_CLAUDE_HOME ?? ''}|${CHROXY_CLAUDE_CONFIG ?? ''}\`` — env-var changes naturally invalidate the previous result
- TTL is 5 seconds (`now + 5_000`)
- Exports `_resetCredsCacheForTest()` so test suites that mutate files under a stable env-var pair can drop the cache; called from the existing `clearKeys()` / `restoreKeys()` helpers in `providers.test.js`

Public `auth.ready` / `auth.source` shape is unchanged.

## Test plan

- [x] `node --experimental-test-module-mocks --test --test-force-exit tests/providers.test.js` — 41/41 pass
- [x] `npm run lint` — 0 new errors (only pre-existing warnings in unrelated files)

Closes #3678.

### DIFF
--- a/packages/server/src/providers.js
+++ b/packages/server/src/providers.js
@@ -346,7 +346,7 @@ function getProviderAuthInfo(name, ProviderClass) {
  *
  * @returns {boolean}
  */
-function _hasClaudeOAuthCreds() {
+function _probeClaudeOAuthCreds() {
   try {
     const claudeHome = process.env.CHROXY_CLAUDE_HOME || join(homedir(), '.claude')
     if (existsSync(join(claudeHome, 'auth.json'))) return true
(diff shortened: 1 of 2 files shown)


blamechris:
## Code Review: PR #3724

### Summary

Wraps `_hasClaudeOAuthCreds()` in a 5s module-scoped TTL cache keyed on the override env vars (`CHROXY_CLAUDE_HOME` | `CHROXY_CLAUDE_CONFIG`). Renames the existing implementation to `_probeClaudeOAuthCreds()` and exports `_resetCredsCacheForTest()` for test isolation. Implementation is exactly what #3678 chartered — no scope creep.

### Strengths

- Minimal blast radius: only `providers.js` (probe + cache) and `providers.test.js` (call reset helper from `clearKeys`/`restoreKeys`)
- Clean private/public split — `_probeClaudeOAuthCreds` is the pure on-disk probe, `_hasClaudeOAuthCreds` is the cached facade
- Cache key built from override env vars means existing tests get free invalidation on `clearKeys()` (each call mints a new tmpdir → new key)
- Defensive belt-and-suspenders: `_resetCredsCacheForTest()` is called from BOTH `clearKeys()` and `restoreKeys()`, so the cache also gets dropped on env-var restore in case a follow-on test in another describe block reuses the same paths
- Code style matches `packages/server/` conventions: ES modules, no semicolons, single quotes, JSDoc on public surface
- Public `auth.ready` / `auth.source` shape preserved — `listProviders()` consumers (dashboard, ws-history) see no change
- 41/41 tests pass including the #3674 OAuth-probe coverage and the #3677 malformed-JSON regression
- Test-only export named with a clear `_…ForTest` suffix per common Node convention — production callers won't mistake it for a regular API

### Issues Found

#### Critical (Must Fix)

None.

| `packages/server/src/providers.js` | ~387 | Consider exposing `_resetCredsCacheForTest` via the existing `_test` namespace if other test-only hooks ever land. Right now this is the only one, so a dedicated export is fine. | Future-proofing only — no action needed on this PR. |

#### Nitpicks (Optional)

- The cache invariant is "key change OR TTL expiry forces a reprobe." A tiny inline comment on the `if (_credsCache.key === key && _credsCache.expiresAt > now)` line could spell out "key check first because env-var change must invalidate even mid-TTL." But the JSDoc above already says this — fine to leave as-is.
- `Date.now()` is wall-clock and could be jumped backwards/forwards by NTP. For a 5s UX-freshness cache this is irrelevant; flagging only for completeness.

None. The suggestion above is a nit and the issue spec ("module-scoped 5-10s TTL cache") is fully implemented.

### Architecture Notes

`providers.js` is a registry module with a single side effect (the new module-scoped `_credsCache` object). That's idiomatic for this codebase — `models.js` also keeps mutable module state for the dynamic model list. No EventEmitter or async coordination needed because Node is single-threaded and the probe is sync.

The cache invalidation strategy (env-var key) is correct because the only runtime mutation paths to credential location are those two env vars; `homedir()` is process-stable. If a future change adds a third override path (e.g. `CHROXY_CLAUDE_CREDENTIALS_FILE`), it must be added to the key — worth a one-line comment on the cache itself, but not blocking.

No WS protocol surface changes, no resilience implications (probe is local fs, no network), no mobile concerns (server-only).

### Verdict

- [x] **Approve** - Ready to merge.

Tight, scoped performance win that exactly matches the issue acceptance criteria. No behaviour change. Test isolation is preserved by both natural key invalidation and an explicit reset hook. CI / Copilot should be the only remaining gates.

PR #3776, opened 2026-05-11
feat(server): forward resultTimeoutMs through JsonlSubprocessSession (#3755)

### LINKED ISSUE #3755: JsonlSubprocessSession: forward resultTimeoutMs to BaseSession (Codex/Gemini consistency)
## Background

PR #3754 added a configurable `resultTimeoutMs` option to `BaseSession`, with `SdkSession` and `CliSession` updated to forward the option in their constructors. `JsonlSubprocessSession` (the BaseSession middle-layer used by `CodexSession` and `GeminiSession`) does NOT forward `resultTimeoutMs` to `super(...)`.

Currently this is harmless because Codex and Gemini sessions don't arm an inactivity timer (no `_armResultTimeout` / `_handleResultTimeout` usage in `jsonl-subprocess-session.js`). The `_resultTimeoutMs` field on the instance still gets the default via the BaseSession initialiser, but no timer ever uses it.

## Why this is worth a follow-up

The middle-layer pattern requires consistent option forwarding (see internal note `feedback_jsonl_subprocess_middle_layer.md`). When/if Codex or Gemini adopt the same inactivity-timer pattern that the Claude providers use, the `resultTimeoutMs` plumbing must already be in place to avoid silently ignoring operator config.

## Proposed change

`packages/server/src/jsonl-subprocess-session.js` — add `resultTimeoutMs` to the destructured constructor args and forward via `super({ ..., resultTimeoutMs })`, mirroring the pattern in `sdk-session.js` and `cli-session.js`.

## Acceptance

- `JsonlSubprocessSession` constructor accepts and forwards `resultTimeoutMs`.
- A test in `codex-session.test.js` (or `jsonl-subprocess-session.test.js` if added) asserts `_resultTimeoutMs` is honoured on construction.

Origin: review of PR #3754.

blamechris:
## Summary

Closes #3755. PR #3754 added the configurable `resultTimeoutMs` BaseSession opt; `SdkSession` and `CliSession` forward it via super(); `JsonlSubprocessSession` (the middle layer used by `CodexSession` / `GeminiSession`) silently dropped it.

This is the canonical "middle-layer drops opts" pattern documented in our internal review notes (`feedback_jsonl_subprocess_middle_layer.md`). Harmless today because Codex/Gemini don't arm an inactivity timer, but the plumbing has to be in place so operator config isn't silently ignored if/when they adopt one.

## Change

```diff
 constructor({
   ...,
   promptEvaluator,
   promptEvaluatorSkipPattern,
+  resultTimeoutMs,
 } = {}) {
   super({
     ...,
     promptEvaluator,
     promptEvaluatorSkipPattern,
+    resultTimeoutMs,
   })
```

Two new constructor tests mirror the SdkSession/CliSession pattern:
- forwards `resultTimeoutMs` to BaseSession when provided
- defaults `_resultTimeoutMs` to 20 min when omitted

## Test plan

- [x] 33/33 jsonl-subprocess-session.test.js passes locally
- [ ] Full server test suite green in CI

Closes #3755

### DIFF
--- a/packages/server/src/codex-session.js
+++ b/packages/server/src/codex-session.js
@@ -219,11 +219,11 @@ export class CodexSession extends JsonlSubprocessSession {
     }
   }
 
-  constructor({ cwd, model, permissionMode, skillsDir, repoSkillsDir, maxSkillBytes, maxTotalSkillBytes, provider, activeManualSkills, providerSkillAllowlist, trustStore, trustMismatchMode, promptEvaluator, promptEvaluatorSkipPattern } = {}) {
+  constructor({ cwd, model, permissionMode, skillsDir, repoSkillsDir, maxSkillBytes, maxTotalSkillBytes, provider, activeManualSkills, providerSkillAllowlist, trustStore, trustMismatchMode, promptEvaluator, promptEvaluatorSkipPattern, resultTimeoutMs } = {}) {
     // `model` may be null/undefined — BaseSession coerces to null and
     // _buildArgs() omits the `-c model=...` flag so Codex CLI defers
     // to its own default from ~/.codex/config.toml.
-    super({ cwd, model: model || DEFAULT_MODEL, permissionMode, skillsDir, repoSkillsDir, maxSkillBytes, maxTotalSkillBytes, provider: provider || 'codex', activeManualSkills, providerSkillAllowlist, trustStore, trustMismatchMode, promptEvaluator, promptEvaluatorSkipPattern })
+    super({ cwd, model: model || DEFAULT_MODEL, permissionMode, skillsDir, repoSkillsDir, maxSkillBytes, maxTotalSkillBytes, provider: provider || 'codex', activeManualSkills, providerSkillAllowlist, trustStore, trustMismatchMode, promptEvaluator, promptEvaluatorSkipPattern, resultTimeoutMs })
   }
 
   // ------------------------------------------------------------------
--- a/packages/server/src/gemini-session.js
+++ b/packages/server/src/gemini-session.js
@@ -188,8 +188,8 @@ export class GeminiSession extends JsonlSubprocessSession {
     }
   }
 
-  constructor({ cwd, model, permissionMode, skillsDir, repoSkillsDir, maxSkillBytes, maxTotalSkillBytes, provider, activeManualSkills, providerSkillAllowlist, trustStore, trustMismatchMode, promptEvaluator, promptEvaluatorSkipPattern } = {}) {
-    super({ cwd, model: model || DEFAULT_MODEL, permissionMode, skillsDir, repoSkillsDir, maxSkillBytes, maxTotalSkillBytes, provider: provider || 'gemini', activeManualSkills, providerSkillAllowlist, trustStore, trustMismatchMode, promptEvaluator, promptEvaluatorSkipPattern })
+  constructor({ cwd, model, permissionMode, skillsDir, repoSkillsDir, maxSkillBytes, maxTotalSkillBytes, provider, activeManualSkills, providerSkillAllowlist, trustStore, trustMismatchMode, promptEvaluator, promptEvaluatorSkipPattern, resultTimeoutMs } = {}) {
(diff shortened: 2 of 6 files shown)


blamechris:
## Chroxy Inspector Review — PR #3776

### Summary

Adds `resultTimeoutMs` to the destructured constructor args and `super(...)` call in `JsonlSubprocessSession`, plus two parallel constructor tests. Closes #3755. The diff is +21/-0 and purely additive on the constructor path — cannot break existing callers.

Pattern parity with the canonical fix at `sdk-session.js:237-238` and `cli-session.js:161-162` is exact: same opt name, same destructure position (last in the list), same `super(...)` forwarding semantics. Test parity with `sdk-session.test.js:68-78` is also exact for the two cases included.

The PR correctly identifies the "middle-layer drops opts" pattern documented in `feedback_jsonl_subprocess_middle_layer.md`. However, while reviewing I noticed the fix is incomplete one layer deeper — flagged below.

### Strengths

- Pattern parity is exact; the change is small, mechanical, and easy to reason about.
- Test comments explicitly cite the issue (#3755) and explain the dormant-bug rationale so future maintainers don't delete the tests as "redundant".
- The PR body and the issue both correctly note that the bug is dormant today (Codex/Gemini don't arm an inactivity timer) — sets clear expectations.
- Net additive change with no risk to existing callers.

### Issues Found

#### Critical

**1. `CodexSession` and `GeminiSession` still drop `resultTimeoutMs` before it reaches `JsonlSubprocessSession`** — `packages/server/src/codex-session.js:222` and `packages/server/src/gemini-session.js:191` each declare an explicit destructure that does NOT list `resultTimeoutMs`, and their `super(...)` calls don't forward it either. `SessionManager.createSession()` does add `providerOpts.resultTimeoutMs` (`session-manager.js:507`), but for Codex/Gemini providers it lands on the implicit rest-bucket of the leaf class's destructure and is silently discarded before this PR's fix in `JsonlSubprocessSession` ever sees it.

In effect, the middle-layer plumbing added here is unreachable in production for the two providers that actually use it. Same "operator config silently ignored" failure mode the PR is trying to prevent — just one layer deeper.

I'd recommend either (a) extending this PR to fix both leaf classes (4 lines, plus a leaf-level test each), or (b) merging this as-is and tracking the leaf fix as a tight follow-up. Filed as #3778 either way.

#### Suggestions

**2. Add the non-positive fallback test (`0`, `-1`, `'oops'`)** — `sdk-session.test.js:80-89` covers this third case to guard the `Number.isFinite(x) && x > 0` check in `base-session.js:154-157`. If that validation regresses, sessions could silently arm a zero-ms inactivity timer. Mirroring this case in `jsonl-subprocess-session.test.js` is cheap and keeps parity with the SdkSession suite. Filed as #3777.

#### Nitpicks

None.

### Deferred Items

| # | Item | Issue |
|---|------|-------|
| 1 | Cover non-positive `resultTimeoutMs` fallback (0, -1, 'oops') in `jsonl-subprocess-session.test.js` | https://github.com/blamechris/chroxy/issues/3777 |
| 2 | `CodexSession` / `GeminiSession` constructors must also forward `resultTimeoutMs` to `super(...)` | https://github.com/blamechris/chroxy/issues/3778 |

### Architecture Notes

The class hierarchy here is `BaseSession` → `JsonlSubprocessSession` → `{CodexSession, GeminiSession}`. `SdkSession` and `CliSession` extend `BaseSession` directly, so they're a two-layer forward; the JSONL-subprocess providers are three layers, and every layer that uses explicit destructure has to list every BaseSession opt it wants to honour. The lesson from `feedback_jsonl_subprocess_middle_layer.md` applies recursively — when a new BaseSession opt is added, every intermediate AND leaf class needs an audit. This PR fixes the middle; #3778 closes the loop at the leaves.

A longer-term option (out of scope for this PR) would be to use a rest pattern in middle/leaf classes — `constructor({ ...opts } = {})` then `super(opts)` after picking off the locally-consumed keys — which would prevent this class of bug entirely. Not suggesting it for this PR since the existing pattern is consistent across all providers and that refactor has its own readability/ergonomics tradeoffs.

### Verdict

**Comment / Approve-with-followup.** The change itself is correct, minimal, and well-tested for the layer it targets. The middle-layer fix is genuinely needed for #3755. The only reason this isn't a clean "Approve" is the leaf-class gap (#3778) — but that's a separate concern from this PR's stated scope, and the dormant-bug rationale (no inactivity timer in Codex/Gemini today) applies equally there. Safe to merge once `Server Tests` finishes green; please prioritise #3778 to actually deliver the end-to-end propagation.

PR #3779, opened 2026-05-11
refactor(protocol): export MAX_SANE_DURATION_MS, import in tests (#3774)

blamechris:
## Summary

Closes #3774. Tiny follow-up to #3773: makes `MAX_SANE_DURATION_MS` a named export of `packages/protocol/src/schemas/server.ts` instead of a module-private constant, and imports it from the three boundary tests in `packages/protocol/tests/schemas.test.js`.

**Why it matters:** if the ceiling ever changes (e.g. dropped to 12h based on operational data), tests would silently keep passing — `MAX-passes / MAX+1-rejects` succeeds against any value as long as both the schema and the inline test constant move together. Inline drift = test no longer guards the actual boundary.

## Change

```diff
- const MAX_SANE_DURATION_MS = 24 * 60 * 60 * 1000
+ export const MAX_SANE_DURATION_MS = 24 * 60 * 60 * 1000
```

Three test sites now use `import { MAX_SANE_DURATION_MS: MAX }` instead of redeclaring it locally.

## Test plan

- [x] 119/119 protocol tests pass
- [x] `npm run build` regenerates `dist/schemas/server.{js,d.ts}` with the new export

Closes #3774

### DIFF
--- a/packages/protocol/dist/schemas/server.d.ts
+++ b/packages/protocol/dist/schemas/server.d.ts
@@ -5,6 +5,16 @@
  * across server, app, and dashboard.
  */
 import { z } from 'zod';
+/**
+ * Sanity ceiling for any ms-typed numeric field (#3768).
+ *
+ * 24 h is well past every legitimate session-timeout / restart-eta /
+ * permission TTL we emit today, and tight enough that an env-var typo
+ * (`CHROXY_RESULT_TIMEOUT_MS=999999999999999`) gets rejected at the
+ * schema boundary instead of corrupting `Date.now() + ms` arithmetic
+ * on the client.
+ */
+export declare const MAX_SANE_DURATION_MS: number;
 export declare const ServerAuthOkSchema: z.ZodObject<{
     type: z.ZodLiteral<"auth_ok">;
     clientId: z.ZodString;
@@ -140,6 +150,7 @@ export declare const ServerPermissionRequestSchema: z.ZodObject<{
     description: z.ZodOptional<z.ZodString>;
     input: z.ZodAny;
     remainingMs: z.ZodOptional<z.ZodNumber>;
+    sessionId: z.ZodOptional<z.ZodString>;
 }, z.core.$strip>;
 export declare const ServerUserQuestionSchema: z.ZodObject<{
     type: z.ZodLiteral<"user_question">;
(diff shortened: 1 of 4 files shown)


blamechris:
## Chroxy Inspector Review — PR #3779

**Verdict:** Approve

Tiny, well-scoped anti-drift refactor. Closes #3774 cleanly with no behavior change.

### Correctness
- Module-private `const` to `export const` is non-breaking — verified zero pre-existing importers of `MAX_SANE_DURATION_MS` outside `server.ts` itself (`grep -rn` across `src/` and `tests/`). The only callers remain the three `.max(MAX_SANE_DURATION_MS)` schema sites at lines 55, 192, 536.
- Because `packages/protocol/src/schemas/index.ts` does `export * from './server.ts'` and `src/index.ts` does `export * from './schemas/index.ts'`, the constant is automatically reachable from the `@chroxy/protocol` package root — no additional plumbing required if a future external consumer wants it. Good.

### Tests
- Local run: 119/119 protocol tests pass, matching the PR description.
- The three boundary tests still exercise the actual schema ceiling (MAX passes, MAX+1 rejects). The whole point of #3774 was preventing silent drift between schema and tests — that hazard is now eliminated.

### Style: `import { MAX_SANE_DURATION_MS: MAX }` rebinding
Minimal-churn approach — the local `MAX` alias was always just for readability in test bodies (avoiding three repetitions of the long name per case). Keeping that alias at the import site is fine; the new tests already used `MAX` throughout and rewriting them to use the long name everywhere would add noise without clarity. Not an anti-pattern at this scale.

### dist/ regeneration
Verified: `dist/schemas/server.js` line 17 and `dist/schemas/server.d.ts` line 17 both reflect the new export. The PR diff also shows two unrelated dist changes (`sessionId` on `ServerPermissionRequest`, `crash` on `ServerShutdown.reason` enum) that appear to be pre-existing dist drift caught by this rebuild — worth a quick sanity check that those match the current `src/schemas/server.ts` source (which they do, given the build was run from current main). Not a defect, just noting.

### Scope discipline
The PR description correctly notes this is plumbing within the protocol package's own tests; not exporting from `index.ts` explicitly is correct — `export *` already handles it transparently. No scope creep.

### Risk
Near zero. Pure refactor, all tests green, no public API removal.

### No deferred items
Nothing worth filing as a follow-up.

Issue #3760, opened 2026-05-11
Broadcast server resultTimeoutMs to clients so ActivityIndicator warns at the right moment


blamechris:
## Context

Identified during review of PR #3759 (#3758 ActivityIndicator).

## Description

The new \`ActivityIndicator\` component on both clients computes the "approaching timeout (Xs left)" warning against a hardcoded \`REFERENCE_TIMEOUT_MS = 20 * 60 * 1000\` — the new \`BaseSession.DEFAULT_RESULT_TIMEOUT_MS\` from #3754.

That works for the default case but is incorrect for any operator who tunes \`config.resultTimeoutMs\` / \`CHROXY_RESULT_TIMEOUT_MS\`. If they set 60 min, the indicator goes red at the wrong moment.

## Approach

Add \`resultTimeoutMs\` to the \`server_status\` broadcast payload (protocolVersion >= 2). Both clients store it on connection state. \`ActivityIndicator\` reads from the store, falls back to the hardcoded reference when missing (older servers).

## Acceptance Criteria

- [ ] Server emits \`resultTimeoutMs\` in \`server_status\`
- [ ] Both clients store it on connection state
- [ ] \`ActivityIndicator\` uses the connection-level value when present, falls back to 20 min otherwise
- [ ] Unit test for the fallback path

PR #3793, opened 2026-05-11
test(app): mirror dashboard resultTimeoutMs parse tests in auth-ok handler (#3764)

### LINKED ISSUE #3764: Mirror dashboard auth_ok resultTimeoutMs parse tests in app auth-ok-handler.test.ts
## Context

Identified during review of PR #3763.

PR #3763 added three `resultTimeoutMs` parse-and-store tests to `packages/dashboard/src/store/auth-ok-handler.test.ts`:

- stores `serverResultTimeoutMs` when present
- leaves it null when older server omits the field
- ignores malformed values (`0`, `-1`, strings, `null`)

The parallel app handler in `packages/app/src/store/message-handler.ts` has the exact same `typeof === 'number' && > 0 ? value : null` parse logic, but the app's `packages/app/__tests__/auth-ok-handler.test.ts` did not receive equivalent test coverage. The app's test file already has symmetric coverage for sibling fields (e.g. `protocolVersion` happy/sad paths), so this is an asymmetry worth closing.

## Acceptance Criteria

- [ ] Add three tests under the existing `server capabilities` describe block (or a new `result timeout` block) mirroring the dashboard tests
- [ ] Verify `mockSetServerInfo` receives the parsed `serverResultTimeoutMs` (or `null` for malformed/missing inputs)

## References

- PR #3763 — feature implementation
- Issue #3760 — original request

blamechris:
## Summary

Closes #3764.

Mirrors the three `resultTimeoutMs` parse-and-store tests added to `packages/dashboard/src/store/auth-ok-handler.test.ts` in PR #3763 into the app's `packages/app/__tests__/auth-ok-handler.test.ts`. The app handler at `message-handler.ts:927` has the same `Number.isFinite + > 0` guard as the dashboard handler; this PR closes the coverage asymmetry.

## Tests added

In the existing `server capabilities` describe block:

- `stores server resultTimeoutMs when present` — happy path, asserts `mockSetServerInfo` receives the parsed value
- `leaves serverResultTimeoutMs null when older server omits the field` — backward-compat with older servers
- `ignores malformed resultTimeoutMs values (non-positive, non-finite, or non-number)` — loops over `[0, -1, NaN, Infinity, -Infinity, 'twenty minutes', null]` and asserts `null` is stored

The malformed-value loop calls `jest.clearAllMocks()` per iteration so each case asserts independently.

## Test plan

- [x] `npx jest __tests__/auth-ok-handler.test.ts` — 29 passed (was 26)
- [x] Coverage assertions match the dashboard's pattern (`toHaveBeenCalledWith(expect.objectContaining(...))`)

### DIFF
--- a/packages/app/__tests__/auth-ok-handler.test.ts
+++ b/packages/app/__tests__/auth-ok-handler.test.ts
@@ -323,6 +323,40 @@ describe('auth_ok handler', () => {
       const state = store.getState();
       expect(state.webFeatures).toEqual({ available: false, remote: false, teleport: false });
     });
+
+    // #3760: server now broadcasts its effective inactivity timeout so the
+    // ActivityIndicator can render its "approaching timeout" warning against
+    // the real configured value instead of a hardcoded 20-min reference.
+    it('stores server resultTimeoutMs when present', () => {
+      const ctx = { url: 'wss://t', token: 'tok', socket: mockSocket, isReconnect: false, silent: false };
+      handleMessage(createAuthOkMessage({ resultTimeoutMs: 45 * 60 * 1000 }), ctx as any);
+
+      expect(mockSetServerInfo).toHaveBeenCalledWith(expect.objectContaining({
+        serverResultTimeoutMs: 45 * 60 * 1000,
+      }));
+    });
+
+    it('leaves serverResultTimeoutMs null when older server omits the field', () => {
+      const ctx = { url: 'wss://t', token: 'tok', socket: mockSocket, isReconnect: false, silent: false };
+      handleMessage(createAuthOkMessage(), ctx as any);
+
+      expect(mockSetServerInfo).toHaveBeenCalledWith(expect.objectContaining({
+        serverResultTimeoutMs: null,
+      }));
+    });
+
+    it('ignores malformed resultTimeoutMs values (non-positive, non-finite, or non-number)', () => {
+      // NaN/Infinity are explicitly rejected to mirror the server's
+      // Number.isFinite guard so the client never stores an unusable timeout.
+      for (const bad of [0, -1, NaN, Infinity, -Infinity, 'twenty minutes', null]) {
+        jest.clearAllMocks();
+        const ctx = { url: 'wss://t', token: 'tok', socket: mockSocket, isReconnect: false, silent: false };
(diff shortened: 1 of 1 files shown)


blamechris:
## Code Review: PR #3793

### Summary
Test-only PR mirroring the dashboard's three `resultTimeoutMs` parse-and-store tests into the app, closing the coverage asymmetry from #3764. +34 lines, no production change.

### Strengths
- Exact symmetry with `packages/dashboard/src/store/auth-ok-handler.test.ts:159-181` — same test names, same `#3760` framing comment, same malformed-value list `[0, -1, NaN, Infinity, -Infinity, 'twenty minutes', null]`, same `Number.isFinite`-guard rationale.
- Mock-assertion translation is sound. Dashboard asserts against the real store via `store.getState().serverResultTimeoutMs`; app correctly asserts against `mockSetServerInfo` because the app routes lifecycle writes through a mocked lifecycle store (line 78/87). The `expect.objectContaining({ serverResultTimeoutMs: ... })` pattern matches existing precedent in the same file (lines 213, 264, 274, 282, 290).
- `jest.clearAllMocks()` inside the malformed-values loop is correct and necessary — `mockSetServerInfo` accumulates calls across iterations, and without the per-iteration reset a stale prior call could satisfy `toHaveBeenCalledWith` on a later (broken) iteration. The dashboard version doesn't need this because it inspects store state, not call history. Pattern matches the existing intra-test reset at lines 278/286 for `protocolVersion`.
- Production guard at `packages/app/src/store/message-handler.ts:927-932` (`typeof === 'number' && Number.isFinite && > 0`) covers exactly the surface the tests exercise.
- Style consistent: semicolons, single quotes, `ctx as any` cast match neighbouring tests.

### Issues Found

#### Critical (Must Fix)
_None_

_None_

#### Nitpicks (Optional)
_None_

_None — no deferred items._

### Architecture Notes
Continues the dashboard/app parity story: PR #3763 added the dashboard tests, this PR closes the matching app gap. Both clients now share guard semantics and test surface for the server-advertised inactivity timeout (#3760), so ActivityIndicator behaviour stays consistent across the two surfaces and a future regression in either handler is caught.

### Verdict

PR #3823, opened 2026-05-12
fix(desktop): strip Bare-runtime prebuilds from server bundle

blamechris:
## Summary

Server bundle was shipping ~3 MB of `.bare` native binaries (`bare-fs`, `bare-url`, `bare-os`, …) that **Node.js never executes**, causing Apple notarization to reject the macOS Tauri build.

## Root cause

```
@chroxy/server → @kubernetes/client-node → tar-fs@3 → bare-fs@4 → bare-url, bare-os, ...
```

`bare-fs` is the Holepunch **Bare runtime** filesystem package. Under the Bare runtime it loads its `.bare` prebuilt natives; under Node.js it... doesn't load at all — `tar-fs` checks the runtime and uses Node's native `fs` directly. Verified by tracing `Module._resolveFilename` during a full server init: zero `bare-*` requires.

## Why it's blocking notarization

From run [25707346237](https://github.com/blamechris/chroxy/actions/runs/25707346237):

```
"path": "Chroxy.zip/Chroxy.app/Contents/Resources/server/
        node_modules/bare-url/prebuilds/darwin-arm64/bare-url.bare",
"message": "The binary is not signed with a valid Developer ID certificate.",
"architecture": "arm64"
```

Apple notarization requires every embedded native binary to be signed with the Developer ID cert + hardened runtime + secure timestamp. Tauri's bundler signs the main app binary but not transitively every `.bare`/`.node` inside `node_modules`.

## Fix

One-line addition to `bundle-server.sh` after `npm install`:

```bash
find "$STAGING/node_modules" -type d -name prebuilds -path "*/bare-*/prebuilds" -prune -exec rm -rf {} +
```

The `bare-*` package metadata stays (npm dependency graph is intact); only the unused native binaries are removed.

## Verification (local)

| Test | Result |
|------|--------|
| `find … -name "*.bare"` after script run | 0 matches |
| `find … -name prebuilds -path "*/bare-*/*"` | 0 matches |
| `require('@kubernetes/client-node')` | ✅ exports OK |
| `tar-fs.pack(dir)` returns Pack stream | ✅ |
| `node --check src/server-cli.js` | ✅ parses |
| `import('./src/ws-server.js')` | ✅ exports OK |
| Bundle size | 306 MB → 303 MB (-3 MB) |

## Test plan

- [x] Local bundle smoke tests (above)
- [ ] CI: dispatch `release.yml` after merge; confirm `Desktop (macOS)` reaches and passes notarization

## Why not deep-sign instead?

Deep-signing every `.bare` binary would also work, but it would still ship 3 MB of dead code through a 5+ minute Apple-server roundtrip for each release. Stripping is the right architectural fix — we shouldn't be shipping binaries we never execute.

### DIFF
--- a/packages/desktop/scripts/bundle-server.sh
+++ b/packages/desktop/scripts/bundle-server.sh
@@ -65,6 +65,18 @@ require('fs').writeFileSync('package.json', JSON.stringify(pkg, null, 2) + '\n')
 echo "[bundle-server] Installing production dependencies..."
 npm install --omit=dev --no-audit --no-fund 2>&1
 
+# Prune Bare-runtime prebuilds.
+#
+# `bare-fs`, `bare-url`, `bare-os` and friends ship `.bare` native binaries for
+# the Holepunch Bare runtime. They arrive as transitive deps of
+# `tar-fs@3 → bare-fs`, pulled in by `@kubernetes/client-node`. When the server
+# runs on Node.js (always, for us), `bare-fs` is never required — `tar-fs`
+# uses Node's native `fs` directly. The prebuilds are pure dead weight (~3 MB)
+# AND they fail Apple notarization because they're unsigned native binaries
+# embedded inside the Tauri app bundle.
+echo "[bundle-server] Pruning Bare-runtime prebuilds (unused under Node.js)..."
+find "$STAGING/node_modules" -type d -name prebuilds -path "*/bare-*/prebuilds" -prune -exec rm -rf {} +
+
 # Copy workspace packages AFTER npm install (npm wipes node_modules during install).
 # Preserve the dist/ directory structure so "main": "./dist/index.js" resolves correctly.
 PROTOCOL_DIR="$REPO_ROOT/packages/protocol"


blamechris:
## Code Review: PR #3823

### Summary
Single-line `find` added after `npm install` in `bundle-server.sh` to strip Bare-runtime `.bare` prebuilds (~3 MB) that block Apple notarization. The fix is correct, minimal, and well-targeted; verification confirms the pattern is precise and the prune is safe under Node.js. Approving.

### Strengths
- **Root cause is correctly diagnosed.** `tar-fs@3` uses the Node `package.json#imports` map with `{"bare": "bare-fs", "default": "fs"}`, so the `default` condition resolves to Node's built-in `fs` under Node.js — `bare-fs` is a pure-Bare optional dep. Verified locally: `require('bare-fs')` under Node throws `Bare is not defined` at module load (it calls `Bare.platform` immediately and `binding.js` is `module.exports = require.addon()` — a Bare-only API). If anything in our server ever did pull it in, the failure would be loud and immediate, not silent. So the prune cannot mask a latent bug.
- **`find` pattern is precise.** Tested on both `bfs` (local) and real BSD `find` (`/usr/bin/find` on macOS, which is what the `macos-latest` CI runner uses) against a fixture covering `bare-fs/prebuilds`, `bare-url/prebuilds`, `some-bare-runtime/prebuilds`, `tar-fs/prebuilds`, `deepbare/sub/bare-fs/prebuilds`, and `only-prebuilds/prebuilds`. Only `bare-*/prebuilds` directories — including the nested-deduped case (`deepbare/sub/bare-fs/prebuilds`) — were matched. `some-bare-runtime/prebuilds` was correctly skipped (the leading slash + literal `bare-` anchor prevents substring matches). `tar-fs/prebuilds` was correctly skipped.
- **Idempotent.** Re-running `find ... -prune -exec rm -rf {} +` on the already-pruned bundle returns 0 and is a no-op. Verified.
- **`-prune` + `-exec rm -rf {} +` is the right combination.** Once `find` matches `…/bare-fs/prebuilds`, `-prune` stops descent into it before `rm -rf` removes it — so `find` never tries to revisit a vanished subtree. (Without `-prune`, `find` would have descended into the dir before `rm` ran, generating "No such file or directory" warnings on `set -e`.)
- **No post-install scripts.** Spot-checked `package.json` for `bare-fs`, `bare-url`, `bare-os`, `bare-events`, `bare-path`, `bare-stream` — none have `postinstall` / `preinstall` / `install` scripts that could resurrect prebuilds. The only `scripts` entries are `format`, `lint`, `test`. Safe across re-installs.
- **No nested `bare-fs` to worry about.** Scanned `server-bundle/node_modules` for any `**/node_modules/bare-*` — npm 10 deduplicates all `bare-*` to the top level. The pattern's recursive nature would catch nested copies anyway (matched the `deepbare/sub/bare-fs/prebuilds` fixture).
- **No other native binaries.** `find … -name "*.node" -o -name "*.dylib" -o -name "*.so" -o -name "*.bare"` against the freshly pruned bundle returns zero matches. The PR's claim that `.bare` are the only problematic binaries holds.
- **Smoke tests post-prune.** `require('@kubernetes/client-node')` and `require('tar-fs')` both load successfully under Node 22 against the pruned bundle.

### Issues Found

#### Critical (Must Fix)
_None._

| `packages/desktop/scripts/bundle-server.sh` | 78 | Optionally print what was pruned for build-log visibility, e.g. `find … -prune -print -exec rm -rf {} +` (BSD/GNU compatible; multi-action). Or capture count: `PRUNED=$(find … -print | tee >(xargs -r rm -rf) \| wc -l); echo "[bundle-server] Pruned $PRUNED bare-* prebuilds dirs"`. | If `bare-*` ever stops shipping prebuilds upstream (or our deps change), a silent 0-match would let dead-code shipping reappear unnoticed. A visible count in the build log makes drift detectable. |
| `packages/desktop/scripts/bundle-server.sh` | 78 | Add a defensive size assertion after the prune, e.g. `! find "$STAGING/node_modules" -name "*.bare" -print -quit \| grep -q .` (fail the build if any `.bare` file remains under `node_modules`). | Defense in depth: if a future dep ships `.bare` outside a `bare-*/prebuilds` path (unlikely but possible), notarization will fail again silently three weeks later. A build-time guard surfaces it immediately. |

#### Nitpicks (Optional)
- Consider following up with a tar-fs upstream issue requesting `bare-fs` move from `optionalDependencies` to a runtime-conditional install (or document the prune step for Node consumers). Not blocking — the workaround is fine.
- The inline comment is good but doesn't link to the failing run. A `# See https://github.com/blamechris/chroxy/actions/runs/25707346237 for the notarization failure that prompted this.` line would make the rationale grep-able from the script.
- One pre-existing CI flake on this PR (`Server Tests`: 1/4785 failed in `WsServer _historyCtx.resultTimeoutMs late-binds to config (#3766)`) — unrelated to this change. Worth a separate flake-tracking issue if not already known.

### Portability / correctness gotchas (none blocking)
- The `-path` glob `*/bare-*/prebuilds` is POSIX-portable: both BSD `find` (macOS `macos-latest` runner) and GNU `find` (Linux dev box) interpret `-path` and shell globbing the same way here. Verified directly with `/usr/bin/find` on macOS.
- `-exec rm -rf {} +` is POSIX-2024 and supported by both BSD and GNU `find`. No quirks observed.
- If `$STAGING/node_modules` somehow doesn't exist (i.e., `npm install` silently produced nothing), `find` exits 1 and `set -euo pipefail` fails the build. Correct behavior — `npm install` failure should not be papered over.

### Verdict
- [x] **Approve**

Ship it. The fix is correct, minimal, well-commented, and unblocks 1.0 notarization. The two suggestions above can land in a follow-up; not blocking.

PR #3885, opened 2026-05-14
fix(scripts): patch Cargo.lock directly on version bumps

blamechris:
Closes #3864

## Summary

- `scripts/bump-version.sh` previously called `cargo generate-lockfile 2>/dev/null` to sync `Cargo.lock` with the new `Cargo.toml` version. The silent stderr suppression meant this step no-opped on any bump-running machine without `cargo` on `PATH`, leaving `Cargo.toml` at the new version but `Cargo.lock` at the old one.
- CI's `cargo test --locked` then refused to regenerate the lockfile and failed with the `cannot update the lock file ... because --locked was passed` error — this is exactly how v0.8.3 shipped broken (fixed for that release by hand-editing the `chroxy-desktop` stanza in `43889e55e`).
- Replace the cargo invocation with an awk-based direct rewrite of the `version` line inside the `chroxy-desktop` package stanza. Mirrors the script's existing pattern for JSON lockfiles and `Info.plist`: no toolchain dependency, minimal diff, hard-fail verification step.

## Test plan

- [x] `./scripts/bump-version.sh 9.9.9` → Cargo.lock `chroxy-desktop` stanza reads `version = "9.9.9"`, `cargo check --locked` passes.
- [x] Idempotent `./scripts/bump-version.sh 0.8.3` → no Cargo.lock diff.
- [x] Reproduced the pre-fix bug locally by stashing the patch and running the old script: `Cargo.toml` went to 9.9.9 but `Cargo.lock` stayed at 0.8.3, `cargo check --locked` failed with the exact error from the issue.
- [x] `shellcheck scripts/bump-version.sh` clean.

### DIFF
--- a/scripts/bump-version.sh
+++ b/scripts/bump-version.sh
@@ -219,12 +219,16 @@ awk -v new="$NEW_VERSION" '
     print
     next
   }
-  in_pkg && /^version = "/ {
+  in_pkg && /^version = "[^"]*"$/ {
     sub(/^version = "[^"]*"/, "version = \"" new "\"")
     in_pkg = 0
     print
     next
   }
+  # Defensive: a malformed chroxy-desktop stanza missing its version line
+  # would otherwise leave in_pkg=1 set into the next package and rewrite
+  # the wrong stanzas version. Resetting on every new package header
+  # bounds the in_pkg window strictly to the current stanza.
   /^\[\[package\]\]/ { in_pkg = 0 }
   { print }
 ' "$CARGO_LOCK" > "$CARGO_LOCK.tmp" && mv "$CARGO_LOCK.tmp" "$CARGO_LOCK"
--- a/scripts/bump-version.sh
+++ b/scripts/bump-version.sh
@@ -20,6 +20,7 @@ STORE_CORE_PKG="$ROOT/packages/store-core/package.json"
 DASHBOARD_PKG="$ROOT/packages/dashboard/package.json"
 TAURI_CONF="$ROOT/packages/desktop/src-tauri/tauri.conf.json"
 CARGO_TOML="$ROOT/packages/desktop/src-tauri/Cargo.toml"
+CARGO_LOCK="$ROOT/packages/desktop/src-tauri/Cargo.lock"
 ROOT_LOCK="$ROOT/package-lock.json"
 SERVER_LOCK="$ROOT/packages/server/package-lock.json"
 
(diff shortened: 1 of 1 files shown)


blamechris:
## Code Review: PR #3885

### Summary

Targeted fix for #3864. Replaces the silently-skippable `cargo generate-lockfile 2>/dev/null` with an awk-based in-place rewrite of the `chroxy-desktop` stanza in `Cargo.lock`, plus a hard-fail verification step. The change is small, well-commented, mirrors the existing `Info.plist` awk pattern, and shellcheck-clean.

I locally verified:

- Awk transform correctly rewrites only the `chroxy-desktop` stanza's `version` line (synthetic fixture with a `bar` package also at `0.8.3` left untouched).
- Idempotency: running with the current version produces a byte-identical file (`diff` empty).
- Missing-stanza failure mode: if `name = "chroxy-desktop"` isn't present, the rewrite is a no-op and the `RS="\n\n"` verify block hard-fails with the intended error.
- Verification still matches even when `chroxy-desktop` is the file's last stanza without trailing blank line (awk treats EOF as a record terminator).

### Strengths

- Comment block ties the change directly to #3864 and references the 43889e55e hand-fix, so the next maintainer to touch this won't be tempted to "simplify" back to `cargo generate-lockfile`.
- The verify step is properly scoped: `awk -v RS="\n\n"` ensures `name =` and `version =` must appear in the same stanza, defeating the obvious false-pass where a transitive dep happens to share `$NEW_VERSION`.
- Pattern is consistent with the script's two other awk transforms (`Info.plist`, `Cargo.toml` version) — no new abstractions introduced for a 40-line fix.
- Pre-flight `[ ! -f "$CARGO_LOCK" ]` guard fails loudly if the lockfile moves or is gitignored on a contributor's machine.
- All four items in the PR test plan are realistic and verifiable; the "reproduced the pre-fix bug locally" step is the right level of rigor for a release-gate script.

### Issues Found

#### Critical (Must Fix)

None.

| `scripts/bump-version.sh` | 230 | Consider `trap 'rm -f \"$CARGO_LOCK.tmp\"' EXIT` near this block (or before awk) | If the awk pipeline ever fails mid-write (disk full, permissions), `set -e` aborts before the `mv` and leaves an orphan `$CARGO_LOCK.tmp`. Low likelihood, but the existing `Info.plist` block has the same pattern — would be worth a follow-up to add cleanup to both rather than just this one. |
| `scripts/bump-version.sh` | 222-227 | Tighten the inner regex slightly: `in_pkg && /^version = "[^"]*"$/` | Currently `/^version = "/` would match a malformed stanza line like `version = "foo` (unterminated). `sub` would then no-op and the verify would catch it, but anchoring the closing quote makes the intent explicit and matches the `Cargo.toml` version regex on line 120. |
| `scripts/bump-version.sh` | 228 | The fallback `/^\[\[package\]\]/ { in_pkg = 0 }` is defensive against a `chroxy-desktop` stanza with no `version` line | Worth a one-line comment explaining what it guards against (malformed stanza), since the inner block already clears `in_pkg`. Otherwise a future reader may delete it as dead. |

#### Nitpicks (Optional)

- The trailing `echo "  $CARGO_LOCK"` (line 252) is good, but it's the full path while the sibling line says `Cargo.lock` previously — the new form is more useful for grep-from-terminal, just noting the cosmetic shift.
- Comment on line 233 (`awk -v RS keeps the per-stanza check scoped`) is great context; consider also noting this same trick would catch the symmetric bug where another package gets renamed *to* `chroxy-desktop` (paranoid, but the comment is already there).

None. Everything above is either inside this PR's scope or trivial enough that filing an issue would be over-engineering.

### Verdict

- [x] **Approve** — ship it. The fix is correct, the verification gate is sound, the comments will keep the next maintainer honest, and the test plan covered the right cases. The three suggestions above are polish, not blockers.

blamechris:
## Review Comments Addressed

Agent-review verdict: **Approve** (0 critical, 3 suggestions, 2 nitpicks). Copilot left a summary-only review with no inline comments.

| # | Source | Comment | Outcome | Reference |
|---|--------|---------|---------|-----------|
| 1 | agent-review | Tighten inner regex to `^version = "[^"]*"$` so malformed unterminated `version = "foo` is not silently accepted | FIX | `c6ba93435` |
| 2 | agent-review | Add comment explaining the defensive `/^\[\[package\]\]/` reset | FIX | `c6ba93435` |
| 3 | agent-review | Trap-cleanup for orphan `.tmp` files on awk failure | FOLLOW-UP | #3886 |
| 4 | agent-review | Nitpick: full vs short path in echo summary | FALSE POSITIVE | Full path is more grep-friendly; cosmetic — agent flagged as just-noting |
| 5 | agent-review | Nitpick: extend RS-scoped verify comment for paranoid rename case | FALSE POSITIVE | Existing comment already covers per-stanza scoping; extra paranoia past diminishing returns |
| 6 | Copilot | Summary review only, no inline comments | FALSE POSITIVE | No comments to address |

**Total:** 5 review items + 1 summary review
- Fixed: 2 (single fold-in commit `c6ba93435`)
- False positives: 3 (cosmetic / no inline comments)
- Follow-up issues created: 1 (#3886)

PR #3976, opened 2026-05-16
feat(server): CHROXY_CODEX_SANDBOX env overrides default sandbox

blamechris:
## Summary

Adds an env-var override for the Codex provider's default sandbox mode. Closes #3847.

- New `resolveCodexSandbox()` reads `CHROXY_CODEX_SANDBOX` at call time and validates against the canonical `read-only` / `workspace-write` / `danger-full-access` set
- `buildCodexArgs()` now routes both first-turn and resume forms through the resolver; the `--sandbox`-before-`resume` ordering invariant from #3865 is preserved
- Unknown values warn to stderr and fall back to `workspace-write` rather than refusing to start — wrong failure mode for a stopgap env knob
- Default behaviour is unchanged for operators who do not set the env var (still `workspace-write`)

Stopgap until the per-session sandbox selector in #3837 lands.

## Test plan

- [x] 16 new unit tests cover: env unset / empty / each of the three valid modes / surrounding whitespace / invalid value with warning / case-sensitivity / first-turn argv / resume argv with ordering invariant / call-time-vs-module-load semantics
- [x] All 96 tests in `codex-session.test.js` pass (`npm test` in `packages/server`)
- [x] `npm run lint` clean (no new warnings, only the 3 pre-existing ones)
- [x] Docs updated in `docs/providers.md` — new "CHROXY_CODEX_SANDBOX env override" subsection with the mode table and the existing planned-in-#3847 reference upgraded to landed-in-#3847

Closes #3847

### DIFF
--- a/docs/providers.md
+++ b/docs/providers.md
@@ -167,7 +167,27 @@ Items 2–4 mean "the workspace" is **broader than the chroxy session
 cwd**. Chroxy's session-trust gate only constrains item 1. If you need
 a hard read-only session, the per-session sandbox selector tracked in
 #3837 is the user-facing override — until that lands, set
-`CHROXY_CODEX_SANDBOX=read-only` server-wide (planned in #3847).
+`CHROXY_CODEX_SANDBOX=read-only` server-wide (#3847).
+
+### `CHROXY_CODEX_SANDBOX` env override (#3847)
+
+Operators on multi-tenant or shared-dev hosts may want Codex to start
+more restrictive than the `workspace-write` default. Set
+`CHROXY_CODEX_SANDBOX` on the server process to one of:
+
+| Value | `--sandbox` passed to `codex exec` | Effect |
+|---|---|---|
+| _unset_ | `workspace-write` | Default. Codex may write to the four surfaces described above. |
+| `read-only` | `read-only` | Codex may read but not write or execute side-effectful commands. |
+| `workspace-write` | `workspace-write` | Same as the default — useful for making the policy explicit in a process manager unit file. |
+| `danger-full-access` | `danger-full-access` | Codex has no sandbox at all. Only set on a host you fully trust. |
+
+Unknown values (e.g. a typo like `readonly`) log a warning to stderr and
+fall back to `workspace-write` rather than refusing to start the server.
+The override is read on each turn, so changing the env applies on the
+*next* message — but already-running `codex exec` subprocesses are
+unaffected. This is a stopgap until the per-session sandbox selector
+(#3837) lands.
 
 Chroxy also unconditionally passes `--skip-git-repo-check` so Codex
 will accept non-git cwds (#3834). This is correct today because
--- a/packages/server/src/codex-session.js
+++ b/packages/server/src/codex-session.js
@@ -55,6 +55,55 @@ const BINARY_CANDIDATES = [
 const CODEX = resolveBinary('codex', BINARY_CANDIDATES)
 
 /**
+ * Codex CLI sandbox modes. Source: `codex exec --sandbox <MODE>` accepts
+ * exactly these three values (verified against codex-cli 0.128.0). Exported
+ * so tests and consumers can pin the canonical list without re-declaring it.
+ */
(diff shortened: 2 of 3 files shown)


blamechris:
## Code Review: PR #3976

### Summary

Implements the `CHROXY_CODEX_SANDBOX` env override proposed in #3847 as a stopgap until the per-session sandbox selector (#3837) lands. Adds three exports (`CODEX_SANDBOX_MODES`, `CODEX_DEFAULT_SANDBOX`, `resolveCodexSandbox`), threads them through `buildCodexArgs()` for both first-turn and resume forms, and updates `docs/providers.md` with an operator-facing table. 16 new unit tests pin the contract.

### Strengths

- **Reads env at call time, not module-load time** — pinned by a dedicated test. Correct choice: caching would break test harnesses and hot reload, and `_buildArgs()` is invoked per `sendMessage()` anyway so there is no perf cost.
- **Fail-soft on invalid values** — warns and falls back to default rather than throwing. Matches the issue's explicit proposal ("Unknown value → log warning, fall back to `workspace-write`") and is the right failure mode for a stopgap env knob; throwing would refuse to start the whole server.
- **Case-sensitive validation** — refuses `Read-Only` rather than silently coercing. Matches Codex CLI's own behavior and surfaces typos that would otherwise have been silently downgraded to the default.
- **Resume-form ordering invariant preserved** — `--sandbox` still precedes the `resume` subcommand on the resume path, with a dedicated test that pins the ordering. The #3865/#3873 regression class can't slip back in via this change.
- **Whitespace trim** — defensive against the common `docker-compose`/shell-export copy-paste foot-gun.
- **Exports the canonical mode list** — `CODEX_SANDBOX_MODES` and `CODEX_DEFAULT_SANDBOX` are exported so future callers (per-session selector under #3837, doctor preflight, dashboard dropdown) reuse the same source of truth rather than re-declaring it.
- **Docs updated in the same PR** — `docs/providers.md` already referenced "planned in #3847" so the change cleanly upgrades to "(#3847)" plus a new mode table. No drift.
- **Security invariant preserved** — env value passes through an explicit allowlist before reaching argv, no shell, no injection vector. Matches the security comment block on `buildCodexArgs`.

### Issues Found

#### Critical (Must Fix)

None.

| `packages/server/src/codex-session.js` | 98 | Consider warning once-per-process for invalid env values | `resolveCodexSandbox()` runs on every `sendMessage()`. A typo like `CHROXY_CODEX_SANDBOX=readonly` will print a `console.warn` for every turn in every session for the lifetime of the server. A `Set<string>` module-level cache of "values already warned about" would keep the loud-on-first-call behavior while avoiding log spam. Tracked as a follow-up below since it does not block correctness. |

#### Nitpicks (Optional)

None.

| Warn-once-per-process for invalid `CHROXY_CODEX_SANDBOX` to avoid log spam | https://github.com/blamechris/chroxy/issues/3981 | Cosmetic. The current behavior is correct (loud, visible) and only affects operators who misconfigured the env. Worth tracking but not gating this stopgap. |

### Architecture Notes

This change sits cleanly inside `codex-session.js` and does not touch the WS protocol, session manager, or any provider-agnostic layer. The per-session selector (#3837) — when it lands — will need to thread a session-scoped sandbox value through `BaseSession`/`SessionManager` and pass it as a third argument to `buildCodexArgs()`; the exported `CODEX_SANDBOX_MODES` constant is already the right validation surface for that follow-up. The env-var path can coexist with the per-session path (env as server-wide default, session as per-call override) without further refactoring.

The integration test `tests/integration/codex-spawn-argv.integration.test.js` calls `buildCodexArgs()` without setting `CHROXY_CODEX_SANDBOX`, so it relies on the documented default (`workspace-write`) — matches the existing assertion at line 304 and continues to work unchanged. Worth noting for anyone adding a future CI matrix that sets the env globally.

### Verdict

- [x] **Approve** — Ready to merge. Clean implementation, comprehensive tests, docs in sync, no regressions to existing invariants. The one suggestion (warn-once) is captured as a follow-up issue and does not block.

Issue #3899, opened 2026-05-14
Replace inactivity-timeout kill with check-in flow (notification + 'Status update?' button)


blamechris:
## Context

#3884 closed by #3892 — CLI timer is now activity-based and the default window is 30 min. But the underlying UX is still wrong: when the window fires we **force-clear busy state, auto-deny pending permissions, and emit a `result_timeout` error**. The session dies.

That's the wrong model. 30 min of silence from an active agent doesn't mean "the session is broken" — it means "the human probably stepped away and the agent is plugging along on a long-running task (or is genuinely stuck)." Either way, the right move is to **prompt the human**, not nuke the session.

## Proposed UX

Replace the kill with a check-in flow:

1. **Server-side**, at the 30-min activity threshold:
   - Emit a new event type (`inactivity_warning` or `check_in_prompt`) — NOT `error`
   - Keep `_isBusy = true`, do NOT clear `_currentMessageId`, do NOT auto-deny pending permissions
   - Send a push notification via existing `push.js` plumbing — text like *"Agent is still working but has been quiet for 30 min. Tap to check in."*
   - Re-arm the timer for another window (the warning event itself counts as activity for the new window)
2. **Client-side** (mobile + dashboard), on `inactivity_warning`:
   - Non-blocking UI affordance — e.g. a chip above the composer: *"Agent quiet for 30+ min — [Status update?]"*
   - Single-click button sends a prefab `Status update?` message through the normal user-input path
   - The agent's response naturally resets the activity timer
3. **Hard cap (optional)** — if the user never check-ins and silence continues for some larger N (say 2h), still emit the existing error to catch genuinely runaway sessions.

## Open design questions

- [ ] Hard cap at 2h/4h, or no cap (let the user decide via check-ins)?
- [ ] Configurable prefab text? Per-user setting? Per-provider override (Claude vs Codex vs Gemini have different idioms)?
- [ ] Warning event in chat history (durable transcript) or transient UI only?
- [ ] Pending permissions when the warning fires — keep pending, or still auto-deny?
- [ ] Scope: SDK + CLI only, or also Codex / Gemini (which have their own timeout paths)?

## Why this matters now

Post-Reddit launch users running marathon agent ops (batch-merge, tackle-issues, big refactors) hit the 30-min threshold semi-regularly. Killing the session at that moment is the single worst UX outcome — agent has 25 min of unsurfaced work in its head, user just stepped away, comes back to *"Response timed out after 30 minutes"* and lost context.

[...]

PR #3983, opened 2026-05-16
test(server): assert cli-session panic timers are cleared and unref'd (#3966)

blamechris:
## Summary

Addresses #3966 — the panic-button test added in PR #3965 armed long-fuse soft + hard inactivity timers but never released them, so the panic flow leaked live `setTimeout` handles into the next test (and against a stale `messageId`).

**Test changes** (`packages/server/tests/cli-session.test.js`):
- Add a panic-block `afterEach` that drains every tracked session's `_resultTimeout`, `_hardTimeout`, `_interruptTimer`, and `_respawnTimer`. Replaces the misleading "so afterEach doesn't dangle" comment in a describe block that had no afterEach.
- `.unref()` the test-armed soft/hard handles so even a future bug that skips the hook cleanup cannot hold the event loop open past test completion.
- Add a new case `panic kill+respawn clears the armed soft + hard inactivity timers` that spies on `global.clearTimeout` and asserts both `_resultTimeout` and `_hardTimeout` are nulled (and their handles passed to `clearTimeout`) after the panic flow runs.

**Implementation change** (`packages/server/src/cli-session.js`):
- Extend `_killAndRespawn()` to clear `_resultTimeout` / `_hardTimeout` (and reset `_resultTimeoutPaused`) alongside the existing `_interruptTimer` / `_respawnTimer` clears. The dropped in-flight turn has no business firing `_handleHardTimeout` against a now-irrelevant `messageId`, and previously the only thing keeping production safe was the next `_armResultTimeout()` call on the respawned process — under test stubs that call never happens.

## Test plan

- [x] `node --test packages/server/tests/cli-session.test.js` — full file (15 suites) passes, panic suite reports 3/3 (was 2/2)
- [x] `node --test packages/server/tests/base-session.test.js` — 86/86 green
- [x] TDD red→green confirmed: with only the test added and `_killAndRespawn` untouched, the new panic-clear assertion fails (`_resultTimeout` still set after kill+respawn)

Closes #3966


blamechris:
## Code Review: PR #3983

### Summary

Closes the test-cleanup gap flagged in #3966 from PR #3965 review. The panic-button suite armed `_resultTimeout` / `_hardTimeout` with 100s fuses but never released them, and `_killAndRespawn()` only cleared `_interruptTimer` / `_respawnTimer` — so the panic flow leaked live `Timeout` objects into the next test and (in production) left the inactivity pair ticking against a stale `messageId` until the next `_armResultTimeout()` happened to fire on the respawned child. The fix is in the right place: both a test-side belt (afterEach + `.unref()`) and an implementation-side fix in `_killAndRespawn()`.

### Strengths

- **Real TDD**: PR description documents the red→green transition; I reproduced it by reverting only the cli-session.js hunk and confirmed the new "panic kill+respawn clears…" case fails with `_resultTimeout` still pointing at a live `Timeout`. The implementation change is justified by a failing test, not by inspection.
- **Three-layer defence in the test file** — `afterEach` drains every tracked session, individual `setTimeout`s are `.unref()`'d, AND the production code now releases the handles. Any one layer being skipped (deleted hook, future test that forgets `sessions.push`, regression in `_killAndRespawn`) still leaves the other two protecting the suite.
- **Comment placement matches behavior**: the misleading "so afterEach doesn't dangle" comment in the non-auto test was removed once a real `afterEach` exists. Good cleanup of stale narration.
- **The `_killAndRespawn` change semantically matches what the function already does** for `_interruptTimer` / `_respawnTimer` — there's no policy invention, just consistency with the existing timer-clearing block.
- **No protocol surface touched** — pure internal lifecycle hygiene.

### Issues Found

#### Critical (Must Fix)

None.

| `packages/server/src/cli-session.js` | 956–971 (`_killAndRespawn`) | Consider also clearing `_pendingPermissionIds` in the same block | The dropped turn's pending permissions belong to a `messageId` that no longer exists; leaving them in the Set means the next `notifyPermissionResolved(staleId)` is a silent no-op (the ID isn't there) but any new permission registered for the *next* turn re-trips `notifyPermissionPending`'s `size === 1` branch correctly, so this is mostly cosmetic — flagging because `_clearMessageState` *does* clear it and the symmetry would be nicer. Out of scope for this PR; defer if it lands. |
| `packages/server/tests/cli-session.test.js` | 939–942 | Direct `global.clearTimeout` reassignment is functional but slightly heavy — `mock.method(global, 'clearTimeout')` from `node:test` would auto-restore on test teardown | Pure ergonomics; the existing `try/finally` is correct. Not worth changing now. |

#### Nitpicks (Optional)

- The new comment in `_killAndRespawn` calls out "panic-button path" and "`setModel` mid-turn" — accurate, but `setModel` is currently gated by the `_isBusy` check in `super.setModel(model)` (see `setModel ignores change when busy` test), so the mid-turn case for `setModel` actually can't happen unless a future change opens it up. Not wrong, but slightly future-tense; either tighten to panic-only or leave as-is for forward-compat narration.

None worth filing. The "_pendingPermissionIds" symmetry note is too small to track as its own issue — would just produce a 3-line PR.

### Architecture Notes

This fits cleanly inside the existing `BaseSession`/`CliSession` timer ownership model. The hard-cap / soft-warning trio (`_resultTimeoutMs` / `_hardTimeoutMs` from #3899) is per-turn lifecycle state, so it belongs to whichever path tears down the turn — `_clearMessageState` already handles the natural completion path, and `_killAndRespawn` is the destructive panic-button counterpart. Adding the same clears to the latter closes a symmetry gap that pre-existed #3965.

No WS protocol changes. No mobile-facing changes. The panic-button externally-visible contract (mid-turn flip to `auto` drops the turn and SIGTERMs the child) is preserved verbatim — the existing test 1 still asserts every observable bit of that.

### Verdict

CI is green across all 12 checks. The implementation change is minimal, the new test pins the contract, and the test-cleanup story is now belt-and-braces.

PR #3986, opened 2026-05-16
fix(dashboard): evict composer refs when sessions vanish from session_list (#3977)

### LINKED ISSUE #3977: Composer paste-collapse: evict ref entries when server removes a session from session_list (#3800 follow-up)
## Context

Identified during review of PR #3973 (which closes #3800).

PR #3973 wires per-session eviction of `inputDraftsRef` / `pastedTextBlocksRef` / `pastedTextNextIdRef` into the two **user-initiated** session-removal paths: `handleCloseSession` and `handleRestartSession`. That closes the leak the user can directly cause by clicking the close (×) button or the StdinDisabledBanner restart button.

It does **not** cover the third path that removes sessions from `sessions[]`: the **`session_list` broadcast** from the server (`packages/dashboard/src/store/message-handler.ts:1958-2006`). The store calls `set({ sessions: sessionList })`, so any session present locally but absent from the broadcast disappears from the React state without `<App />` knowing — and its three ref entries stay behind for the lifetime of `<App />`.

This was flagged in the original #3800 description (\"and possibly subscribe to the store's session-list updates to evict orphaned entries\") and explicitly deferred from #3973 to keep that PR small and reviewable.

## When can this actually leak?

Any path where the server (or another client) removes a session from the list without `<App />` having clicked the close button locally:

- **Another connected dashboard / app** closes a session — `session_list` broadcast removes it.
- **Server-side supervisor** culls a session (crash, timeout, manual `destroy_session` over CLI).
- **Server cold restart** rebuilds `session_list` from disk; sessions destroyed during the downtime never reappear.
- **Multi-server switching** (`switchServer` flow) — old server's sessions vanish, new server's sessions arrive.

Each of those leaves the pasted-text content for the disappeared sessionIds keyed in the Map forever.

## Suggested fix

Subscribe to `sessions[]` changes inside `<App />` and reconcile the three refs:

```ts
// Reconcile per-session composer refs against the live sessions list.
// Any ref entry whose sessionId is no longer present in `sessions` was
// destroyed by something other than handleCloseSession/handleRestartSession
// (e.g. server-side cull, another client closing, supervisor restart) —
// evict it so the maps don't grow unbounded over a long dashboard session.
useEffect(() => {
  const liveIds = new Set(sessions.map(s => s.sessionId))
  for (const id of inputDraftsRef.current.keys()) {
    if (!liveIds.has(id)) inputDraftsRef.current.delete(id)
  }
  for (const id of pastedTextBlocksRef.current.keys()) {
    if (!liveIds.has(id)) pastedTextBlocksRef.current.delete(id)
  }
  for (const id of pastedTextNextIdRef.current.keys()) {
    if (!liveIds.has(id)) pastedTextNextIdRef.current.delete(id)
  }
}, [sessions])
```

That single effect makes the eviction in `handleCloseSession` and `handleRestartSession` strictly redundant — but they're cheap and they keep the eviction synchronous with the user action, so it's reasonable to leave them as belt-and-braces. Alternatively, delete them once the subscription lands.

## Original review note

> Evict orphaned ref entries when sessions disappear from the `session_list` broadcast (server-side close, another client closing a session, supervisor restart culling stale entries).
>
> The current fix closes the two user-initiated paths; the broadcast-driven path is a strictly larger surface that needs a store subscription, and shipping it alongside this small fix would muddy review.

## Acceptance Criteria

- [ ] `<App />` subscribes to `sessions[]` and evicts ref entries for any sessionId not in the current list.
- [ ] Test: render `<App />` with sessions s1, s2; paste-collapse in s1; replace `sessions` with `[s2]` only (simulating a server-driven removal); verify the next `setActiveSessionId('s1')` (with s1 re-appearing) does NOT rehydrate the paste chip.
- [ ] Test: same flow for `inputDraftsRef` — a draft typed into s1, server-removed, then s1 reappearing must start with an empty draft.
- [ ] Document the reconciliation invariant ("any entry not in `sessions[]` is evicted") near the ref declarations so future per-session refs follow the same pattern.

## Related

- PR #3973 — the user-initiated half of this cleanup.
- #3797 — the PR that introduced the per-session refs.
- #3800 — the parent leak issue (now closed by #3973).

blamechris:
## Summary

Closes #3977. Related to #3973, #3800.

PR #3973 wired per-session composer-ref eviction into the two user-initiated removal paths (`handleCloseSession`, `handleRestartSession`). This closes the third path: server-driven removals via the `session_list` broadcast (another client closes a session, supervisor culls a wedged one, server cold-restart rebuilds `session_list` from disk, multi-server switching). Each of those left the per-session `inputDraftsRef`, `pastedTextBlocksRef`, and `pastedTextNextIdRef` entries keyed by the now-defunct sessionId for the lifetime of `<App />`.

## Changes

- Extract `evictSessionComposerState(sessionId)` helper. The three existing call sites (close, restart, send) now route through it instead of duplicating the three-`.delete` pattern.
- Add a `useEffect` subscribed to the `sessions` array that builds a Set of live sessionIds and evicts any ref entry whose key is no longer present. Runs once per sessions-array change, O(refs) + O(sessions).
- Document the reconciliation invariant near the ref declarations so future per-session refs are added to the helper.

## Test plan

- [x] New tests in `App.test.tsx` cover the broadcast-driven removal path:
  - Paste an oversized block into s1, simulate a `session_list` broadcast that drops s1 (NOT via `handleCloseSession`), then re-present a session with id s1 — the paste chip must not resurrect from the ref.
  - Same shape for the typed draft (verified via the `<textarea>` value on rebind).
  - Negative test: ref entries for sessions that remain in the broadcast are preserved.
- [x] All 1704 existing dashboard tests still pass.
- [x] `tsc --noEmit` clean.
- [ ] Manual verification (optional): open the dashboard with two sessions, paste oversized text into session A, then close session A from another connected client / kill the session server-side. Reconnect session A's id — composer must start clean.

### DIFF
--- a/packages/dashboard/src/App.test.tsx
+++ b/packages/dashboard/src/App.test.tsx
@@ -871,4 +871,140 @@ describe('App', () => {
       expect(screen.queryByTestId('pasted-text-chips')).not.toBeInTheDocument()
     })
   })
+
+  describe('Composer ref reconciliation against session_list (#3977)', () => {
+    // #3977: the server can remove a session from `sessions[]` without the
+    // dashboard having called `handleCloseSession` locally — another client
+    // closes the session, the supervisor culls it, the server cold-restarts
+    // and rebuilds session_list from disk without the dead entries, or the
+    // user switches servers. In every one of those paths the per-session
+    // composer refs (`inputDraftsRef`, `pastedTextBlocksRef`,
+    // `pastedTextNextIdRef`) would leak the dead sessionId's entries for the
+    // lifetime of <App />. The fix subscribes to `sessions[]` and reconciles
+    // the refs whenever an entry disappears from the list. These tests prove
+    // the eviction is observable: a session whose composer had a paste chip,
+    // once removed from `sessions[]` by anything other than the local close
+    // handler, must not "remember" it when its sessionId reappears.
+    function bigText(lines: number, charsPerLine = 100): string {
+      const line = 'x'.repeat(charsPerLine)
+      return Array(lines).fill(line).join('\n')
+    }
+
+    const twoSessions = [
+      {
+        sessionId: 's1', name: 'One', cwd: '/tmp', type: 'cli' as const,
+        hasTerminal: true, model: null, permissionMode: null, isBusy: false,
+        createdAt: 1, conversationId: null, provider: 'claude-sdk', worktree: false,
+      },
+      {
+        sessionId: 's2', name: 'Two', cwd: '/tmp', type: 'cli' as const,
+        hasTerminal: true, model: null, permissionMode: null, isBusy: false,
+        createdAt: 2, conversationId: null, provider: 'claude-sdk', worktree: false,
+      },
+    ]
+
+    it('evicts paste blocks when a session vanishes from session_list (server-driven removal)', () => {
+      stateOverrides = {
+        connectionPhase: 'connected',
+        sessions: twoSessions,
+        activeSessionId: 's1',
+      }
+      const { rerender } = render(<App />)
+
+      // Stage a collapsed paste on s1.
+      const textarea = screen.getByRole('textbox', { name: /message input/i })
+      fireEvent.paste(textarea, {
+        clipboardData: {
+          files: [],
+          items: [],
+          getData: (type: string) => (type === 'text/plain' ? bigText(30) : ''),
+        },
+      })
+      expect(screen.getByTestId('pasted-text-chips')).toBeInTheDocument()
+
+      // Simulate a server-driven `session_list` broadcast that drops s1 —
+      // NOT the local close handler. Active id moves to s2.
+      stateOverrides = {
+        ...stateOverrides,
+        sessions: twoSessions.filter(s => s.sessionId !== 's1'),
+        activeSessionId: 's2',
+      }
+      rerender(<App />)
+      expect(screen.queryByTestId('pasted-text-chips')).not.toBeInTheDocument()
+
+      // Re-present a session with id s1 (the dashboard rebinds to the same
+      // id, or a new session happens to reuse it). If the ref entry leaked,
+      // the stale paste chip would resurrect here — the failure mode #3977
+      // describes for the broadcast-driven removal path.
+      stateOverrides = {
+        ...stateOverrides,
+        sessions: twoSessions,
+        activeSessionId: 's1',
+      }
+      rerender(<App />)
+      expect(screen.queryByTestId('pasted-text-chips')).not.toBeInTheDocument()
+    })
+
+    it('evicts the typed draft when a session vanishes from session_list', () => {
+      stateOverrides = {
+        connectionPhase: 'connected',
+        sessions: twoSessions,
+        activeSessionId: 's1',
+      }
+      const { rerender } = render(<App />)
+
+      // Type a draft into s1.
+      const textarea = screen.getByRole('textbox', { name: /message input/i })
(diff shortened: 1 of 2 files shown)


blamechris:
## Code Review: PR #3986

### Summary

Closes the broadcast-driven half of the per-session composer-ref leak first surfaced in #3800. PR #3973 evicted on the two user-initiated removal paths (`handleCloseSession`, `handleRestartSession`); this adds the `sessions[]` reconciliation effect that catches every other removal path (server-side cull, another client closing, supervisor restart, multi-server switch). Extraction of `evictSessionComposerState(sessionId)` consolidates the eviction logic that was duplicated across three call sites in #3973.

### Strengths

- Clean helper extraction — the three pre-existing `.delete` triples now route through one function, the new reconciliation effect uses the same primitive, and the invariant comment near the ref declarations tells future contributors where to plug in additional per-session refs.
- Effect deps are correct: `[sessions]` only. The helper reads stable refs so it doesn't need to participate in the dep graph, and the comment explains why useCallback wasn't used.
- Test pattern mirrors the existing #3800 tests in the same file — same `bigText` helper, same `twoSessions` fixture shape, same "remove → re-present id → chip must not resurrect" probe. Easy for the next maintainer to extend.
- Negative test ("preserves composer state for sessions that remain in session_list") proves the diff isn't over-eager — without it a future regression that evicts ALL keys on every `sessions` change would still pass the two main tests.
- All three per-session refs (`inputDraftsRef`, `pastedTextBlocksRef`, `pastedTextNextIdRef`) are scanned independently with their own key iterators, so a paste-without-draft or draft-without-paste edge case can't leak past the eviction.

### Issues Found

#### Critical (Must Fix)

None.

| `packages/dashboard/src/App.tsx` | ~993 | `evictSessionComposerState` is captured by `handleCloseSession`/`handleRestartSession`/`handleSend` useCallbacks but is not in their deps lists. Currently safe (the helper only touches stable refs), but a future change that gives the helper non-ref state would silently capture a stale version. Adding an eslint-disable-next-line + comment, or wrapping the helper in `useCallback([])`, would document the intent. | Defensive hardening for future maintainers — not a correctness issue today. |

#### Nitpicks (Optional)

- The three `for...of` loops could be unified into one iteration over `new Set([...inputDraftsRef.current.keys(), ...pastedTextBlocksRef.current.keys(), ...pastedTextNextIdRef.current.keys()])`, but the helper is idempotent and the loops are O(refs) which is small. The current shape is clearer about which ref is being scanned. Leave as-is.

None — the only suggestion above is borderline defensive and not worth a tracked issue. The reconciliation invariant comment already nudges future contributors toward the helper.

### Architecture Notes

The fix lives entirely in `<App />` and consumes the same `useConnectionStore(s => s.sessions)` selector that the rest of the component already uses. No store-layer changes, no protocol changes, no new WS messages. The reconciliation is a pure React-side concern — the store doesn't need to know that App keeps per-session refs. This keeps the cleanup co-located with the refs themselves, which matches the codebase's pattern of "the thing that owns the state owns its cleanup."

The effect runs once per `sessions[]` reference change, which Zustand emits on every `set({ sessions: ... })` from `message-handler.ts` (the `session_list` case at line 2006, plus the `sessions` updates at 1150 and 2779). Worst case the loop iterates over the entire ref maps on every WS `session_list` broadcast — which is O(refs) per change, dominated by the O(sessions) Set build. With <100 sessions on a busy dashboard this is sub-microsecond.

### Verdict

Issue #3736, opened 2026-05-10
permissionSessionMap leaks entries on internal permission resolutions (timeout/abort/auto-allow/clearAll)


blamechris:
## Context

Identified during Copilot review of PR #3730.

## Problem

\`WsServer._permissionSessionMap\` (\`Map<requestId, sessionId>\`) is populated when a \`permission_request\` is broadcast (via \`ws-forwarding.executeRegistrations\` reading \`registrations\` from the EventNormalizer). It is cleared on:

- WS \`permission_response\` from the user (\`settings-handlers.js:305\`)
- HTTP \`/permission-response\` resolution (\`ws-permissions.js:313, 353\`)
- HTTP \`/permission\` request close handler (\`ws-permissions.js:175\`)
- Session destroy (\`ws-server.js:612\`, filtered by sessionId)
- Server destroy (\`ws-permissions.js:469\`, full clear)

It is NOT cleared when permission requests resolve internally without a user response:

- \`PermissionManager.clearAll()\` — auto-deny on message completion
- Permission timeout (5-min auto-deny)
- Abort signal
- New \`autoAllowPending()\` introduced in PR #3730 (panic-button switch to auto)

The unified pipeline emits \`permission_resolved\` for each of these paths and the EventNormalizer broadcasts the message, but neither layer deletes the \`permissionSessionMap\` entry. Bounded only by session destroy.

## Severity

Memory leak per long-lived session. Each entry is \`<request-id-string, session-id-string>\` ≈ ~80 bytes. A session that auto-resolves 1000 permissions accumulates ~80KB of dead map entries. Not catastrophic, but unbounded growth and a code smell.

\`autoAllowPending()\` (the new path in PR #3730) makes this leak more frequent for users who flip to auto mode mid-turn — which is exactly the scenario the panic-button serves.

## Original Comment

> \`autoAllowPending()\` resolves pending permission requests internally and emits \`permission_resolved\`, but nothing clears the WsServer \`permissionSessionMap\` entry created for each \`permission_request\` (EventNormalizer registers it). When prompts resolve via non-user paths (timeout/abort/autoAllowPending), no \`permission_response\` arrives to trigger \`permissionSessionMap.delete(requestId)\`, so the map can grow unbounded for long-lived sessions. Consider deleting the mapping when forwarding \`permission_resolved\` (or otherwise adding a cleanup hook keyed by \`requestId\`).

## Proposed approach

[...]

PR #3987, opened 2026-05-16
fix(server): include toolUseId on PermissionManager.clearAll _pendingUserAnswer emit (#3975)

blamechris:
## Summary

PR #3971 plugged the `questionSessionMap` leak for every internal-resolution path that carries a `toolUseId` — except `clearAll()`, which emitted `{ reason: 'cleared' }` with no `toolUseId`. The sdk-session re-emit gate at `sdk-session.js:281` requires `data.requestId || data.toolUseId`, so the question-variant cleared event was silently dropped and the unified EventNormalizer cleanup never pruned the routing map.

The leak fires at most once per message-completion-while-`AskUserQuestion`-pending event (~80 bytes per stale entry) and is bounded by `session_destroyed` cleanup at `ws-server.js:621-623`, so this only affects long-lived sessions.

## Changes

- Track `toolUseId` on `_pendingUserAnswer` at capture time in `_handleAskUserQuestion()`.
- `clearAll()` now captures the pending-answer object (renamed `hadUserAnswer` -> `clearedUserAnswer`) and emits `{ toolUseId, reason: 'cleared' }` so the unified pipeline prunes `questionSessionMap`.

The SdkSession timeout-pause listener only decrements a counter on `permission_resolved` and ignores payload fields, so the extra `toolUseId` is harmless there (acceptance-criteria item from the issue).

## Test plan

- [x] New unit test in `permission-manager.test.js` asserts the cleared emit carries `toolUseId` matching the `user_question` toolUseId.
- [x] New end-to-end pipeline test in `permission-session-map-cleanup.test.js` wires a real `PermissionManager` through `setupForwarding` and asserts `questionSessionMap` is empty after `clearAll()`.
- [x] Both tests verified failing before the fix and passing after.
- [x] Existing 207 permission-related tests still pass (the 5 pre-existing failures match `origin/main` baseline and are unrelated).

Closes #3975
Related to #3971

### DIFF
--- a/packages/server/src/permission-manager.js
+++ b/packages/server/src/permission-manager.js
@@ -45,7 +45,7 @@ export class PermissionManager extends EventEmitter {
     this._sessionRules = [] // [{ tool, decision }]
 
     // AskUserQuestion handling
-    this._pendingUserAnswer = null // { resolve, input } when waiting for user answer
+    this._pendingUserAnswer = null // { resolve, input, toolUseId } when waiting for user answer
     this._questionTimer = null
     this._waitingForAnswer = false
   }
@@ -229,9 +229,15 @@ export class PermissionManager extends EventEmitter {
     return new Promise((resolve) => {
       const questionInput = input || {}
       this._waitingForAnswer = true
-      this._pendingUserAnswer = { resolve, input: questionInput }
-
       const toolUseId = `ask-${++this._permissionCounter}-${Date.now()}`
+      // #3975: stash toolUseId on the pending entry so clearAll() can
+      // emit it on the cleared-variant permission_resolved. Without
+      // toolUseId the sdk-session re-emit gate at sdk-session.js:281
+      // drops the event and the unified pipeline never prunes the
+      // questionSessionMap entry — small leak (~80 bytes) per
+      // message-completion-while-question-pending event, bounded only by
+      // session_destroyed cleanup.
+      this._pendingUserAnswer = { resolve, input: questionInput, toolUseId }
       this._logInfo(`AskUserQuestion detected (${toolUseId})`)
 
       this.emit('user_question', {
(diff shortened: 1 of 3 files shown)


blamechris:
## Code Review: PR #3987

### Summary

Tight, surgical fix for the small leak window identified in the original review of PR #3971. Adds `toolUseId` to the `_pendingUserAnswer` entry at capture time and propagates it on the `clearAll()` cleared emit so the unified EventNormalizer pipeline can prune `questionSessionMap`.

### Strengths

- **TDD discipline.** Test added in both the unit suite (`permission-manager.test.js`) and the integration suite (`permission-session-map-cleanup.test.js`). The integration test bridges a real `PermissionManager` through `setupForwarding` with the production re-emit gate (`data.requestId || data.toolUseId`) reproduced verbatim — that's exactly the right shape of regression guard for this bug class.
- **Race-aware ordering.** The `clearAll()` change captures the pending entry into `clearedUserAnswer` BEFORE the null-out, and the inline comment explicitly calls out the synchronous-listener race window. Same pattern as the existing `pendingIds = Array.from(...)` snapshot at line 429.
- **Naming.** `hadUserAnswer` (bool) → `clearedUserAnswer` (object) reads naturally and the diff makes the semantic shift obvious.
- **Backwards-compatible event shape.** Existing listeners (#2831 SdkSession timeout-pause counter) only care that the event fires, not its payload — adding `toolUseId` is purely additive on the wire.
- **Inline comment hygiene.** Every behavioral comment references its issue number (#3975, #2831) and explains the *why* — load-bearing for the next contributor who reads this code with no context.

### Issues Found

#### Critical (Must Fix)
None.

None blocking. One symmetry observation deferred below.

#### Nitpicks (Optional)
None.

| `respondToQuestion()` at `permission-manager.js:348` also emits `{ reason: 'answered' }` without `toolUseId`. Not a leak — the user-response handler at `input-handlers.js:451` explicitly deletes from `questionSessionMap` before this fires. But symmetric inclusion of `toolUseId` would match the other 4 question-variant emit sites (`aborted`, `timeout`, `cleared`, future) and make the pipeline contract uniform. | [#3988](https://github.com/blamechris/chroxy/issues/3988) | Out-of-scope; not a leak. Codifies symmetry as a defensive consistency improvement. |

### Architecture Notes

This is the last unfixed corner of the `#3736` → `#3971` leak class. The unified pipeline contract (any `permission_resolved` emit with a `requestId` or `toolUseId` triggers routing-map cleanup) is now closed across all 4 internal-resolution code paths in `PermissionManager` (timeout, abort, autoAllowPending, clearAll). Combined with the explicit deletes in `handleUserQuestionResponse` (user-response) and `WsServer.session_destroyed` cleanup (terminal fallback), `questionSessionMap` should no longer leak under any path.

The fix shape — capture object reference before null-out, emit identifying field on cleared event — is the right template for future internal-resolution paths added to this class.

### Verdict

Clean, minimal, well-tested fix that closes the leak window correctly. Tests follow the project's `node:test` + `assert/strict` patterns. ES modules, no semicolons, single quotes — all in line with `CLAUDE.md` server style. No protocol changes, no breaking changes.

Issue #3979, opened 2026-05-16
RateLimiter: unbounded per-IP map can be exhausted by source-IP rotation


blamechris:
## Context

Identified during review of PR #3978.

## Description

`RateLimiter` (`packages/server/src/rate-limiter.js`) stores a per-clientId timestamp array in `this._clients`. The map is only pruned when `remove()` or `clear()` is called explicitly.

- The WebSocket limiter (`_rateLimiter`) hooks `remove()` on disconnect, so its map naturally bounds at concurrent client count.
- The HTTP limiters (`_permissionRateLimiter`, the new `_diagnosticsRateLimiter` from #3737, and `_httpPermissionLimiter` inside ws-permissions.js) are stateless — there's no disconnect to hook. The map grows for every unique source IP that ever hits the endpoint.

An attacker rotating source IPs (or just normal Cloudflare traffic over weeks) can grow these maps unbounded. Each entry is small (a few timestamps), but at scale this is real memory pressure and an OOM vector.

Same problem already exists for the pre-existing permission limiters; #3737 just inherits the pattern.

## Original Review Comment

> `RateLimiter._clients` is never pruned for the HTTP-keyed limiters (no disconnect hook). A long-lived server accumulates one entry per source IP forever. Bound it (LRU at N entries, or periodic GC of entries whose newest timestamp is older than windowMs) the same way `ws-auth.js` evicts `authFailures`.

## Acceptance Criteria

- [ ] `RateLimiter` evicts entries whose newest timestamp is older than `windowMs` on `check()` access (low-cost amortised), OR an LRU cap with a documented `maxEntries` option.
- [ ] Tests cover the eviction path so the bound doesn't silently regress.
- [ ] Audit existing HTTP callers (`_permissionRateLimiter`, `_diagnosticsRateLimiter`, `_httpPermissionLimiter`) — all benefit; no caller should opt out.
- [ ] No behaviour change to the WS limiter (already bounded by `remove()` on disconnect).

PR #3998, opened 2026-05-16
fix(server): per-CF-IP bucketing for /permission rate limiter (#3980)

blamechris:
## Summary

- The POST `/permission` per-source-IP rate limiter keyed on `req.socket.remoteAddress` directly. Behind the Cloudflare tunnel the TCP peer is always `127.0.0.1` (the local cloudflared process), so every real client collapsed into a single shared bucket — a single noisy mobile could rate-limit every other tunneled user.
- Switches to the shared `getRateLimitKey()` helper introduced in #3978 for `/diagnostics`: trust `CF-Connecting-IP` / `X-Forwarded-For` only when the TCP peer is loopback; otherwise key off the kernel-supplied socket address so a direct attacker cannot spoof the header to share or exhaust another IP's bucket.
- Adds an optional `rateLimit` opt to `createPermissionHandler` for tests; production default (30 + 10 burst per minute) is unchanged.

## Test plan

- [x] New unit test: two different `CF-Connecting-IP` values from the loopback peer get separate buckets (one can hit max, the other still works).
- [x] New unit test: direct (non-loopback) peer keys off the socket address — header is NOT trusted, so a forged `CF-Connecting-IP` cannot grant a fresh bucket.
- [x] `node --test tests/ws-permissions.test.js tests/ws-server-diagnostics.test.js tests/rate-limiter.test.js` — 95/95 pass.
- [x] Full `npm test` — only pre-existing environmental failures (`TunnelManager Integration` needs `cloudflared` binary; `WebTaskManager` needs `claude` CLI). All permission and rate-limit suites green.

Closes #3980
Related to #3978

### DIFF
--- a/packages/server/src/ws-permissions.js
+++ b/packages/server/src/ws-permissions.js
@@ -1,6 +1,6 @@
 import { randomUUID } from 'crypto'
 import { createLogger } from './logger.js'
-import { RateLimiter } from './rate-limiter.js'
+import { RateLimiter, getRateLimitKey } from './rate-limiter.js'
 import { buildSessionTokenMismatchPayload } from './handler-utils.js'
 
 const log = createLogger('ws')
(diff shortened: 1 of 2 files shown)


blamechris:
## Code Review: PR #3998

### Summary

Fixes a per-IP rate-limiter collapse bug on `POST /permission` where the local cloudflared TCP peer (`127.0.0.1`) caused every tunneled mobile client to share a single 30+10 req/min bucket. Switches to the shared `getRateLimitKey()` helper introduced in #3978 for `/diagnostics`, so:

- Loopback peer (cloudflared) -> trust forwarded `CF-Connecting-IP` / `X-Forwarded-For`
- Direct peer (non-loopback) -> ignore the header, key off `req.socket.remoteAddress`

Adds a `rateLimit` opt to `createPermissionHandler` so tests can use tight `1+0` limits to assert per-key isolation without spamming 40 requests; production default is unchanged.

### Strengths

- Reuses the existing `getRateLimitKey()` helper rather than forking the loopback-trust logic — single source of truth for the proxy header trust policy.
- Comment block at the call site is specific: cites #3980 for the bug, #3978 for the precedent, and explicitly notes the spoofing surface that's still protected for direct peers.
- TDD: both tests fail without the implementation change (verified locally) and cover both halves of the contract (loopback per-header isolation; direct-peer header rejection).
- New `rateLimit` opt is additive and defaults to the previous literal, so no production behaviour change and no other call sites need updating.
- The diagnostic `[session-binding-create]` log's `sourceIp=...` field now naturally reports the real forwarded client IP instead of `127.0.0.1` — incidental win for the #2832 triage runbook.

### Issues Found

#### Critical (Must Fix)

None.

| `packages/server/src/ws-permissions.js` | 78 | `req.socket?.remoteAddress \|\| ''` falls through to `getRateLimitKey('', req)` which returns `'unknown'` only if no proxy header is present. Behavior matches `/diagnostics` (`http-routes.js:232`), so this is consistent — but worth a brief comment that the empty-string sentinel is intentional, mirroring the diagnostics helper contract. | Minor — future readers may wonder why this isn't `\|\| 'unknown'` like the pre-fix code. |

#### Nitpicks (Optional)

- The JSDoc for `opts.rateLimit` could mirror the style of `diagnosticsRateLimit` in `ws-server.js:50` (full `{ windowMs, maxMessages, burst }` shape) — not load-bearing.

None. The related concern about unbounded per-IP map growth (now slightly amplified because more distinct keys exist) is already tracked in #3979 ("RateLimiter: unbounded per-IP map can be exhausted by source-IP rotation") and explicitly calls out `_httpPermissionLimiter` as one of the three HTTP limiters that benefit. No new issue needed.

### Architecture Notes

This is the third HTTP endpoint to converge on `getRateLimitKey()` (the WS upgrade path and `/diagnostics` are the others). The pattern is now established as the project's standard for "this endpoint is reachable through the tunnel; do not key on the loopback peer." Any future HTTP rate-limited endpoint should follow this template by default.

The `rateLimit` opt is a minor expansion of the `createPermissionHandler` API surface but is in line with how `WsServer` accepts `diagnosticsRateLimit` for the same reason (test-time tightness). Symmetric and self-explanatory.

### Verdict

Tight, targeted security/availability fix. Tests cover both the bug and the spoofing surface. No production behavior change for legitimate clients; previously starved tunneled clients now get their own buckets.

PR #4001, opened 2026-05-16
fix(desktop): harden command_drift parser against multi-byte UTF-8 (#3992)

blamechris:
## Summary

`strip_cfg_attrs` in `packages/desktop/src-tauri/tests/command_drift.rs` iterated `input.as_bytes()` and pushed `bytes[i] as char`, which truncates each byte to Latin-1. Today the `generate_handler!` body in `src/lib.rs` is pure ASCII so the bug is latent — but a future contributor adding a comment like `// tile across displays — wide` inside the macro body would silently emit garbage chars that the downstream split-on-comma pipeline turns into either a missing real command or a phantom identifier (both fail the drift test in misleading ways).

Switch to char-based iteration via `input.chars().collect()` so multi-byte codepoints are preserved intact. The `#`/`[`/`]` markers we scan for are all ASCII, so they always sit on char boundaries — no boundary-handling complexity.

## Changes

- `tests/command_drift.rs` — `strip_cfg_attrs` now collects `chars()` and indexes by char position instead of byte position
- Added three regression tests:
  - `strip_cfg_attrs_preserves_multibyte_em_dash` — 3-byte UTF-8 (U+2014)
  - `strip_cfg_attrs_preserves_multibyte_emoji` — 4-byte UTF-8 (U+1F5C2)
  - `parse_generate_handler_extracts_identifiers_with_multibyte_in_cfg_string` — end-to-end fixture with a multi-byte char inside a `#[cfg(...)]` attribute string

## Test plan

- [x] `cargo test --test command_drift` — all 5 tests pass (3 new + 2 existing)
- [x] `cargo test` (full desktop suite) — 131 passed, 0 failed
- [x] `cargo clippy --tests` — no new errors or warnings on the modified file
- [x] Confirmed the new tests fail against the old byte-cast implementation before the fix

Closes #3992
Related to #3989, #3742

### DIFF
--- a/packages/desktop/src-tauri/tests/command_drift.rs
+++ b/packages/desktop/src-tauri/tests/command_drift.rs
@@ -101,19 +101,26 @@ fn find_matching_close_bracket(s: &str) -> Option<usize> {
 
 /// Strip every `#[cfg(...)]` attribute from a snippet, balancing brackets
 /// so nested parens inside `cfg(...)` don't confuse the parser.
+///
+/// Iterates by `char` (not byte) so multi-byte UTF-8 in the input — e.g. an
+/// em-dash inside a comment, or a unicode string literal inside a `cfg(...)`
+/// attribute — is preserved as-is rather than split into Latin-1 garbage
+/// (issue #3992). The `#`/`[`/`]` markers we scan for are all ASCII, so they
+/// always sit on char boundaries and never collide with UTF-8 continuation
+/// bytes.
 fn strip_cfg_attrs(input: &str) -> String {
     let mut out = String::with_capacity(input.len());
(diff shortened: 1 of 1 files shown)


blamechris:
## Review Comments Addressed

No actionable review comments. Copilot left a `COMMENTED` review with an overview only (no inline findings). The agent-review (issue-level comment, not inline) flagged 0 critical / 1 suggestion / 1 nitpick — neither requires a code change here.

**Status:** all CI green except `Server Tests` pending (unrelated to this PR — runs the JS server suite).
- Desktop Rust Tests: pass (5/5 in `command_drift.rs`, including the 3 new UTF-8 regression tests)
- Conversation threads: 0 (no inline comments)

PR #4082, opened 2026-05-22
Add BYOK MCP config discovery

Baijack-star:
Fixes #4076.

This adds the foundation-only BYOK MCP config discovery slice:

- reads a Claude-style config path (`opts.mcpConfigPath`, `opts.claudeConfigPath`, `CHROXY_CLAUDE_CONFIG`, or `~/.claude.json`)
- parses `mcpServers` entries with `{ command, args, env }`
- surfaces safe read-only metadata on `session.mcpServers`
- keeps full parsed env values internal on `_mcpServerConfigs` for future wiring
- warns and continues with an empty list for malformed config
- does not spawn MCP child processes or materialize tools yet

I kept env values out of public session metadata and expose only sorted `envKeys`.

Verification:

- `node --experimental-test-module-mocks --test --test-force-exit tests/byok-mcp-config.test.js tests/byok-session.test.js`
- `npx eslint --no-ignore src/byok-mcp-config.js src/byok-session.js tests/byok-mcp-config.test.js tests/byok-session.test.js`
- `git diff --check`

Notes:

- My first `npm install` attempt failed against the default user npm cache with a local permission/EEXIST cache error, so I retried successfully with `--cache /tmp/chroxy-npm-cache`.
- I also tried `npm test -- --test-force-exit tests/byok-mcp-config.test.js tests/byok-session.test.js`, but the package script always expands `./tests/**/*.test.js`, so that ran the full server suite. The new BYOK MCP tests passed in that run; unrelated full-suite HTTP/WS/doctor failures remained, so I used the direct Node test command above for the targeted signal.

Prepared by an AI coding agent.

### DIFF
--- a/packages/server/src/byok-mcp-config.js
+++ b/packages/server/src/byok-mcp-config.js
@@ -1,0 +1,102 @@
+/**
+ * MCP config discovery for the claude-byok provider.
+ *
+ * Foundation only for #4048/#4076: parse Claude-style mcpServers blocks and
+ * expose safe read-only metadata. This module deliberately does not spawn MCP
+ * children or wire tools.
+ */
+
+import { existsSync, readFileSync } from 'node:fs'
+import { homedir } from 'node:os'
+import { join } from 'node:path'
+
+export function defaultClaudeConfigPath() {
+  return process.env.CHROXY_CLAUDE_CONFIG || join(homedir(), '.claude.json')
+}
+
+function coerceStringArray(value) {
+  if (!Array.isArray(value)) return []
+  return value.filter((item) => typeof item === 'string')
+}
+
+function coerceEnv(value) {
+  if (!value || typeof value !== 'object' || Array.isArray(value)) return {}
+  const env = {}
+  for (const [key, raw] of Object.entries(value)) {
+    if (typeof key !== 'string' || key.length === 0) continue
+    if (typeof raw === 'string') env[key] = raw
+  }
+  return env
+}
+
+/**
+ * Parse a Claude-style MCP config object.
+ *
+ * @param {unknown} raw
+ * @returns {{ servers: Array<{ name: string, command: string, args: string[], env: Record<string, string> }>, warnings: string[] }}
+ */
+export function parseClaudeMcpConfig(raw) {
+  const warnings = []
+  const servers = []
+  if (!raw || typeof raw !== 'object' || Array.isArray(raw)) {
+    return { servers, warnings: ['MCP config root must be an object'] }
+  }
+  const block = raw.mcpServers
+  if (block == null) return { servers, warnings }
+  if (typeof block !== 'object' || Array.isArray(block)) {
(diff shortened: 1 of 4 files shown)


blamechris:
Thanks for this — sorry for the slow turnaround on review!

Quick request before we land it: **could you rebase onto current `main`?** The PR is currently in a conflicting state. Since the branch was opened, `packages/server/src/byok-session.js` was touched by #4383 (`dangerouslySkipPermissions` rename / deprecation alias), which is what's causing the merge conflict. The conflict should be small — the import block plus the constructor body where you added `loadClaudeMcpConfig`.

Once rebased and force-pushed, CI will re-run and we can take another look.

### Cursory review

I read through the diff — overall this is solid and the scope matches **#4076** exactly (foundation only; no child-process spawn, no tools materialized yet). A few notes:

**Looks good**
- Code style matches the project conventions (ES modules, no semicolons, single quotes, JSDoc) — no nits.
- Defensive parsing: every malformed entry returns a `warnings[]` entry instead of throwing. Nice operator UX.
- Security: `mcpServers` metadata exposes `envKeys` (sorted, frozen) but never the raw `env` values — full env stays on the internal `_mcpServerConfigs` field. That's exactly the boundary I'd want for a `session.mcpServers` surface.
- Tests cover the right shapes: happy path, multiple entries with one malformed, missing file, malformed JSON, env redaction, malformed config doesn't block session startup.
- `node:test` + `mkdtempSync` for temp dirs ✓ (matches the rest of the test suite).

**Minor things worth considering** (none are blockers)

1. **Operator visibility on the no-config path.** When `~/.claude.json` is missing, the session starts silently. A single debug log like `BYOK MCP config: no config found at ${filePath}` (when `missing: true`) would help operators distinguish "no MCP servers configured" from "config exists but has zero entries" without diving into a dump. Optional — feel free to defer.

2. **`existsSync` + `readFileSync` race.** Strictly speaking, the file could disappear between the two calls. In practice this fires once at startup, so the risk is essentially zero, but catching `ENOENT` inside the existing `try`/`catch` and treating it as `missing` would let you drop the `existsSync` check entirely. Cosmetic.

3. **`coerceEnv` silently drops non-string values.** The test pins this behavior (`COUNT: 12` is dropped, no warning). That's a reasonable choice — but if a config writer puts numbers in by mistake, they'll get no signal. A warning would help, e.g. `Skipping MCP server X env key COUNT: env values must be strings`. Optional.

4. **Path-traversal on `opts.mcpConfigPath`.** This is server-controlled (caller is `byok-session.js`), so it's not a user-input boundary today — but if the dashboard ever exposes a "choose your config file" UI, this becomes the entry point. Worth a short JSDoc note that the caller is responsible for validating the path. Optional.

5. **Squash commit message.** When this lands, we squash-merge. The PR description is good as-is, but feel free to tighten the commit message before merge if you'd like (e.g., add `Closes #4076` to the body so the issue auto-closes).

That's it. Once rebased and CI is green, this is a clean merge.

PR #4149, opened 2026-05-23
test(server): real _executeToolBlock end-to-end coverage (#4065)

blamechris:
## Summary

Closes #4065. The existing PR 2 tool-dispatch tests in `byok-session.test.js` stub `session._executeToolBlock` entirely — they pass even if the dispatch / permission-gate / result-shape contract between the agent loop and the real executor (`byok-tool-executor.js`) silently broke.

This PR adds three end-to-end tests under "end-to-end real executor coverage (#4065)" that exercise the unstubbed path: real `cwd`, real auto-permission gate, real `executeBuiltinTool`, real Read tool, real file ops.

### What the tests cover

1. **Happy path**: write a file into a temp workspace, stream a `tool_use` for `Read`, assert the `tool_result` block the agent feeds back on round 2 contains the **real on-disk content** (not a stubbed value).
2. **Error path**: stream a `Read` for a missing file, assert `is_error: true` propagates to the next round's history.
3. **Security boundary**: stream a `Read` for `/etc/passwd`, assert the `validatePathWithinCwd` realpath check rejects it as `is_error: true` with `"outside workspace"` — proves the path-safety seam is wired through end-to-end even in `auto` permission mode.

## Test plan

- [x] 3 new tests pass on Node 22
- [x] Full `byok-session.test.js` suite: 33/33
- [x] No source changes — test-only PR

### DIFF
--- a/packages/server/tests/byok-session.test.js
+++ b/packages/server/tests/byok-session.test.js
@@ -1,6 +1,6 @@
 import { describe, it, beforeEach, afterEach } from 'node:test'
 import assert from 'node:assert/strict'
-import { mkdtempSync, rmSync } from 'node:fs'
+import { mkdtempSync, rmSync, writeFileSync } from 'node:fs'
 import { tmpdir } from 'node:os'
 import { join } from 'node:path'
 import { APIUserAbortError } from '@anthropic-ai/sdk'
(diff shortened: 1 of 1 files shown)


blamechris (commented):
## Verdict: LGTM (cannot self-approve via gh API)

Test-only PR that lands exactly the coverage #4065 mandated. Three new tests, fresh temp workspace per test, the unstubbed dispatch + permission gate + executor + tool_result return path is exercised end-to-end. All 33 tests in the file pass on Node 22 locally.

## Strengths

1. **The path-traversal test is the standout.** Asserting that `/etc/passwd` is rejected with `is_error: true` and `/outside workspace/i` even in `auto` permission mode is a strong invariant — `auto` short-circuits the permission gate (`permission-manager.js:144-146`), so this proves the second line of defence (`validatePathWithinCwd` in `safeResolve`) is wired through the real executor. Worth flagging as a regression sentinel for any future refactor of the auto-mode short-circuit.

2. **No stub leakage.** `session._client` is replaced (the only fakeable boundary above the SDK), but `_executeToolBlock`, `_permissions`, the executor, `validatePathWithinCwd`, and the file ops all run for real. The inline comment at line 1047-1048 ("IMPORTANT: do NOT stub session._executeToolBlock") spells out the intent so a future contributor doesn't undo it.

3. **Isolation is clean.** `mkdtempSync` per test in nested `beforeEach`, `rmSync({ recursive: true, force: true })` in nested `afterEach`. The round counter and `toolResultBlock` capture are declared inside each `it` block, so no state can leak across tests.

4. **Determinism.** `auto` mode resolves the permission promise synchronously (`Promise.resolve({ behavior: 'allow', ... })`), so there is no UI-handshake timing dependency. The model-stub only emits two rounds — no risk of unbounded loops.

5. **Result-shape assertions pin the contract.** `tool_use_id === 'toolu_a'`, `is_error === false`, content regex matching disk content. If the agent loop ever serialised the tool_result block with the wrong key names (e.g. `toolUseId` instead of `tool_use_id`), the API would 400 — these tests would now catch that in CI.

## Observations (not blockers)

- **Read-only coverage.** This is what #4065 asked for, but `Write`, `Edit`, `Bash`, `Glob`, `Grep` flow through the same seam. Filed #4150 to extend coverage; the secret-redaction (Bash) and pattern-metachar-rejection (Glob) seams in particular are worth proving end-to-end.

- **Permission-gate-throw catch is still uncovered.** `byok-session.js:502-510` converts a rejected `handlePermission` to a `Permission gate error: ...` `is_error: true` tool_result, but no test forces the catch. Also no end-to-end test for `acceptEdits` mode or session-rule allow/deny through the unstubbed executor. Filed #4151.

- **One-tool-per-round assumption.** The round-counter pattern works because the model stub emits exactly one `tool_use` per round-1 response. Real models can emit multiple. The existing "synthetic-completion" test (#4108) already covers multi-block + abort, so this is fine for these three additions, but worth noting if anyone extends the round-counter pattern to a multi-block end-to-end test in the future.

## Follow-ups

- #4150 — Extend real-executor coverage to Write/Edit/Bash/Glob/Grep
- #4151 — Cover permission-gate throw, acceptEdits, and session-rule seam paths

No changes requested. Ready to merge once CI clears (Server Tests currently pending).

PR #4168, opened 2026-05-23
test(server): pin MAX_TOOL_ROUNDS summary-failure + abort event sequences (#4147)

blamechris:
## Summary

Closes #4147.

Two new byok-session tests that cover the MAX_TOOL_ROUNDS cap-handler paths flagged during #4145 review:

### 1. Sync stream-init failure — pin event sequence

The existing test already asserted the synthetic instruction is popped on init failure. New test pins the event-sequence side:
- exactly one MAX_TOOL_ROUNDS_REACHED (fatal: false)
- zero STREAM_ERROR
- exactly one result event with `stopReason: 'tool_use'` (the cap-hit round's reason — no summary ran)
- session released (`_isBusy === false`)

A future refactor that accidentally escalates init failure to STREAM_ERROR or double-emits result will fail loudly.

### 2. Abort during summary for-await — distinguish ABORT vs STREAM_ERROR

The existing async-rejection test throws a plain Error and asserts STREAM_ERROR. The new test throws `APIUserAbortError` from the for-await iterator and pins `_emitTurnError`'s instanceof check that routes to ABORT instead. Also covers:
- MAX_TOOL_ROUNDS_REACHED + ABORT both fire
- STREAM_ERROR does NOT fire alongside ABORT
- history rolls back to `historyLengthBeforeSend`
- session is usable for the next turn

## Test plan

- [x] Both new tests pass alongside the existing 38
- [x] No production code changes — tests-only

### DIFF
--- a/packages/server/tests/byok-session.test.js
+++ b/packages/server/tests/byok-session.test.js
@@ -1216,7 +1216,12 @@ describe('ClaudeByokSession', () => {
         'history must roll back to pre-send length on abort')
       assert.equal(session._isBusy, false, 'session must be released so the user can send again')
 
-      // Session usable after: stub a clean next-turn stream and send another message.
+      // Session usable after: the real bug this guards against is
+      // _isBusy left true / _abortController not nulled after abort
+      // (then the next sendMessage short-circuits with 'Already
+      // processing' per byok-session.js:_finishTurn). Swap the stream
+      // stub so the assertion isolates session reusability, not the
+      // cap-hit path's stub behaviour.
       session._client.messages.stream = () => fakeStream(
         [
           { type: 'content_block_delta', index: 0, delta: { type: 'text_delta', text: 'ok' } },
--- a/packages/server/tests/byok-session.test.js
+++ b/packages/server/tests/byok-session.test.js
@@ -1121,6 +1121,115 @@ describe('ClaudeByokSession', () => {
       await session.destroy()
     })
 
+    it('emits exactly one result + no STREAM_ERROR on summary stream-init failure (#4147)', async () => {
+      // Pre-#4147 we tested that the synthetic instruction was popped,
+      // but didn't pin the event sequence. After the cap-hit break the
+      // outer try-block still reaches `emit('result', ...)`, so the
+      // turn ends cleanly: ONE non-fatal MAX_TOOL_ROUNDS_REACHED error
+      // (fatal: false), ONE result event with the cap-hit round's
+      // stop_reason (tool_use), and NO STREAM_ERROR event. Pin both
+      // shape and count so a future refactor that accidentally double-
+      // emits or escalates to STREAM_ERROR fails loudly.
+      const session = new ClaudeByokSession({ cwd: '/tmp' })
+      session.setPermissionMode('auto')
+      session._executeToolBlock = async function ({ block }) {
+        return { type: 'tool_result', tool_use_id: block.id, content: 'r', is_error: false }
+      }
+      session._client = {
+        messages: {
+          stream: ({ tools }) => {
+            if (!tools || tools.length === 0) {
+              throw new Error('summary-init failed')
+            }
+            return fakeStream(
+              [{ type: 'message_delta', delta: { stop_reason: 'tool_use' } }],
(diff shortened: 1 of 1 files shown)


blamechris:
## Code Review: PR #4168

### Summary

Pure test additions (109 lines, single file) targeting the two cap-handler gaps flagged in #4147. The MAX_TOOL_ROUNDS branch has been a recurring source of regressions (#4063, #4109, #4118, #4145, #4146), so locking the event-sequence contract here is genuinely valuable — not over-pinning. Both tests pass; all 40 byok-session tests green.

The chosen assertions trade some refactor flexibility for high-signal failures: an accidental "emit STREAM_ERROR on cap-hit init failure" or "double-emit result" regression will fail loudly, which is exactly what we want. The abort test correctly isolates the `instanceof APIUserAbortError` routing path inside `_emitTurnError` rather than relying on the `signal.aborted` fallback (mirrors the design of the existing #4057 test).

**Recommendation: approve.**

### Strengths

- **`APIUserAbortError({ message })` construction** matches the established pattern at line 716 / 746 and the real SDK's exported class. No invented signature.
- **No cleanup leaks.** Both tests `await session.destroy()`. The abort test's iterator throws BEFORE `finalMessage()` is awaited, so no dangling promise. `_finishTurn` resets `_isBusy / _currentMessageId / _abortController` to null, which the assertions verify (`session._isBusy === false`).
- **Event-count brittleness is intentional, not accidental.** `capErrs.length === 1`, `streamErrs.length === 0`, `results.length === 1` are exactly the contract — one cap event, no escalation, one result per turn. The cap branch falls through to a single `emit('result', ...)` at L540; the assertions pin that invariant. A defensive refactor that, say, double-emitted on the failure path SHOULD break this test — that's the point.
- **`results[0].payload.stopReason === 'tool_use'`** correctly pins the semantic that `lastStopReason` reflects the last *successful* stream — the summary never ran, so the cap-hit round's reason wins. This is a non-obvious behaviour worth documenting via test.
- **History rollback assertion** in the abort test correctly captures `historyBefore` after `start()` but before `sendMessage()`, comparing post-rollback to that baseline. Math checks out: `historyLengthBeforeSend = max(0, length - 1) = 0`, truncate to 0, equal to `historyBefore`.

### Issues

#### Critical
None.

#### Suggestions

1. **Reusability assertion is meaningful but narrow.** The abort test's "session usable after" branch swaps `_client.messages.stream` for a new function. That's a fair simplification — `_finishTurn` already nulled the AbortController, so a clean SDK call returning a clean stream is the actual contract we want to verify (busy flag cleared, `_messageCounter` increments, no stale state blocks the turn). Worth adding a one-line comment saying so, since on first read it can look like we're "not really testing post-error reusability of the same client" — but on closer inspection, the bug it catches (session left busy → next send `emit('error', 'Already processing a message')`) is real and important.

2. **Second `captureEvents` call** at L1227 attaches a new set of listeners but does NOT detach the originals — so the first `captured` array keeps growing across the second `sendMessage`. No assertion is made against `captured` after that point, so this is benign, but a future maintainer adding such an assertion would be confused. Tiny code-comment fix.

3. **Length-only history rollback assertion** verifies the count but not the contents. A hypothetical bug that truncated the wrong end would still pass if the lengths matched. Belt-and-braces would be `assert.deepEqual(session._history, [])`, which here is equivalent. Optional.

#### Nitpicks

- L1163's `aborts.length === 0` is good defensive coverage — the cap-hit-init-failure path should never look like an abort even though both emit `'error'`. Worth keeping.
- The two tests share enough scaffolding (`session._executeToolBlock` stub, cap-hit round stream stub) that they could share a small helper. Not blocking; the repetition reads fine.

### What's NOT covered (deferred to follow-ups)

Two adjacent paths are surfaced but intentionally out of scope for this PR. Filed as follow-ups:

- **#4169** — `summaryStream.finalMessage()` async rejection (the existing async-rejection test throws from the for-await iterator, not from finalMessage)
- **#4170** — abort BETWEEN cap-hit emit and summary stream-init (no `signal.aborted` check in that window; current behaviour silently breaks to the standard result emit — worth pinning whichever contract we want)

### Verdict

Approve. Tight scope, real coverage, no production code touched, all tests green. The event-count assertions are appropriately strict for a region that has regressed before.

PR #4228, opened 2026-05-24
fix(scripts): bump-version.sh syncs Cargo.lock with Cargo.toml

blamechris:
## Summary

Replace the silent `cargo generate-lockfile 2>/dev/null` step in `scripts/bump-version.sh` with a surgical `cargo update -p chroxy-desktop --precise <NEW_VERSION>`, plus an awk-in-place fallback for machines without a Rust toolchain. Both paths verify the `chroxy-desktop` stanza of `packages/desktop/src-tauri/Cargo.lock` lands on the new version and fail loudly otherwise.

## Why

`Desktop Rust Tests` CI runs `cargo test --locked`, which refuses to regenerate the lockfile and fails when the `chroxy-desktop` entry doesn't match `Cargo.toml`. Every version bump silently broke CI until someone patched the lockfile by hand — done for v0.8.3 in commit `43889e55e`. The prior `cargo generate-lockfile 2>/dev/null` would have helped, but it swallowed errors (e.g. cargo not on PATH) and could churn unrelated dep versions.

## What changes

- Prefer `cargo update --workspace -p chroxy-desktop --precise "$NEW_VERSION"` (offline first, retries online) so only the workspace stanza moves and transitive pins are preserved.
- Fall back to a scoped awk edit of the single `version = "..."` line inside the `name = "chroxy-desktop"` stanza when cargo is missing — mirrors the existing JS lockfile patching idiom.
- Post-hoc verify the chroxy-desktop entry equals `$NEW_VERSION`; exit non-zero on any mismatch.
- Track the temp file via the existing `track_tmp` cleanup trap.
- Update the "Updated:" output to reference the full Cargo.lock path.

## Test plan

- [x] `bash -n scripts/bump-version.sh` clean
- [x] Ran `./scripts/bump-version.sh --no-changelog 0.8.7` with cargo on PATH; `Cargo.lock` chroxy-desktop stanza moved to 0.8.7
- [x] Re-ran with cargo stripped from PATH; awk fallback patched the same stanza to 0.8.7
- [x] `cargo metadata --locked` validates the resulting lockfile (i.e. CI's `--locked` would pass)
- [x] Reverted the test bump before commit

Closes #3864


blamechris (commented):
## Agent review — PR #4228

Solid surgical fix for the long-standing bump → CI breakage. The cargo+awk dual-path is the right shape, and the post-verify guard is what the PR description claims. But there's a blocker the CI is already screaming about.

### Blocker

**Scripts Tests CI is red — 7 of 10 tests fail because the test harness wasn't updated alongside this PR.**

`scripts/__tests__/bump-version.test.sh` installs a neutered copy of the script with:

```sh
sed 's|cargo generate-lockfile|true|g' "$BUMP" > "$dir/scripts/bump-version.sh"
```

That substitution targeted the old single-line `cargo generate-lockfile` step. This PR removed that exact string and replaced it with a multi-line block (`cargo update --workspace -p chroxy-desktop --precise ...` + awk fallback + `cargo metadata`-style awk verify). The sed match no longer hits anything, so the full lockfile-sync block runs against the fake-repo skeleton — which has no real `[workspace]`, no `Cargo.lock`, no buildable Cargo project. Cargo errors out (`no targets specified in the manifest`), the online fallback fails, and the post-verify `awk` returns empty against the missing `Cargo.lock`, so the script exits non-zero. Every test that exercises a successful bump (7 of them) fails on exit code. The 3 passing tests (`blocks when prior TODO unresolved`, `--help`, `rejects unknown flags`) all exit before reaching the lockfile step.

Reproduced locally on darwin with cargo on PATH: same 7 failures, identical to CI.

**Fix in the harness, not the production script** — either:

1. Extend `install_bump_script` to also strip the new block, e.g. replace the entire `# Sync Cargo.lock...` ... `exit 1\nfi` region with `:`, or
2. Have `build_fake_repo` write a real (minimal) `Cargo.lock` containing the `chroxy-desktop` stanza so the awk path works without invoking cargo, and stub cargo to a no-op (e.g. prepend a fake `cargo` to `PATH`).

Option 2 is preferable: it exercises the new awk fallback for real instead of stubbing it out, which is the whole point of having it.

### Awk fallback correctness

Verified against the real `Cargo.lock` stanza (line 521 onwards: `[[package]]` / `name = "chroxy-desktop"` / `version = "0.8.7"` / `dependencies = [...]`). The state machine correctly:

- Resets `in_pkg` / `is_target` on each `[[package]]` and blank line
- Only rewrites the first `^version = "` line in the matching stanza (clears `is_target` immediately after substitution)
- Anchors the name match to `^name = "chroxy-desktop"$` exactly — won't false-match `chroxy-desktop-foo` or whatever future package

One minor: `\[\[package\]\]` works under POSIX awk, gawk, and mawk (the three you'll see on operator machines and CI). Good.

### Verify-or-fail check

The PR description mentions `cargo metadata --locked` as the validation. The actual script does NOT call `cargo metadata`; it parses the lockfile with awk and string-compares to `$NEW_VERSION`. That's actually fine (no cargo dependency, works on machines without a Rust toolchain), but the PR body is misleading — consider tightening the wording before merge. The awk-based verify is the right call.

### Backward compat

`cargo generate-lockfile 2>/dev/null` silently rewrote unrelated transitive pins (and silently swallowed all errors, including "cargo not installed"). The new surgical approach loses that broader rewrite — but that broader rewrite was a bug, not a feature: it churned dep pins on every patch bump, which is exactly the noise a `--locked` CI is designed to prevent. No real-world workflow relied on it.

### Shell quality

- `set -euo pipefail`: yes, at line 19. Good.
- Variables quoted everywhere they matter.
- `command -v cargo` is the right portable check (not `which`).
- `track_tmp` is reused for the awk tmp file — good cleanup hygiene.
- The `(cd ... && cargo ... --offline) || (cd ... && cargo ...)` retry chain is bash-friendly subshells; no surprises.

### Test coverage

No new shell-level tests for the lockfile-sync block specifically. Given the harness already fakes the repo and the awk path is straightforward to test in isolation, a follow-up that:

1. Fixes the harness regression (blocker above)
2. Adds 2-3 cases for the awk fallback (cargo absent, stanza shape, verify mismatch → non-zero exit)

would be valuable. Tracking as a follow-up below.

### No production code touched outside scripts/

Confirmed. Diff is `scripts/bump-version.sh` only.

---

**Verdict: changes-requested**. The production code is good. The CI failure is the harness, not this script's logic, but it has to be green before merge — fix in the same PR (small) or as a follow-up immediately before this lands.

PR #4229, opened 2026-05-24
refactor(dashboard): single source of truth for client-estimated-cost providers

blamechris:
Closes #4206

## Summary
Pre-#4206 the list of providers whose `cost` is computed client-side lived in two places that stayed in sync by accident — Codex and Gemini were the only providers emitting `cost: null` + usage, so the bare `if (cost === null && contextUsage)` fallback in `message-handler.ts` agreed with the hardcoded `new Set(['codex', 'gemini'])` in `status-tooltips.ts`. The moment a new such provider landed, the message-handler would estimate the cost and the tooltip would still call it server-authoritative.

## Fix
- New `lib/client-estimated-cost-providers.ts` exports the set; both sites import it.
- `message-handler.ts` cost fallback now also gates on the provider being in the set — a server-priced provider that momentarily emits `cost: null` (race, edge case) no longer gets a wrong client-side estimate written into its state.

## Tests
- **costTooltip single source of truth (#4206)** — loop test asserts every entry in the exported set triggers the `estimated client-side` wording. Adding a provider in one place is the single edit that satisfies acceptance criterion 1.
- **Negative test in message-handler** — `claude-byok` with `cost: null` + usage leaves `lastResultCost` null. Regression guard for the new gate.
- Existing 4 cost-fallback tests updated to seed sessions with the matching provider id (the gate now requires it). All pass with the same expected costs.

## Test plan
- [x] `npx vitest run src/lib/status-tooltips.test src/store/message-handler.test` — 137/139 pass.
- [x] The 2 failures (byok_credentials_status TypeError on `safeParse`) are pre-existing on main — reproduced with `git stash`. Unrelated to this change; appears tied to a bad `ServerByokCredentialsStatusSchema` import at line 97.
- [x] No production runtime behaviour change for Codex/Gemini sessions (gate matches what fired by accident before).

## Out of scope
- Wording for `claude-byok` being table-priced rather than API-authoritative — called out in the issue, left to a focused follow-up.

### DIFF
--- a/packages/dashboard/src/lib/client-estimated-cost-providers.ts
+++ b/packages/dashboard/src/lib/client-estimated-cost-providers.ts
@@ -1,0 +1,29 @@
+/**
+ * Shared source of truth for the provider ids whose `cost` field is
+ * computed client-side from token usage rather than reported by the
+ * server.
+ *
+ * #4206: pre-fix this list lived twice — once in `status-tooltips.ts`
+ * (to gate the "estimated client-side" wording in `costTooltip`) and
+ * implicitly in `message-handler.ts` (where the only providers that
+ * emit `cost: null` + a usage payload happened to be Codex and Gemini,
+ * so the two sites agreed by accident). The moment a NEW provider
+ * lands that emits `cost: null` + non-null `contextUsage` —
+ * a Gemini fork, a future BYOK variant, an MCP provider — the
+ * message-handler fallback would silently run the client-side
+ * estimation and the cost-tooltip would call it server-authoritative.
+ *
+ * Both sites now import this set. Adding a provider here is the single
+ * edit that keeps the fallback and the tooltip wording in lockstep.
+ *
+ * Keep this list narrow: a provider belongs here ONLY if its server
+ * really cannot return a cost number. `claude-byok` does NOT belong
+ * here — it computes cost server-side via the byok-session pricing
+ * table — even though the value is table-priced rather than
+ * API-billed-authoritative. (Refining the byok wording is a separate
+ * concern; see #4206 issue body.)
+ */
+export const CLIENT_ESTIMATED_COST_PROVIDERS: ReadonlySet<string> = new Set([
+  'codex',
+  'gemini',
+])
--- a/packages/dashboard/src/lib/status-tooltips.test.ts
+++ b/packages/dashboard/src/lib/status-tooltips.test.ts
@@ -15,6 +15,7 @@ import {
   modelTooltip,
   agentCountTooltip,
 } from './status-tooltips'
+import { CLIENT_ESTIMATED_COST_PROVIDERS } from './client-estimated-cost-providers'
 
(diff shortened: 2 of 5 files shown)


blamechris (commented):
## Agent review — APPROVE

Pipeline review of #4229. CI is fully green (13/13 checks), Copilot generated no inline comments, and there are zero unresolved review threads.

### Correctness — behavior change is safe and defensive

The pre-PR bug was real: `message-handler.ts` ran the client-side cost fallback for **any** session whose `result.cost === null && contextUsage` — it agreed with the hardcoded tooltip set in `status-tooltips.ts` only because Codex and Gemini were the sole providers emitting that shape. A new provider with a transient `cost: null` (race, edge case) would have:
- silently received a wrong client-side estimate in `lastResultCost`, and
- shown the tooltip wording "server-authoritative" because `status-tooltips.ts` only flagged the original two.

Post-PR, both sites import from `client-estimated-cost-providers.ts`. The message-handler gate now requires `sessionModel && sessionProvider && CLIENT_ESTIMATED_COST_PROVIDERS.has(sessionProvider)` — the strictly defensive side of the change. Verified `SessionInfo.provider` is `optional` in `packages/protocol/src/schemas/server.ts:356`, so the `session?.provider ?? null` chain correctly no-ops for legacy/incomplete sessions rather than throwing — those sessions get a `null` cost (acceptable; they wouldn't have priced correctly anyway).

Cross-checked provider IDs against `packages/server/src/providers.js:29-35` — `codex` / `gemini` / `claude-sdk` / `claude-byok` all match.

### Tests

- Existing 4 cost-fallback tests correctly updated to seed `provider` on the mock `SessionInfo` (without this they'd no-op silently against the new gate — good that they would fail loudly if anyone removed the seed).
- New loop test pins the single-source-of-truth contract for every entry.
- Non-empty-floor assertion is good belt-and-suspenders — a future refactor that empties the set would silently make the loop vacuously pass without it.
- New `claude-byok` negative test pins the gate against regression.

### Acceptance criterion 1 (#4206) met

Adding a new client-priced provider is now a single edit in `client-estimated-cost-providers.ts`. Both the fallback and the tooltip wording move in lockstep.

### Non-blocking notes

1. **`ReadonlySet<string>` is a compile-time hint only.** At runtime the `Set` is still mutable — `.add()` / `.delete()` will work. The TS hint is sufficient for the single-edit acceptance criterion, but if true immutability matters the cheap upgrade is `Object.freeze(new Set([...]))` plus the existing type. Not a blocker — leaving it gives flexibility for tests that want to monkey-patch.
2. **Out of scope (correctly deferred):** The `claude-byok` "table-priced vs API-authoritative" wording refinement called out in the issue body is left for a follow-up. The header comment in the new file explicitly explains the boundary, which is helpful for the next person.

### Verdict

Ready to merge. Clean refactor, real bug fix (defensive against future providers), good test coverage, no behavior change for current Codex/Gemini sessions.

PR #4304, opened 2026-05-26
feat(dashboard): pluggable sidebar panel slot + token view v0 (#4303 PR 1/3)

blamechris:
First of three PRs for #4303. Lands the **slot infrastructure** plus a **minimal token usage view** as the first occupant. PRs 2 and 3 add time-series storage and the sparkline / API-equiv toggle / per-tool breakdown respectively.

### Slot infrastructure (`SidebarPanelSlot`)

- View registry: each entry is `{ id, label, render, collapsedHeaderMetric? }` — adding a new view is a one-line append, no slot code changes
- Tab strip with `role=\"tablist\"` / `aria-selected` / roving tabindex
- Collapse toggle (header bar stays visible, body + resize handle hide)
- Drag-resize on the top edge with clamped `[minHeight, maxHeight]`
- Arrow-key resize (16px steps) when the handle is focused
- `role=\"separator\"` + aria-orientation / min / max / now on the handle
- When collapsed, the active view can supply a single live metric to the header bar via `collapsedHeaderMetric()` (decision #4 in #4303)
- Stale-view-id fallback: when `selectedViewId` doesn't match any registered view, falls back to the first registered view rather than rendering nothing

### Token view v0 (`SidebarTokenView`)

- Reads existing in-memory `cumulativeUsage` off `SessionInfo` — no new wire / event work
- `aggregateUsage(sessions)` pure helper: per-provider rollup with untracked-provider isolation (unit-testable, no React deps)
- claude-tui rows surface `—` with a tooltip explaining the PTY-only limitation; their zero-block does NOT pull cross-provider totals toward zero (decision #1)
- Today totals row shows input + output; BYOK cost row only renders when `costUsd > 0` (hidden for pure-subscription runs)
- Per-provider list sorted tokens-desc, untracked rows last
- `tokenViewCollapsedMetric()` exposed for the collapsed-header

### Persistence

- New keys: `chroxy_persist_sidebar_panel_{height,view,collapsed}` mirroring existing `KEY_SIDEBAR_WIDTH` pattern
- Round-trip + edge-case tests: invalid / non-positive heights, removal-on-null

### SERVERS placement

Kept BELOW the slot for v1. The Sidebar.tsx code comment + the issue's decision section explain the rationale: ServerPicker has its own always-visible add button and live status — forcing it into the slot's one-view-at-a-time model loses functionality you'd want simultaneously with whichever slot view is selected. Migrating SERVERS into the slot as a registered view is a follow-up if it proves desirable.

## What this PR does NOT do (PR 2 & 3 scope)

- No `~/.chroxy/usage-history.json` write path (PR 2)
- No 7-day sparkline (depends on history file from PR 2)
- No \"show as API-equivalent cost\" toggle (PR 3)
- No per-tool breakdown for BYOK sessions (PR 3)
- No cache-hit ratio display (PR 3)

The slot is ready to host PR 3's enhancements without further infrastructure work — the registry pattern means PR 3 just extends the token view's render function.

## Tests

- 16 new tests for `SidebarPanelSlot` (view selection, collapse, resize, a11y)
- 18 new tests for `SidebarTokenView` (aggregate, per-provider, untracked handling, render)
- 8 new tests for sidebar panel persistence helpers
- 1899/1899 dashboard tests pass (up 42)
- `tsc --noEmit` clean

## Test plan

- [ ] Run the dashboard locally, confirm the new bottom panel renders below the sessions tree and above the SERVERS section
- [ ] Click the Tokens tab — body shows; click the collapse arrow — body hides, tab strip + collapsed metric remain
- [ ] Drag the resize handle up/down — height changes, persists across reload
- [ ] Reload — view, height, collapsed state restored
- [ ] Provider mix (BYOK + SDK + CLI + TUI) — TUI row shows `—` with tooltip; others show real numbers
- [ ] Pure-subscription session — no BYOK cost row visible
- [ ] Keyboard a11y — Tab into resize handle, press ArrowUp/ArrowDown — panel resizes

## Related

- Closes part of #4303 (slot infrastructure + first occupant)
- Builds on the BYOK cost suite: #4072, #4083, #4088, #4119, #4126, #4128

### DIFF
--- a/packages/dashboard/src/App.tsx
+++ b/packages/dashboard/src/App.tsx
@@ -64,7 +64,7 @@ import { isTauri } from './utils/tauri'
 import { startServer, revealInFinder } from './hooks/useTauriIPC'
 import { usePermissionNotification, type PermissionPromptInfo } from './hooks/usePermissionNotification'
 import { SplitPane, type SplitDirection } from './components/SplitPane'
-import { persistSidebarWidth, loadPersistedSidebarWidth, persistSplitMode, loadPersistedSplitMode, persistShowConsoleTab, loadPersistedShowConsoleTab } from './store/persistence'
+import { persistSidebarWidth, loadPersistedSidebarWidth, persistSplitMode, loadPersistedSplitMode, persistShowConsoleTab, loadPersistedShowConsoleTab, loadPersistedSidebarPanelHeight, loadPersistedSidebarPanelView, loadPersistedSidebarPanelCollapsed } from './store/persistence'
 import { DiffViewerPanel } from './components/DiffViewerPanel'
 import { AgentMonitorPanel } from './components/AgentMonitorPanel'
 import { SessionLoadingSkeleton } from './components/SessionLoadingSkeleton'
@@ -1745,6 +1745,10 @@ export function App() {
           searchQuery={searchQuery}
           searchConversations={searchConversations}
           clearSearchResults={clearSearchResults}
+          sessions={sessions}
+          initialPanelHeight={loadPersistedSidebarPanelHeight() ?? 200}
+          initialPanelView={loadPersistedSidebarPanelView()}
+          initialPanelCollapsed={loadPersistedSidebarPanelCollapsed()}
         />
       )}
 
--- a/packages/dashboard/src/components/Sidebar.tsx
+++ b/packages/dashboard/src/components/Sidebar.tsx
@@ -4,12 +4,19 @@
  * Shows repos with active/resumable sessions, filter, status footer.
  * Collapsible with Cmd+B toggle.
  */
-import { useState, useCallback, useRef } from 'react'
-import type { CumulativeUsage, SessionVisualStatus } from '@chroxy/store-core'
+import { useState, useCallback, useRef, useMemo } from 'react'
+import type { CumulativeUsage, SessionInfo, SessionVisualStatus } from '@chroxy/store-core'
 import { formatCostBadge, formatCostBreakdown } from '@chroxy/store-core'
 import { ConversationSearch } from './ConversationSearch'
 import { ServerPicker } from './ServerPicker'
+import { SidebarPanelSlot, type SidebarPanelView } from './SidebarPanelSlot'
+import { SidebarTokenView, tokenViewCollapsedMetric } from './SidebarTokenView'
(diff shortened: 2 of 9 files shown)


blamechris:
## Code Review: PR #4304

### Verdict
**Approve with suggestions** — slot abstraction is clean, test coverage is solid (42 new tests), staging across the 3-PR trio is sensible. Three items worth fixing before merge; none block.

### Critical
None — nothing blocks merge.

### Suggestions (will address in this PR)

1. **Duplicate `PROVIDER_LABELS` table** in `SidebarTokenView.tsx:14-21`. `@chroxy/store-core/provider-labels.ts` already exports `getProviderLabel()`. The local copy is missing `docker-cli`, `docker-sdk`, and `docker` — those sessions render with raw provider strings. Will replace with the store-core helper for single source of truth.

2. **Tablist missing ArrowLeft/ArrowRight handlers.** Roving tabindex is correctly set but no keyboard navigation between tabs. With one tab today it's silently broken; the moment a second view lands (PR 2/3 or future MCP/skills), keyboard users are stuck. WAI-ARIA tablist pattern expects Left/Right + Home/End. Cheap to add now.

3. **Drag listeners leak if slot unmounts mid-drag.** `handleResizeMouseDown` adds `mousemove`/`mouseup` to `document`; the cleanup `useEffect` only flips a ref. If the user starts dragging then triggers an unmount (Cmd+B sidebar collapse is plausible), listeners stay attached to a dead component. Will lift the listeners to refs and remove in the effect cleanup.

### Nitpicks (also fixing this PR while there)

- `formatTokenCount(999999)` returns `'1000.0K'` instead of crossing into M. Real boundary bug, will round up to `'1.0M'` for values ≥995000 (where the K abbreviation would round to 1000.0K).

### Architecture confirmed sound

- Cross-device desync: one-shot read on mount, each setter mirrors to localStorage → last write wins on next persist. Acceptable for v0.
- Stale-view fallback (`views.find ?? views[0] ?? null`) correctly handles "view shipped then removed."
- Untracked-provider isolation logic correct: TUI's zero-block excluded from `crossTotal`, still counted in `totalSessions`.
- Slot extensibility: registry shape sound for planned future occupants.
- CSS scoping verified — no conflict with existing `.sidebar-*` selectors.

Fixes incoming in the next commit.

PR #4326, opened 2026-05-26
test(server): add cross-provider transition test for sendSessionInfo (#4315)

blamechris:
## Summary

Adds a test under `describe('available_models on session switch (#4302)', ...)` that verifies the cross-provider transition scenario — the exact wire sequence the original #4302 bug reproduces against.

The two tests added in #4310 verify the end-state of `sendSessionInfo` for a single session, but they don't cover the actual bug scenario: a client whose `availableModelsProvider` was tagged for one provider then sees a `switch_session` to a different-provider session.

The new test:

- Creates two mock sessions tagged `claude-tui` and `claude-cli`
- Calls `sendSessionInfo` for the TUI session and asserts the emitted `available_models` carries `provider: 'claude-tui'`
- Clears the recorded sends, calls `sendSessionInfo` for the CLI session, and asserts the new `available_models` is re-tagged with `provider: 'claude-cli'`

Without re-tagging on every switch, the dashboard's `availableModelsProvider` stays pinned to the first provider and `modelsMatchProvider` (App.tsx) suppresses the model picker. A future refactor that suppresses the push when the provider hasn't changed must still fire one when it has — this test pins that behaviour.

## Test plan

- [x] `node --experimental-test-module-mocks --test packages/server/tests/ws-history.test.js` — 61/61 pass (60 existing + 1 new)
- [x] New test runs under the expected `available_models on session switch (#4302)` block

Closes #4315

### DIFF
--- a/packages/server/tests/ws-history.test.js
+++ b/packages/server/tests/ws-history.test.js
@@ -583,6 +583,41 @@ describe('sendSessionInfo', () => {
       assert.ok(modelsMsg, 'available_models was not sent on session switch')
       assert.equal(modelsMsg.provider, null)
     })
+
+    // #4315 — follow-up to #4310/#4302. The two tests above verify the
+    // end-state for a single sendSessionInfo call, but the original bug
+    // scenario is a *transition* between providers across two consecutive
+    // switches. Without re-tagging on every switch, the dashboard's
+    // `availableModelsProvider` would stay pinned to whichever provider
+    // the client saw first and `modelsMatchProvider` (App.tsx) would
+    // suppress the model picker for the second session. A future refactor
+    // that suppresses the push when the provider hasn't changed must
+    // still fire one when it has — this test pins that behaviour.
+    it('re-tags available_models when switching between different-provider sessions', () => {
+      const { manager, sessionsMap } = createMockSessionManager([
+        { id: 'sess-tui', name: 'TUI', cwd: '/tui' },
+        { id: 'sess-cli', name: 'CLI', cwd: '/cli' },
+      ])
+      sessionsMap.get('sess-tui').provider = 'claude-tui'
+      sessionsMap.get('sess-cli').provider = 'claude-cli'
+      const ws = makeFakeWs()
+      const ctx = makeCtx({ sessionManager: manager })
+      registerClient(ctx, ws)
+
+      // First switch: TUI session
+      sendSessionInfo(ctx, ws, 'sess-tui')
+      const firstModels = ctx._sends.find(m => m.type === 'available_models')
+      assert.ok(firstModels, 'available_models was not sent for first session')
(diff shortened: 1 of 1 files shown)


blamechris (commented):
## Code Review: PR #4326

### Summary

Adds a single test under the existing `available_models on session switch (#4302)` describe block that pins the cross-provider transition behaviour of `sendSessionInfo`. The test creates two mock sessions tagged with different providers (`claude-tui` and `claude-cli`), calls `sendSessionInfo` for the first, asserts the `available_models` payload carries `provider: 'claude-tui'`, clears `ctx._sends`, switches to the CLI session, and asserts the re-tagging to `provider: 'claude-cli'`. Closes #4315.

### Strengths

- **Targets the exact wire sequence the original bug reproduces against.** The two prior tests in #4310 only assert the end-state of a single `sendSessionInfo` call. This one exercises the transition — the precise path where a future refactor that suppresses redundant pushes could silently regress the dashboard's `modelsMatchProvider` guard in `App.tsx`.
- **Reuses the existing harness** (`createMockSessionManager`, `makeFakeWs`, `makeCtx`, `registerClient`) — no new test scaffolding, no helper drift.
- **Clear inline comment** above the test explains both the bug history and the regression-prevention intent. This is the kind of context that survives the next refactor.
- **Style adherence** — ES modules, no semicolons, single quotes, arrow callbacks consistent with the surrounding file.
- **Provider pairing is well chosen** — `claude-tui` and `claude-cli` are both real provider tags in `providers.js`, not synthetic strings. The transition mirrors a realistic user flow (TUI session, then CLI session).
- **Assertion messages are specific** — `'cross-provider switch must re-tag available_models'` makes the failure mode self-documenting if this test breaks in the future.

### Issues Found

#### Critical (Must Fix)

None.

None — the test is tight, scoped, and matches the acceptance criteria in #4315.

#### Nitpicks (Optional)

- The acceptance criteria in #4315 proposes `claude-sdk → claude-cli` whereas the PR uses `claude-tui → claude-cli`. Both are valid cross-provider transitions and the comment explicitly calls out "TUI/SDK" together. Not worth changing — the underlying invariant (re-tag on switch) is identical.

None.

### Architecture Notes

This is a tests-only change. Single file, +35/-0, no production code touched. CI is green across all 13 checks (Server Tests, Server Lint, Dashboard Tests, App Tests, Protocol Tests, Store Core Tests, Desktop Rust Tests, etc.). The new test runs under the production wire path: `sendSessionInfo` is the real exported function, and the assertions inspect the actual `available_models` payload emitted to a fake WS client. No mocking of the function under test.

This reinforces the regression net around #4302 without coupling to implementation details (it asserts the observable wire message, not internal state).

### Verdict

- [x] **Approve** — Ready to merge. Closes #4315 and tightens the regression coverage around #4302.

Issue #4262, opened 2026-05-24
fix(server): byok-session tool_start should use per-tool content_block.id, not turn messageId


blamechris:
## Context

Identified during review of PR #4257. Pre-existing in `byok-session.js`; called out in the PR description as deferred.

## Problem

`packages/server/src/byok-session.js` `tool_start` emit reuses the turn-level `messageId`:

```js
this.emit('tool_start', {
  messageId,            // <-- turn-level id, NOT the content_block id
  toolUseId: t.toolUseId,
  tool: t.toolName,
  input: null,
})
```

Compare with `sdk-session.js:635-641` and `cli-session.js:708-714`, which use the tool's own `content_block.id`:

```js
const toolId = event.content_block.id || `${messageId}-tool`
const toolStartData = {
  messageId: toolId,
  toolUseId: toolId,
  tool: event.content_block.name,
  input: null,
}
```

The sdk-session comment is explicit about why:

> Use the tool's content_block.id as the tool_start messageId so each tool in a multi-tool turn has a distinct id. Sharing the turn-level messageId across tools collides with the post-tool stream_start id and corrupts client message state.

## Symptom

In multi-tool turns from a BYOK provider, every `tool_start` shares the same `messageId`, and that same id then collides with the post-tool `stream_start` for the same turn. `store-core/handlers/handleToolStart` uses `messageId` as the `ChatMessage.id`:

```ts
const toolId = messageId || nextMessageId('tool')
// ...
const chatMessage: ChatMessage = { id: toolId, type: 'tool_use', ... }
```

So:
- Each tool in a multi-tool turn overwrites the previous tool's `ChatMessage` (same id → dedupe collision in the replay path).
- The follow-up `stream_start` for the same turn reuses the id again, potentially mutating an existing tool bubble into a text bubble (#stream_id_collision class of bug).

## Why not in #4257

#4257 is a one-line shape fix for the canonical `tool_start` keys. Switching to per-tool ids is a behavior change that needs its own regression tests (multi-tool turn fixture, replay-dedupe pinning, stream_start id-collision pinning).

## Acceptance Criteria

- [ ] `byok-session.js` `tool_start` uses the per-tool content_block id (or `t.toolUseId` if the SDK translator surfaces it) as `messageId`, with the `${messageId}-tool` fallback for parity
- [ ] Multi-tool turn fixture in `byok-session.test.js` asserts distinct `messageId`s per `tool_start`
[...]

PR #4381, opened 2026-05-28
fix(server): byok-session tool_start fallback sets toolUseId to match messageId

blamechris:
## Summary

- `byok-session.js` `tool_start` fallback now sets `toolUseId` to the same derived `${messageId}-tool` string as `messageId`, mirroring `sdk-session.js:635-641`.
- Pre-fix: `toolUseId: undefined` leaked through whenever the SDK omitted `content_block.id` — a silent schema violation (`ServerToolStartSchema.toolUseId: z.string()`), no runtime impact today because no broadcast-side validator enforces the contract.
- Extends the existing `#4262: falls back to \`${turnMessageId}-tool\`` test to also assert `typeof toolUseId === 'string'` and `toolUseId === messageId`.

## Diff

```diff
-              const toolId = t.toolUseId || `${messageId}-tool`
+              // #4364: Mirror sdk-session.js — reuse derived toolId for both fields
+              const toolId = t.toolUseId || `${messageId}-tool`
               this.emit('tool_start', {
                 messageId: toolId,
-                toolUseId: t.toolUseId,
+                toolUseId: toolId,
                 tool: t.toolName,
                 input: null,
               })
```

## Test plan

- [x] Test added asserts `toolUseId` is a string and equals `messageId` on the fallback path
- [x] Test fails on the pre-fix `toolUseId: undefined` shape
- [x] Existing #4262 multi-tool test stays green (non-fallback path unaffected)
- [x] Full `byok-session.test.js`: 70/70 pass

Closes #4364


blamechris:
## Code Review: PR #4381

### Summary

One-line parity fix for the `byok-session.js` `tool_start` fallback path: when the Anthropic SDK omits `content_block.id`, emit `toolUseId` equal to the derived `${messageId}-tool` string instead of `undefined`. Mirrors `sdk-session.js:635-641` and `cli-session.js:706-711`. Extends the existing `#4262: falls back to ${turnMessageId}-tool` test to pin the new shape.

### Strengths

- **Minimum-surface fix.** Two-line code change, no behavioral change on the happy path (`t.toolUseId` present), no broadcast/wire impact today.
- **TDD discipline.** Test was added/extended before the fix; failure was verified (`actual: 'undefined'`) before applying the change.
- **Test extends instead of duplicates.** Picks up the existing `#4262` fallback test rather than adding a parallel test that would drift on future refactors. Matches the acceptance criteria in #4364 exactly.
- **Schema-rooted rationale.** Both the code comment and the test docstring cite `ServerToolStartSchema.toolUseId: z.string()` as the contract — the *why* survives even if the fallback string format changes.
- **Cross-session parity is now total.** `byok-session.js`, `sdk-session.js`, and `cli-session.js` all converge on the same `toolId = ... || \`${messageId}-tool\`` → both fields shape. Future readers won't have to wonder which one is right.
- **Test isolation honored.** `tmpHome`/`HOME` swap in `beforeEach`/`afterEach` (lines 107-121) prevents the test from touching real `session-state.json` (per memory `feedback_test_state_contamination.md`).
- **No conflict with sibling marathon PRs.** PR #4378 (issue #4247, abort hardening) touches `byok-session.js` at lines ~695-770, this PR touches ~390-410 — different code paths. PR #4374 (issue #4363, strip stale toolName from stubs) touches `byok-session.test.js` at lines ~462, 511, 567, 619, 673, 940, 1545, 1629, 1710, 1783 — none overlap with the extended test at line 1050.

### Issues Found

#### Critical (Must Fix)
None.

None blocking.

#### Nitpicks (Optional)

| File | Line | Note |
|------|------|------|
| `packages/server/src/byok-session.js` | 385-400 | The new `#4364:` comment block sits directly under the existing `#4262:` block that already says "Mirrors sdk-session.js:635-641". Strictly speaking the two are different historical references (#4262 = "why use content_block.id at all"; #4364 = "why mirror both fields on the fallback"), and the comment is short. Defensible as-is. Optional micro-edit: collapse the two comments into a single paragraph that references both issue numbers, or drop the redundant "Mirror sdk-session.js" phrase from the #4364 block since #4262 already says it. Not worth a follow-up — call it on first edit only. |

None. This PR is single-purpose and the fix is complete.

### Architecture Notes

This PR closes the last gap in three-session tool_start fallback parity. The fix matters more for *future* changes than for today's runtime:
- **Today:** No broadcast-side validator runs `ServerToolStartSchema.parse(...)` on the outbound payload, so the `undefined` slip was silently dropped/serialized.
- **Future:** If/when issue #4141-style schema gating expands to broadcast events (or if a downstream consumer parses tool_start payloads with strict mode), the BYOK fallback path would have been the lone violator. Now it isn't.

The fix also tightens the invariant that `ws-server.js` and downstream renderers (dashboard `ToolBubble`, store-core `handleToolStart`) can rely on: `tool_start.toolUseId` is *always* a string. That assumption was previously safe-by-accident; now it's safe-by-design.

### Verdict

Single-purpose parity fix, two-line implementation, TDD'd, no observable runtime impact, no risk of regression on the happy path. Comments are explicit and cite the schema contract. Tests pass: 70/70 in `byok-session.test.js`.

PR #4390, opened 2026-05-28
fix(dashboard): focus sidebar row before opening context menu (#4372)

blamechris:
## Summary

- PR #4369 added focus restoration on `SessionContextMenu` unmount: the menu captures `document.activeElement` at open time and restores focus to it on Esc/outside-click. The gap was on the consumer side — a **right-click does not move focus by default**, so the captured "trigger" was whatever the user last clicked (often the composer textarea), and Esc landed focus there instead of the right-clicked row.
- `handleSidebarContextMenu` in `App.tsx` now calls `event.currentTarget.focus()` before opening the menu, so `SessionContextMenu`'s mount-effect capture lands on the row and unmount focus-restoration returns there.
- Added a `:focus-visible` outline on session, resumable, and repo-header sidebar rows so the focused row is visible after dismiss. The existing WAI-ARIA roving-tabindex pattern (focused row = `tabIndex={0}`, others = `-1`) already makes every row programmatically focusable — no Sidebar.tsx changes needed for that.

## Test plan

- [x] New Sidebar tests assert rows are focusable (`tabindex` attribute present) and that `event.currentTarget` in `onContextMenu` is the row.
- [x] New App tests cover (a) `handleSidebarContextMenu` invokes `focus()` on the row, and (b) end-to-end: composer textarea is focused → right-click row → Esc → focus lands on the row, not the composer (the regression the issue describes).
- [x] `npm test -- --run` in `packages/dashboard` — 2065/2065 pass.
- [x] `npm run typecheck` in `packages/dashboard` — clean.

Closes #4372

### DIFF
--- a/packages/dashboard/src/App.test.tsx
+++ b/packages/dashboard/src/App.test.tsx
@@ -1083,16 +1083,53 @@ describe('App', () => {
       const row = screen.getByTestId('session-item-s1')
       fireEvent.contextMenu(row)
       // Sanity: the menu is open and its first item should be focused.
-      expect(screen.getByRole('menu')).toBeInTheDocument()
+      const menu = screen.getByRole('menu')
+      expect(menu).toBeInTheDocument()
 
       // Esc dismisses the menu. SessionContextMenu listens at document scope
-      // for keydown — fire on the menu element so the event bubbles.
-      fireEvent.keyDown(document, { key: 'Escape' })
+      // for keydown today; firing on the menu element (rather than directly
+      // on `document`) lets the event bubble through the standard user-facing
+      // path, so this test stays green if the listener is later scoped to
+      // the menu itself.
+      fireEvent.keyDown(menu, { key: 'Escape' })
 
       // The menu should be gone, and focus restored to the row (NOT the
       // textarea, which would indicate the pre-#4372 bug).
       expect(screen.queryByRole('menu')).not.toBeInTheDocument()
       expect(document.activeElement).toBe(row)
+    })
+
+    // Parallel to the session-row case above, but for repo headers — the
+    // case the first pass of #4372 missed because the onContextMenu listener
+    // used to be on the inner `.sidebar-repo-header` (no tabIndex, no role)
+    // instead of the outer `.sidebar-repo` treeitem. Without the Sidebar
+    // structural fix, this test would fail: focus would stay on the textarea
+    // (because focus() on the unfocusable header is a no-op) and Esc would
+    // land focus on document.body or the textarea, not the treeitem.
+    it('returns focus to the repo treeitem when the menu is dismissed (Esc)', () => {
+      stateOverrides = connectedState
+      render(<App />)
+      const textarea = screen.getByRole('textbox', { name: /message input/i }) as HTMLTextAreaElement
+      textarea.focus()
+      expect(document.activeElement).toBe(textarea)
+
+      // App derives repos from sessions; our s1 session cwd is /tmp/repo,
+      // which produces a `repo-header-/tmp/repo` testid on the header.
+      const header = screen.getByTestId('repo-header-/tmp/repo')
+      const treeitem = header.closest('[role="treeitem"]') as HTMLElement | null
+      expect(treeitem).not.toBeNull()
+
+      // Right-click on the inner header. The listener now lives on the
+      // outer treeitem, so the contextmenu event bubbles up and the
+      // handler's event.currentTarget is the focusable treeitem.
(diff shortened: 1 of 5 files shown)


blamechris:
## Code Review: PR #4390

### Summary
Implements #4372 a11y fix: `handleSidebarContextMenu` now focuses `event.currentTarget` before opening `SessionContextMenu`, so PR #4369's unmount focus-restoration lands back on the right-clicked row instead of the composer textarea. Adds `:focus-visible` outlines on session/resumable/repo-header sidebar rows and includes both Sidebar-level (5) and App-level (2) tests covering the regression.

### Strengths

- **Tight scoping.** Single-line behavioral change + matching CSS + tests; no overreach. Builds cleanly on the existing roving-tabindex pattern instead of forcing `tabIndex={0}` on every row (which would have broken the WAI-ARIA tree pattern).
- **Test discipline.** The "Esc returns focus to row, not composer" App-level test reproduces the exact real-world regression #4372 describes — composer focused → right-click → Esc → focus on row. The Sidebar tests carefully document why `currentTarget` must be read synchronously (React event pooling).
- **Comment quality.** Inline comments cite both PRs (#4369 mount-effect capture, #4248 the original a11y issue) so a future reader can reconstruct the chain.
- **CSS pattern match.** `outline: 2px solid var(--border-focus)` mirrors the existing dashboard convention (`session-tab`, `view-tab`, `tool-group-header`, etc.).

### Issues Found

#### Critical (Must Fix)

| `packages/dashboard/src/components/Sidebar.tsx` | 360–367 | **Repo headers fail the fix.** `onContextMenu` is bound to `.sidebar-repo-header` (the inner clickable div), not the outer `<div className="sidebar-repo" role="treeitem" tabIndex={…}>`. So `event.currentTarget` for repo right-clicks is the header — which has **no `tabIndex` and no `role`** — and `.focus()` on it is a no-op. `document.activeElement` stays at `<body>`, and `SessionContextMenu` captures `<body>` as `previouslyFocused`. Net effect: Esc returns focus to body (worse than the pre-#4372 composer case, since no focus ring). I confirmed this by instrumenting a probe test: `currentTarget className: sidebar-repo-header / tabIndex attr: null / activeElement after focus: BODY`. | Either (a) move `onContextMenu` up to the outer `.sidebar-repo` treeitem wrapper, or (b) add `tabIndex={-1}` to `.sidebar-repo-header`. Option (a) is more consistent — session and resumable rows already have `onContextMenu` on the treeitem element, which is exactly why their case works. |

| `packages/dashboard/src/App.test.tsx` | 1085–1090 | The "Esc dismisses" assertion fires `keyDown` on `document` directly. `SessionContextMenu`'s keyboard listener attaches at `document` scope (line ~234 there), so this works — but a future refactor that scopes the listener to the menu element would silently break this test without changing the user-visible behavior. Consider `fireEvent.keyDown(screen.getByRole('menu'), { key: 'Escape' })` so the test exercises the actual user-facing path. | Defensive; not blocking. |
| `packages/dashboard/src/components/Sidebar.test.tsx` | 358–365 | The "passes currentTarget" test is a useful contract guard but doesn't actually catch the repo-header bug above (the test uses a session row). A parallel test for the repo header would have caught it. | Add a repo-header variant once the Critical fix lands. |
| `packages/dashboard/src/theme/components.css` | 600–606 | `.sidebar-repo:focus-visible` adds `border-radius: 3px` only on focus. If the wrapper has no border-radius normally, the outline corners and the content corners won't match. Verify visually that this looks right when a repo treeitem is keyboard-focused. | Minor visual polish. |

#### Nitpicks (Optional)

- None.

| Keyboard invocation of the context menu (ContextMenu key / Shift+F10) on focused sidebar rows | will create after Critical fix | Discoverable a11y gap noticed during this review but out of scope for #4372 (which is solely about right-click focus); deserves its own ticket. |

### Architecture Notes

The fix slots cleanly into the focus contract established by #4369: `SessionContextMenu`'s mount-effect captures `document.activeElement` and the unmount cleanup restores it. The right place for the consumer-side `.focus()` is the open path of `handleSidebarContextMenu` — that's what this PR does, and there's no alternative location that would work.

The roving-tabindex tree pattern means every row already has an explicit `tabIndex` (0 for focused, -1 for others). Programmatic `.focus()` works on `tabIndex={-1}` elements, so the existing Sidebar.tsx structure was already fit for purpose **for session and resumable rows**. The repo-header case (Critical above) is the one place this assumption breaks because the listener is on a child of the treeitem.

### Verdict

- [x] Request Changes - Repo-header case must be fixed; the rest is polish

Issue #3138, opened 2026-04-28
Replace hand-rolled handler validation with Zod schemas across @chroxy/store-core


blamechris:
## Context

Identified during review of PR #3136.

## Description

Many `handle*` functions in `packages/store-core/src/handlers/index.ts` perform hand-rolled `typeof === 'string' && trim()` validation against incoming WS message payloads. The repo already uses Zod schemas in `@chroxy/protocol` for outgoing/incoming message envelopes. We should consolidate handler-side validation behind those (or new) Zod schemas to:

1. Remove duplicated `typeof` checks
2. Catch protocol drift earlier (server adds a field; schema either tolerates or rejects)
3. Improve TypeScript narrowing inside handlers

This is cross-cutting and should be tackled as a sweep across all handlers, not piecemeal per migration. PR #3136 is in scope for the migration but not the schema refactor.

## Original Review Comment

> Add a Zod schema (or move into the existing `protocol` schemas) for the `available_models` envelope. The hand-rolled `typeof === 'string' && trim()` validation is idiomatic in this file but is a near-duplicate of what Zod could express in 5 lines.

## Acceptance Criteria

- [ ] Audit existing `handle*` functions and decide which payloads warrant Zod schemas
- [ ] Add schemas in `@chroxy/protocol` (or `@chroxy/store-core` if they're handler-internal shapes)
- [ ] Refactor handlers to parse via Zod and remove the hand-rolled validation
- [ ] Tests updated; `handlers.test.ts` continues to pass
- [ ] No behaviour change for valid inputs; invalid inputs continue to return the same fallback shape

blamechris:
Addressed in PR #3180 (commit `d4e465351`) — included in the 16-issue from-review batch cleanup. GitHub's auto-close only matched the first issue after the `Closes` keyword; closing manually with cross-reference.

PR #4416, opened 2026-05-28
fix(server): track pending background shells so waiting sessions aren't reaped (#4307)

blamechris:
Closes #4307

## Summary

Chroxy now models backgrounded `Bash` shells (`run_in_background: true`) per session so a session waiting on background work is **distinct from idle** and is **never reaped** by `CHROXY_SESSION_TIMEOUT`. The dashboard / app get both a live `background_work_changed` event and a `pendingBackgroundShells` field on every `session_list` entry so late joiners catch up without a delta protocol.

This is the server / lifecycle half of #4307 (companion #4308 — the dashboard activity indicator — already shipped via PR #4399).

### Server
- New `background-shells.js` helper: parses the canonical `Command running in background with ID: <id>` tool_result text, recognises `Bash` + `run_in_background: true`, and `BashOutput`'s `bash_id`.
- `BaseSession`:
  - Adds `_pendingBackgroundShells: Map<shellId, { shellId, command, startedAt }>` and an ephemeral `_pendingBackgroundCommands` (tool_use_id → command).
  - `trackBackgroundShell` / `clearBackgroundShell` / `getPendingBackgroundShells` with idempotent semantics; both push + clear emit `background_work_changed` carrying the full snapshot.
  - **`isRunning` now returns `_isBusy || pendingBackgroundShells.size > 0`** — the canonical change that prevents `SessionTimeoutManager` reaping.
  - `_clearMessageState` deliberately does **not** clear pending shells (the whole point — they survive turn-end). The only paths that clear are matching `BashOutput` or `destroy()`.
- `SdkSession`: `_handleToolUseBlock` stashes the command on `run_in_background:true` and clears on `BashOutput`; new `_recordBackgroundShellsFromToolResults` registers the shell when the matching tool_result lands. `destroy()` calls `_destroyPendingBackgroundShells`.
- `ClaudeTuiSession`: parity via `_emitToolHookEvent` — `PreToolUse` stashes / handles `BashOutput`, `PostToolUse` parses the canonical shellId text.
- `SessionManager`:
  - `background_work_changed` added to `builtinTransient` so the event proxies transparently.
  - `listSessions()` includes `pendingBackgroundShells: session.getPendingBackgroundShells()` (defaults to `[]` for providers without the getter).
- `event-normalizer.js`: new normaliser route for `background_work_changed` that emits the wire message **and** a `session_list` side effect (refreshes the SessionInfo snapshot for clients that render off the list).
- `ws-server.js`: documents the new event in the Server -> Client doc-comment (satisfies `handler-coverage.test.js`).

### Protocol
- New `ServerPendingBackgroundShellSchema` (`shellId`, `command`, `startedAt`).
- New `ServerBackgroundWorkChangedSchema` (`sessionId`, `pending[]`).
- `ServerSessionListEntrySchema.pendingBackgroundShells` optional array.

### store-core
- New `PendingBackgroundShell` type.
- `BaseSessionState.pendingBackgroundShells: PendingBackgroundShell[]` — defaulted to `[]` in `createEmptyBaseSessionState`.
- `SessionInfo.pendingBackgroundShells?` (optional for pre-#4307 servers).
- New `handleBackgroundWorkChanged` builder (snapshot replacement) with reference-equality short-circuit when next === current.

### Dashboard + App
- Both message-handlers dispatch `background_work_changed` to the shared handler and seed the slot from `session_list`. Dashboard adds a stable `EMPTY_PENDING_BACKGROUND_SHELLS` for the flat-state fallback (matches the #4308 pattern).

## Design decisions

- **Event + snapshot field (both)** — per the issue's "recommend doing both". Late joiners get `pendingBackgroundShells` on every `session_list` snapshot; live updates flow through `background_work_changed`.
- **Full snapshot on each event (no delta protocol)** — pending sets are tiny (typically 0 or 1 entries) and events fire rarely; the wire cost is negligible and the consumer logic is a flat replace.
- **SDK no-next-turn behaviour: entry persists until `destroy()` or a matching `BashOutput` call.** Documented as the issue requested. We deliberately do NOT try to tail the output file ourselves — that's a separate sidecar lifecycle problem the issue body explicitly flagged as out of scope.
- **TUI parity shipped in this PR.** The PreToolUse / PostToolUse hook flow already exists for tool_start / tool_result; layering the same parser surface was small and avoided a split-PR follow-up.
- **Pending shells survive `_clearMessageState` (turn-end) by construction** — clearing them there would defeat the whole purpose of #4307. The only clear paths are the agent's `BashOutput` (acknowledged the shell) or `destroy()` (no leak).
- **No per-turn time decay.** The agent owns the lifecycle signal; if the agent forgets to `BashOutput`, the entry sits there until destroy. That's the documented contract.

## Test plan

- [x] `node --test packages/server/tests/background-shells.test.js` — 21 new tests
- [x] `node --test packages/server/tests/sdk-session.test.js` — 8 new SDK-side tests
- [x] `node --test packages/server/tests/claude-tui-session.test.js` — 4 new TUI-side tests
- [x] `node --test packages/server/tests/session-manager.test.js` — 2 new listSessions tests
- [x] `vitest run packages/store-core/src/handlers/handlers.test.ts` — 9 new handler tests
- [x] `tsc --noEmit` clean across protocol / store-core / dashboard
- [x] Full server test suite: 5739/5745 pass (the 2 failures — `WebTaskManager` feature detection and `encryption integration` — are pre-existing flaky tests; both pass when run standalone)
- [x] Full dashboard suite: 2171/2171 pass
- [x] Full store-core suite: 865/865 pass
- [x] Full app suite: 1299/1299 pass
- [x] Full protocol suite: 139/139 pass (handler-coverage contract enforced)

Key invariants pinned by the new tests:
- Parser: canonical text → shellId; whitespace-only / non-matching → null
- `isRunning` is `true` when `_isBusy=false` AND `pendingBackgroundShells.size > 0`
- `_clearMessageState` (turn-end) does NOT clear pending shells
- `destroy()` clears the pending map (no leak)
- `SessionTimeoutManager` skips waiting sessions
- `background_work_changed` event fires on push AND clear
- `session_list` snapshot includes `pendingBackgroundShells`
- TUI + SDK paths produce parity behaviour

### DIFF
--- a/packages/app/src/store/message-handler.ts
+++ b/packages/app/src/store/message-handler.ts
@@ -91,6 +91,7 @@ import {
   handleGitCommitResult as sharedGitCommitResult,
   handleAgentSpawned as sharedAgentSpawned,
   handleAgentCompleted as sharedAgentCompleted,
+  handleBackgroundWorkChanged as sharedBackgroundWorkChanged,
   handleAvailableModels as sharedAvailableModels,
   handleMcpServers as sharedMcpServers,
   handleCostUpdate as sharedCostUpdate,
@@ -1219,6 +1220,24 @@ export function handleMessage(raw: unknown, ctxOverride?: ConnectionContext): vo
             });
           }
         }
+        // #4307: seed pendingBackgroundShells from the snapshot so the
+        // app catches up on any sessions already waiting on background
+        // work without waiting for the next `background_work_changed`
+        // event. The shared handler's reference-equality short-circuit
+        // suppresses no-op re-renders.
+        for (const s of sessionList) {
+          if (!get().sessionStates[s.sessionId]) continue;
+          const builder = sharedBackgroundWorkChanged(
+            { sessionId: s.sessionId, pending: s.pendingBackgroundShells ?? [] },
+            get().activeSessionId,
+          );
+          updateSession(s.sessionId, (ss) => {
+            const next = builder.applyTo(ss.pendingBackgroundShells);
+            return next === ss.pendingBackgroundShells
+              ? {}
+              : { pendingBackgroundShells: next };
+          });
+        }
         // Persist the active session's conversationId for auto-resume on server restart.
         // Use the activeSessionId if set, otherwise fall back to the first session in the list.
         const resolvedActiveId = get().activeSessionId ?? (sessionList[0]?.sessionId ?? null);
@@ -1777,6 +1796,22 @@ export function handleMessage(raw: unknown, ctxOverride?: ConnectionContext): vo
         updateSession(builder.sessionId, (ss) => {
           const next = builder.applyTo(ss.activeAgents);
           return next === ss.activeAgents ? {} : { activeAgents: next };
+        });
+      }
+      break;
+    }
+
+    case 'background_work_changed': {
+      // #4307 — pending-background-shells snapshot updated for a
+      // session. Full-snapshot protocol; the shared handler returns
+      // the same reference when next === current to skip re-renders.
+      const builder = sharedBackgroundWorkChanged(msg, get().activeSessionId);
+      if (builder.sessionId && get().sessionStates[builder.sessionId]) {
+        updateSession(builder.sessionId, (ss) => {
+          const next = builder.applyTo(ss.pendingBackgroundShells);
+          return next === ss.pendingBackgroundShells
+            ? {}
+            : { pendingBackgroundShells: next };
         });
       }
       break;
--- a/packages/dashboard/src/store/connection.ts
+++ b/packages/dashboard/src/store/connection.ts
@@ -208,6 +208,9 @@ const EMPTY_DEV_PREVIEWS: never[] = [];
 // #4308: stable empty reference for `activeTools` in the flat-state
 // fallback. Same `useShallow` stability rationale as the others above.
 const EMPTY_ACTIVE_TOOLS: never[] = [];
+// #4307: stable empty reference for `pendingBackgroundShells` —
+// same `useShallow` stability rationale as the others above.
+const EMPTY_PENDING_BACKGROUND_SHELLS: never[] = [];
 
 /** Delay before auto-reconnecting after an unexpected socket close (ms) */
 const AUTO_RECONNECT_DELAY = 1500;
@@ -498,6 +501,9 @@ export const useConnectionStore = create<ConnectionState>((set, get) => ({
       // #4308: parity with the BaseSessionState shape; no live tool tracking
       // in flat-state fallback (only populated once a real SessionState lands).
       activeTools: EMPTY_ACTIVE_TOOLS,
+      // #4307: parity with the BaseSessionState shape; no live
+      // background-shell tracking in the flat-state fallback.
+      pendingBackgroundShells: EMPTY_PENDING_BACKGROUND_SHELLS,
       isPlanPending: false,
       planAllowedPrompts: EMPTY_PROMPTS,
       primaryClientId: null,
--- a/packages/dashboard/src/store/message-handler.ts
+++ b/packages/dashboard/src/store/message-handler.ts
@@ -74,6 +74,7 @@ import {
   handleGitStatusResult as sharedGitStatusResult,
   handleAgentSpawned as sharedAgentSpawned,
   handleAgentCompleted as sharedAgentCompleted,
+  handleBackgroundWorkChanged as sharedBackgroundWorkChanged,
   handleEnvironmentList as sharedEnvironmentList,
   handleEnvironmentError as sharedEnvironmentError,
   handleAvailableModels as sharedAvailableModels,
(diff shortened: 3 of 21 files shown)


blamechris:
## Code Review: PR #4416

### Summary

Implements the server-side half of #4307: tracks `Bash` shells dispatched with `run_in_background: true` per session so a waiting session is treated as not-idle (preventing `CHROXY_SESSION_TIMEOUT` reaping) and surfacing the pending list to clients via a new `background_work_changed` event + a `pendingBackgroundShells` field on the `session_list` snapshot. Parity across SDK and TUI providers, store-core handler + type, dashboard + app dispatch + late-joiner seed. 44+ new tests across 5 test files.

### Strengths

- **Architectural fit is excellent.** The `_pendingBackgroundShells` Map lives on `BaseSession` so both providers share one state model, and `isRunning` is the single integration point with `SessionTimeoutManager` — no new wiring needed in `session-manager.js`'s skip-check path. Mirrors the `_activeAgents` pattern from prior work.
- **Parser is correctly conservative.** `[A-Za-z0-9_-]+` (not `\S+`) avoids swallowing trailing punctuation, and the explicit comment about that decision is exactly the kind of footnote future maintainers need.
- **The "survives turn-end" invariant is pinned in code AND in a test** (`_clearMessageState does NOT clear pending shells`). This is the load-bearing #4307 semantic and would be easy to regress; the test is well-placed.
- **TUI parity in the same PR** keeps the user-facing behaviour identical regardless of provider — avoids a follow-up issue and the temporary "works on SDK, broken on TUI" gap the issue body warned about.
- **`_recordBackgroundShellsFromToolResults` runs on the FULL raw block content**, not the post-truncation result string. So even if a `tool_result` exceeds `MAX_TOOL_RESULT_SIZE` the shellId still registers (the canonical message lives at the front of the response, so this is mostly defensive — but correctly so).
- **Idempotent `trackBackgroundShell`**: re-registering the same shellId is a no-op AND does NOT re-emit `background_work_changed`. Avoids broadcast amplification on duplicate `tool_result` events.
- **Defensive parsing in `handleBackgroundWorkChanged`**: malformed entries are filtered out fail-soft so one bad row can't make the whole list disappear. Matches the established Zod-handler resilience pattern (#3138).
- **Late-joiner seed is wired in BOTH dashboard and app** message handlers, off the same `session_list` event — clients that reconnect mid-flight catch up without needing a fresh `background_work_changed`.
- **Documentation is thorough**: every code addition has rationale comments tying back to #4307 and explaining WHY (e.g. why pending shells survive `_clearMessageState`, why no delta protocol).
- **Test coverage is genuinely deep**: tests for the actual integration paths (SDK tool_use → tool_result flow, TUI Pre + Post hook pair, SessionTimeoutManager skip check, listSessions snapshot shape).

### Issues Found

#### Critical (Must Fix)

None.

| `packages/server/src/event-normalizer.js` | 158-167 | Each `background_work_changed` triggers a full `session_list` side-effect broadcast. If an agent rapid-fires N background shells in one turn, that's N+1 session_list re-broadcasts to ALL clients (each encrypted per-client). Consider debouncing or dropping the side-effect — clients subscribed to the event already get the canonical pending list. | Steady-state cost is negligible; pathological case (10+ backgrounds in one turn) could thrash mobile clients on slow links. Same trade-off `permission_resolved` / `inactivity_warning` already make, so this is consistent — but session_list payload is heavier. |
| `packages/server/src/base-session.js` | 596-607 | `_destroyPendingBackgroundShells` clears the map but does NOT emit `background_work_changed`. If a session is destroyed administratively, dashboards subscribed to the event don't see the clear. | The `session_destroyed` event removes the whole session state from the client, so this is correct in practice. Leaving as-is is fine; the rationale is worth a one-line comment so future readers don't add a spurious emit. |
| `packages/server/src/session-manager.js` (persistence) | — | `pendingBackgroundShells` is not persisted to `~/.chroxy/session-state.json`. A server restart loses the pending list. | Tracked in #4417. |

#### Nitpicks (Optional)

- `packages/server/src/base-session.js:604` — `_destroyPendingBackgroundShells` is invoked from subclass `destroy()` (not internal), so the `_` prefix is misleading. Either rename to `destroyPendingBackgroundShells` (public) or accept the convention. Trivial.
- `packages/store-core/src/handlers/index.ts:2340-2356` — the same-reference short-circuit could be extracted to a small helper if other handlers acquire the same shape. Not worth doing in this PR.
- `packages/server/tests/sdk-session.test.js` — one test calls `session = createSession()` after `session.destroy()` so the `afterEach` doesn't double-destroy. A cleaner pattern would be a `keepDestroyed` flag, but the inline recreation is fine.

| Persist `pendingBackgroundShells` across server restart, OR document as transient | [#4417](https://github.com/blamechris/chroxy/issues/4417) | Worth tracking — restart semantics are user-visible. Out of scope for the lifecycle fix. |
| ActivityIndicator (dashboard) renderer integration to surface pending background shells in the chip | [#4418](https://github.com/blamechris/chroxy/issues/4418) | UI half — companion to this lifecycle work. The existing `TODO(#4307)` marker in `ActivityIndicator.tsx:223-229` already flags it. |

### Architecture Notes

This PR cleanly slots into the existing BaseSession state machine. The chain is:

1. Provider's tool_use handler stashes the command in `_pendingBackgroundCommands` (turn-local).
2. Provider's tool_result handler calls `trackBackgroundShell(...)` — promotes to `_pendingBackgroundShells` (session-lifetime).
3. `_clearMessageState` clears turn-local, preserves session-lifetime — the #4307 invariant.
4. `isRunning` reflects both `_isBusy` and pending size — feeds into `SessionTimeoutManager`'s skip path with zero new wiring.
5. `background_work_changed` event flows through `EventNormalizer` → `ws-forwarding.js` → clients with a `session_list` side-effect for snapshot reconciliation.
6. Clients dispatch via shared `handleBackgroundWorkChanged` to the `pendingBackgroundShells` slot on `BaseSessionState`.

The state slot mirrors `activeTools` (#4308) but with the critical distinction that pending shells survive turn-end while active tools clear on `agent_idle`/`result`. The TODO comment that #4308 left in `BaseSessionState.activeTools` is now actionable for the renderer follow-up.

### Verdict

**Approve** — Ready to merge.

This is a textbook #4307 implementation: the lifecycle gap is closed, the protocol surface is small and well-modelled (event + snapshot field, both), parity ships in the same PR, and the test coverage is solid against the genuinely tricky cases (turn-end-preservation, destroy-clears, isRunning-when-busy-and-pending, snapshot equality short-circuits). The two follow-ups (restart persistence, dashboard renderer) are real but properly deferred.

The pre-existing flaky test failures (`WebTaskManager` feature detection + `encryption integration` end-to-end) noted in the PR body are unrelated to these changes and pass standalone.

PR #4424, opened 2026-05-29
docs(server): document pendingBackgroundShells as transient (#4417)

blamechris:
## Summary

Closes #4417.

PR #4416 (which closed #4307) added `_pendingBackgroundShells: Map<shellId, {shellId, command, startedAt}>` to `BaseSession`. The map is in-memory only — it is not written to `~/.chroxy/session-state.json` and is dropped on server restart. #4417 asked us to either persist it or explicitly document why we don't.

This PR takes the document-as-transient path (option A in #4417). Rationale captured in the new JSDoc:

- OS-level background shells are owned by the claude TUI / SDK runtime, not chroxy. After a restart their PIDs are either dead or orphaned beyond chroxy's ability to correlate — we never held the PID, only the opaque shellId Claude printed.
- Persisting the map would re-surface a "waiting on `<command>`" activity chip that chroxy can never clear, because the shellId is meaningless to the new claude process. That's a worse UX than briefly showing idle until the agent's next turn re-syncs reality.
- Documenting the choice in-code stops a future PR from re-adding "persist for safety" without first solving the claude-side reconciliation problem.

## Changes

- `packages/server/src/base-session.js` — extended the JSDoc block on `_pendingBackgroundShells` with a `#4417: TRANSIENT BY DESIGN` section explaining the non-persistence rationale and the user-facing consequence (briefly-stale activity chip, not data loss).
- `packages/server/src/background-shells.js` — added a short module-level pointer to the BaseSession comment so the helper module is self-documenting too.

## Test plan

- [x] `node --test tests/background-shells.test.js tests/base-session.test.js` — 115 tests pass
- [x] Diff is comment-only, no behaviour change

### DIFF
--- a/packages/server/src/background-shells.js
+++ b/packages/server/src/background-shells.js
@@ -15,6 +15,13 @@
  * We do not depend on the format of the surrounding sentence — only on
  * the canonical `Command running in background with ID: <token>` opener,
  * which has been stable across the claude versions chroxy targets.
+ *
+ * #4417: the pending-shells map populated via these helpers lives in
+ * memory only and is intentionally NOT persisted across server
+ * restart. See `BaseSession._pendingBackgroundShells` for the full
+ * rationale — short version: the OS-level shells are owned by claude,
+ * not chroxy, so a restored map would be stale and the activity
+ * indicator would lie. Dropping the map on restart is the safe choice.
  */
 
 // `[A-Za-z0-9_-]+` is intentionally tight — every shell id observed so
--- a/packages/server/src/base-session.js
+++ b/packages/server/src/base-session.js
@@ -204,6 +204,28 @@ export class BaseSession extends EventEmitter {
     // sidecar lifecycle problem, out of scope).
     // TUI provider: same parity — the agent's next turn surfaces the
     // BashOutput call and clears the entry naturally.
+    //
+    // #4417: TRANSIENT BY DESIGN — this map is in-memory only and is
+    // NOT serialized to `~/.chroxy/session-state.json`. A server
+    // restart drops every entry. This is the correct behaviour, not an
+    // oversight:
+    //   - The actual OS-level background shells are owned by the
+    //     claude TUI / SDK runtime, not by chroxy. On server restart
+    //     those processes have either died with the parent or have
+    //     been orphaned beyond chroxy's ability to correlate (we never
+    //     held the PID, only the opaque shellId Claude printed).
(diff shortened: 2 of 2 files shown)


blamechris:
## Code Review: PR #4424

### Summary

Comment-only change. Extends the JSDoc on `BaseSession._pendingBackgroundShells` and the header of `background-shells.js` to explicitly document the in-memory / non-persisted lifetime introduced in #4416 (closes #4307). Closes #4417 by choosing path (a) — document as transient — over path (b) — persist + reconcile.

### Strengths

- **Correct verdict.** The rationale is sound: chroxy never owns the OS-level background shell PIDs; Claude's TUI / SDK runtime owns them. ShellIds are opaque tokens minted by Claude per-process, so after a restart the new claude has no way to reuse them. A persisted map would re-surface an activity chip that no `BashOutput` call from the post-restart agent could ever clear — a permanent stuck-indicator UX bug. Dropping the map on restart is strictly better than restoring stale state.
- **Anti-footgun.** The "if a future change ever needs to persist this state, it MUST also reconcile with the post-restart claude side — blindly restoring a stale Map will lock the indicator on forever" line is exactly the guard rail #4417 asked for.
- **Style match.** Inline block comment inside the constructor, in the same shape as the existing `_pendingBackgroundCommands` and `_messageIdPrefix` blocks (`base-session.js:208-216`, `:218-229`). No new convention introduced.
- **Cross-reference.** The `background-shells.js` module header points to `BaseSession._pendingBackgroundShells` for the full rationale — keeps the canonical explanation in one place without leaving the helper module silent.
- **No semicolons, single quotes, ES module style** preserved (vacuously — comments only).

### Issues Found

#### Critical (Must Fix)

None.

None — the diff is targeted and minimal.

#### Nitpicks (Optional)

None worth raising.

None.

### Architecture Notes

This PR settles a design question rather than adding behaviour. The transient-by-design choice is consistent with how chroxy handles other "agent-side ownership" state — chroxy mirrors what Claude tells it but does not try to outlive the underlying process. Persisting `_pendingBackgroundShells` would have inverted that contract for a single field and introduced a class of stale-state bugs (the comment correctly calls these out).

The two existing clear paths — matching `BashOutput` and `destroy()` — remain the only ways an entry leaves the map. Server restart implicitly goes through the latter (process exit), so the in-memory lifetime ends cleanly.

### Verdict

Comment-only, no behaviour change, 115 related tests still pass (`tests/background-shells.test.js`, `tests/base-session.test.js`).

PR #4479, opened 2026-05-29
fix(server): docker-namespaced log on exec exit + tighten respawn race test (#4473, #4474)

### LINKED ISSUE #4474: cli-session test: harden _killAndRespawn double-listener race coverage (#4472 follow-up)
## Context

#4472's test at \`packages/server/tests/cli-session.test.js:399\` pins the critical invariant that \`_killAndRespawn\` emits \`stream_end\` + \`result\` BEFORE setting \`_respawning=true\`. That's the load-bearing fix for #4471.

What the test does NOT cover: in production, \`_spawnPersistentProcess\` attaches \`child.on('close', (code) => this._handleChildClose(code))\` at line 311. So \`oldChild\` carries that listener when \`_killAndRespawn\` runs. \`_killAndRespawn\` then also attaches its own closure callback at line 972. When \`oldChild\` actually closes, BOTH listeners fire — \`_handleChildClose\` (which sees \`_respawning=true\` and short-circuits at line 1077) AND the closure \`respawn()\`.

The test's mock child has no \`_handleChildClose\` listener attached (because \`_spawnPersistentProcess\` is never called in unit-test setup). So the \`oldChild.emit('close', 0)\` at test line 421 only invokes the closure, never the \`_handleChildClose\` short-circuit path. The race the \`_respawning\` guard exists to protect against is never exercised.

## Impact

A future refactor of \`_handleChildClose\` that drops the \`_respawning\` guard (or moves the helper call above it) would silently double-emit \`result\` on every \`_killAndRespawn\` and the test suite wouldn't catch it.

## Suggested approach

Either:

1. **Attach the real listener**: In the test, after \`createReadySession\`, do \`session._child.on('close', (code) => session._handleChildClose(code))\` (mirroring \`_spawnPersistentProcess\`). Then \`oldChild.emit('close', 0)\` triggers BOTH listeners. Assert that no SECOND \`result\` event was observed.
2. **Explicit second-call assertion**: After the closure runs (\`oldChild.emit('close', 0)\`), manually invoke \`session._handleChildClose(0)\` and assert the event-stream snapshot is unchanged. Less natural but more surgical.

(1) more closely mirrors production wiring.

## Reference

PR #4472 — review comment thread.

blamechris:
## Summary

Two follow-ups to the #4472 agent review, bundled because both are tiny and tightly related.

### #4473 — Restore docker-specific log namespace

#4472 deleted the docker close handler and made DockerSession inherit \`_handleChildClose\` from CliSession. Correctness-wise that was right — but it regressed the log line operators rely on for \`docker logs <ctr>\` correlation:

| Before #4472 | After #4472 | After this PR |
|--------------|-------------|---------------|
| \`[docker-session] Container exec exited\` | \`[cli-session] Process exited\` | \`[docker-session] Container exec exited\` + \`[cli-session] Process exited\` |

The override lands the container-specific line first, then \`super._handleChildClose(code)\` runs the real recovery path. Guarded against firing during destroy / respawn so we don't double-log the supervisor restart path.

### #4474 — Tighten the _killAndRespawn race test

The existing test pinned the emit ordering (\`stream_end\` + \`result\` before \`_respawning=true\`) but never exercised the actual race the guard protects against. In production, \`_spawnPersistentProcess\` attaches \`_handleChildClose\` as a child-close listener, so when \`_killAndRespawn\` fires the \`oldChild\` carries TWO listeners. The old test's mock child had none — so the guard's short-circuit at \`cli-session.js:1077\` was untested.

New test attaches the real listener (mirroring production wiring) and asserts no double \`result\` emit when the close event fires:

\`\`\`js
oldChild.on('close', (code) => session._handleChildClose(code))
// ...
session._killAndRespawn()
oldChild.emit('close', 0)  // BOTH listeners fire
assert.equal(results.length, 1)  // guard MUST short-circuit
\`\`\`

A future refactor that drops the \`_respawning\` guard would now flag.

## Test plan

- [x] \`node --test --test-name-pattern \"_killAndRespawn|delegates to CliSession|inherits CliSession close-handler\"\` — 15/15 pass
- [x] Existing #4469 behavioural test (\"emits stream_end + result + error on mid-turn close via inherited handler\") still passes — confirms the override correctly delegates

## Closes

- Closes #4473
- Closes #4474

## Related

- #4472 — the original sibling-bug fixes that triggered these follow-ups

### DIFF
--- a/packages/server/src/docker-session.js
+++ b/packages/server/src/docker-session.js
@@ -281,6 +281,17 @@ export class DockerSession extends CliSession {
     }
   }
 
+  // #4473: log the container-specific exit line under the docker-session
+  // namespace before delegating to the inherited handler (#4469). The
+  // inherited code emits its own "Process exited" line; this override
+  // lands first so `docker logs <ctr>` correlation by operators stays easy.
+  _handleChildClose(code) {
+    if (!this._destroying && !this._respawning) {
+      log.info(`Container exec exited (code ${code}), scheduling respawn`)
+    }
+    super._handleChildClose(code)
+  }
+
   /**
    * Destroy the session: stop the exec process, remove the container,
    * then call super.destroy() to clean up CliSession state.
--- a/packages/server/tests/cli-session.test.js
+++ b/packages/server/tests/cli-session.test.js
@@ -447,32 +447,49 @@ describe('CliSession._killAndRespawn (#4471: panic-button drops dashboard recove
   // _killAndRespawn fires, the oldChild carries BOTH listeners: the
   // production one (which would re-emit result if the `_respawning` guard
   // were ever dropped) and the closure-scoped respawn() callback.
-  // This test attaches the real listener and confirms that emit('close')
-  // does NOT double-emit `result`.
+  //
+  // Crucially: we do NOT set _destroying. The `_destroying` check in
+  // _handleChildClose precedes the `_respawning` check (cli-session.js:
+  // 1117-1118), so using `_destroying` as a respawn-suppression hatch
+  // would mask the `_respawning` guard the comment claims to pin (#4480).
+  // Instead we stub `start()` to a no-op so the closure respawn()
+  // callback doesn't actually spawn a child in the unit test.
   it('does NOT double-emit result when the production close listener also fires (#4474)', () => {
     const session = createReadySession()
     session._isBusy = true
     session._currentMessageId = 'msg_dl'
     session._currentCtx = { hasStreamStarted: true }
     session._sessionId = 'sess_dl'
-    session._destroying = true
+    // Stub spawn so the closure respawn() callback inside _killAndRespawn
+    // doesn't actually try to spawn a real `claude` process.
+    session.start = () => {}
 
     const oldChild = session._child
     // Mirror _spawnPersistentProcess's wiring — this is what makes the
-    // _respawning guard at cli-session.js:1077 load-bearing.
+    // _respawning guard at cli-session.js:1118 load-bearing.
     oldChild.on('close', (code) => session._handleChildClose(code))
 
     const results = []
+    const errors = []
     session.on('result', (p) => results.push(p))
+    session.on('error', (p) => errors.push(p))
 
     session._killAndRespawn()
-    // Now emit 'close' — BOTH listeners fire: the closure respawn() AND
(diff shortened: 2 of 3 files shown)


blamechris (commented):
## Chroxy Inspector review — PR #4479

**Verdict: APPROVE with one test-tightening follow-up.**

Tiny, focused bundle. Both changes are correct.

### #4473 — DockerSession override

Looks good. Override logs `[docker-session] Container exec exited (code X)` then delegates to `super._handleChildClose(code)`. Guard `if (!this._destroying && !this._respawning)` mirrors the inherited handler's own short-circuits, so the docker line is suppressed exactly when the inherited code is also a no-op — no spurious operator log during destroy/respawn paths.

Order is right: log first, super second. If super were ever to throw (it can't today — all sync emits), the docker breadcrumb is already out for `docker logs <ctr>` correlation. Prototype-chain delegation composes cleanly with any future CliSession override.

The updated `docker-session.test.js` assertion (`notStrictEqual`) correctly captures the new invariant, and the behavioural test below it still passes through the delegated handler (verified: `['stream_end', 'result', 'error']` fires unchanged, docker log line precedes).

### #4474 — Race test for `_respawning` guard

The intent is right, but the test as written doesn't actually isolate the guard it claims to pin. See #4480 — the test sets `_destroying=true` to suppress respawn, but `_handleChildClose` checks `_destroying` BEFORE `_respawning`, so the manually-attached listener short-circuits on `_destroying` and never reaches the `_respawning` guard. The assertion `results.length === 1` would still pass if the `_respawning` guard were deleted.

That said, the assertion still proves the load-bearing invariant ("no double `result` emit when both listeners fire"), and the new test does cover ground the original didn't (real production-wired listener now exercised). Net positive — just file the follow-up to harden the isolation later.

### Reliability lens (cellular + tunnel)

No new failure modes. The added log is an extra `log.info` per container exec exit; negligible. The override's guard means we don't spam logs during the supervisor's intentional restart paths.

### Test plan

- [x] `node --test --test-name-pattern="_killAndRespawn|delegates to CliSession|inherits CliSession close-handler"` → 15/15 pass locally on Node 22

### Follow-ups

- #4480 — tighten #4474 test so `_destroying` doesn't mask the `_respawning` guard it claims to pin

### Findings

- 1 minor (test isolation, filed as #4480)
- 0 blocking

Issue #3758, opened 2026-05-11
Surface agent progress vs timeout — show "still working" with elapsed idle time so users can distinguish slow turns from frozen ones


blamechris:
## Problem

When an agent turn runs longer than expected, today's UI gives users no way to tell:

- **(a)** The agent is genuinely working — streaming tokens, calling tools, making progress
- **(b)** The agent has stalled — SDK hung, network wedged, last token was N seconds ago

Both look identical until the server's hardcoded inactivity timeout fires (`RESULT_TIMEOUT_MS = 300_000` in `packages/server/src/sdk-session.js:514` — see #3749 / #3754 for the configurable-default work) and the user sees "Response timed out".

Real dogfood incident (this session): the desktop agent was mid-tool-loop on a long marathon task. Process died at the 5-min mark. From the chat UI's perspective it just stopped — no indication whether it was still working at minute 4:59 or whether it had stalled at minute 0:30 and the timer finally caught up.

## Signals already available server-side

The server emits enough events to drive a "last activity" indicator:

- `stream_start` / `stream_delta` / `stream_end` — token streaming
- `tool_start` / `tool_result` — tool calls
- `task_started` / `task_progress` / `task_notification` — Anthropic-side task lifecycle (visible in `chroxy.log`)
- `result` — turn completion
- `busy=true/false` in `/diagnostics` snapshot

The CLI server also has `busy=true mode=auto pause=false count=0` in the diagnostics snapshot for the active session.

## Proposed UX

A non-intrusive footer/banner on the chat composer area, **only shown while the active session is busy**:

\`\`\`
🟢 Working… last activity 12s ago
\`\`\`

- Updates every second. Resets on any incoming server event for the active session.
- Color escalates as the silent gap widens, mirroring the inactivity-timeout window:
  - 0–30s: green
  - 30–60s: amber
  - 60s–(timeout − 60s): orange
  - last 60s before timeout: red + "approaching timeout (Ns left)"
- On timeout fires: replaces with "Timed out after Ns of inactivity — last activity at HH:MM:SS" instead of a bare "Response timed out".

## Acceptance criteria

- [ ] Indicator visible whenever active session \`busy=true\`
- [ ] Resets on every \`stream_delta\`, \`tool_start\`, \`tool_result\`, \`task_progress\`, \`message\`, \`stream_start\`, \`stream_end\` for the active session
- [ ] Color escalates as the silent gap widens
[...]

Issue #4493, opened 2026-05-29
fix(dashboard): cross-session activity bump leaks when replay flag is held


blamechris:
## Context

Identified during review of PR #4490. The new `_receivingHistoryReplay` gate is keyed on a module-level boolean, but the replay is per-session.

## Description

`packages/dashboard/src/store/message-handler.ts` line 404 declares `_receivingHistoryReplay` at module scope. When `history_replay_start` for session A fires, the flag is `true` until the matching `history_replay_end`. `replayHistory()` chunks history sends over `setImmediate` (`packages/server/src/ws-history.js` line 350), which yields the event loop between chunks — so live broadcasts from OTHER sessions can interleave.

If session B broadcasts a live `tool_start` (or any activity event) while session A's replay is in-flight, the pre-handler bump at line 1882:

```ts
if (isActivityEvent(msg.type) && !_receivingHistoryReplay) {
```

short-circuits because the flag is true for session A. Session B's live activity does NOT bump `lastClientActivityAt`, even though it's a genuine real-time event.

Impact: one or more missed activity bumps per active background session per replay chunk window. Visible as a brief "stale" activity indicator on background sessions while another session is mid-switch-replay. Not critical (replays are typically <200ms), but worth tracking.

The fix is straightforward: key the flag on session id (e.g. `_replayingSessions: Set<string>`) and gate per-target:

```ts
if (isActivityEvent(msg.type) && !_replayingSessions.has(targetId)) {
```

## Original Review Comment

> The `_receivingHistoryReplay` flag is module-level, not per-session. Since `replayHistory` chunks over `setImmediate`, a live event from a non-replaying session B can interleave between chunks of session A's replay. The pre-handler gate then drops session B's legitimate activity bump.

## Acceptance Criteria

- [ ] Convert `_receivingHistoryReplay` to a `Set<string>` of session ids currently replaying
- [ ] Update `history_replay_start` / `history_replay_end` / `auth_ok` to add/remove session ids from the set
- [ ] Gate `isActivityEvent` check on `!_replayingSessions.has(targetId)` instead of the module-wide flag
- [ ] Add regression test: replay flag set for session A → live tool_start for session B bumps B's `lastClientActivityAt`

PR #4501, opened 2026-05-29
fix(app): gate pre-handler activity bump on receivingHistoryReplay (#4492)

### LINKED ISSUE #4492: fix(app): mirror #4466 activity-bump replay gate to mobile message-handler
## Context

Identified during review of PR #4490 (closes #4466). The dashboard's `isActivityEvent` pre-handler bump was gated on `!_receivingHistoryReplay`, but the mobile app has the exact same pre-handler logic ungated.

## Description

`packages/app/src/store/message-handler.ts` line 915 contains:

```ts
if (isActivityEvent(msg.type)) {
  const targetId = (typeof msg.sessionId === 'string' && msg.sessionId) || get().activeSessionId;
  if (targetId && get().sessionStates[targetId]) {
    updateSession(targetId, (ss) => {
      const patch: Partial<SessionState> = { lastClientActivityAt: Date.now() };
      if (ss.inactivityWarning) patch.inactivityWarning = null;
      return patch;
    });
  }
}
```

Mobile's `_ctx.receivingHistoryReplay` is the per-connection equivalent of the dashboard's `_receivingHistoryReplay` module flag. Mobile also receives `history_replay_start → events → history_replay_end` on reconnect or session switch (line 1373 sets `_ctx.receivingHistoryReplay = true`), so the same symptom applies: replayed activity events bump `lastClientActivityAt` and dismiss `inactivityWarning` exactly as in the dashboard.

Mobile does NOT have the `activeTools` clear in its `history_replay_start` handler (line 1390-1398), so that half of #4466 doesn't affect mobile. But the `lastClientActivityAt` and `inactivityWarning` halves do.

## Original Review Comment

> Mobile mirrors the dashboard's pre-handler bump logic (`packages/app/src/store/message-handler.ts` line 915) without the `!_receivingHistoryReplay` gate. Same symptom on session switch / reconnect replay — `lastClientActivityAt` resets, `inactivityWarning` is wiped. Should get the same gate as the dashboard fix.

## Acceptance Criteria

- [ ] Add `&& !_ctx.receivingHistoryReplay` to the `isActivityEvent` check in `packages/app/src/store/message-handler.ts` line 915
- [ ] Add the same three regression tests from PR #4490 to the mobile test file

blamechris:
## Summary
- Mirrors dashboard fix #4466 / PR #4490 to the mobile app: gates the pre-handler `lastClientActivityAt` bump (#3758) and `inactivityWarning` dismiss (#3899) on `!_ctx.receivingHistoryReplay` so replayed events on reconnect / session switch no longer reset activity timers.
- Adds 3 regression tests in `packages/app/src/__tests__/store/message-handler.test.ts` covering replay-doesn't-bump, replay-doesn't-dismiss, and live-activity-after-replay still bumps.

## Background
Mobile's `_ctx.receivingHistoryReplay` is the per-connection equivalent of the dashboard's module-level `_receivingHistoryReplay` flag, already used elsewhere in this handler. The dashboard's `activeTools`-clear half of #4466 does not apply on mobile because mobile's `history_replay_start` case doesn't clear `activeTools`.

## Test plan
- [x] RED → GREEN: 3 new tests at `packages/app/src/__tests__/store/message-handler.test.ts` failed before the gate change, pass after.
- [x] Full `packages/app` message-handler suite (125 tests) passes.
- [x] No regressions across the broader app jest suite (1156 tests pass; 3 pre-existing suite failures are missing `xterm-bundle.generated.ts` postinstall artifact, unrelated).

Closes #4492

### DIFF
--- a/packages/app/src/__tests__/store/message-handler.test.ts
+++ b/packages/app/src/__tests__/store/message-handler.test.ts
@@ -3175,6 +3175,84 @@ describe('set_permission_mode CAPABILITY_NOT_SUPPORTED rejection', () => {
   });
 });
 
+// #4492: switching tabs sends `switch_session`, which causes the server to
+// dispatch history_replay_start → all past events → history_replay_end.
+// Pre-fix, the mobile pre-handler logic at message-handler.ts ran the
+// lastClientActivityAt bump (#3758) and inactivityWarning dismiss (#3899) for
+// EVERY replayed event. Visible symptoms mirror the dashboard's #4466:
+// "Working… last activity Ns ago" reset to 1s and the "Agent quiet for
+// Nm Ns" chip disappeared. The fix gates the pre-handler bump on the same
+// per-connection `_ctx.receivingHistoryReplay` flag used elsewhere in this
+// handler — mobile's equivalent of the dashboard's module-level
+// `_receivingHistoryReplay`.
+describe('history replay must not reset activity timers (#4492)', () => {
+  function seedSession(extra: Partial<ReturnType<typeof createEmptySessionState>> = {}) {
+    const store = createMockStore({
+      activeSessionId: 's1',
+      sessions: [{ sessionId: 's1', name: 'S1' } as any],
+      sessionStates: { s1: { ...createEmptySessionState(), ...extra } },
+    });
+    setStore(store as any);
+    _testMessageHandler.setContext(createMockContext() as any);
+    return store;
+  }
+
+  afterEach(() => {
+    resetReplayFlags();
+  });
+
+  it('does not bump lastClientActivityAt for replayed activity events', () => {
+    const store = seedSession({ lastClientActivityAt: 100 });
+    _testMessageHandler.handle({ type: 'history_replay_start', sessionId: 's1' });
+    // tool_start is in ACTIVITY_EVENT_TYPES — pre-fix this bumped to Date.now().
+    _testMessageHandler.handle({
+      type: 'tool_start',
+      messageId: 'tool-1',
+      tool: 'Bash',
+      toolUseId: 'tu-1',
+      input: { command: 'ls' },
+      sessionId: 's1',
+    });
+    // The pre-seeded "stale" 100ms timestamp must survive — replay is NOT
+    // fresh activity. Without this guard, every tab switch resets the
+    // "last activity Ns ago" pill to "1s ago" no matter how long the
+    // session has actually been idle.
+    const ss = store.getState().sessionStates.s1;
+    expect(ss.lastClientActivityAt).toBe(100);
+  });
+
+  it('does not dismiss inactivityWarning for replayed activity events', () => {
+    // "Agent quiet for 46m 32s · Status update?" chip is mid-display when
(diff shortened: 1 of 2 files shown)


blamechris (commented):
## Agent review (Phase 1)

**Verdict:** Approve. Tight, accurate mirror of dashboard PR #4490; CI green across all 13 checks.

### Correctness audit

**Gate placement** — matches dashboard exactly. Same line / same site / same condition. `_ctx.receivingHistoryReplay` is typed as `boolean` (line 208) and initialized to `false` in `createDefaultContext()` (line 237), so the `undefined` risk on `!_ctx.receivingHistoryReplay` is nil — no nullish path can pass the gate incorrectly. The flag is also explicitly reset at three other sites (`auth_ok` reset line 933, `resetReplayFlags()` line 490, and `history_replay_end` line 1404), so no flag-stuck-true scenarios.

**Behavioral parity with dashboard** — accurate caveat in PR description: mobile has no `activeTools` symbol (`grep -c activeTools packages/app/src/store/message-handler.ts` → 0), so the dashboard's second-half fix (preserving activeTools across replay, gating `case 'result'` activeTools clear) genuinely doesn't apply.

**Test discipline** — 3 tests, jest (not vitest, correct for mobile). Verified RED → GREEN against this branch:

- With gate reverted: tests 1 & 2 fail with the expected assertion (pre-seeded `100` clobbered to `Date.now()`, `inactivityWarning` clobbered to `null`).
- With gate restored: 3/3 pass; full suite 125/125 pass.

Tests use `_testMessageHandler.handle({type: 'history_replay_start', ...})` to set the real `_ctx.receivingHistoryReplay = true` flag — same code path the server triggers. Faithful simulation, not a contrived mock.

### Findings

1. **Per-session scoping (follow-up, not blocking)** — Dashboard PR #4502 (#4493) just landed converting the same `_receivingHistoryReplay` from module-level boolean to per-session `Set<string>`. Mobile's `_ctx` is also module-level — same cross-session interleave risk (`packages/server/src/ws-history.js` line 363 chunks via `setImmediate`, so session B's live event can interleave with session A's replay window and be wrongly suppressed). Filed as #4513 for mobile parity. Not blocking this PR — the core single-session fix here is correct and shippable.

2. **Minor test parity gap (non-blocking)** — Dashboard #4490 ships 5 regression tests including a `live-activity-after-replay still dismisses inactivityWarning` guard. Mobile only ships the `lastClientActivityAt` resumption guard, not the `inactivityWarning` resumption guard. Not load-bearing — both fields share the same gate, so the existing guard covers the gate semantics. Could be added defensively but optional.

3. **`targetId` resolution unchanged** — line 916 still uses `msg.sessionId || activeSessionId` fallback. Matches dashboard PR #4490's deliberate scope choice (called out in #4490's PR body under "Why not also drop the fallback"). Same call applies here.

### Pipeline status

- **CI:** all 13 checks SUCCESS (App Tests, App Type Check, Dashboard Tests, Server Tests, etc.)
- **Mergeable:** yes, MERGEABLE / CLEAN
- **Open review comments:** none
- **Follow-up created:** #4513

PR #4502, opened 2026-05-29
fix(dashboard): scope receivingHistoryReplay per-session (#4493)

### LINKED ISSUE #4493: fix(dashboard): cross-session activity bump leaks when replay flag is held

blamechris:
## Summary

- Convert dashboard's `_receivingHistoryReplay` from module-level boolean to `_replayingSessions: Set<string>`, gated per target session id.
- `replayHistory()` chunks over `setImmediate`, so a live `tool_start` / `message` / `result` for session B could interleave with A's replay and be wrongly suppressed by the module-wide flag (missed activity bump, missed activeTools sweep).
- All four call sites updated: pre-handler activity bump (line 1882), `case 'result'` activeTools clear (line 2474), and the per-session dedup booleans passed to `sharedToolStart` / `sharedMessageHandler`.

## Behavior

- `history_replay_start`: adds the resolved session id to the set.
- `history_replay_end`: removes that session id (server emits `sessionId` on both — confirmed in `packages/server/src/ws-history.js` line 353/365).
- `auth_ok`: `clear()` — reconnect means clean slate.
- Activity gate: `if (isActivityEvent(msg.type)) { ... if (!_replayingSessions.has(targetId)) bump }`.

## Test plan

- [x] RED: 3 of 4 new tests fail against the pre-fix module flag (verified locally).
- [x] GREEN: all 148 message-handler tests pass; all 320 tests across the affected store test files pass (`message-handler`, `message-handler-evaluator`, `auth-ok-handler`, `dispatch-activity`, `store`).
- [x] The existing #4466 + #4491 regression suite still passes — per-session scoping does not regress the single-session case.
- [x] TypeScript: no new errors introduced in `message-handler.ts` / `message-handler.test.ts`.

Closes #4493

### DIFF
--- a/packages/dashboard/src/store/message-handler.test.ts
+++ b/packages/dashboard/src/store/message-handler.test.ts
@@ -3334,4 +3334,108 @@ describe('dashboard message-handler dispatch', () => {
       expect(ss.activeTools).toEqual([])
     })
   })
+
+  // #4493 — the replay flag must be scoped per-session. Pre-fix it was a
+  // module-level boolean: once `history_replay_start` fired for session A,
+  // every interleaved live event from session B (replayHistory chunks over
+  // setImmediate, so live broadcasts can land between chunks) was treated
+  // as replay and dropped its activity bump / activeTools clear.
+  describe('replay flag is per-session (#4493)', () => {
+    function seedTwoSessions(
+      sA: Partial<ReturnType<typeof createEmptySessionState>> = {},
+      sB: Partial<ReturnType<typeof createEmptySessionState>> = {},
+    ) {
+      store = createMockStore(baseState({
+        activeSessionId: 'sA',
+        sessions: [
+          { sessionId: 'sA', name: 'A' } as any,
+          { sessionId: 'sB', name: 'B' } as any,
+        ],
+        sessionStates: {
+          sA: { ...createEmptySessionState(), ...sA },
+          sB: { ...createEmptySessionState(), ...sB },
+        },
+      }))
+      setStore(store)
+    }
+
+    it('live tool_start for session B bumps B during session A replay', () => {
+      // Session A is mid-replay; an interleaved live tool_start for B must
+      // still bump B.lastClientActivityAt. Pre-fix the module flag was true
+      // (for A) so B's bump was suppressed.
+      seedTwoSessions({ lastClientActivityAt: 100 }, { lastClientActivityAt: 100 })
+      handleMessage({ type: 'history_replay_start', sessionId: 'sA' }, ctx() as any)
+      handleMessage(
+        {
+          type: 'tool_start',
+          messageId: 'tool-b-1',
+          tool: 'Bash',
+          toolUseId: 'tu-b-1',
+          input: { command: 'echo hi' },
+          sessionId: 'sB',
+        },
+        ctx() as any,
+      )
+      const ssB = (store.getState() as any).sessionStates.sB
+      expect(ssB.lastClientActivityAt).toBeGreaterThan(100)
+    })
+
+    it('replayed tool_start for session A still does NOT bump A', () => {
+      // Sanity: per-session scoping must still suppress replayed events for
+      // the actually-replaying session (the #4466 guarantee).
+      seedTwoSessions({ lastClientActivityAt: 100 }, { lastClientActivityAt: 100 })
+      handleMessage({ type: 'history_replay_start', sessionId: 'sA' }, ctx() as any)
+      handleMessage(
+        {
+          type: 'tool_start',
+          messageId: 'tool-a-1',
+          tool: 'Bash',
+          toolUseId: 'tu-a-1',
+          input: { command: 'sleep' },
+          sessionId: 'sA',
+        },
+        ctx() as any,
+      )
+      const ssA = (store.getState() as any).sessionStates.sA
+      expect(ssA.lastClientActivityAt).toBe(100)
+    })
+
+    it('live result for session B still clears B.activeTools during A replay', () => {
+      // Mirror of the #4491 gate, per-session: a live `result` for session
+      // B during A's replay window must still run the #4308 turn-boundary
+      // sweep on B.activeTools.
+      const inFlightToolB = { toolUseId: 'tu-b-1', tool: 'Bash', input: { command: 'sleep' }, startedAt: 100 }
+      seedTwoSessions({}, { activeTools: [inFlightToolB] })
+      handleMessage({ type: 'history_replay_start', sessionId: 'sA' }, ctx() as any)
+      handleMessage(
+        { type: 'result', sessionId: 'sB', usage: {}, cost: 0, duration: 0 },
+        ctx() as any,
+      )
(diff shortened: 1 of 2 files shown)


blamechris (commented):
## Agent review (deep)

**Verdict: Approve** — clean, correct, well-tested per-session refinement of the #4466/#4491 replay gate. CI green across the board, no merge conflicts.

### Correctness

All four module-scope `_receivingHistoryReplay` consumers updated to `_replayingSessions.has(targetId)`:

1. `handleToolStart` (L1483) — gates `sharedToolStart`'s replay dedup against the resolved target id, not a global flag.
2. Pre-handler `isActivityEvent` gate (L1896-1904) — moved `!_replayingSessions.has(targetId)` inside the inner `if` so the gate is per-target.
3. `case 'message'` (L2411) — dedup boolean derived from set membership of the message's session id.
4. `case 'result'` activeTools sweep (L2505) — gate is per `targetId` (the resolved `normalized.sessionId`), so a live `result` for B during A's replay still runs the #4308 sweep on B.

Mutation sites:

- `history_replay_start` (L2315): `_replayingSessions.add(replayTargetId)` — `replayTargetId` already falls back to `activeSessionId` inside `sharedHistoryReplayStart`, so the resolved id matches the gate's `targetId` resolution exactly.
- `history_replay_end` (L2354): inline resolver `(typeof msg.sessionId === 'string' && msg.sessionId) || get().activeSessionId` mirrors start's resolution, then `_replayingSessions.delete(endTargetId)`. The empty-string fallback is identical (`'' || active` → `active`). Server emits `sessionId` on both ends (`packages/server/src/ws-history.js` L353/365, `handlers/conversation-handlers.js` L121/136) — confirmed.
- `auth_ok` (L1920): `_replayingSessions.clear()` — matches the prior reset-on-reconnect behavior of `_receivingHistoryReplay = false`.
- `resetReplayFlags()` (L413): clears the set; called by `connection.ts` line 891 on disconnect — so no leak across disconnects either.

`grep -rn "_receivingHistoryReplay" packages/dashboard/src` after the diff returns only two hits, both in pre-existing test comments describing historical (pre-fix) behavior — neither are load-bearing.

### Edge cases

- **`history_replay_end` for an unknown session id:** `Set.delete()` is a no-op — safe.
- **`auth_ok` mid-replay:** clears the entire set, including the in-flight session. Subsequent replay events for that session would be treated as live. This **matches** the previous `_receivingHistoryReplay = false` behavior exactly.
- **Two concurrent replays:** Covered by the 4th regression test. Set scoping is correct.
- **Disconnect mid-replay:** `connection.ts` line 891 calls `resetReplayFlags()` → `_replayingSessions.clear()`. No leak.
- **Undefined targetId on `case 'result'`:** `_replayingSessions.has(undefined as any)` returns `false` — sweep proceeds (matches prior global-flag behavior when the flag was false).
- **`history_replay_end` arrives with stale `activeSessionId` (server-side bug):** server always emits `sessionId`, so the fallback path is dead in practice. If it ever fired, worst case is the wrong id is removed — a future event for the truly-replaying session would be treated as live (false negative on dedup), not a stuck state.

### Test discipline

4 new tests in the existing `#4466` describe block. They actually exercise the bug:

1. `live tool_start for session B bumps B during session A replay` — the direct bug scenario from #4493's repro.
2. `replayed tool_start for session A still does NOT bump A` — single-session regression guard (the #4466 guarantee).
3. `live result for session B still clears B.activeTools during A replay` — the #4491 follow-up gate, per-session.
4. `history_replay_end for A clears only A from the replaying set` — concurrent-replay coverage; ending A doesn't un-gate B.

Ran `cd packages/dashboard && npx vitest run`:
- All 144 message-handler tests pass.
- Full dashboard suite: 124 files, 2235 tests pass.
- `npx tsc --noEmit` clean.

### Acceptance criteria match (issue #4493)

- [x] Convert `_receivingHistoryReplay` to a `Set<string>` of session ids currently replaying — L410.
- [x] Update `history_replay_start` / `history_replay_end` / `auth_ok` to add/remove/clear — L2315, L2354, L1920.
- [x] Gate `isActivityEvent` on `!_replayingSessions.has(targetId)` — L1898.
- [x] Regression test: replay flag set for session A → live tool_start for session B bumps B's `lastClientActivityAt` — test #1.

All four AC items covered.

### Out-of-scope but worth noting

The mobile app (`packages/app/src/store/message-handler.ts`) has the same conceptual bug at a different scope. The recently-merged PR #4492 added `!_ctx.receivingHistoryReplay` to the pre-handler bump, but `_ctx.receivingHistoryReplay` is **per-connection**, not per-session — so the cross-session leak still applies on mobile. Will track as a separate follow-up issue.

### Style / minor

- The `case 'history_replay_end'` block introduces a bare `{ … }` block scope just to localize `endTargetId`. Functionally fine; minor stylistic nit that the `case` could instead use a `const` declared before `sharedHistoryReplayEnd()` is invoked. Not worth a fix.
- The discarded return value from `sharedHistoryReplayEnd()` is intentional now that the flag is no longer a single boolean. Comment at L1442-1443 of `store-core/handlers/index.ts` still says "Returns the new flag value" — a future PR could collapse the shared handler to a no-op or just remove the return type. Out of scope here.

PR #4504, opened 2026-05-29
feat(app): stream-stalled chip with retry (#4496)

blamechris:
## Summary

Adopts the dashboard stream-stall chip (#4476 / PR #4494) on the mobile chat surface so `error{code:'stream_stall'}` from server PR #4475 renders as a distinct amber/yellow recoverable affordance instead of the generic red error bubble.

- New `StreamStallChip` component (mobile parallel to `packages/dashboard/src/components/StreamStallChip`): yellow palette using `accentYellow500`, "Stream stalled — retry?" headline, one-tap Retry `Pressable` with `hitSlop` reaching the 44pt accessibility minimum.
- Retry is only wired when the stall is the tail message (ChatView walks `messages` once for tail id + last `user_input`, mirrors dashboard `isTail` gate). Historical replayed stalls receive no `onRetry` and render text only.
- Raw server error remains accessible: exposed via `accessibilityHint` for assistive tech and revealed inline by long-pressing the chip body (toggleable detail row).
- `MessageBubble` special-cases `type==='error' && code==='stream_stall'` before the existing red-bubble path; legacy errors without `code` still fall through unchanged.

## Acceptance criteria

- [x] Distinct chip (not red bubble) for `error.code === 'stream_stall'`
- [x] Retry resends most recent `user_input` when stall is the tail message
- [x] Historical/replayed stalls render chip text only (no Retry button)
- [x] Raw error text accessible (long-press + `accessibilityHint`)
- [x] Uses `accentYellow500` from `constants/colors`
- [x] Touch target reaches 44pt via `hitSlop`

## Test plan

- [x] `StreamStallChip.test.tsx` — 8 tests (render, accessibility, retry gating, 44pt target)
- [x] `MessageBubble.streamStall.test.tsx` — 4 tests (chip-vs-bubble routing, retry forwarding)
- [x] Full app test suite: 100 suites / 1321 tests pass
- [x] `tsc --noEmit` clean
- [ ] Visual verification deferred — Expo/Maestro can't run in this sandboxed worktree; type-check + unit tests are the bar.

Closes #4496

### DIFF
--- a/packages/app/src/components/ChatView.tsx
+++ b/packages/app/src/components/ChatView.tsx
@@ -20,6 +20,7 @@ import { ActivityGroup } from './chat/ActivityGroup';
 import { ToolDetailModal } from './chat/ToolDetailModal';
 import { MessageBubble } from './chat/MessageBubble';
 import { groupMessages, applyStreamingOverlay } from '@chroxy/store-core';
+import { useConnectionStore } from '../store/connection';
 
 // -- Props --
 
@@ -200,6 +201,23 @@ export function ChatView({
     [baseGroups, streamingMessageId, messages],
   );
 
+  // #4496: tail message id + last user input — used to gate the
+  // stream-stall chip's Retry button. Mirrors the dashboard's `isTail`
+  // check (only the most recent stall retry-button-wires; historical
+  // replayed stalls render chip text only). Walking messages once here
+  // is cheaper than recomputing per bubble.
+  const sendInput = useConnectionStore((s) => s.sendInput);
+  const chatTailMessageId = useMemo(
+    () => (messages.length > 0 ? messages[messages.length - 1].id : null),
+    [messages],
+  );
+  const lastUserInputContent = useMemo(() => {
+    for (let i = messages.length - 1; i >= 0; i--) {
+      if (messages[i].type === 'user_input') return messages[i].content;
+    }
+    return null;
+  }, [messages]);
+
   const handleScroll = (event: NativeSyntheticEvent<NativeScrollEvent>) => {
     const { contentOffset, contentSize, layoutMeasurement } = event.nativeEvent;
     const threshold = 100;
@@ -304,6 +322,17 @@ export function ChatView({
                   onPress={() => onToggleSelection(msg.id)}
                   onOpenDetail={handleOpenDetail}
                   onImagePress={setViewerUri}
+                  onRetryStreamStall={
+                    // #4496: only wire retry when this stall is the tail
+                    // bubble AND there's a user_input to resend. Mirrors
+                    // dashboard App.tsx renderMessage gating.
+                    msg.type === 'error' &&
+                    msg.code === 'stream_stall' &&
+                    msg.id === chatTailMessageId &&
+                    lastUserInputContent != null
+                      ? () => sendInput(lastUserInputContent)
+                      : undefined
+                  }
                 />
               </AnimatedMessage>
             </View>
--- a/packages/app/src/components/StreamStallChip.tsx
+++ b/packages/app/src/components/StreamStallChip.tsx
@@ -1,0 +1,149 @@
+/**
+ * StreamStallChip — #4496
+ *
+ * Mobile companion to packages/dashboard/src/components/StreamStallChip
+ * (#4476). Replaces the generic red error bubble when the server emits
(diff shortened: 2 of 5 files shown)


blamechris (commented):
## agent-review — PR #4504 (mobile stream-stall chip)

Reviewed the diff against CLAUDE.md's mobile constraints (44pt touch targets, RN accessibility) and the dashboard parity reference (PR #4494, packages/dashboard/src/components/StreamStallChip + App.tsx renderMessage).

### Correctness — looks good

- **Render-path ordering:** `MessageBubble.tsx` puts the `isError && code === 'stream_stall'` chip branch BEFORE both `showAsPill` and the default `errorBubble` TouchableOpacity render. Legacy errors with no `code` still fall through to the red bubble — exactly the parity behaviour the dashboard ships.
- **isTail gate:** `ChatView` computes `chatTailMessageId` from the current `messages` array, so the gate works identically for live streams and replayed-from-session_messages history. Historical stalls that aren't the most recent bubble correctly render text-only.
- **`accentYellow500` is the right token.** `#eab308` is what `CheckInChip` and `ActivityIndicator`'s "quiet" escalation use; the `rgba(217, 165, 12, 0.12)` background is a deliberate parity with CheckInChip's existing surface convention (`packages/app/src/components/CheckInChip.tsx:121`), not a hex typo.
- **Memoization not at risk.** Neither `MessageBubble` nor `AnimatedMessage` are wrapped in `React.memo`, so the inline `onRetryStreamStall={() => sendInput(...)}` lambda allocated per render isn't breaking any memo barrier. If a future PR adds `memo` to `MessageBubble`, the lambda would need a `useCallback` (likely keyed on `chatTailMessageId` + `lastUserInputContent`) — worth a comment but not a blocker today.

### Mobile/RN

- **hitSlop math checks out for 44pt.** Button is `paddingVertical: 6` + 13pt text ≈ ~25pt visible; `hitSlop.top + bottom = 22` brings the tappable region to ~47pt. The test asserts `>= 22` which is the right inequality.
- **`Pressable` pressed state:** Retry button has `({ pressed }) => [styles.button, pressed && styles.buttonPressed]` (opacity 0.85). Good visual feedback. Outer chip Pressable has no pressed style — acceptable since long-press is a power-user affordance.

### Nit (non-blocking) — accessibilityHint misuse

The outer `Pressable` sets `accessibilityHint={errorText}`. Per RN's accessibility docs `accessibilityHint` should describe what happens on activation (e.g. "double-tap to expand details"), not convey arbitrary content. VoiceOver reads it as a follow-up announcement after the label, so users will hear something like "Stream stalled, retry? — Stream stalled, no response for 5 minutes (provider=claude-sdk session=abc)", which works in practice but isn't the prop's intent. Two cleaner options:

1. Hint becomes the affordance description: `accessibilityHint="Long-press to view the underlying server error"`, and the raw text is only revealed by the long-press detail row.
2. Append the raw text to `accessibilityLabel` directly, leaving `accessibilityHint` for the long-press affordance description.

Either is a tiny tweak. Happy to leave as-is for now since the diagnostic IS reachable.

### Test discipline

- 12 tests across two files (8 chip-unit, 4 integration). Coverage of: chip-vs-bubble routing on `code`, retry-button gating on `onRetry` presence, long-press detail toggle, accessibilityRole + liveRegion, hitSlop sum, chip-vs-generic error fall-through, and forwarded callback invocation.
- Tests use `react-test-renderer` + `act` (mirrors `CheckInChip.test.tsx` convention — no `@testing-library/react-native` added). Locally: `npm test -- --testPathPattern="StreamStallChip|streamStall"` → 12/12 passing.
- One untested path worth flagging in a follow-up (not this PR): real-device visual verification via Maestro — the in-repo flows can't exercise the stream-stall path because it requires the server to emit `error{code:'stream_stall'}`. Tracking this separately so it doesn't gate the chip landing.

### Acceptance criteria

All six criteria in the PR body are met by code:
- distinct chip (not red bubble) ✓
- retry resends most recent user_input when stall is tail ✓
- historical stalls render text-only ✓
- raw error accessible (long-press + accessibilityHint) ✓ (with hint-semantics nit above)
- `accentYellow500` from `constants/colors` ✓
- 44pt touch target via hitSlop ✓

### Verdict

**Approve** — ship it. The `accessibilityHint` semantics tweak is a non-blocking polish that can land as a follow-up if you agree.

Issue #3749, opened 2026-05-11
Server: hardcoded 5-min inactivity timeout kills long-running responses (RESULT_TIMEOUT_MS)


blamechris:
## Summary

Chroxy's server kills an in-flight response after **5 minutes of inactivity** (no stream events from the SDK). Responses that legitimately need longer — long tool executions, slow MCP calls, big web fetches, extended thinking phases — get force-aborted with `Response timed out after 5 minutes of inactivity`, and the user sees the Claude subprocess \"exit unexpectedly\" / restart.

## Root cause

`packages/server/src/sdk-session.js:514`

\`\`\`js
const RESULT_TIMEOUT_MS = 300_000 // 5 min of inactivity
const resetResultTimeout = () => {
  if (this._resultTimeout) clearTimeout(this._resultTimeout)
  this._resultTimeout = null
  if (this._resultTimeoutPaused) return
  this._resultTimeout = setTimeout(() => {
    this._handleResultTimeout(messageId, streamState.hasStreamStarted)
  }, RESULT_TIMEOUT_MS)
}
\`\`\`

What fires on timeout (`sdk-session.js:1122` `_handleResultTimeout`):

1. Emits `stream_end` if streaming
2. Emits `permission_expired` for every outstanding permission
3. Aborts the SDK query generator (`_abortActiveQuery`)
4. Clears message state
5. Emits `error: 'Response timed out after 5 minutes of inactivity'`

The timer is reset on every SDK event (stream delta, tool_start, tool_result, etc.) and paused while permission/AskUserQuestion prompts are outstanding (#2831). So the bug doesn't fire when Claude is *visibly* working — it fires when there's a stretch >5 min with **no SDK event at all** (e.g., a long single Bash, a slow MCP tool, an extended-thinking phase).

## Repro

- Trigger a tool call that legitimately takes >5 min (large build, long network fetch, deep agent run that doesn't emit intermediate updates).
- Watch the server log for `Result timeout (5 min inactivity) — force-clearing busy state` and the client receive the `Response timed out after 5 minutes of inactivity` error.

## Suggested fix

1. **Raise the default.** 5 min is too aggressive for an agent that uses tools — 15–20 min would cover the realistic long-tail.
[...]

PR #4506, opened 2026-05-29
test(server): cover session-manager providerOpts timeout forwarding (#4487)

blamechris:
## Summary

Adds session-manager-level tests for the four operator-tunable timeout knobs that `SessionManager` forwards through `providerOpts` to the provider constructor:

- `resultTimeoutMs` (#3749)
- `hardTimeoutMs` (#3899)
- `streamStallTimeoutMs` (#4467)
- `mcpToolCallTimeoutMs` (#4482)

For each knob, the new tests assert:

1. A configured positive value flows through verbatim as `providerOpts.<key>`.
2. A `null` value is OMITTED from `providerOpts` (key absent — not set to `null`) so each provider's BaseSession-level default applies.

Consumer-side coverage already existed in `byok-session` / `base-session`; this PR closes the forwarding-side gap noted during review of #4486.

## Implementation notes

- Registers a single `CapturingProvider` that records its constructor `opts` into a shared buffer.
- A `captureProviderOpts(configKey, configValue)` helper spins up a `SessionManager` with the knob configured, drives `createSession()` through the capturing provider, and returns the captured opts.
- An `assertForwardingPattern({ configKey, providerOptsKey, setValue })` helper collapses each knob to a single assertion, satisfying the issue's "reasonable shared helper" criterion.
- Tests-only — no behaviour changes.

## Test plan

- [x] `node --test packages/server/tests/session-manager.test.js` — 164/164 pass (added 4)
- [x] Full server suite via `npm test -w packages/server` — only pre-existing flakes in `tunnel-manager.integration.test.js` and `web-task-manager.test.js` fail (confirmed present on `main` without this change)

Closes #4487

### DIFF
--- a/packages/server/tests/session-manager.test.js
+++ b/packages/server/tests/session-manager.test.js
@@ -3200,3 +3200,129 @@ describe('SessionManager cost-threshold soft warning (#4075)', () => {
     assert.equal(crossings.length, 2)
   })
 })
+
+// SessionManager forwards four operator-tunable timeout knobs via providerOpts
+// using a "forward only when set" pattern: a configured positive value is
+// passed through verbatim, while a null/unset value is OMITTED from
+// providerOpts (rather than set to null) so each provider's BaseSession-level
+// default applies. The consumer side (byok-session / base-session) has direct
+// coverage for each knob; these tests close the forwarding-side gap (#4487).
+describe('SessionManager providerOpts timeout forwarding (#4487)', () => {
+  // Each captured opts snapshot from a test provider constructor is appended
+  // to this array via the shared helper below.
+  let captured
+
+  // Bare-bones provider that records the providerOpts handed to its
+  // constructor. Mirrors the FailingProvider/TestProvider shapes used
+  // elsewhere in this file so registerProvider's validateProviderClass
+  // contract is satisfied.
+  class CapturingProvider extends EventEmitter {
+    constructor(opts) {
+      super()
+      captured.push(opts)
+      this.cwd = opts.cwd
+      this.model = opts.model || null
+      this.permissionMode = opts.permissionMode || 'approve'
+      this.isRunning = false
+      this.resumeSessionId = null
+    }
+    static get capabilities() { return {} }
+    start() {}
+    destroy() {}
+    interrupt() {}
+    sendMessage() {}
+    setModel() {}
+    setPermissionMode() {}
+  }
+
+  // Register the capturing provider once and reuse it across every assertion.
+  // The provider registry is process-global so re-registering under the same
+  // name is a no-op after the first import.
+  let providerRegistered = false
+  async function ensureProvider() {
+    if (providerRegistered) return
+    const { registerProvider } = await import('../src/providers.js')
+    registerProvider('test-timeout-capture', CapturingProvider)
+    providerRegistered = true
+  }
+
+  beforeEach(() => {
+    captured = []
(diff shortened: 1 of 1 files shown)


blamechris (commented):
## Agent review — PR #4506

### Verdict: Approve

Tests-only PR. Implementation follows the established pattern in this file, all four issue-acceptance-criteria knobs are covered with proper isolation, and `npm test -w packages/server` passes 164/164 on the PR head.

### What I checked

**1. Buffer-reset robustness** (resolves the bug flagged mid-implementation)
- `captureProviderOpts` calls `captured.length = 0` at the top of every invocation, BEFORE constructing the SessionManager.
- `assertForwardingPattern` calls `captureProviderOpts` twice per test (set + null) — each call gets a fresh buffer.
- `beforeEach(() => { captured = [] })` resets between tests as well.
- Per-call `assert.equal(captured.length, 1, ...)` guards against silent leakage if a future change causes double-instantiation.

The fix is robust. Belt + suspenders here is appropriate.

**2. Mocking depth — adequate, not over-mocked**
`CapturingProvider` mirrors the existing `FailingProvider` / `TestProvider` shapes (constructor sig, `static capabilities`, `start/destroy/interrupt/sendMessage/setModel/setPermissionMode`). It goes through the real path:

- real `registerProvider()` → real `validateProviderClass()` (line 61 of `providers.js`) — so a missing required method would still throw
- real `new SessionManager({...})` → real constructor's `Number.isFinite() && >0 ? value : null` normalization (lines 259-282)
- real `createSession()` → real `providerOpts` construction block (lines 581-666) including the four `if (this._XxxTimeoutMs != null) providerOpts.xxxTimeoutMs = ...` lines

The only things stubbed are post-construction lifecycle (start/destroy as no-ops), which is correct because we are testing what gets HANDED to the constructor, not what the session DOES with it. Refactoring `providerOpts` construction without updating these tests would still surface a regression.

**3. Coverage isolation — no bleeding**
Each test sets exactly ONE knob via `[configKey]: configValue` spread on the SessionManager options. The other three knobs are not passed, so SessionManager's constructor defaults them to `null` (lines 259-282), which means they would NOT be forwarded (lines 596-599). Each test then asserts ONLY its own key. No cross-knob coupling.

**4. Server style** — ES modules, no semicolons, single quotes. Matches.

**5. Behavior-change risk** — Only the test file is touched (126 additions, 0 deletions). Pure tests-only.

### Minor observations (non-blocking, follow-up tracked separately)

1. **`streamStallTimeoutMs = 0` is a valid value but uncovered.** SessionManager (line 274) and BaseSession (line 304) both gate on `>= 0` for this knob (unlike the others which use `> 0`), so `0` is a legitimate operator config meaning "disable stall recovery". The test uses `120_000` (positive nonzero), which exercises the happy path but not the `0`-is-forwarded edge. Not blocking — the issue's acceptance criteria are met as written.
2. **Helper could be reused if a 5th knob lands.** `assertForwardingPattern` is well-shaped for this. If/when another forwarding-only-when-set knob is added (the comment block in `session-manager.js:581-666` already has skill-budget, allowlist, trustMismatchMode etc. with the same pattern), extracting the helper to `tests/helpers/` would be worth doing. Tracked as a follow-up.

### CI

All 13 checks pass on the current head (commit `a6d78ffd9`).

### Recommendation

Ship it. The two observations above are non-blocking and tracked in follow-up issues.

PR #4518, opened 2026-05-29
fix(server): apply providerMeta label to 1M variant synthesis

blamechris:
## Summary

- Variant synthesis in `updateModels()` was the last `humanizeModelId()` call site that didn't consult `getModelMetadataFn()` first; fix brings it in line with the other five call sites.
- Today only Claude crosses the >=1M context threshold so the divergence is unreachable, but the moment a non-Claude provider ships a >=1M model (or an operator hand-edits a provider cache), the synthesized `[1m]` chip would get a mangled label in the same way #4434 described.

## Changes

- `packages/server/src/models.js` — consult `getModelMetadataFn(variantFullId)?.label` before falling back to `humanizeModelId(variantId)` in the variant-synthesis branch.
- `packages/server/tests/models-per-provider.test.js` — registry-level regression test for both a hypothetical non-Claude 1M model (uses provider label) and the default Claude path (unchanged: `humanizeModelId` fallback).

## Test plan

- [x] `node --test packages/server/tests/models-per-provider.test.js` — 20 pass
- [x] `node --test packages/server/tests/models.test.js` — 76 pass (no regressions)

Closes #4441

### DIFF
--- a/packages/server/src/models.js
+++ b/packages/server/src/models.js
@@ -578,9 +578,15 @@ export function createModelsRegistry(hooks = {}) {
         const variantFullId = `${m.fullId}${ONE_M_SUFFIX}`
         if (seenFullIds.has(variantFullId)) continue
         const variantId = `${m.id}${ONE_M_SUFFIX}`
+        // Consult the provider's metadata hook first so non-Claude registries
+        // that eventually ship a >=1M model get their authoritative label
+        // instead of the humanizeModelId mangling (#4441 follow-up to #4438).
+        // Currently unreachable for codex/gemini (no 1M models today) but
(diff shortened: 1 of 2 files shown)


blamechris (commented):
cc copilot for review

blamechris:
## check-pr summary

| Comment | Author | Status | Action | Reference |
|---------|--------|--------|--------|-----------|
| _(no inline review comments — Copilot did not post a review during the 5-min poll window; no other reviewers)_ | — | — | — | — |

**Threads:** 0 review threads on this PR; 0 unresolved.
**Inline replies:** none required (no comments to respond to).
**Resolved follow-ups:** none.

Phase 2 (check-pr) complete — no agent input to process.

PR #4527, opened 2026-05-29
refactor(server): parameterize byok-mcp-client protocolVersion + clientInfo.version

### LINKED ISSUE #4452: byok-mcp-client: parameterize protocolVersion + clientInfo.version (#4450 follow-up)
## Context

Identified during review of PR #4450.

## Description

`byok-mcp-client.js _handshake()` hard-codes two values that should derive from real sources:

```js
const initResult = await this._request('initialize', {
  protocolVersion: '2024-11-05',
  capabilities: {},
  clientInfo: { name: 'chroxy-byok', version: '1' },
}, HANDSHAKE_TIMEOUT_MS)
```

- `protocolVersion: '2024-11-05'` is the current MCP spec version. The MCP spec evolves; if a server requires a newer version we'll silently fall behind. Negotiation is preferable — accept whatever the server returns from `initialize` if it's in a compatible window, otherwise warn.
- `clientInfo.version: '1'` is a placeholder. Should derive from `package.json` (`packages/server/package.json` version) so MCP-server-side logs/debugging name a real chroxy version.

Neither is blocking — today's MCP servers accept `2024-11-05` and clientInfo.version is informational. But both will get harder to fix later when more tooling depends on the values.

## Original Review Comment

> The `protocolVersion: '2024-11-05'` is hard-coded in `_handshake`. MCP spec evolves — should we version this or accept whatever the server returns? Flag as a follow-up (small risk, easy fix later).
>
> The handshake's `clientInfo: { name: 'chroxy-byok', version: '1' }` — version is a placeholder string. Could derive from `package.json`. Flag as a follow-up.

## Acceptance Criteria

- [ ] `protocolVersion` is a module constant with a comment naming the spec date; bump as a deliberate change.
- [ ] After `initialize`, read the server's `result.protocolVersion` and log-warn (don't error) if it disagrees with the requested one.
- [ ] `clientInfo.version` is derived from `packages/server/package.json` at module load (e.g. via `import pkg from '../package.json' with { type: 'json' }` or `createRequire`).
- [ ] Test fixture asserts the version flows through.

blamechris:
## Summary

- Extract `MCP_PROTOCOL_VERSION` as a module constant so MCP spec bumps are a deliberate, single-edit-point change instead of a hard-coded literal in `_handshake`.
- Derive `MCP_CLIENT_VERSION` from `packages/server/package.json` at module load — MCP-server-side logs now name a real chroxy semver instead of the `'1'` placeholder.
- Log-warn (not error) when the server replies with a different `protocolVersion` than we requested. Per spec the server's value wins and both sides interoperate at it; the warn surfaces drift so a future spec mismatch is debuggable.
- Test stub fixture echoes initialize params on stderr (gated by `MCP_STUB_ECHO_INITIALIZE`) so a wire-shape test asserts what the client actually sent without re-implementing JSON-RPC framing.

Closes #4452

## Test plan

- [x] `node --test packages/server/tests/byok-mcp-client.test.js` — 19/19 pass (was 15, added 4)
- [x] `node --test packages/server/tests/byok-mcp-fleet.test.js` — 15/15 still pass (stub fixture change is additive + gated by env var)

### DIFF
--- a/packages/server/src/byok-mcp-client.js
+++ b/packages/server/src/byok-mcp-client.js
@@ -17,6 +17,9 @@
 
 import { spawn } from 'node:child_process'
 import { EventEmitter } from 'node:events'
+import { readFileSync } from 'node:fs'
+import { fileURLToPath } from 'node:url'
+import { dirname, join } from 'node:path'
 import { createLogger } from './logger.js'
 
 const MAX_RESTART_ATTEMPTS = 3
@@ -24,6 +27,35 @@ const RESTART_DELAY_MS = 1000
 const KILL_GRACE_MS = 1000
 const HANDSHAKE_TIMEOUT_MS = 5000
 export const DEFAULT_TOOL_CALL_TIMEOUT_MS = 30_000
+
+// #4452: MCP spec version we request on initialize. Bumping this is a
+// deliberate change — when MCP releases a new spec date and we adopt it,
+// update this constant + verify against the negotiation log-warn output.
+// Servers that don't recognise it should still accept the handshake
+// (per spec) and reply with their own protocolVersion; the negotiation
+// warn surfaces the mismatch without erroring.
+export const MCP_PROTOCOL_VERSION = '2024-11-05'
+
+// #4452: clientInfo.version on initialize. Derived from this package's
+// package.json so MCP-server-side logs/debugging name a real chroxy
+// version instead of the previous '1' placeholder. Synchronous read at
+// module load is fine because the file is tiny and never changes
+// mid-process.
+function readPackageVersion() {
+  try {
+    const here = dirname(fileURLToPath(import.meta.url))
+    const pkgPath = join(here, '..', 'package.json')
+    const raw = readFileSync(pkgPath, 'utf8')
+    const parsed = JSON.parse(raw)
+    if (typeof parsed?.version === 'string' && parsed.version.length > 0) {
+      return parsed.version
+    }
+  } catch {
+    // Fall through — never block startup on a missing/unreadable package.json.
+  }
+  return '0.0.0'
+}
+export const MCP_CLIENT_VERSION = readPackageVersion()
 
 export const MCP_STATES = Object.freeze({
   IDLE: 'idle',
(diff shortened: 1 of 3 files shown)


blamechris (commented):
## Summary

Cleans up `_handshake()` so `protocolVersion` and `clientInfo.version` are no longer hard-coded literals. Adds module constants `MCP_PROTOCOL_VERSION` and `MCP_CLIENT_VERSION` (derived from `packages/server/package.json` via sync `readFileSync` at module load), and a one-shot warn when the server replies with a different `protocolVersion`. Test stub gains an `MCP_STUB_ECHO_INITIALIZE` env knob plus a `MCP_STUB_PROTOCOL_VERSION` override so the new tests can pin the wire shape and exercise the negotiation-warn branch.

## Strengths

- The sync `readFileSync` pattern matches what already exists in `supervisor.js`, `ws-server.js`, `server-cli.js`, and `http-routes.js`. No new precedent; just one more file using the established idiom.
- Comments at the constant declarations explain why each value is a deliberate-bump (`MCP_PROTOCOL_VERSION`) vs. derived-from-source (`MCP_CLIENT_VERSION`) — future maintainers will know what to do.
- The `readPackageVersion` helper is fully defensive: try/catch around the read+parse, validates `parsed?.version` shape, falls back to `'0.0.0'` so a missing/malformed `package.json` never blocks startup.
- The negotiation warn is `log.warn` not `error`, matches the spec's "server's value wins" rule, and only fires on mismatch — happy path stays silent, verified by the dedicated `does NOT warn when matching` test.
- Stub-fixture change is additive and gated by env vars (`MCP_STUB_ECHO_INITIALIZE`, `MCP_STUB_PROTOCOL_VERSION`), so the 15 existing fleet tests stay green without touching them.
- The wire-shape test spawns the stub directly via `spawn` and drives one raw JSON-RPC round-trip rather than going through `MCPClient`, which isolates the `initialize` params from `tools/list` noise — good isolation.

## Issues

### Critical
None.

### Suggestions
- `readPackageVersion()` swallows all errors silently. A debug log on the catch arm (`could not read package.json, falling back to 0.0.0`) would help if `'0.0.0'` ever shows up in production MCP-server logs as a mystery. Non-blocking — the existing pattern in the four sibling files doesn't do this either, so this is consistency-vs-debuggability and I'd leave it.
- The mismatch warn uses string equality only. If we ever speak two MCP spec versions (e.g. accept both `2024-11-05` and `2025-03-26`), this will warn on the supported newer one. Today not a concern — single supported version, no negotiation range — but worth a comment if/when we add a compatible-version set.

### Nitpicks
- `MCP_PROTOCOL_VERSION = '2024-11-05'` could be more discoverable as `// see https://modelcontextprotocol.io/specification/2024-11-05` so a future bumper has a one-click link to the spec date. Tiny — comment already says "MCP spec date", which is enough.

## Deferred
None — issue scope is fully addressed.

## Architecture notes

The new module-load `readFileSync` adds two relative path segments worth of fragility (`../package.json`) but it's the same fragility four sibling files already have, so any future fix (e.g. moving to `import pkg from '../package.json' with { type: 'json' }` once we drop support for older Node) should be a cross-cutting sweep, not a per-file change. Don't split it here.

The PR description claims `createRequire`-style sync read; the implementation actually uses `readFileSync` + `JSON.parse`, which is functionally equivalent and what the other four files use. Description vs. implementation mismatch is cosmetic.

## Verdict

LGTM. Ready to merge after CI.

Numbers: +122 / -4 across 3 files, byok-mcp-client.test.js 19/19 pass, byok-mcp-fleet.test.js 15/15 still pass, all CI green.

PR #4530, opened 2026-05-29
feat(server): exponential restart backoff in byok-mcp-client

blamechris:
## Summary

- Replace `RESTART_DELAY_MS = 1000` with `RESTART_BACKOFF_MS = [1000, 2000, 4000]` so the restart loop can ride out "wedged dependency that needs a moment to recover" failure modes that the fixed 1s schedule couldn't cover.
- Derive `MAX_RESTART_ATTEMPTS` from `RESTART_BACKOFF_MS.length` so a future tune of one can't silently desync the two.
- Switch the death-counter gate from `>=` to `>` so the **full** schedule actually fires. Pre-fix the 3rd backoff entry was unused — only 2 delays were applied (1s + 1s = ~2s budget). Now 3 delays apply (1s + 2s + 4s = ~7s budget).
- Total budget before DEAD is now ~7s, up from ~2s. First restart still fires at ~1s (no regression on the existing acceptance test).
- Add tests asserting the 2nd attempt fires ~2s after the 2nd failure and the 3rd attempt ~4s after the 3rd failure.

Closes #4453

## Test plan

- [x] `node --test packages/server/tests/byok-mcp-client.test.js` — 17/17 pass (was 15, added 2)
- [x] `node --test packages/server/tests/byok-session.test.js` — 90/90 still pass

### DIFF
--- a/packages/server/src/byok-mcp-client.js
+++ b/packages/server/src/byok-mcp-client.js
@@ -2,12 +2,14 @@
  * MCP child-process lifecycle for the claude-byok provider (#4077).
  *
  * One MCPClient owns one MCP server child. Handles spawn, JSON-RPC handshake
- * (initialize → initialized → tools/list), crash detection with a constant 1s
- * backoff up to MAX_RESTART_ATTEMPTS, and SIGTERM/SIGKILL grace on destroy.
+ * (initialize → initialized → tools/list), crash detection with exponential
+ * backoff between restart attempts (1s/2s/4s, #4453), and SIGTERM/SIGKILL
+ * grace on destroy.
  *
  * Per #4077 scope:
  *   - Lazy spawn: caller decides when to call start().
- *   - Crash detection on child exit / spawn error → schedule restart at 1s.
+ *   - Crash detection on child exit / spawn error → schedule restart per the
+ *     RESTART_BACKOFF_MS schedule (1s/2s/4s).
  *   - After MAX_RESTART_ATTEMPTS failures, state=dead and tools=[].
  *   - Session destroy sends SIGTERM, escalates to SIGKILL at KILL_GRACE_MS.
  *
@@ -19,8 +21,19 @@ import { spawn } from 'node:child_process'
 import { EventEmitter } from 'node:events'
 import { createLogger } from './logger.js'
 
-const MAX_RESTART_ATTEMPTS = 3
-const RESTART_DELAY_MS = 1000
+// #4453: exponential restart backoff. The first restart still fires at ~1s
+// after the death (no regression on the existing acceptance test); the
+// second waits ~2s after that failure; the third waits ~4s. Total budget
+// before DEAD is ~7s, up from the previous fixed 3s. This better matches
+// "wedged dependency that needs a moment to recover" failure modes (e.g.
+// a transient port conflict that resolves in ~5s would have been blown
+// past by the previous 1/1/1 schedule).
+//
+// The array length is the authoritative max-attempt count — derive
+// MAX_RESTART_ATTEMPTS from it so an operator who edits the schedule
+// can't accidentally desync the two.
+const RESTART_BACKOFF_MS = [1000, 2000, 4000]
+const MAX_RESTART_ATTEMPTS = RESTART_BACKOFF_MS.length
 const KILL_GRACE_MS = 1000
 const HANDSHAKE_TIMEOUT_MS = 5000
 export const DEFAULT_TOOL_CALL_TIMEOUT_MS = 30_000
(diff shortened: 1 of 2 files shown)


blamechris (commented):
## Summary

Replaces the fixed `RESTART_DELAY_MS = 1000` with `RESTART_BACKOFF_MS = [1000, 2000, 4000]` and derives `MAX_RESTART_ATTEMPTS` from the array length. Most importantly, fixes a real off-by-one bug — the death gate was `>=` against `MAX=3`, which only ever applied **two** restart delays (the third post-death went straight to DEAD with the third schedule entry unused). Switching to `>` consumes the full schedule and matches the issue's framing of "3 restart attempts before declaring dead". Total budget moves from ~2s to ~7s, matching the issue spec.

## Strengths

- The `>=`-vs-`>` bug is correctly identified — the pre-fix gate fired after 2 applied delays, not 3. With the schedule `[1000, 2000, 4000]` and the corrected `>` gate, deaths-vs-attempts cleanly maps to:
  - death 1 → `attempts=1`, `1>3` false → delay `[0]=1s`
  - death 2 → `attempts=2`, `2>3` false → delay `[1]=2s`
  - death 3 → `attempts=3`, `3>3` false → delay `[2]=4s`
  - death 4 → `attempts=4`, `4>3` true → DEAD
  So 3 restarts actually fire, total 7s of backoff applied. PR description's math is correct.
- `MAX_RESTART_ATTEMPTS = RESTART_BACKOFF_MS.length` is the right pattern — operators editing the schedule can't desync the two. The `Math.min(..., RESTART_BACKOFF_MS.length - 1)` defensive index even guards against a future refactor that diverges them. Belt-and-suspenders.
- First-attempt timing (`~1s`) is preserved verbatim — the existing acceptance test was renamed and the assertion kept (`elapsed >= 800 && elapsed <= 1500`). No regression for fast-recovery flakes.
- The two new backoff-timing tests use a fresh fixture (`process.exit(2)`) and capture events with a pre-`start()` listener — the comment correctly notes that `start()` only resolves on DEAD, so without the pre-listener the early STARTING/RESTARTING transitions would be missed. Good catch.
- Two separate tests for the 2nd and 3rd gaps (rather than one combined) means a regression in just one step surfaces clearly. The 3rd-gap test cleverly measures `RESTARTING#3 → DEAD` instead of `RESTARTING#3 → STARTING#4` (which never fires because the loop terminates).
- The two pre-existing tests (`declares dead after MAX_RESTART_ATTEMPTS`, `restart event accounting`) had their deadlines bumped from 8s to 10s to absorb the new 7s budget — annotated in comments so the bump isn't mystery.
- Log line update is correct: `dead after ${this._restartAttempts - 1} failed attempts` reports 3 attempts (`4-1`), not 4 — the user-visible count matches the schedule.

## Issues

### Critical
None.

### Suggestions
- The 2nd-backoff test's range is `1700-2500ms`, the 3rd is `3700-4800ms`. Tight on the low end if CI is faster than the 2s setTimeout, but generous on the high end. Real-world timer drift on busy CI usually overshoots not undershoots, so this is probably fine — the 1st-attempt test (`800-1500ms`) has lived through CI for a while at similar tightness. Flag only if a flake actually shows up.
- The 3rd-gap test depends on the spawn-exit-cycle being `<=750ms` (`4800ms - 4050ms`). On a heavily loaded CI runner this might overshoot. Not a current concern — the existing `declares dead` test has lived at `8s → 10s` margins — but worth a follow-up if flakes appear.

### Nitpicks
- The comment on the death gate (`#4453: gate on > (not >=)...`) is verbose but earns its keep: it explains why the seemingly-cosmetic change is a real bug fix. Future maintainers reviewing a "tighten this" suggestion will know to leave it alone.
- `delayIdx` could just be `this._restartAttempts - 1` since the `>` gate already prevents reaching the out-of-bounds index. The `Math.min` is defensive against future refactor — comment already says so. Leave it.

## Deferred

## Architecture notes

**Cluster conflict warning**: This PR and #4527 both edit `packages/server/src/byok-mcp-client.js` around the constants block (lines 17-58 after both PRs). Merging in either order will require a rebase on the second PR — the conflict will be at the constants region where 4527 adds `MCP_PROTOCOL_VERSION` / `MCP_CLIENT_VERSION` declarations right where 4530 changes `RESTART_DELAY_MS` to `RESTART_BACKOFF_MS`. Mechanical, but flag for the merge step.

The pre-`start()` event listener pattern in the new tests is worth documenting — any future test that asserts on intermediate state transitions of the restart loop will need the same setup. Maybe a follow-up to extract a `collectStateEvents(client)` helper if a third test needs it.

## Verdict

LGTM. Ready to merge after CI. Will need a rebase if #4527 lands first.

Numbers: +105 / -15 across 2 files, byok-mcp-client.test.js 17/17 pass, byok-mcp-fleet.test.js 15/15 still pass, byok-session.test.js 90/90 still pass, all CI green.

PR #4533, opened 2026-05-29
test(server): cover byok-mcp-client handshake timeout + tunable HANDSHAKE_TIMEOUT_MS

blamechris:
## Summary

- Export `DEFAULT_HANDSHAKE_TIMEOUT_MS` so downstream tooling (#4078) can read the default without duplicating it.
- Accept `handshakeTimeoutMs` via constructor opts AND per-server `config.handshakeTimeoutMs` (the path `~/.claude.json` → `byok-mcp-config` will use). Precedence: **opts > config > module default**. Same defensive guard as `resultTimeoutMs` so NaN / Infinity / 0 / negative / non-number values fall back instead of silently coercing to 0 ms.
- Add stub fixture knobs `MCP_STUB_INITIALIZE_HANG` and `MCP_STUB_TOOLS_LIST_HANG` so the two distinct handshake-timeout branches can be exercised. The existing `MCP_STUB_HANG` couldn't drive these — it also blocks `initialize`, leaving the `tools/list` branch unreachable.
- Add tests covering: the constant export, opts/config precedence (including override-wins), defensive-fallback for bogus values, and the timeout-then-restart-then-DEAD path with no leaked timers or `_pending` entries.

Closes #4454

## Test plan

- [x] `node --test packages/server/tests/byok-mcp-client.test.js` — 22/22 pass (was 15, added 7)

### DIFF
--- a/packages/server/src/byok-mcp-client.js
+++ b/packages/server/src/byok-mcp-client.js
@@ -22,7 +22,16 @@ import { createLogger } from './logger.js'
 const MAX_RESTART_ATTEMPTS = 3
 const RESTART_DELAY_MS = 1000
 const KILL_GRACE_MS = 1000
-const HANDSHAKE_TIMEOUT_MS = 5000
+// #4454: handshake-request timeout. Some MCP servers (sandboxed containers,
+// servers that download packages on startup) legitimately take >5s to reply
+// to initialize / tools/list. Override per-config via `handshakeTimeoutMs`
+// (preferred — sourced from ~/.claude.json → byok-mcp-config) or per-
+// MCPClient via `opts.handshakeTimeoutMs` (used by tests). The export
+// makes #4078 / future tooling able to read the default without
+// re-deriving it. Both export and the override remain on the same
+// "absolute upper bound for ONE handshake request" semantics — total
+// handshake budget is up to 2× this (initialize + tools/list).
+export const DEFAULT_HANDSHAKE_TIMEOUT_MS = 5000
 export const DEFAULT_TOOL_CALL_TIMEOUT_MS = 30_000
 
 export const MCP_STATES = Object.freeze({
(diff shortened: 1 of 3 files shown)


blamechris (commented):
## Summary

Exports `DEFAULT_HANDSHAKE_TIMEOUT_MS` (renamed from the private `HANDSHAKE_TIMEOUT_MS`) so #4078 / future tooling can read the default without re-deriving it. Adds a per-instance `_handshakeTimeoutMs` field sourced via `opts > config > module default`, with the same defensive guard that `resultTimeoutMs` already uses (rejects NaN / Infinity / 0 / negative / non-number). Adds two new stub knobs — `MCP_STUB_INITIALIZE_HANG` and `MCP_STUB_TOOLS_LIST_HANG` — so the previously unreachable second handshake-timeout branch (tools/list) can be exercised; the existing `MCP_STUB_HANG` also blocks `initialize` so the post-initialize-pre-tools-list path was untestable.

## Strengths

- The precedence chain `opts > config > module default` matches the established pattern in `byok-session` (`resultTimeoutMs`, `mcpToolCallTimeoutMs`). One less convention to learn.
- Defensive guard uses `Number.isFinite() && > 0 &&` chained with `||` for the falsy fall-through — same shape as elsewhere. The `bogus` array in the precedence test (`NaN, Infinity, 0, -1, '5s', null, undefined`) covers everything that would silently coerce to 0ms via `setTimeout` and break every handshake.
- New stub knobs are deterministic: both branches early-`return` (no reply emitted), stub stays alive (no exit), so the client is forced to rely on its own timeout rather than EOF. That's the right test signal — flaky timing only enters via the `200ms` test override, which is large enough to ride out CI load on the client side.
- Tests verify post-DEAD invariants (`_restartTimer === null`, `_pending.size === 0`) — these are the bugs you'd actually want to catch. A leaked timer or stranded pending entry would surface as memory growth in long-lived sessions.
- Tooling-readable default (`export const`) means #4078 can do `import { DEFAULT_HANDSHAKE_TIMEOUT_MS } from './byok-mcp-client.js'` instead of redeclaring `5000`.
- Reuse of `stubConfig({ env: { MCP_STUB_TOOLS_LIST_HANG: '1' } })` across multiple tests is clean — no per-test fixture explosion.

## Issues

### Critical
None.

### Suggestions
- The opts/config precedence test has `opts=200, config=60_000` — if precedence were reversed, the test would hang for ~60s × 3 attempts = ~3 minutes before timing out. Probably OK because `node --test` will kill the suite at its default deadline, but a tighter `config=10_000` would surface the regression in ~30s and keep CI signal cleaner.
- `bogus` test loops without `await` on `client.destroy()` and never calls `start()`. That's fine because the test only reads `_handshakeTimeoutMs`, but it does leak 7 IDLE-state MCPClient instances. They have no spawned children and no timers so this is harmless, but a `client.destroy()` in the loop body would be tidy.

### Nitpicks
- `_handshakeTimeoutMs` initialization uses three chained `||` short-circuits across a single multi-line statement. Readable, but the alternative (a tiny `pickPositiveFinite(opts, config, fallback)` helper) would make the precedence visually obvious and DRY against `resultTimeoutMs`. Not a blocker.
- The hang-stub knob names are clear (`MCP_STUB_INITIALIZE_HANG`, `MCP_STUB_TOOLS_LIST_HANG`) — good. The existing `MCP_STUB_HANG` could grow a comment noting the new alternatives so future maintainers don't reach for the wrong knob.

## Deferred

## Architecture notes

**Cluster conflict warning**: This PR renames `HANDSHAKE_TIMEOUT_MS → DEFAULT_HANDSHAKE_TIMEOUT_MS` and converts it to an export at the top of `byok-mcp-client.js`. PR #4527 inserts `MCP_PROTOCOL_VERSION` / `MCP_CLIENT_VERSION` declarations into the same constants block. PR #4530 changes `RESTART_DELAY_MS → RESTART_BACKOFF_MS` in the same block. All three touch lines 22-29 in different ways — whichever lands first is fine, the second and third will need a rebase that interleaves the three constant additions/renames. Each is mechanical but flag for merge sequencing.

`_handshakeTimeoutMs` is read once at construction. If `config.handshakeTimeoutMs` is updated at runtime (e.g. config reload), the client won't pick it up — needs a fresh client. Consistent with how `resultTimeoutMs` works, so non-issue.

## Verdict

LGTM. Ready to merge after CI. Will need a rebase if either #4527 or #4530 lands first.

Numbers: +137 / -4 across 3 files, byok-mcp-client.test.js 22/22 pass (was 15, +7), byok-mcp-fleet.test.js 15/15 still pass, all CI green.

PR #4536, opened 2026-05-29
fix(server): log orphan JSON-RPC responses in byok-mcp-client at debug level

blamechris:
## Summary

- Split `_onStdoutChunk` message handling into three explicit branches:
  1. **Notification** (`id == null`) — silently dropped (normal protocol).
  2. **Matched response** (`id` in `_pending`) — settle the pending promise.
  3. **Orphan response** (`id` set, no `_pending` entry) — debug log with the unmatched id and shape (`result` | `error`).
- Previously orphan responses were silently dropped alongside notifications. A buggy MCP server emitting a stale id (post-restart) or a duplicate response left no audit trail — debugging started with "why doesn't my MCP tool respond" with nothing to grep for. A single debug-level log fixes that without spamming the happy path.
- Stub fixture knobs `MCP_STUB_EMIT_ORPHAN` and `MCP_STUB_EMIT_NOTIFICATION` drive the two paths from tests. Happy-path test asserts zero log noise.

Closes #4455

## Test plan

- [x] `node --test packages/server/tests/byok-mcp-client.test.js` — 18/18 pass (was 15, added 3)

### DIFF
--- a/packages/server/src/byok-mcp-client.js
+++ b/packages/server/src/byok-mcp-client.js
@@ -192,10 +192,28 @@ export class MCPClient extends EventEmitter {
         this._log.warn(`MCP server ${this.name}: non-JSON line: ${line.slice(0, 80)}`)
         continue
       }
-      if (msg.id != null && this._pending.has(msg.id)) {
+      if (msg.id == null) {
+        // Notification from the server (per JSON-RPC spec: `id` is required
+        // on responses, absent on notifications). Ignored silently — they
+        // are a normal part of the protocol (e.g. `notifications/cancelled`,
+        // server-initiated progress events) and would only add log noise.
+        continue
+      }
+      if (this._pending.has(msg.id)) {
         const settle = this._pending.get(msg.id)
         if (msg.error) settle(new Error(msg.error.message || 'MCP RPC error'))
         else settle(null, msg.result)
+      } else {
+        // #4455: orphan response — server emitted a response with an id we
+        // never sent (or one we already settled). Could be a buggy server
+        // emitting a stale id after restart, a state-machine bug on the
+        // server side, or a duplicate. Not necessarily an error (don't
+        // warn); but without a log we'd have no audit trail when debugging
+        // "why doesn't my MCP tool respond". Capture the unmatched id and
+        // a small shape marker so future investigation has something to
+        // grep for.
+        const shape = msg.error ? 'error' : 'result'
+        this._log.debug(`MCP server ${this.name}: orphan JSON-RPC response id=${msg.id} shape=${shape}`)
       }
     }
   }
--- a/packages/server/tests/byok-mcp-client.test.js
+++ b/packages/server/tests/byok-mcp-client.test.js
@@ -61,6 +61,56 @@ describe('MCPClient', () => {
     })
   })
 
+  describe('orphan responses (#4455)', () => {
+    it('logs orphan JSON-RPC response at debug level', async () => {
+      const debugs = []
+      const log = { info: () => {}, warn: () => {}, debug: (m) => debugs.push(m), error: () => {} }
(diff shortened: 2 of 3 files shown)


blamechris (commented):
## Summary

Splits the previously-collapsed `if (msg.id != null && this._pending.has(msg.id))` into three explicit branches: (1) `id == null` → silent notification drop, (2) matched response → settle, (3) orphan response → debug log with id + shape. Adds two stub knobs (`MCP_STUB_EMIT_ORPHAN`, `MCP_STUB_EMIT_NOTIFICATION`) so all three branches can be exercised, plus a happy-path test asserting zero log noise.

## Strengths

- The 3-branch split reads cleanly — each branch has a comment naming the semantics (notification, matched, orphan). Future maintainer doesn't have to deduce why the conditions are what they are.
- Debug-level (not warn) is the right choice. Per the comment: "could be a buggy server, could be a duplicate, not necessarily an error". Warn-level for orphans would create false-positive noise; debug gives the audit trail without spamming.
- Log format `id=999999 shape=result` is greppable — exactly what the linked-issue's "nothing to grep for" complaint asked for. The `shape=result|error` distinction is small but useful — orphan errors are more suspicious than orphan results (suggests server-side cancellation race) and surfacing the difference at the log lets the operator skip log-archaeology.
- Happy-path test asserts **zero** orphan logs — locks in the no-noise contract, which is exactly the bug to guard against in future refactors.
- Notification test specifically asserts that `notifications/progress` (a real MCP wire-shape) does NOT appear as an orphan. Important — without this someone could "fix" the orphan branch in a way that also caught notifications.
- Stub uses `999_999` for the orphan id — deliberately high so it can't collide with anything the client sends in the short-lived test. Defensive without being magic.

## Issues

### Critical
None.

### Suggestions
- The two new env knobs only fire when `msg.method === 'initialize'`. If a test ever needed an orphan after `tools/list`, the stub would need a second knob. Probably fine — both code paths (orphan after initialize, orphan after tools/list) hit the same `_onStdoutChunk` branch — so the existing knob covers the coverage need.
- The `setTimeout(r, 100)` await is a bit fragile (timing-dependent flush). Could be replaced with a deterministic await on the `state` event after READY plus a microtask flush, but 100ms × 18 tests = ~1.8s of pure wait time is small enough not to matter. Leave it.

### Nitpicks
- Comment on the orphan branch is verbose (8 lines). Earns its keep — explains _why_ debug and not warn — but could be trimmed if you wanted. Not a blocker.
- "shape" terminology is unusual; "kind" or "type" is more common. But `shape=result` reads as "result-shape" which is fine and unambiguous.

## Deferred

## Architecture notes

**Cluster conflict warning**: This PR modifies `_onStdoutChunk` (lines ~192-216 in main). The other 4 wave-1 PRs touch other regions of `byok-mcp-client.js`:
- #4527: constants block (lines 17-29) and `_handshake()` (lines ~129+)
- #4530: constants block, `_onExit` (lines ~215-258)
- #4533: constants block, `_handshake()`, constructor

#4536 should be the cleanest of the wave to merge — `_onStdoutChunk` is independent of the constants, constructor, and `_onExit` changes. Only conflicts will come from import lines if the other PRs also touch them (none currently do). Mechanically lowest-risk PR to land last in the wave; alternatively can go early as the smallest, surface-area-minimum diff.

The shape marker (`result|error`) is an extension point — if MCP ever adds a `partial` response shape (it currently doesn't) the marker would catch the new case as `result` (truthy) without specially classifying it. Acceptable for now.

## Verdict

LGTM. Ready to merge after CI. Lowest cross-cluster conflict risk of the wave.

Numbers: +87 / -1 across 3 files, byok-mcp-client.test.js 18/18 pass (was 15, +3), byok-mcp-fleet.test.js 15/15 still pass, all CI green.

PR #4537, opened 2026-05-29
fix(server): cap MCPFleet.start() wall-clock to fast-fail broken MCP configs

### LINKED ISSUE #4456: byok-session: start() blocks ~3s on a fully-broken MCP config (#4450 follow-up)
## Context

Identified during review of PR #4450.

## Description

`byok-session.start()` awaits `fleet.start()`, which awaits each `MCPClient.start()`, which resolves only on the first stable state (`READY` or `DEAD`). For a fully-broken MCP server, that's `MAX_RESTART_ATTEMPTS (3) × RESTART_DELAY_MS (1000)` = ~3 seconds before the client gives up and the session can emit `ready`.

For the common case of one or two well-configured servers this is invisible. The cost surfaces in two scenarios:

1. **User has one broken MCP server in their `~/.claude.json`** (e.g. command-not-found from a deleted binary). Session-start now takes 3s longer than it used to, with no UI signal that something's pending.
2. **User has multiple MCP servers and one is broken.** `Promise.all` in `fleet.start()` means the slow path bounds the wait, so a single bad server delays *all* session starts by 3s.

Two reasonable fixes (prefer #2):

1. Make `client.start()` resolve on the *first* `RESTARTING` transition (i.e. "child has crashed at least once, treat as not-yet-ready"), then let the restart loop continue in the background. Trade-off: a transiently-flaky server that recovers on attempt 2 wouldn't have its tools available on the first turn.
2. Add a wall-clock cap on `fleet.start()` (e.g. 1500ms): wait for clients to stabilize OR for the cap, whichever is first. Clients still in `STARTING`/`RESTARTING` at the cap contribute zero tools to the first turn but get included on subsequent turns (#4078 will need to re-query `fleet.tools` per turn anyway, so this composes cleanly).

Option 2 keeps the happy path identical (sub-second for ready servers) and bounds the worst case at a tunable cap.

## Original Review Comment

> `start()` resolves on first stable state (READY or DEAD) — this means `byok-session.start()` BLOCKS for up to ~3s on a bad config (3 restart attempts × 1s). Is that acceptable for UX? Consider: a real user with one broken MCP server in their config would see a 3s delay on session ready.

## Acceptance Criteria

- [ ] Pick a strategy (preference: wall-clock cap on `fleet.start()`).
- [ ] Add a test asserting session-ready emits within the cap even when one server is permanently broken.
- [ ] Document the trade-off in a comment on `MCPFleet.start()`.
- [ ] When #4078 lands, ensure tools list is re-queried per turn so late-arriving clients contribute on subsequent turns.

blamechris:
## Summary

Pre-fix: `byok-session.start()` awaited `MCPFleet.start()`, which awaited each `MCPClient.start()`, which only resolved on the first stable state (READY or DEAD). A single broken MCP server in `~/.claude.json` added the full restart budget (~7 s after #4453's exponential backoff) to every session-start — with no UI signal that anything was pending. `Promise.all` in `fleet.start()` meant a single bad server delayed *all* multi-server sessions too.

This PR implements the issue's **Option 2** (wall-clock cap on `fleet.start()`):

- Add `DEFAULT_FLEET_START_CAP_MS = 1500` and a `startCapMs` constructor opt on `MCPFleet`. `start()` now resolves on whichever happens first: every client reaches a stable state, OR the wall-clock cap fires.
- Clients still in `STARTING` / `RESTARTING` at the cap remain alive in the background. `byok-session._buildTools()` already re-queries `fleet.tools` per turn, so a late-arriving client lands naturally on the next turn — no manual re-trigger needed.
- Forward the cap end-to-end: `session-manager` → `providerOpts` → `byok-session` → `MCPFleet`. Same defensive-guard shape as `mcpToolCallTimeoutMs` (only positive finite values; bogus values fall back to the fleet's default).
- Two existing fleet tests asserted "broken server is DEAD by the time `start()` returns" — kept those invariants by passing `startCapMs: 60_000` (legacy wait-for-full-convergence behavior).

Closes #4456

## Test plan

- [x] `node --test packages/server/tests/byok-mcp-fleet.test.js` — 20/20 pass (was 15, added 5)
- [x] `node --test packages/server/tests/byok-session.test.js` — 91/91 pass (was 90, added 1 end-to-end test)
- [x] `node --test packages/server/tests/byok-mcp-client.test.js` — 15/15 still pass

### DIFF
--- a/packages/server/src/byok-mcp-fleet.js
+++ b/packages/server/src/byok-mcp-fleet.js
@@ -16,6 +16,19 @@ import { loadTrustStore, recordTrust, isTrusted } from './byok-mcp-trust.js'
 import { MCP_PREFIX, parseMcpToolName } from './mcp-tools.js'
 
 export const FLEET_KILL_GRACE_MS = 2000
+
+// #4456: wall-clock cap on fleet.start() to bound the worst-case session-
+// start latency on a misconfigured MCP server. The fleet awaits each
+// client's first stable state (READY or DEAD); with the post-#4453 1/2/4s
+// restart backoff a single broken server adds up to ~7s to session-start.
+// At the cap, clients still in STARTING/RESTARTING contribute zero tools
+// to the first turn but remain alive and will surface on subsequent turns
+// (the fleet re-queries .tools per turn, so a late-arriving client lands
+// naturally). 1500 ms is enough for a fast happy path (the common case
+// completes in ~50–200 ms) without making the user feel the bad-config
+// tax. Operators can override via the `startCapMs` constructor opt.
+export const DEFAULT_FLEET_START_CAP_MS = 1500
+
 // #4451: re-export under the legacy name from this module so existing
 // callers (byok-session, tests) keep working. Canonical source is
 // mcp-tools.js to prevent silent parser drift.
@@ -40,6 +53,15 @@ export class MCPFleet {
     // (session passes its _permissions in) is wired in byok-session.
     const permissionManager = opts.permissionManager || null
     const trustStorePath = opts.trustStorePath  // tests override; undefined falls through to default
+    // #4456: per-fleet wall-clock cap on start(). Same defensive guard
+    // pattern as resultTimeoutMs in byok-session: non-finite / non-positive
+    // values fall back to the module default. Setting opts.startCapMs to
+    // Infinity (legacy behavior — wait for full convergence) requires
+    // explicitly opting in below.
+    this._startCapMs =
+      Number.isFinite(opts.startCapMs) && opts.startCapMs > 0
+        ? opts.startCapMs
+        : DEFAULT_FLEET_START_CAP_MS
     this._clients = configs.map((cfg) => {
       const clientOpts = { ...opts }
       if (permissionManager) {
@@ -62,8 +84,46 @@ export class MCPFleet {
 
   get clients() { return this._clients }
 
+  /**
+   * Spawn every client and resolve when either (a) every client has reached
+   * a stable state (READY or DEAD), or (b) the wall-clock cap fires —
+   * whichever is first (#4456).
+   *
+   * Trade-off: a session with one broken MCP server (e.g. command-not-found
+   * from a deleted binary) used to wait the FULL ~7s restart budget on
+   * session start. With the cap, the user gets a ready session in ≤
+   * `startCapMs` ms even on broken configs; clients still in STARTING /
+   * RESTARTING at the cap contribute zero tools to the first turn but get
+   * picked up automatically on the next turn (byok-session re-queries
+   * `fleet.tools` per turn). A transiently-flaky server that would have
+   * recovered on attempt 2 also loses its first-turn contribution under
+   * the cap — that's the cost of bounding the worst case.
+   *
+   * The per-client `start()` promises are *not* dropped at the cap — they
+   * continue running in the background and update `client.state` as they
+   * settle. We just stop *awaiting* them so the caller can move on.
+   */
   async start() {
-    await Promise.all(this._clients.map((c) => c.start().catch(() => {})))
+    const startPromises = this._clients.map((c) => c.start().catch(() => {}))
+    if (!Number.isFinite(this._startCapMs) || this._startCapMs <= 0) {
+      // Defensive — should be unreachable given the constructor guard, but
+      // an explicit Infinity opt-in would land here and we just wait.
+      await Promise.all(startPromises)
+      return
+    }
+    let capTimer = null
+    const capPromise = new Promise((resolve) => {
+      capTimer = setTimeout(resolve, this._startCapMs)
+      // Don't keep the event loop alive solely for this timer — if every
+      // client settles before the cap, the timer's still pending and would
+      // otherwise hold the process open.
+      if (typeof capTimer.unref === 'function') capTimer.unref()
+    })
+    try {
+      await Promise.race([Promise.all(startPromises), capPromise])
+    } finally {
+      if (capTimer) clearTimeout(capTimer)
+    }
   }
 
   get tools() {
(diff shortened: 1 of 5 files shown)


blamechris (commented):
## Summary

Implements Option 2 of #4456: wall-clock cap on `MCPFleet.start()` so a single broken MCP server in `~/.claude.json` no longer adds the full ~7s restart budget (post-#4453) to every session start. Adds `DEFAULT_FLEET_START_CAP_MS = 1500` plus a `startCapMs` opt threaded end-to-end (`session-manager → byok-session → MCPFleet`). At the cap, clients still in `STARTING`/`RESTARTING` stay alive in the background and surface on the next turn (byok-session re-queries `fleet.tools` per turn). Two pre-existing fleet tests that asserted "broken server is DEAD by the time start() returns" are kept honest by passing `startCapMs: 60_000` (legacy wait-for-full-convergence behavior).

## Strengths

- The cap is implemented as `Promise.race([Promise.all(startPromises), capPromise])` with a `setTimeout` that uses `.unref()` — happy-path doesn't hold the event loop open if every client settles first. Cleanup in `finally` block clears the timer either way.
- Per-client `start()` promises are NOT dropped at the cap (the comment explicitly calls this out) — they continue running in the background and `client.state` updates as they settle. The "late-arriving client lands on next turn" claim is real and verified by the existing per-turn re-query in `byok-session._buildTools()`.
- Defensive guard pattern (`Number.isFinite(...) && > 0`) is the same shape used by `mcpToolCallTimeoutMs`, `resultTimeoutMs`, `handshakeTimeoutMs` (#4533) — one less convention to learn. The `bogus` test loop covers `NaN, 0, -1, '500', null, undefined`.
- End-to-end forwarding is correct: `session-manager` accepts `mcpStartCapMs` in its constructor opts, stores `_mcpStartCapMs`, threads via `providerOpts` to `byok-session`, which builds `fleetOpts` and only sets `startCapMs` if the override is non-null (so the fleet's own default still wins when nothing is configured). Three-tier override doesn't leak nulls into the fleet's defensive guard.
- The two updated tests (`excludes tools from dead servers`, `excludes anthropicTools from dead servers`) are correctly fixed with `startCapMs: 60_000` — they're testing the DEAD-by-start invariant, not the cap behavior, so opting into "legacy wait-for-full-convergence" is exactly right. Comment explains why.
- New end-to-end byok-session test seeds both a healthy stub and a broken server in `~/.claude.json`, asserts `ready` emits in <1000ms under a 400ms cap, and confirms broken is NOT DEAD yet — proves the cap fires before the restart loop completes. Exercises the production wire path.
- The 1500ms default lands in a defensible spot: happy-path handshakes typically complete in 50-200ms (so no user-visible latency from the cap), 1500ms is small enough that the user doesn't feel a "stuck" connection on broken configs, and it's well under the post-#4453 7s budget so the cap actually bites.

## Issues

### Critical
None.

### Suggestions
- The cap-fires-early test asserts `elapsed < 800ms` under a 300ms cap. Tight upper bound on a busy CI runner — 500ms slop is OK for happy-path but might flake under heavy load. If a flake shows up, bumping to `< 1500ms` would still prove the cap is doing real work (no-cap baseline would be ~7s).
- `_startCapMs` is read once at construction. If `~/.claude.json` is hot-reloaded with a new cap, the fleet won't pick it up — needs a fresh fleet. Consistent with how the other fleet/client opts work, but worth noting if a settings-panel feature ever wants live tuning.
- The 1500ms default is undocumented at the operator-facing level (no mention in user-visible config docs, no session-manager docstring callout). Currently fine because the cap is auto-tuned and "just works", but if support starts seeing "my MCP server takes 2s to start" issues, the discoverability gap will bite. Not a blocker — flag as a doc-debt follow-up if the bug-hunt finds it.

### Nitpicks
- `if (typeof capTimer.unref === 'function') capTimer.unref()` — Node 22's `setTimeout` always returns a `Timeout` with `unref()`, the typeof guard is paranoid but harmless. Browser-environment defensiveness is overkill here (this is server-side ESM) but the cost is one line. Leave it.
- The cap-fires-early test name "returns within ~startCapMs even when one server is permanently broken" — the "even when" framing is good, captures the regression we're guarding against.

## Deferred
- Operator-facing doc on `mcpStartCapMs` + the cap behavior — not blocking, but worth a follow-up if/when support sees the friction.

## Architecture notes

**Cluster conflict potential**: This PR touches `byok-mcp-fleet.js`, `byok-session.js`, `session-manager.js`, plus two test files. The other 4 wave-1 PRs all touch `byok-mcp-client.js` only — so #4537 is **conflict-free** with the rest of the wave. Mechanically it can land in any order relative to the others.

The "session-ready fires before all MCP clients converge" semantic is a real behavior change. Today's callers of `fleet.tools` get a snapshot at session-start; after this PR they get a snapshot at the cap. The `byok-session._buildTools()` per-turn re-query is what makes this safe — if a future caller starts caching `fleet.tools` outside the per-turn re-query, it would silently miss late-arriving servers. Worth a doc note on the `MCPFleet.start()` JSDoc — and the JSDoc actually already says "clients still in STARTING / RESTARTING at the cap contribute zero tools to the first turn but get picked up automatically on the next turn (byok-session re-queries fleet.tools per turn)". Good.

The `Infinity` opt-in path (legacy wait-for-full-convergence) is reachable only by explicitly passing a non-finite value, which the constructor guard converts to the default. So the comment "Setting opts.startCapMs to Infinity (legacy behavior) requires explicitly opting in below" is slightly misleading — the guard rejects `Infinity` as not-finite. The two tests that need legacy behavior do so by passing `60_000`, not `Infinity`. Either: (a) update the constructor to accept `=== Infinity` as an explicit opt-in, (b) reword the comment to say "use a large finite value like 60_000". (b) is the smaller fix and matches what the tests actually do.

## Verdict

LGTM. Ready to merge after CI. Conflict-free with the other 4 wave-1 PRs.

Numbers: +208 / -8 across 5 files, byok-mcp-fleet.test.js 20/20 pass (was 15, +5), byok-session.test.js 91/91 pass (was 90, +1), byok-mcp-client.test.js 15/15 still pass, all CI green.

Issue #4509, opened 2026-05-29
Internal timeout guards (session-manager / server-cli / base-session) also lack MAX_SANE_DURATION_MS ceiling


blamechris:
## Context

Identified during review of PR #4503 (which fixed the wire-side guards in `packages/server/src/ws-history.js sendPostAuthInfo`).

## Description

PR #4503 added the `<= MAX_SANE_DURATION_MS` ceiling to the three timeout guards that emit values on the `auth_ok` wire, fixing a handshake breakage when an operator set e.g. `CHROXY_STREAM_STALL_TIMEOUT_MS=99999999999`.

The same `Number.isFinite(x) && x [>|>=] 0` pattern (without the ceiling) still exists internally at:

- `packages/server/src/session-manager.js:260-275` — `_resultTimeoutMs`, `_hardTimeoutMs`, `_streamStallTimeoutMs` (forwarded to provider sessions via `providerOpts`).
- `packages/server/src/server-cli.js:373-414` — both the `SessionManager(...)` constructor args and the `effective*` log-line block.
- `packages/server/src/base-session.js:289-304` — the per-session BaseSession assignment.

These are **internal** timer-config sites — they don't directly hit the protocol schema, so an over-ceiling value doesn't break the handshake the way the `ws-history.js` gap did. But they silently produce a `>24h` internal inactivity timer that an operator almost certainly didn't intend (a typo or wrong unit assumption), with no log warning.

## Acceptance criteria

- [ ] All three guards above mirror the `<= MAX_SANE_DURATION_MS` ceiling and fall back to `null` (or the appropriate default) when exceeded.
- [ ] When the ceiling falls back, emit a single warning log line so operators can spot the bad config (`CHROXY_*` env-var typos are the realistic source).
- [ ] Tests cover one over-ceiling case per timeout per file (or a shared parameterized helper).
- [ ] No behavioral change for in-range values.

## Out of scope

Other duration fields gated by `.max(MAX_SANE_DURATION_MS)` in `packages/protocol/src/schemas/server.ts` (e.g. `remainingMs`, `idleMs`, `restartEtaMs`) — only the three operator-facing inactivity timeouts share this pattern. If audit turns up more, file separately.

PR #4538, opened 2026-05-29
fix(server): clamp mcpToolCallTimeoutMs to MAX_SANE_DURATION_MS (#4517)

blamechris:
## Summary

PR #4516 added the 24h `MAX_SANE_DURATION_MS` ceiling guard to the three sibling internal timeouts (`resultTimeoutMs` / `hardTimeoutMs` / `streamStallTimeoutMs`); `mcpToolCallTimeoutMs` was explicitly out of scope at the time. This PR closes the audit by applying the same `isOperatorTimeoutInRange()` gate to the three internal sites where the knob lives.

## Sites fixed

1. **`packages/server/src/session-manager.js`** — providerOpts forwarding guard now uses the shared helper; `_mcpToolCallTimeoutMs` falls back to `null` (BaseSession default) on over-ceiling values.
2. **`packages/server/src/server-cli.js`** — `resolveStartupTimeouts()` gains a fourth knob. The companion `effectiveMcpToolCallTimeoutMs` resolves to `DEFAULT_TOOL_CALL_TIMEOUT_MS` (30s) so the existing pattern (`*TimeoutMs` for SessionManager arg side, `effective*` for log line) extends cleanly.
3. **`packages/server/src/byok-session.js`** — per-session BYOK assignment uses the helper; defends against an over-ceiling value reaching here via an embedder that constructs `ClaudeByokSession` directly.

## Site 4 decision (config.js)

Per the issue, two options:
- (a) Leave the existing 10-min validator (1s-10min) in place — it's **strictly tighter** than the 24h ceiling, so tightening to 24h would *loosen* it, not strengthen it.
- (b) Tighten validator to `MAX_SANE_DURATION_MS`.

Chose **(a)** — `config.js` is unchanged. The 10-min validator catches file-loaded values; the three sites above act as defense-in-depth for programmatic instantiation paths that bypass config validation (embedders, direct `SessionManager` construction, future env-var paths that route differently).

## Tests

- `session-manager.test.js` — extends the `#4509` `TIMEOUT_SPECS` table to cover the 4th knob with the existing spec-driven pattern (clamps + exact boundary).
- `server-cli-startup-timeouts.test.js` — shape, in-range, bad-input, over-ceiling, and exact-boundary cases for `mcpToolCallTimeoutMs`.
- `byok-session.test.js` — over-ceiling falls back to fleet default; exact boundary passes through.

All 273 tests across the three affected files pass; full server suite has 4 unrelated pre-existing flakes (cloudflared/encryption/web-task-manager/feature-detection) that fail identically on `main`.

## Test plan

- [x] `node --test packages/server/tests/session-manager.test.js packages/server/tests/server-cli-startup-timeouts.test.js packages/server/tests/byok-session.test.js` — 273 pass, 0 fail
- [x] RED → GREEN proven per site (test added, ran, confirmed fail, then implemented)

Closes #4517

### DIFF
--- a/packages/server/src/byok-session.js
+++ b/packages/server/src/byok-session.js
@@ -20,6 +20,7 @@ import { homedir } from 'os'
 import { BaseSession } from './base-session.js'
 import { PermissionManager } from './permission-manager.js'
 import { createLogger } from './logger.js'
+import { isOperatorTimeoutInRange } from './duration.js'
 import {
   FALLBACK_MODELS,
   ALLOWED_MODEL_IDS,
@@ -251,8 +252,12 @@ export class ClaudeByokSession extends BaseSession {
     // non-finite / non-positive (NaN, Infinity, 0, -1, strings) falls back
     // to null because setTimeout coerces those to 0 ms and every MCP tool
     // would look broken.
+    // #4517: ceiling check via `isOperatorTimeoutInRange` (same as the three
+    // sibling timeouts in #4509) — defends per-session BYOK assignment
+    // against an over-24h value that would survive session-manager (e.g.
+    // an embedder constructing ClaudeByokSession directly with a typoed opt).
     this._mcpToolCallTimeoutMs =
-      Number.isFinite(opts.mcpToolCallTimeoutMs) && opts.mcpToolCallTimeoutMs > 0
+      isOperatorTimeoutInRange(opts.mcpToolCallTimeoutMs, { name: 'mcpToolCallTimeoutMs', log })
         ? opts.mcpToolCallTimeoutMs
         : null
   }
--- a/packages/server/src/server-cli.js
+++ b/packages/server/src/server-cli.js
@@ -1,5 +1,6 @@
 import { SessionManager } from './session-manager.js'
 import { DEFAULT_RESULT_TIMEOUT_MS, DEFAULT_HARD_TIMEOUT_MS, DEFAULT_STREAM_STALL_TIMEOUT_MS } from './base-session.js'
+import { DEFAULT_TOOL_CALL_TIMEOUT_MS } from './byok-mcp-client.js'
 import { formatIdleDuration } from './session-timeout-manager.js'
 import { isOperatorTimeoutInRange } from './duration.js'
 import { WsServer, TUNNEL_STATUS_MIN_PROTOCOL_VERSION } from './ws-server.js'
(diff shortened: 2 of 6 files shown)


blamechris:
## Agent review — PR #4538

**Verdict: APPROVE with one minor follow-up suggestion.**

Closes the audit gap from #4516 by extending the `MAX_SANE_DURATION_MS` ceiling guard to `mcpToolCallTimeoutMs` at the three sites that matter. RED → GREEN testing per site is solid (273/273 pass locally on this branch).

### Site-by-site

1. **`session-manager.js:289-292`** — drop-in swap from `Number.isFinite + > 0` to `isOperatorTimeoutInRange(... { name, log })`. Mirrors the surrounding `_resultTimeoutMs` / `_hardTimeoutMs` / `_streamStallTimeoutMs` lines exactly. `_mcpToolCallTimeoutMs` does NOT need `allowZero: true` — comment on line 285 is right (a 0-ms callTool timeout fires immediately and makes every MCP tool look broken).

2. **`server-cli.js:258-283`** — extending `resolveStartupTimeouts()` to a fourth knob fits the helper's shape: same `*TimeoutMs` (clamped null) + `effective*TimeoutMs` (resolved default) duo. The doc comment is updated.

3. **`byok-session.js:254-257`** — defense-in-depth for direct `ClaudeByokSession` construction (embedders that bypass SessionManager). Uses the module-level `log` (line 39) so no risk of an undefined reference. The over-ceiling test passes a value via the SessionManager → providerOpts → byok-session.opts path, and the test fixture covers the embedder-direct case via `buildSessionWithMockFleet`.

### Site 4 (config.js) — keep the 1s-10min validator. Right call.

The PR description option (a) is correct: `config.js`'s 1s-10min validator is **strictly tighter** than the 24h ceiling, so tightening it to `MAX_SANE_DURATION_MS` would *loosen* the file-loaded path, not strengthen it. The three sites above act as defense-in-depth for the programmatic paths (env-var, embedder, direct construction) that don't flow through `loadConfig()`. Splitting the validation across layers reflects the genuine difference between "file-config sane range" and "anything-non-pathological". This is well-justified in the PR body.

### `resolveStartupTimeouts()` extension — shape preserved?

Yes. The function's contract is "operator-tunable timeouts get a uniform `{value, effective}` pair so the SessionManager arg side stays in sync with the startup log line." Adding the fourth knob is a one-line addition per output field and one boundary line in the bad-input loop. Doc JSDoc updated. Spec-driven `TIMEOUT_SPECS` table extension keeps the per-knob assertions DRY.

### Minor finding — `effectiveMcpToolCallTimeoutMs` is unused

`resolveStartupTimeouts()` now returns `effectiveMcpToolCallTimeoutMs`, but `server-cli.js:465` destructures only `effectiveResultTimeoutMs`, `effectiveHardTimeoutMs`, `effectiveStreamStallTimeoutMs` and the `log.info` on line 469 omits the MCP value:

```js
log.info(`Inactivity soft-warning: ... hard-cap: ... stream-stall: ${stallLabel}`)
```

This is consistent with the existing pre-PR behaviour (the MCP timeout has never been surfaced at startup), but the new `effectiveMcpToolCallTimeoutMs` is genuinely dead until a future PR adds it to the log line. Two options:

- **(a) Leave as-is** — the field is harmless, matches the helper's uniform shape, and a follow-up issue can add the log-line surface.
- **(b) Extend the log line in this PR** — e.g. `... stream-stall: ${stallLabel}; mcp-tool-call: ...`. Tiny addition, makes the `effective*` field load-bearing immediately.

Either is fine. If you go with (a), worth filing a small follow-up issue to track the dead field — "extend startup log to include mcpToolCallTimeoutMs alongside the other three".

### Test coverage

- `session-manager.test.js` — extends `TIMEOUT_SPECS` (4th entry). 170/170 pass.
- `server-cli-startup-timeouts.test.js` — shape, in-range, bad-input loop, dedicated over-ceiling test, exact-boundary on all four, defaults test. Comprehensive.
- `byok-session.test.js` — over-ceiling falls back to undefined (BaseSession default applies); exact boundary passes through verbatim.

All comments explain *why*, not just *what*. The `#4517` references thread the chain back to #4516 / #4509 / #4503 cleanly.

### CI / mergeability

- All 13 CI checks pass.
- `MERGEABLE` / `CLEAN` on `main`.
- 273/273 tests across the three affected files pass locally.
- Independent of #4519/#4522 — these can ship in any order.

Ready to merge. The `effectiveMcpToolCallTimeoutMs` observation is non-blocking — surfacer can ship in a follow-up if not addressed here.

Issue #4542, opened 2026-05-30
feat(dashboard,app): per-category notification opt-in/out UI


blamechris:
## Context

Sub-issue of #4349. Blocked by #4541 (notification-prefs store + WS protocol).

## Description

Once #4541 ships the foundation, expose per-category notification toggles in:
- Mobile app Settings screen (categories: `permission`, `question`, `error`, `result`, `inactivity`, etc.)
- Dashboard SettingsPanel (same set; same store)

Both surfaces hit the same `notification_prefs_set` WS message — the server is the single source of truth.

## Acceptance Criteria

- [ ] Toggle a category off → no OS push delivered for that category, on any device
- [ ] Server-side RATE_LIMITS remain as defensive lower bound (cannot be overridden up by user prefs)
- [ ] Settings round-trip persists across app/dashboard restart
- [ ] Tests cover the resolve-decision path and a toggle round-trip

PR #4565, opened 2026-05-30
feat(server,dashboard,app): notification quiet-hours window

blamechris:
## Summary

Server-side gate inside `PushManager.send()` drops Expo pushes that land inside a per-device quiet-hours window. Operator-blocking categories (permission prompts, session errors) bypass the gate by default so the agent never silently stalls overnight.

- **Server logic** — `notification-prefs.js` `isInQuietHoursIn` is now real (replaces the #4541 stub). Timezone-aware via `Intl.DateTimeFormat` so DST + non-Pacific zones work without re-implementing zone math. `PushManager.send` filters tokens through category-mute + quiet-hours + bypass list, with the defensive `RATE_LIMITS` gate untouched as the spam ceiling.
- **Schema** — `quietHours` carries an IANA `timezone` (required when present; loader drops the window otherwise so the gate never mis-mutes). New top-level `bypassCategories` array; per-device entries may carry their own `quietHours` and `bypassCategories` — both REPLACE the global value entirely.
- **Dashboard** — `QuietHoursEditor` sub-component inside the existing Notifications section: enable toggle, HH:MM time inputs, IANA timezone select (browser default + curated short list), per-category bypass checkboxes.
- **Mobile** — equivalent `QUIET HOURS` section in `SettingsScreen`: same fields with a modal timezone picker and TextInput HH:MM validation.

## Design decisions (per #4544 brief)

- **Override resolution** — per-device REPLACES the global window. Simpler mental model than shadowing: one device can express \"my entire policy is X\" without thinking about which global fields it would inherit.
- **Default bypass categories** — `permission` + `activity_error`. Operator-blocking by nature; muting them would stall the agent. User opts every category out by sending `bypassCategories: []`.
- **Failure mode** — defensive fail-OPEN. Any structural error (missing timezone, malformed times, unknown IANA zone, missing window) returns `false` so the push goes through. Silently swallowing every notification on a misconfiguration is the worst failure mode for a notification system; the layered `RATE_LIMITS` gate + per-category mute still throttle the worst-case spam.

## Scope notes

- The dashboard QuietHoursEditor patches the GLOBAL window. Per-device override UI (the wire schema supports it) lands in #4543's per-device notification panel; this PR keeps that scope intentionally separate to avoid colliding with the parallel work.
- Existing `interceptPermissionPrompt` path, per-category mute (#4542), and `RATE_LIMITS` gate all still fire — quiet-hours is layered ON TOP of them, never replaces.
- Older servers that emit a `notification_prefs` snapshot without `bypassCategories` still parse; clients fall back to the documented defaults at the gate.

## Test plan

- [x] **Server quiet-hours unit tests** (12 new suites in `notification-prefs-quiet-hours.test.js`):
  - midnight-wrap window inclusive/exclusive boundaries
  - DST spring-forward + fall-back in `America/Los_Angeles`
  - per-device override REPLACE precedence (including `null` opt-out)
  - per-device different-timezone resolution (Tokyo vs Pacific)
  - bypass-category gating (default list, per-device override, empty opt-out)
  - PushManager.send filters tokens correctly + RATE_LIMITS still fires first
- [x] **Server existing tests** (`notification-prefs.test.js`, `notification-prefs-handlers.test.js`, `ws-schemas.test.js`) updated for the extended schema (timezone required, bypassCategories accepted) — all pass.
- [x] **Server full suite**: 6044/6050 pass; the 6 fails are pre-existing environmental flakes (cloudflared-needs-installed + a Chroxy server happening to be running on the test machine), reproduced on unmodified `main`.
- [x] **Dashboard SettingsPanel tests**: 66/66 pass (8 new quiet-hours editor tests covering loading state, enable toggle defaults, Save round-trip, bypass checkboxes, snapshot fallback when older server omits the field).
- [x] **Dashboard full suite**: 2272/2272 pass; `tsc --noEmit` clean.
- [x] **Mobile suite**: 1363/1363 pass (35 tests in `SettingsScreenNotificationPrefs.test.ts` including 16 new for quiet-hours editor + types + connection actions); `tsc --noEmit` clean.
- [x] **Protocol package**: 143/143 pass after the schema bump.

Closes #4544
Related to #4541 (foundation), #4549 (foundation PR), #4542 (per-category UI), #4557 (per-category UI PR), parent #4349.

### DIFF
--- a/packages/app/src/__tests__/components/SettingsScreenNotificationPrefs.test.ts
+++ b/packages/app/src/__tests__/components/SettingsScreenNotificationPrefs.test.ts
@@ -127,12 +127,102 @@ describe('message-handler.ts — notification_prefs WS message (#4542)', () => {
   });
 
   it('handles the notification_prefs case and stores the parsed snapshot', () => {
-    expect(messageHandlerSource).toMatch(/case 'notification_prefs'[\s\S]{0,800}ServerNotificationPrefsSchema\.safeParse[\s\S]{0,300}notificationPrefs:/);
+    expect(messageHandlerSource).toMatch(/case 'notification_prefs'[\s\S]{0,1500}ServerNotificationPrefsSchema\.safeParse[\s\S]{0,800}notificationPrefs:/);
   });
 
   it('logs and skips when the payload fails schema validation', () => {
     expect(messageHandlerSource).toMatch(
       /case 'notification_prefs'[\s\S]{0,600}!parsed\.success[\s\S]{0,200}console\.warn\(\s*'notification_prefs:/,
+    );
+  });
+
+  // #4544: the wire snapshot now carries an optional bypassCategories
+  // array. The handler must forward it when present so the UI sees the
+  // current gate state; absent means "use documented defaults".
+  it('forwards bypassCategories from the parsed snapshot when present (#4544)', () => {
+    expect(messageHandlerSource).toMatch(/bypassCategories\s*=\s*\(prefs\s*as[\s\S]{0,150}\.bypassCategories/);
+    expect(messageHandlerSource).toMatch(/Array\.isArray\(bypassCategories\)/);
+  });
+});
+
+describe('SettingsScreen — Quiet hours editor section (#4544)', () => {
+  it('renders a QUIET HOURS section header', () => {
+    expect(settingsSource).toMatch(/QUIET HOURS/);
+  });
+
+  it('imports the quiet-hours store actions from the connection store', () => {
+    expect(settingsSource).toMatch(/setNotificationPrefsQuietHours\s*=\s*useConnectionStore/);
+    expect(settingsSource).toMatch(/setNotificationPrefsBypassCategories\s*=\s*useConnectionStore/);
+  });
+
+  it('renders the QuietHoursEditor sub-component inside the QUIET HOURS section', () => {
+    expect(settingsSource).toMatch(/<QuietHoursEditor[\s\S]{0,800}window=\{notificationPrefs\.quietHours\}/);
+  });
+
+  it('defines the QuietHoursEditor component with the documented props shape', () => {
+    expect(settingsSource).toMatch(/function QuietHoursEditor\(props:\s*\{[\s\S]{0,400}window:\s*\{\s*start:\s*string;\s*end:\s*string;\s*timezone:\s*string\s*\}\s*\|\s*null/);
+    expect(settingsSource).toMatch(/onWindowChange:\s*\(w:[\s\S]{0,150}timezone:\s*string\s*\}\s*\|\s*null\)\s*=>\s*void/);
+    expect(settingsSource).toMatch(/onBypassChange:\s*\(categories:\s*string\[\]\)\s*=>\s*void/);
+  });
+
+  it('emits the documented quiet-hours testIDs', () => {
+    expect(settingsSource).toMatch(/testID="quiet-hours-editor"/);
+    expect(settingsSource).toMatch(/testID="quiet-hours-enabled-toggle"/);
+    expect(settingsSource).toMatch(/testID="quiet-hours-start-input"/);
+    expect(settingsSource).toMatch(/testID="quiet-hours-end-input"/);
+    expect(settingsSource).toMatch(/testID="quiet-hours-timezone-picker"/);
+    expect(settingsSource).toMatch(/testID="quiet-hours-save-button"/);
+    expect(settingsSource).toMatch(/testID=\{`quiet-hours-bypass-toggle-\$\{cat\}`\}/);
+  });
+
+  it('validates HH:MM before round-tripping', () => {
+    expect(settingsSource).toMatch(/function isValidHHMM/);
+    expect(settingsSource).toMatch(/Invalid time[\s\S]{0,100}HH:MM/);
+  });
+
+  it('declares a sensible curated timezone list', () => {
+    expect(settingsSource).toMatch(/QUIET_HOURS_TIMEZONE_CHOICES/);
+    expect(settingsSource).toMatch(/America\/Los_Angeles/);
+    expect(settingsSource).toMatch(/Europe\/London/);
+  });
+
+  it('falls back to DEFAULT_BYPASS_CATEGORIES when the snapshot omits the list', () => {
+    expect(settingsSource).toMatch(/bypassCategories=\{notificationPrefs\.bypassCategories\s*\?\?\s*DEFAULT_BYPASS_CATEGORIES\}/);
+  });
+});
+
+describe('ConnectionState — quiet-hours actions (#4544)', () => {
+  it('declares setNotificationPrefsQuietHours with the documented signature', () => {
+    expect(typesSource).toMatch(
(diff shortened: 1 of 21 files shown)


blamechris:
## Code Review: PR #4565

### Summary

Server-side quiet-hours enforcement for push notifications. `PushManager.send()` now filters tokens through a layered gate (`RATE_LIMITS` → per-category mute → quiet hours + bypass list). Timezone-aware via `Intl.DateTimeFormat`, with per-device REPLACE precedence and a documented fail-open posture on misconfiguration.

Touches 21 files across server, protocol, dashboard, and mobile-app. Adds ~2,100 lines including 12 new test suites for the quiet-hours boundary cases (midnight wrap, DST, per-device overrides, bypass categories).

### Strengths

- **Boundary coverage is excellent** — DST spring-forward and fall-back in `America/Los_Angeles`, midnight wrap with inclusive start / exclusive end, per-device Tokyo vs Pacific resolution. The intent is well-encoded in the tests.
- **Defensive fail-OPEN is the right call** for a notification system and is clearly documented at the call site (`isInQuietHoursIn` header comment) and explained in the PR body. Silently swallowing every push on misconfiguration would be the worst possible failure mode.
- **REPLACE precedence is consistent** across `quietHours` and `bypassCategories` (per-device wins entirely, including `null` opt-out). Mental model is simple and the `hasOwnProperty` distinction between absent and null is preserved end-to-end.
- **Layered gate ordering preserved** — `RATE_LIMITS` still fires first as the spam ceiling, then per-category mute (#4542), then quiet-hours. Existing `interceptPermissionPrompt` path untouched.
- **Wire schema is open-ended** — `bypassCategories` is `z.array(z.string())` without whitelisting against `ALL_CATEGORIES`, so a forward-compat client (older binary, newer prefs mentioning a category the binary doesn't ship) preserves the stored value.
- **Per-token filtering inside `send()`** correctly handles the mixed case where one device is awake and another asleep — only the eligible tokens are messaged, and the prune-on-error path still receives the full token set so Expo-side errors clean up correctly.

### Issues Found

| `packages/server/src/notification-prefs.js` | `sanitizeQuietHours` | Add range validation for HH:MM (00–23, 00–59) | The regex `/^\d{2}:\d{2}$/` accepts `25:99`. I verified — a hand-edited file with `start: '25:99'` survives the loader. At gate-eval time `_parseHHMM` returns `null` (it does the 23/59 check) and `isInQuietHoursIn` fails open, so this is silently-disabled muting rather than a security issue, but a stricter check at sanitisation time would surface the typo via the existing `log.warn` path instead of pretending the window is configured. |
| `packages/server/src/notification-prefs.js` | `isInQuietHoursIn` entry | Add `Number.isFinite` guard for `now` | I tested `isInQuietHoursIn(prefs, null, 'tok')` and it returns `true` because `new Date(null)` → 1970-01-01 UTC midnight, which is inside a `22:00–07:00` UTC window. All in-tree callers pass `Date.now()` so this is theoretical, but a defensive `Number.isFinite(now)` guard would prevent a future caller bug from silently activating quiet hours. |
| `packages/server/src/notification-prefs.js` | `_wallClockMinutes` | Memoize `Intl.DateTimeFormat` per timezone | A new `DateTimeFormat` instance is allocated on every gate evaluation (~few KB each). Push rates are bounded by `RATE_LIMITS` so this isn't a real perf issue today, but a small `Map<timezone, formatter>` cache would future-proof a hot path. |
| `packages/server/src/push.js` | 273 | Replace `defaultNotificationPrefs().bypassCategories` with `[...DEFAULT_BYPASS_CATEGORIES]` | Allocates a full prefs object just to copy the bypass list. Direct import would be cleaner and avoid the unnecessary categories/devices/quietHours allocation. |
| `packages/app/src/screens/SettingsScreen.tsx` | TextInput HH:MM | Consider a stricter input mask | `keyboardType="numbers-and-punctuation"` exposes the colon on iOS but on Android the colon is one tap deeper. A custom 4-digit masked input (auto-inserting `:` after 2 digits) would match the dashboard's `<input type="time">` ergonomics. Acceptable as-is — `isValidHHMM` rejects bad input before save. |

#### Nitpicks (Optional)

- `QuietHoursEditor` (dashboard) Save button's dirty check compares against `win?.start ?? ''`. On first enable, `win` is briefly null while the broadcast round-trips, so Save shows enabled with no actual diff. Self-corrects within ~50ms and doesn't cause incorrect saves; cosmetic only.
- Mobile + dashboard `QuietHoursEditor` re-syncs draft state on every `win` snapshot via `useEffect([win])` — if the user is mid-edit and a broadcast lands from another client, their typed-but-unsaved text gets clobbered. Acceptable for v1 (broadcasts are rare).
- `tzOptions` curated list mirrors mobile↔dashboard manually.

| Tighten HH:MM range validation in `sanitizeQuietHours` | [#4566](https://github.com/blamechris/chroxy/issues/4566) | Out of scope — the bad-window case fails open at gate-eval time so functionality is correct; this is a UX improvement to surface typos via log.warn. |
| Defensive `Number.isFinite` guard in `isInQuietHoursIn` | [#4567](https://github.com/blamechris/chroxy/issues/4567) | Out of scope — all in-tree callers pass `Date.now()` so this is purely future-proofing against a hypothetical caller bug. |
| Memoize `Intl.DateTimeFormat` per timezone | [#4568](https://github.com/blamechris/chroxy/issues/4568) | Out of scope — `RATE_LIMITS` bounds push frequency so the allocation isn't a real perf issue today. |
| Share quiet-hours timezone list across mobile + dashboard | [#4569](https://github.com/blamechris/chroxy/issues/4569) | Out of scope — duplication is small (~17 IANA names) and the lists are identical at PR time; consolidation is a small refactor for later. |
| Preserve quiet-hours editor drafts across snapshot broadcasts | [#4570](https://github.com/blamechris/chroxy/issues/4570) | Out of scope — broadcasts are rare and acceptable for v1; better UX requires a dirty-flag pattern across both editors. |

### Architecture Notes

This PR composes cleanly with the foundation:

- **#4541 (foundation)** declared the wire shape + `isInQuietHoursIn` stub. This PR fills the stub and tightens the schema (`timezone` now required when window present, `bypassCategories` added).
- **#4542 (per-category UI)** added the Notifications section in both dashboards. This PR extends the section with the `QuietHoursEditor` sub-component without restructuring the existing per-category checkboxes.
- **#4543 (per-device UI, parallel)** — the wire schema's per-device `quietHours` / `bypassCategories` fields are landed by this PR but no UI surfaces them yet. The two PRs can merge in either order; this one is deliberately scoped to the global window.
- Layered gate (rate limit → per-category mute → quiet hours) is the established push pattern. Nothing breaks the existing `interceptPermissionPrompt` flow.

The fail-open posture for misconfiguration is the correct trade-off — a notification system that silently mutes everything on a typo is worse than one that fires the missed pushes. Layered defences (`RATE_LIMITS` ceiling, per-category mute) still throttle the worst case.

### Verdict

- [x] **Approve** — Ready to merge. The deferred items above are polish; none block the merge. Test coverage is solid for the boundary cases that historically break time-window code (DST, midnight wrap, per-device timezone mismatch). The defensive fail-open posture is the right call for the failure mode this code is in the path of.

Issue #4567, opened 2026-05-30
fix(server): defensive Number.isFinite guard in isInQuietHoursIn


blamechris:
## Context

Identified during review of PR #4565 (notification quiet-hours window).

## Description

`isInQuietHoursIn(prefs, now, pushToken)` in `packages/server/src/notification-prefs.js` trusts `now` to be a valid epoch ms. I tested `isInQuietHoursIn(prefs, null, 'tok')` with a `22:00-07:00 UTC` window and it returns `true` because `new Date(null)` → 1970-01-01T00:00:00Z, which is inside the window.

All in-tree callers pass `Date.now()` so this is theoretical, but a defensive `Number.isFinite(now)` guard at the entry would prevent a future caller bug from silently activating quiet hours for every push.

## Original Review Comment

> Add `typeof now === 'number'` guard. I tested `isInQuietHoursIn(prefs, null, 'tok')` and it returns `true` because `new Date(null)` → 1970-01-01 UTC midnight, which is inside a `22:00-07:00` UTC window. All in-tree callers pass `Date.now()` so this is theoretical, but a defensive `Number.isFinite(now)` guard would prevent a future caller bug from silently activating quiet hours.

## Acceptance Criteria

- [ ] `isInQuietHoursIn` returns `false` when `now` is not a finite number (null, undefined, NaN, Infinity, string)
- [ ] Add a defensive test case in `notification-prefs-quiet-hours.test.js`

PR #4575, opened 2026-05-30
fix(server): tighten quiet-hours HH:MM range validation in sanitizeQuietHours

blamechris:
## Summary

Closes #4566.

`sanitizeQuietHours` in `packages/server/src/notification-prefs.js` previously used the regex `/^\d{2}:\d{2}$/` to validate `start`/`end`, which accepted out-of-range values like `25:99`. A hand-edited prefs file with `start: '25:99'` survived the loader and landed in the in-memory prefs. At gate-eval time `_parseHHMM` returned `null` and `isInQuietHoursIn` fail-opened, so the bad window was silently disabled rather than mis-muting — but the operator had no signal that their config was wrong.

## Changes

- Tighten `_HHMM_RE` to `^(?:[01]\d|2[0-3]):[0-5]\d$` so HH must be 00-23 and MM must be 00-59. Two-digit zero-padded form unchanged.
- `loadPrefs` warns via `log.warn` when an on-disk `quietHours` block was present but rejected by `sanitizeQuietHours`. Silent for the "absent" and "explicit null" cases — only a present, non-null, rejected block fires the warn.
- `sanitizeDevices` threads the same `log` down so per-device rejections also warn, naming the offending token.
- Doc-comments on `_HHMM_RE` and `sanitizeQuietHours` updated to call out the #4566 contract.

## Tests

Added `#4566` block in `notification-prefs-quiet-hours.test.js` covering:
- Rejected: `25:00`, `12:99`, `25:99`, `ab:cd`, `9:00`, empty string
- Accepted: `00:00`, `23:59` boundaries plus typical `09:00` / `23:59`
- Warn fires for rejected global block, stays silent when block is absent
- Per-device rejection warns and identifies the token

All 94 notification-prefs tests pass; the one unrelated full-suite failure (`WebTaskManager > feature detection`) requires the `claude` CLI binary and is pre-existing.

## Test plan

- [x] `cd packages/server && PATH="/opt/homebrew/opt/node@22/bin:\$PATH" node --experimental-test-module-mocks --test tests/notification-prefs-quiet-hours.test.js` — 57/57 pass
- [x] All notification-prefs test files — 94/94 pass
- [ ] CI green

### DIFF
--- a/packages/server/src/notification-prefs.js
+++ b/packages/server/src/notification-prefs.js
@@ -105,7 +105,11 @@ export const CATEGORY_DEFAULTS = Object.freeze(
 export const DEFAULT_BYPASS_CATEGORIES = Object.freeze(['permission', 'activity_error'])
 
 const _ALL_CATEGORIES_SET = new Set(ALL_CATEGORIES)
-const _HHMM_RE = /^\d{2}:\d{2}$/
+// Two-digit zero-padded HH:MM with HH in 00-23 and MM in 00-59 (#4566).
+// The narrower regex replaces the older `/^\d{2}:\d{2}$/`, which accepted
+// out-of-range values like `25:99` and let a hand-edited typo land in the
+// in-memory prefs before fail-opening at gate-eval time.
+const _HHMM_RE = /^(?:[01]\d|2[0-3]):[0-5]\d$/
 
 export function defaultNotificationPrefsPath() {
   return process.env.CHROXY_NOTIFICATION_PREFS_PATH || join(homedir(), '.chroxy', 'notification-prefs.json')
@@ -155,7 +159,7 @@ function sanitizeCategoryMap(raw) {
  * of muting on this device, `{ start, end, timezone }` = device-specific
  * window.
  */
-function sanitizeDevices(raw) {
+function sanitizeDevices(raw, { log } = {}) {
   if (!raw || typeof raw !== 'object') return {}
   const out = {}
   for (const [token, entry] of Object.entries(raw)) {
@@ -167,6 +171,14 @@ function sanitizeDevices(raw) {
     // "fall back to global". hasOwnProperty distinguishes absent from null.
     if (Object.prototype.hasOwnProperty.call(entry, 'quietHours')) {
       cleaned.quietHours = sanitizeQuietHours(entry.quietHours)
+      // #4566: surface a typo on a per-device window — the original block
+      // existed on disk but failed range/shape validation, so the device
+      // silently inherits "no muting" (null). Warn so the operator notices
+      // rather than discovering it the next time a notification fires
+      // outside their expected window.
+      if (cleaned.quietHours === null && entry.quietHours !== null) {
+        log?.warn?.(`notification-prefs: device ${token} quietHours rejected (invalid shape or out-of-range HH:MM)`)
+      }
     }
     if (Object.prototype.hasOwnProperty.call(entry, 'bypassCategories')) {
       const bypass = sanitizeBypassList(entry.bypassCategories)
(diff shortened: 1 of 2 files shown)


blamechris:
## Code Review: PR #4575

### Summary

Tightens HH:MM range validation in `sanitizeQuietHours` and adds a `log.warn` signal in `loadPrefs` when an on-disk quietHours block is rejected. Closes #4566.

The change replaces the loose `/^\d{2}:\d{2}$/` regex with `/^(?:[01]\d|2[0-3]):[0-5]\d$/`, threads the existing `log` option from `loadPrefs` down through `sanitizeDevices` so per-device rejections also warn, and updates doc-comments to call out the new contract.

### Strengths

- **Minimal, focused diff** — 43 lines of source + 155 lines of tests for a well-scoped behaviour change. No incidental refactoring.
- **Regex is correct and clear** — `(?:[01]\d|2[0-3])` is a standard two-digit 0-23 idiom, `[0-5]\d` for 0-59 minutes. Readable inline without a parser helper.
- **Warn-fire predicate is exactly right** — only fires for "block present but rejected"; stays silent for absent (default first-run) and explicit `null` (operator-disabled). The `rawQuietHours !== null && rawQuietHours !== undefined` guard at `loadPrefs` matches the per-device `entry.quietHours !== null` guard in `sanitizeDevices`. Both cases tested.
- **Per-device warn includes the token** — the operator can identify which device override failed validation rather than chasing through the file by hand.
- **Doc-comments updated in lockstep** — both `_HHMM_RE` and `sanitizeQuietHours` docstrings reference #4566 and explain the pre-/post-change behaviour. The block on `sanitizeQuietHours` explicitly notes that returning `null` triggers the warn in `loadPrefs`, which keeps the two sites linked.
- **TDD path is visible** — boundary tests (`00:00`, `23:59`) are pinned alongside the headline rejection cases (`25:00`, `12:99`, `25:99`), so a future regex refactor that accidentally clamps the boundary is caught.
- **No regression risk** — `_HHMM_RE` is used by both `sanitizeQuietHours` and `_parseHHMM`. Pre-#4566 `_parseHHMM` had a redundant `h > 23 || m > 59` check after the regex; that's still in place and now redundant but harmless. Worth leaving — defence in depth.

### Issues Found

#### Critical (Must Fix)

None.

None. The change is appropriately scoped — broader cleanups (e.g. removing the now-redundant `h > 23 || m > 59` check in `_parseHHMM`, deduping warn message format with the existing unreadable-file warn) would expand scope without functional benefit and are not worth a follow-up issue.

#### Nitpicks (Optional)

- **Per-device warn message includes the raw token.** ExponentPushToken values are ~50 chars and end up in operator logs. Not a leak concern (the operator already has access to the prefs file), but worth noting if there's a future "truncate tokens in logs" convention. No action needed today.

### Architecture Notes

The PR sits cleanly in the existing notification-prefs sanitisation layer (#4541, #4544). It preserves the documented invariants:

- Fail-open at the gate (`isInQuietHoursIn` still returns `false` for any structural error) — the stricter sanitiser only short-circuits the bad-window-survives-load path; the layered RATE_LIMITS gate in `push.js` remains the spam ceiling.
- Tri-state semantics on per-device `quietHours` (undefined = inherit, null = opt out, window = override) are intact. The warn is gated on "present and non-null", so the `null` opt-out path stays silent.
- `log` option threading mirrors the existing unreadable-file warn — no new logging convention introduced.

The companion #4567 PR (defensive `Number.isFinite` guard in `isInQuietHoursIn`) touches a different function in the same file; the two changes are independent and conflict-free at this diff.

### Verdict

Test coverage matches the acceptance criteria in #4566 line-for-line. No suggestions, no nitpicks requiring action, no deferred items. CI gate is the only remaining check.

Issue #4570, opened 2026-05-30
fix(ui): preserve quiet-hours editor drafts across snapshot broadcasts


blamechris:
## Context

Identified during review of PR #4565 (notification quiet-hours window).

## Description

Both `QuietHoursEditor` components (dashboard `SettingsPanel.tsx` + mobile `SettingsScreen.tsx`) re-sync their draft state on every `notificationPrefs.quietHours` snapshot via `useEffect([win])`. If the user is mid-edit (typing in start/end, picking a timezone) and another client saves a quiet-hours change, the incoming broadcast clobbers the user's typed-but-unsaved text.

For v1 this is acceptable — broadcasts are rare and a user would notice quickly — but a "you have unsaved changes — refresh?" UX (or a "dirty" guard that skips the re-sync when the form is dirty) would be more polite.

## Original Review Comment

> Mobile `QuietHoursEditor` re-syncs draft state on every `win` snapshot via `useEffect([win])` — if the user is mid-edit and a broadcast lands from another client, their typed-but-unsaved text gets clobbered. The dashboard editor has the same behaviour. Acceptable for v1 (broadcasts are rare), but a "you have unsaved changes — refresh?" UX would be more polite.

## Acceptance Criteria

- [ ] Editor tracks a `dirty` flag (true after any field is edited; cleared on Save / Disable)
- [ ] When `dirty === true`, an incoming snapshot does NOT overwrite the draft fields
- [ ] User gets a visible signal (e.g. "Snapshot updated from another client — discard your edits?") with explicit accept/discard
- [ ] Add tests for both dashboard + mobile that cover the dirty-then-broadcast case

Issue #4559, opened 2026-05-30
fix(ui): surface inline error when notification-prefs / BYOK writes fire while WS is closed


blamechris:
## Context

Identified during review of PR #4557 (per-category notification opt-in/out UI for #4542). Pre-existing gap shared with the BYOK credentials actions (#4052).

## Description

When `setNotificationPrefsCategory` (or `setByokCredentials` / `clearByokCredentials` / `refreshByokCredentialsStatus`) fires while `socket.readyState !== OPEN`, the action silently no-ops:

```ts
setNotificationPrefsCategory: (category, enabled) => {
  const { socket } = get();
  if (socket && socket.readyState === WebSocket.OPEN) {
    wsSend(socket, { type: 'notification_prefs_set', prefs: { categories: { [category]: enabled } } });
  }
  // else: silent drop, no UI feedback, no queue, no retry
},
```

On a cellular reconnect, the user toggles a category, sees the switch refuse to stay flipped, and has no idea why. Same for BYOK Save / Remove.

## Original Review Comment

> When `socket.readyState !== OPEN`, the action silently no-ops. The toggle reverts (because the snapshot didn't change), leaving the user unsure whether the click registered. Either queue locally + reapply on reconnect, or surface "Settings save failed — server disconnected" inline.

## Proposed Approach

Pick one:

**Option A: Queue + retry on reconnect** — Keep a small in-memory queue of pending settings writes. On `auth_ok`, drain the queue. Add a "Pending — will save when reconnected" inline hint. Best UX, more state to manage.

**Option B: Fail-loud inline error** — When the socket is closed, surface "Settings save failed — server disconnected." next to the control. User has to retry after reconnect. Simpler, less magic.

The right answer depends on how often users toggle settings while offline. Option B is the safer floor; Option A is a refinement on top.

## Acceptance Criteria

- [ ] No silent no-op for notification-prefs toggle, BYOK Save, BYOK Remove, or BYOK Refresh when WS is closed.
- [ ] Either: (A) the action queues and surfaces "Pending — will save when reconnected", drained on next `auth_ok`; OR (B) an inline error appears next to the control.
- [ ] Dashboard + mobile both implement the same pattern.
- [ ] Tests cover the disconnected-toggle path for at least notification-prefs (BYOK can land in the same PR if cheap).

PR #4578, opened 2026-05-30
feat(ui): optimistic notification-prefs toggle to mask WS round-trip latency

blamechris:
## Summary

Per #4558 (review of #4557). The per-category notification toggles in
both the dashboard `SettingsPanel` and the mobile `SettingsScreen` are
currently fully server-of-truth: the click → `notification_prefs_set` →
server merge + persist → broadcast → store overwrite → re-render
round-trip is hundreds of milliseconds on a cellular → Cloudflare
tunnel path, long enough that the Switch / checkbox visibly does not
move when the user clicks it.

This PR adds an optimistic local apply in front of the wire send,
keeping the server-of-truth contract intact for reconciliation.

## What changes

The four notification-prefs setter actions in both stores
(`packages/dashboard/src/store/connection.ts` and
`packages/app/src/store/connection.ts`) now:

1. Patch `notificationPrefs` locally with the requested change BEFORE
   calling `wsSend` (when the snapshot is non-null and the socket is
   open).
2. Ship the same `notification_prefs_set` message as before.
3. Rely on the existing message-handler case `notification_prefs` to
   overwrite the optimistic value with whatever the server actually
   broadcasts back. **Server wins** — rejected or coerced toggles
   visibly revert on the next snapshot.

Applies to:

- `setNotificationPrefsCategory` — per-category checkbox / Switch
- `setNotificationPrefsDevice` — per-device override (shallow-merges
  other devices and other categories under THIS device)
- `setNotificationPrefsQuietHours` — quiet-hours window set / clear
- `setNotificationPrefsBypassCategories` — global bypass list
  replacement

## Edge cases

- `notificationPrefs == null` → ship the WS message (so the server's
  reply seeds the snapshot) but do **not** mint a synthetic snapshot
  locally. The UI gates Switch / checkbox rendering on a non-null
  snapshot, so this branch is only hit by a race; minting a fake
  snapshot would lock in an incorrect "everything but this one
  enabled" baseline.
- Socket closed → no optimistic patch either. A local-only flip would
  never reconcile and would drift on the next reconnect snapshot, so
  the existing no-op contract is preserved for both wire and local
  state.

## Out of scope (intentionally, separate sub-issues)

- Inline error UI on a 5s timeout — #4559
- Editor-draft preservation across reconnect — #4570

## Test plan

- [x] `packages/dashboard/src/store/notification-prefs-optimistic.test.ts`
  (new, 9 tests):
  - Immediate local apply for each of the four actions
  - WS `notification_prefs_set` payload shape unchanged
  - Server-snapshot reconciliation via the real `handleMessage` —
    asserts the optimistic `false` is overwritten by a broadcast `true`
  - Null-snapshot guard (no synthetic snapshot, WS still sent)
  - Per-device shallow-merge preserves other category overrides under
    THIS device
- [x] `packages/app/src/__tests__/store/notification-prefs-optimistic.test.ts`
  (new, 8 tests) — mirrored for the mobile app, plus socket-closed and
  empty-`deviceKey` defensive guards
- [x] Two existing static-analysis regex windows in
  `SettingsScreenNotificationPrefs(.|Device).test.ts` widened to cover
  the new `set({...})` block now sitting between the action's
  declaration and its `wsSend` call (comment-documented)
- [x] Full dashboard suite: **2289 / 2289** passing
- [x] Full mobile suite: **1384 / 1384** passing
- [x] `tsc --noEmit` clean for both `packages/dashboard` and
  `packages/app`

Closes #4558

### DIFF
--- a/packages/app/src/__tests__/components/SettingsScreenNotificationPrefs.test.ts
+++ b/packages/app/src/__tests__/components/SettingsScreenNotificationPrefs.test.ts
@@ -115,8 +115,11 @@ describe('connection.ts — notification prefs actions (#4542)', () => {
   });
 
   it('setNotificationPrefsCategory sends a single-category notification_prefs_set patch', () => {
+    // Window widened to 1200 chars in #4558 to accommodate the optimistic
+    // local-state patch (~350 chars of `set({...})` + comment) that now
+    // sits between the function declaration and the `wsSend` call.
     expect(connectionSource).toMatch(
-      /setNotificationPrefsCategory[\s\S]{0,400}notification_prefs_set[\s\S]{0,200}\[category\]:\s*enabled/,
+      /setNotificationPrefsCategory[\s\S]{0,1200}notification_prefs_set[\s\S]{0,200}\[category\]:\s*enabled/,
     );
   });
 });
--- a/packages/app/src/__tests__/components/SettingsScreenNotificationPrefsDevice.test.ts
+++ b/packages/app/src/__tests__/components/SettingsScreenNotificationPrefsDevice.test.ts
@@ -102,15 +102,22 @@ describe('connection.ts — per-device action wiring (#4543)', () => {
   });
 
   it('setNotificationPrefsDevice sends a single-device notification_prefs_set patch', () => {
+    // Window widened to 1500 chars in #4558 to accommodate the optimistic
+    // local-state patch (~500 chars of nested `set({...})` mirroring the
+    // server's shallow-merge) that now sits between the function declaration
+    // and the `wsSend` call.
     expect(connectionSource).toMatch(
-      /setNotificationPrefsDevice[\s\S]{0,400}notification_prefs_set[\s\S]{0,300}devices:\s*\{[\s\S]{0,200}\[deviceKey\]:[\s\S]{0,200}categories:[\s\S]{0,80}\[category\]:\s*enabled/,
+      /setNotificationPrefsDevice[\s\S]{0,1500}notification_prefs_set[\s\S]{0,300}devices:\s*\{[\s\S]{0,200}\[deviceKey\]:[\s\S]{0,200}categories:[\s\S]{0,80}\[category\]:\s*enabled/,
     );
   });
 
   it('short-circuits the per-device action when deviceKey is empty', () => {
     // Defensive: even if the UI rail ever drops, the action MUST refuse to
     // ship a `devices[""]` patch which would pollute the map indefinitely.
-    expect(connectionSource).toMatch(/setNotificationPrefsDevice[\s\S]{0,300}if\s*\(!deviceKey\)\s*return/);
+    // Window widened in #4558 — the `if (!deviceKey) return` lives before
+    // the optimistic patch + wire-send, but downstream changes may push it
+    // further into the body.
+    expect(connectionSource).toMatch(/setNotificationPrefsDevice[\s\S]{0,500}if\s*\(!deviceKey\)\s*return/);
   });
 });
 
--- a/packages/app/src/__tests__/store/notification-prefs-optimistic.test.ts
+++ b/packages/app/src/__tests__/store/notification-prefs-optimistic.test.ts
@@ -1,0 +1,183 @@
+/**
+ * Mobile-app notification-prefs optimistic update tests (#4558)
+ *
+ * Mirrors the dashboard suite at
+ * packages/dashboard/src/store/notification-prefs-optimistic.test.ts —
+ * the contract is the same on both clients:
+ *
+ *   1. Clicking the per-category Switch calls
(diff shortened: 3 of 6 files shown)


blamechris:
## Code Review: PR #4578

### Summary

Adds optimistic UI updates for the four notification-prefs setter actions in both the dashboard and mobile stores, so per-category / per-device / quiet-hours / bypass toggles flip the moment the user clicks rather than waiting for the WS → server → broadcast round-trip. The local apply happens BEFORE `wsSend`; the existing `notification_prefs` message-handler then reconciles by overwriting with the server snapshot. Test coverage is solid: one new spec file per package plus minor regex-window widening on two pre-existing static-analysis tests.

### Strengths

- **Symmetric implementation across dashboard and mobile.** Same shape, same edge-case gates, same comment headers (`#4558`). Easy to keep in sync.
- **Server-wins reconciliation comes for free.** The existing `case 'notification_prefs'` handler already overwrites local state with the broadcast snapshot — no new reconciliation code, no race window where local and server states could diverge silently. The third required behavior (server snapshot overrides local on disagreement) is verified end-to-end in the dashboard suite by feeding a real handler message and asserting the optimistic `false` flips back to broadcast `true`.
- **Defensive null-snapshot guard is correct.** `notificationPrefs == null` skips the optimistic apply but still ships the WS message — minting a synthetic snapshot would have locked in an incorrect "everything but this one enabled" baseline. The comment calls out the rationale clearly.
- **Defensive socket-closed guard is correct.** Old contract (server-of-truth, no-op when socket closed) is preserved end-to-end for both wire and local state. A local-only flip with no socket would never reconcile and would drift on the next reconnect snapshot.
- **Per-device shallow-merge mirrors the server's `setPrefs` semantics exactly.** `existingDevice` and `existingCats` spreads ensure other devices and other categories under THIS device survive a single-category patch — verified by the "preserves existing per-device categories" test.
- **Scope discipline is excellent.** PR sticks tightly to optimistic update + server-snapshot reconciliation. Inline error UI (#4559) and editor-draft preservation (#4570) are explicitly deferred in the PR description.
- **Static-analysis regex widening is comment-documented.** `// Window widened to 1200 chars in #4558 ...` explains the why so future code spelunkers don't wonder.

### Issues Found

#### Critical (Must Fix)

None.

| `packages/dashboard/src/store/connection.ts` | 523-554 | The four setter actions share the same `if (socket && socket.readyState === WebSocket.OPEN)` gate and the same conditional local-apply branch. Could be extracted into a private `_optimisticAndSend(localPatch, wirePayload)` helper, but only if you find yourself adding a fifth setter. Two clients × four actions = 8 sites today and that's fine. | Minor DRY observation, not worth doing now. |
| `packages/dashboard/src/store/notification-prefs-optimistic.test.ts` | 122 | `setStore(useConnectionStore as any)` works but masks a type mismatch — the message-handler `setStore` contract presumably expects `{ getState, setState }` only, and Zustand's store exposes more. Fine as-is for a test fixture; if you ever decouple the contract, just pin the cast to that minimal shape instead of `any`. | Cosmetic. |

#### Nitpicks (Optional)

- `packages/dashboard/src/store/connection.ts:528` — the comment block on `setNotificationPrefsCategory` (#4558 rationale + edge cases) is 19 lines; the four siblings repeat condensed variants. Considered: extracting the long-form rationale to a top-of-file constant comment and pointing each action at it. Decided against suggesting — duplicated comments at call sites are easier to discover than indirect references during code review.

None — no findings warrant out-of-scope tracking. Issues #4559 (inline error UI on 5s timeout) and #4570 (editor-draft preservation) are correctly already-deferred upstream.

### Architecture Notes

This PR slots cleanly into the existing pattern at `packages/dashboard/src/store/permission-mode-optimistic.test.ts` (#3693 reference). The four affected actions remain Zustand `set({ ... })` invocations and the existing `notification_prefs` message-handler does the reconciliation — no new wire messages, no new state fields, no new reconnect contract. The acceptance-criteria item from #4558 about "no broadcast within 5s → revert + surface error" is explicitly out of scope for this PR per the description; that path is owned by sibling issue #4559.

The local-shape semantics match the server's `setPrefs` (push.js) shallow-merge — verified in tests by the "preserves existing per-device categories" assertion, which would catch a regression where the local patch wiped a sibling category that the server would have preserved. This means local optimistic state stays internally consistent with what the eventual server broadcast will look like, so the "reconciliation" step is usually a no-op for the happy path and only does real work when the server rejects / coerces a value.

### Verdict

- [x] **Approve** — ready to merge after CI + any Copilot pass.

The implementation is minimal, focused, well-tested, and follows the prior optimistic-update reference (#3693) in this codebase. Acceptance criteria 1, 2, and 3 from #4558 are met; criteria 4 (5s timeout + inline error) is correctly deferred to #4559.

PR #4579, opened 2026-05-30
perf(server): memoize Intl.DateTimeFormat per timezone in quiet-hours gate

### LINKED ISSUE #4568: perf(server): memoize Intl.DateTimeFormat per timezone in quiet-hours gate
## Context

Identified during review of PR #4565 (notification quiet-hours window).

## Description

`_wallClockMinutes` in `packages/server/src/notification-prefs.js` allocates a new `Intl.DateTimeFormat` instance on every gate evaluation. Each instance is a few KB of internal locale data. Push send rates are bounded by `RATE_LIMITS` so this isn't a real perf issue today, but a small `Map<timezone, formatter>` cache would future-proof a code path that runs once per registered token per `PushManager.send()` call (multiplied by however many `send()` calls fire during a busy session).

## Original Review Comment

> Memoize `Intl.DateTimeFormat` per timezone. A new `DateTimeFormat` instance is allocated on every gate evaluation (~few KB each). Push rates are bounded by `RATE_LIMITS` so this isn't a real perf issue today, but a small `Map<timezone, formatter>` cache would future-proof a hot path.

## Acceptance Criteria

- [ ] Module-local `Map<timezone, Intl.DateTimeFormat>` cache populated lazily
- [ ] Cache is keyed on timezone string; identical formatter options reused across calls
- [ ] Existing test suite still passes (no functional change)

blamechris:
## Summary

`_wallClockMinutes` in `packages/server/src/notification-prefs.js` allocated a fresh `Intl.DateTimeFormat` on every quiet-hours gate evaluation. Each instance carries a few KB of locale data, and the gate runs once per registered push token per `PushManager.send()` call.

This PR memoizes the formatter in a module-local `Map<timezone, Intl.DateTimeFormat>`, keyed by IANA zone, populated lazily.

## Why this shape

- **Plain `Map`, no eviction.** The set of timezones in real prefs is tiny (one global window + at most one per device) and stable across the process lifetime. LRU bookkeeping would be more code than the cache it protects.
- **Cache only successful constructions.** A `RangeError` from a bad IANA zone returns `null` upstream so the gate fail-opens. We deliberately do NOT cache the failure — a later prefs reload that fixes the typo reconstructs cleanly with the same key.
- **Test-only escape hatch.** `_resetDateTimeFormatCacheForTests` lets spy-on-constructor tests deterministically count invocations. Production code never calls it.

Closes #4568.

## Test plan

- [x] New `isInQuietHoursIn — Intl.DateTimeFormat memoization (#4568)` suite (4 tests):
  - construct exactly once per timezone across repeated gate calls
  - distinct timezones get distinct formatters
  - failed constructions are NOT cached (re-attempts every call)
  - DST spring-forward + boundary results still correct across cache hits
- [x] `npm test` (packages/server): 6094 pass / 5 pre-existing unrelated failures (`TunnelManager Integration (requires cloudflared)`, `WebTaskManager` claude-CLI feature detection, encryption-integration nonce flake — all environmental, unaffected by this change)
- [x] Notification-prefs + push focused run: 123/123 pass

### DIFF
--- a/packages/server/src/notification-prefs.js
+++ b/packages/server/src/notification-prefs.js
@@ -393,6 +393,55 @@ export function shouldBypassQuietHours(prefs, category, pushToken) {
 }
 
 /**
+ * Memoized `Intl.DateTimeFormat` instances keyed by IANA timezone (#4568).
+ *
+ * The quiet-hours gate runs once per registered push token per
+ * `PushManager.send()` call. Each `Intl.DateTimeFormat` instance carries a
+ * few KB of locale data, so re-allocating on every call wastes work in a hot
+ * path that scales with token count.
+ *
+ * Bounded memory: the set of timezones in real prefs is tiny (one global
+ * window plus at most one per device) and stable for the process lifetime,
+ * so a plain `Map` with no eviction is safe — no LRU bookkeeping needed.
+ * If a future feature lets users construct ad-hoc timezone strings (e.g.
+ * from untrusted input), revisit this; today every timezone arrives via
+ * `sanitizeQuietHours`, which only persists strings the constructor accepts.
+ *
+ * Failure-cache policy: we only cache SUCCESSFUL constructions. A
+ * `RangeError` from an unrecognised IANA zone returns `null` upstream so
+ * the gate fail-opens; never caching the failure means a later prefs reload
+ * that fixes the typo will reconstruct cleanly with the same key.
+ */
+const _dtfCache = new Map()
+
+/**
+ * Test-only escape hatch: clear the cached formatters so a test can
+ * deterministically count constructor invocations (e.g. by spying on
+ * `Intl.DateTimeFormat`). Production code never needs this — the cache is
+ * correct for the process lifetime.
+ */
+export function _resetDateTimeFormatCacheForTests() {
+  _dtfCache.clear()
+}
+
+function _getDateTimeFormat(timezone) {
+  let dtf = _dtfCache.get(timezone)
+  if (dtf) return dtf
+  try {
+    dtf = new Intl.DateTimeFormat('en-US', {
+      timeZone: timezone,
+      hour: '2-digit',
+      minute: '2-digit',
+      hourCycle: 'h23',
+    })
+  } catch {
+    return null
+  }
+  _dtfCache.set(timezone, dtf)
+  return dtf
+}
+
+/**
  * Format `now` (epoch ms) in the given IANA timezone and return its
  * minute-of-day (0..1439). Uses `Intl.DateTimeFormat` with a 24-hour
  * `h23` cycle so DST transitions and non-Pacific zones are handled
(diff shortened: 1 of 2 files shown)


blamechris:
## Code Review: PR #4579

### Summary

Memoizes `Intl.DateTimeFormat` in a module-local `Map<timezone, formatter>` so the quiet-hours gate stops re-allocating a few KB of locale data on every push. Adds 4 new tests pinning the contract (one construction per zone, distinct zones get distinct formatters, failed constructions are NOT cached, DST results still correct across cache hits) plus a `_resetDateTimeFormatCacheForTests` escape hatch for the spy-on-constructor pattern.

### Strengths

- **Doc comment carries its weight.** The block above `_dtfCache` explains why (hot path, locale-data cost), why the `Map` shape is fine (small + stable timezone set, no LRU bookkeeping needed), and the failure-cache policy with the rationale for never caching `RangeError`. Future maintainers won't have to reverse-engineer the choice.
- **Failure-cache policy is the right call.** Caching a sentinel on `RangeError` would defeat a prefs reload that swaps in a valid zone with the same key. The test pins this so a "let's optimize the throw path" refactor breaks loudly.
- **Test instrumentation is careful.** `Object.setPrototypeOf(SpyDTF, OriginalDTF)` preserves static methods (`supportedLocalesOf` etc.) so any incidental caller during the test window keeps working. Cache reset in both `beforeEach` and `afterEach` defends against suite-ordering surprises.
- **Convention adherence.** `_` prefix on `_dtfCache`, `_getDateTimeFormat`, `_resetDateTimeFormatCacheForTests` matches the existing `_wallClockMinutes` / `_parseHHMM` / `_HHMM_RE` pattern in the same file.
- **Behaviour preserved.** The DST spring-forward smoke test in the new suite calls the gate 5× with hot cache and asserts both inside/outside-window invariants — guards against any "did we capture stale offset" worry.

### Issues Found

#### Critical (Must Fix)

None.

None worth blocking on.

#### Nitpicks (Optional)

| File | Line | Issue |
|------|------|-------|
| `packages/server/tests/notification-prefs-quiet-hours.test.js` | ~717-719 | Comment says "First call may also construct one extra DTF inside the test's `dateInZone` helper" — but that test doesn't call `dateInZone` (it uses `Date.now()`). The filter-by-timezone defence is still good practice (any incidental construction from elsewhere in the suite gets ignored), but the rationale text is misleading. Either drop the comment or rewrite it as "filter by the gate's timezone so any incidental DTF construction during the test window is ignored." |

### Architecture Notes

The change stays inside `notification-prefs.js` — no new exports outside the test escape hatch, no signature changes, no callers updated. `_wallClockMinutes` keeps the same return-shape contract (`number | null`) so `isInQuietHoursIn` is untouched. The hot-path improvement scales linearly with `token-count × send-rate × distinct-timezones`, which means the wins compound as users add more devices but stay bounded by the (already tiny) timezone set.

The test-only export is a sharp tool — production code that calls it would silently invalidate the cache mid-process, which is fine semantically (next call rebuilds) but wastes the optimisation. The `ForTests` suffix is the right signal; consider a follow-up to add an ESLint rule banning imports of `_*ForTests` outside `tests/**` if this pattern repeats.

### Verdict

- [x] **Approve** — Ready to merge after the one nit (or as-is; the misleading comment doesn't change behaviour).

The contract is well-pinned by the tests, the doc comment captures the design decisions, and the fail-open posture of the surrounding `isInQuietHoursIn` is preserved unchanged. Closes #4568 cleanly.

PR #4600, opened 2026-05-30
refactor(dashboard): consolidate duration helpers (#4510)

blamechris:
## Summary
- Extract `formatDurationTerse` and `formatDurationVerbose` to `packages/dashboard/src/utils/duration.ts`
- Replace inlined helpers in `ActivityIndicator.tsx` and `StreamStallChip.tsx`
- Add `Math.max(1, seconds)` guard on verbose path

## Test plan
- [x] `cd packages/dashboard && npx vitest run src/utils/duration.test.ts` (unit coverage)
- [x] `cd packages/dashboard && npx vitest run` (full dashboard suite stays green)

Closes #4510

### DIFF
--- a/packages/dashboard/src/components/ActivityIndicator.tsx
+++ b/packages/dashboard/src/components/ActivityIndicator.tsx
@@ -25,6 +25,7 @@ import { useEffect, useId, useMemo, useRef, useState } from 'react'
 import { useShallow } from 'zustand/react/shallow'
 import { formatToolName } from '@chroxy/store-core'
 import { useConnectionStore } from '../store/connection'
+import { formatDurationTerse } from '../utils/duration'
 
 /**
  * #4420 — Tail-ellipsis truncation for the pending-shell command label. User-
@@ -104,17 +105,11 @@ function formatElapsed(ms: number): string {
 }
 
 // #4308 — duration without the "ago" suffix, used for the "Running X · 12s"
-// label (the named tool is current, not past, so "ago" is wrong).
-function formatDuration(ms: number): string {
-  const s = Math.max(0, Math.floor(ms / 1000))
-  if (s < 60) return `${s}s`
-  const m = Math.floor(s / 60)
-  const remS = s % 60
-  if (m < 60) return remS === 0 ? `${m}m` : `${m}m ${remS}s`
-  const h = Math.floor(m / 60)
-  return `${h}h ${m % 60}m`
-}
-
+// label (the named tool is current, not past, so "ago" is wrong). #4510 — the
+// implementation lives in `utils/duration.ts` (`formatDurationTerse`) so the
+// terse register is shared with any future consumer; ActivityIndicator
+// continues to consume the terse form because the chip's compact pill layout
+// can't accommodate "5 minutes 12 seconds".
 function statusClass(elapsedMs: number, timeoutMs: number): string {
   if (elapsedMs >= timeoutMs - 60_000) return 'activity-indicator--red'
   if (elapsedMs >= 60_000) return 'activity-indicator--orange'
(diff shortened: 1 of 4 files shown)


blamechris:
## Code Review: PR #4600

### Summary
Extracts the two inlined duration formatters from `ActivityIndicator.tsx` (#4308) and `StreamStallChip.tsx` (#4497) into a single `packages/dashboard/src/utils/duration.ts` module with named `formatDurationTerse` / `formatDurationVerbose` exports, plus a dedicated unit test file. Adds a `Math.max(1, seconds)` guard and a non-finite-input guard on the verbose path so the helper is safe for future reuse without per-call-site gating.

### Strengths
- **API surface choice is right.** Two separate named exports beat a boolean flag because (a) terse uses `Math.floor`, verbose uses `Math.round` — they're semantically distinct, not a presentation toggle, and (b) the call site reads its intent (`formatDurationTerse(…)` vs `formatDurationVerbose(…)`) without parameter-spelunking. Easy to grep, hard to misuse.
- **Zero behavior drift at call sites.** `ActivityIndicator` already passes `Math.max(0, now - startedAt)` to every consumer, so the terse path can't see negatives in production. `StreamStallChip` gates with `typeof timeoutMs === 'number' && Number.isFinite(timeoutMs) && timeoutMs > 0` before calling the verbose path, so the new `Math.max(1, …)` / non-finite guard never activates in current code. Hardening is purely defensive for future consumers.
- **Latent-bug hardening is well-justified.** The original `humanizeDuration(0)` returned the meaningless prose `"0 seconds"`; the new `Math.max(1, …)` floor makes the helper safe to reuse without re-implementing the pre-call gate. Documented in the JSDoc and locked in by tests.
- **Test coverage is comprehensive.** Terse: 0/499/999 sub-second floor, 1s/30s/59s, exact 60s and 120s minute boundaries, mixed `1m 5s` / `5m 7s`, exact 1h/`1h 2m`/`2h 30m`, negative-input clamp. Verbose: sub-500ms → "1 second", exact 1s/2s/30s/59s, exact 60s singular minute, 2/5/59 minutes plural, exact 1h singular, 2h/5h plural, NaN guard, Infinity guard. Every acceptance criterion in #4510 is locked in.
- **Doc comments explain the why.** The two-register rationale, the `Math.max(1, …)` rationale, and the non-finite fallback are all called out — future readers won't need archaeology.
- **Module placement matches sibling conventions.** `utils/duration.ts` + `utils/duration.test.ts` mirrors the existing pattern (`attachment-utils.ts`, `clipboard-image.ts`, `platform.ts`).

### Issues Found

#### Critical (Must Fix)
None.

None blocking.

#### Nitpicks (Optional)
| `ActivityIndicator.tsx` | 105-112 | The retained explanatory comment block (originally a docstring for the removed `formatDuration`) now sits floating above `statusClass`. Consider moving the prose alongside the import on line 28 or trimming to a one-liner — currently reads like a docstring for `statusClass`, which it isn't. | Pure readability nit; the JSDoc on `duration.ts` already covers the register-choice rationale. |
| `duration.ts` | 22-30 | `formatDurationTerse` does not get the same `Math.max(1, …)` hardening as the verbose form. Intentional (chip layout wants `"0s"` for sub-second), and documented as "mirrors the original byte-for-byte" — but if future consumers also want "never zero" semantics, a small `{ floor?: boolean }` option would beat a third helper. Defer until a second consumer asks. | Forward-looking only; current state is correct. |

None — no out-of-scope suggestions worth a tracker entry. The refactor is self-contained and the hardening is captured in tests.

### Architecture Notes
Sits cleanly inside the existing `packages/dashboard/src/utils/` conventions. No protocol changes, no store touchpoints, no WS surface area. The duo of exports gives a natural home for a third "Nh Nm Ns" register if monitoring/uptime widgets ever need it, without forcing today's two consumers to pass any options.

### Verdict

Refactor is faithful, hardening is justified, tests are thorough, and both call sites read more cleanly than before. Recommend merging once Copilot weighs in.

blamechris:
## Review Comments Addressed

No inline review comments or bot reviews were posted on this PR — Copilot review is not currently active on this repo and no human reviewers left inline feedback. The agent-review pass found no critical or blocking issues (verdict: Approve), only two optional nitpicks documented in the review comment above.

| — | No review comments to process | n/a | n/a |

- Fixed: 0
- False positives: 0
- Follow-up issues created: 0
- Existing issues closed: 0

**CI:** all 13 checks passing.
**Threads:** 0 unresolved.
**Ready for merge:** yes (pending user approval).

PR #4606, opened 2026-05-30
fix(server): validate device-token format in notification_prefs_set (#4551)

blamechris:
## Summary
- `handleNotificationPrefsSet` now iterates `patch.prefs.devices` keys and validates each via `PushManager.isValidPushTokenFormat`
- Malformed keys produce `INVALID_REQUEST` and skip the persist call entirely
- Prevents a buggy/malicious client from bloating `~/.chroxy/notification-prefs.json`

## Test plan
- [x] Test with malformed key → INVALID_REQUEST + setPrefs NOT called
- [x] Test with valid 64-char hex key → succeeds
- [x] Full server suite stays green

Closes #4551

### DIFF
--- a/packages/server/src/handlers/input-handlers.js
+++ b/packages/server/src/handlers/input-handlers.js
@@ -8,6 +8,7 @@
 import { randomUUID } from 'node:crypto'
 import { validateAttachments, resolveFileRefAttachments, resolveSession, sendError } from '../handler-utils.js'
 import { evaluateDraft as defaultEvaluateDraft, shouldSkipEvaluator } from '../prompt-evaluator.js'
+import { PushManager } from '../push.js'
 import { createLogger } from '../logger.js'
 
 const log = createLogger('ws')
@@ -484,6 +485,22 @@ function handleNotificationPrefsSet(ws, client, msg, ctx) {
   if (!patch || typeof patch !== 'object') {
     sendError(ws, msg?.requestId, 'INVALID_REQUEST', 'prefs object is required')
     return
+  }
+  // #4551 — validate every device-key against the same gate
+  // register_push_token uses (>= 20 chars, no whitespace/punctuation).
+  // The Zod schema caps the devices map size as a DoS guard but does
+  // not check key format, so without this an authenticated client can
+  // stuff arbitrary strings into ~/.chroxy/notification-prefs.json and
+  // have them re-served on every notification_prefs_get. Reject the
+  // whole patch on the first bad key (no partial-apply) so the
+  // on-disk state stays clean.
+  if (patch.devices && typeof patch.devices === 'object') {
+    for (const key of Object.keys(patch.devices)) {
+      if (!PushManager.isValidPushTokenFormat(key)) {
+        sendError(ws, msg?.requestId, 'INVALID_REQUEST', 'Invalid device token format in notification_prefs_set')
+        return
+      }
+    }
   }
   let next
   try {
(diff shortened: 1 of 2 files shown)


blamechris:
## Agent Review: PR #4606

### Summary
Fix-PR for #4551. Adds device-key format validation to `handleNotificationPrefsSet` so the same `isValidPushTokenFormat` gate used by `register_push_token` is enforced on every key in `patch.devices`. Failure short-circuits with `INVALID_REQUEST` before `setPrefs` is called, so the on-disk prefs file is never touched on a bad patch.

### Strengths
- **No partial-apply.** The validator loop runs before any persist call, and the first bad key emits `INVALID_REQUEST` + `return` — `setPrefs` is never invoked. Confirmed in the negative tests via a `setPrefs` spy plus `existsSync(prefsPath) === false`.
- **Frame matches existing conventions.** `sendError(ws, msg?.requestId, 'INVALID_REQUEST', ...)` mirrors the `prefs object is required` branch directly above; error code, requestId threading, and signature are all consistent with neighbour handlers.
- **Reuses the canonical validator.** Importing `PushManager.isValidPushTokenFormat` rather than rolling a second regex avoids drift with `register_push_token` (`packages/server/src/push.js:472`). Any future relaxation/tightening of the format flows to both code paths automatically.
- **Fixture upgrade preserves original test intent.** The `tok-a`/`tok-b` -> `ExponentPushToken[...]` 45-char swap keeps the per-device-override-defeats-global-mute assertion shape intact; only the literal key strings change. Pre-existing coverage is not weakened.
- **Tests are well-isolated.** All three new cases reuse the temp `prefsPath` from `beforeEach`, so they cannot contaminate `~/.chroxy/notification-prefs.json` (memorized hazard from prior incidents).
- **Defensive `typeof patch.devices === 'object'` guard** prevents the `Object.keys(null)` throw when `devices` is omitted / explicitly null — a partial patch that only changes `categories` still passes through unchanged.

### Issues Found

#### Critical (Must Fix)
None.

| `packages/server/src/handlers/input-handlers.js` | 497 | Consider also guarding against array-typed `devices` (`Array.isArray(patch.devices)` -> reject) | `typeof [] === 'object'` so a client sending `devices: ['foo']` reaches the loop; `Object.keys(['foo'])` returns `['0']` which then fails the validator and (correctly) rejects, so the current behavior is safe-by-accident. Explicit rejection with a clearer error would document intent. **Not blocking** — current path still rejects, just with the same generic message. |
| `packages/server/src/handlers/input-handlers.js` | 490 | Comment says register_push_token requires "no whitespace/punctuation" but `isValidPushTokenFormat` also rejects quotes/brackets/shell-metas | Minor doc drift; the JSDoc on `isValidPushTokenFormat` itself is the source of truth. Not worth a follow-up commit. |

#### Nitpicks (Optional)
- `'a'.repeat(64)` in the accept-valid test is technically valid hex but not representative of any real platform token format. Using a representative `ExponentPushToken[...]` or FCM-shaped string would make the test self-documenting. Cosmetic.

### Architecture Notes
The validation lives at the handler boundary, which is the right layer — Zod schema already caps the `devices` map cardinality (DoS guard) but cannot encode the per-key format rule without duplicating `isValidPushTokenFormat`'s logic, and pushing the check into `PushManager.setPrefs` would force every direct-call site (tests, future internal callers) to learn the rule. Gating at the WS handler matches the threat model: untrusted client input.

### Test Coverage
- New tests: 3 (malformed-with-spaces, too-short, valid-64-char-hex)
- Existing test adjusted: 1 (per-device override fixture upgraded to real-format keys)
- Coverage delta: positive — the rejection path was previously untested
- All tests use temp `prefsPath`; no real state file risk

### Verdict
**Approve** — ready to merge. Tight, scoped, well-tested fix that closes a real CVE-shape gap (authenticated client could bloat `~/.chroxy/notification-prefs.json`). Suggestions are non-blocking polish items.

PR #4619, opened 2026-05-31
fix(server): sweep unresolved tool_starts on session restore (#4617)

### LINKED ISSUE #4617: fix(server,store-core): session restore re-injects unresolved tool_starts → permanent activeTools zombies
## Symptom (v0.9.18, 2026-05-30 live)

After restarting Chroxy.app cleanly with sessions persisted (SIGTERM → session-state.json flush → fresh launch → restore), the dashboard footer pill on a session that had a wedged AskUserQuestion before the restart shows \`Running AskUserQuestion · 4h 26m\` — where the elapsed time corresponds to the ORIGINAL pre-restart tool_use_id's wire timestamp.

The wedged tool's PreToolUse fired at ~13:00 PDT in the pre-restart session. The user-initiated restart at ~16:31 PDT spawned a fresh claude TUI process. By ~17:26 PDT the chip reads 4h 26m → ~13:00 PDT origin → the OLD tool_use_id is the source.

## Root cause

`session-state.json` serializes each session's `history` array including every `tool_start` event. On restore, the server's history-replay path re-emits those `tool_start` events to subscribers (including the dashboard's `applyToActiveTools` path), with their ORIGINAL wire timestamps (per #4612's correct fix).

**Tools whose matching `tool_result` was never emitted before the shutdown** (because the session was wedged on AskUserQuestion, or any other unresolved tool) re-enter `activeTools` as zombies with no path to ever clear. They keep ticking in the footer pill forever — across restart after restart.

## Reproduction

1. Trigger any unresolvable tool (e.g. multi-question AskUserQuestion — #4604) on a session
2. Wait for the wedge — server has `_isBusy=true` indefinitely
3. Graceful Chroxy quit (SIGTERM): `session-state.json` flushes with the unresolved `tool_start` in history
4. Relaunch Chroxy
5. Session restores; dashboard renders the old `tool_start` in activeTools; chip starts ticking from the OLD wire timestamp

Live evidence from 2026-05-30 — chip shows \`Running AskUserQuestion · 4h 26m\` even though the user's NEW message + NEW AskUserQuestion fired 9 minutes ago.

## Fix options

1. **Sweep at restore time** (cleanest): when SessionManager replays history during restore, drop `tool_start` entries that don't have a matching `tool_result` later in the same history array. Per-session, single linear scan.

2. **Synthesize tool_result on the unresolved entries** during restore: instead of dropping the tool_start, prepend a synthetic `tool_result` with `interrupted: true, reason: 'session_restored'` so the dashboard can render the bubble in a \"interrupted\" state rather than silently dropping it.

3. **Dashboard-side timeout** for activeTools entries older than the session's connect time: if a `tool_start` has a wire timestamp older than the current WebSocket connection's open time, treat it as orphaned and don't add to activeTools. Hackier; doesn't help users who actually were watching a long-running tool across reconnects.

Recommend **option 2** — preserves the bubble for context, makes the interruption visible, gives the user closure.

## Acceptance

- [ ] On session restore, `tool_start` events without matching `tool_result` get a synthetic `tool_result{interrupted: true, reason: 'session_restored'}` injected (or some equivalent path that clears activeTools)
- [ ] Test: serialize a session with an unresolved tool_start, restore, assert activeTools is empty (or the tool is marked interrupted)
- [ ] Visual: dashboard bubble for the synthesised interrupted tool renders distinctly from \"in progress\" — perhaps muted/strikethrough or a small \"⚠ interrupted\" badge
- [ ] Verify by repro: after Chroxy restart, no \`Running X · Nh Mm\` zombie pill

## Related

- #4604 — parent (TUI wedge that exposed this)
- #4612 — #4607's fix correctly honours wire timestamp; this issue is the natural follow-on for the case where the wire timestamp shouldn't even be honoured because the tool is orphaned
- The companion issue I'm filing for the watchdog → tool_result gap covers the live (non-restart) case; this issue covers the post-restart case

blamechris:
## Summary
- Before this PR: if a session was wedged on a tool when chroxy shut down, the unresolved tool_start was persisted to session-state.json. On next restore, history replay re-emitted it to the dashboard and activeTools picked it up — but no path ever cleared it. Footer pill showed "Running X · 4h+" forever.
- After: session restore scans history and synthesizes a tool_result{interrupted:true, synthetic:true, reason:'session_restored'} for any tool_start that doesn't have a matching tool_result. Dashboard's tool-pairing path clears activeTools normally.

## Test plan
- [x] Unit tests cover the sweep logic + edge cases
- [x] Live verification will land when v0.9.19 is built — restart with a wedged session, confirm activeTools is empty on dashboard reconnect

Closes #4617

### DIFF
--- a/packages/server/src/session-manager.js
+++ b/packages/server/src/session-manager.js
@@ -1334,9 +1334,16 @@ export class SessionManager extends EventEmitter {
             if (n > this._sessionCounter) this._sessionCounter = n
           }
         }
-        // Restore message history if present (v1+)
+        // Restore message history if present (v1+).
+        // Sweep any `tool_start` that lacks a matching `tool_result` and
+        // splice in a synthetic interrupted result (#4617) BEFORE seeding
+        // history so the subsequent dashboard history replay never sees a
+        // dangling tool_start. Without this, an unresolved tool_use from
+        // before shutdown re-enters `activeTools` on replay and never gets
+        // cleared — the footer shows "Running X · Nh Mm" forever.
         if (hasVersion && Array.isArray(saved.history) && saved.history.length > 0) {
-          this._history.setHistory(sessionId, saved.history)
+          const swept = SessionMessageHistory.sweepUnresolvedToolStarts(saved.history)
+          this._history.setHistory(sessionId, swept)
         }
         if (typeof saved.lastActivityAt === 'number' && Number.isFinite(saved.lastActivityAt) && saved.lastActivityAt > 0) {
           this._sessionLastActivityAt.set(sessionId, saved.lastActivityAt)
--- a/packages/server/src/session-message-history.js
+++ b/packages/server/src/session-message-history.js
@@ -152,6 +152,15 @@ export class SessionMessageHistory extends EventEmitter {
         const baseTs = typeof entry.timestamp === 'number' && Number.isFinite(entry.timestamp)
           ? entry.timestamp
           : Date.now()
+        // `synthetic` / `interrupted` / `isError` / `reason` are diagnostic
+        // hints only — no client code branches on them today, and the wire
+        // schema (`ServerToolResultSchema` in packages/protocol) strips
+        // unknown fields on parse. The behaviour that clears the dashboard's
+        // activeTools entry is driven purely by the `tool_result` type +
+        // matching `toolUseId` going through `handleToolResult.applyToActiveTools`
+        // (store-core/handlers/index.ts). Keep these fields anyway so the
+        // synthetic stays grep-able on disk and a future renderer can show
+        // a distinct "interrupted" badge without a protocol change.
         out.push({
           type: 'tool_result',
           toolUseId: entry.toolUseId,
--- a/packages/server/src/session-message-history.js
+++ b/packages/server/src/session-message-history.js
@@ -110,6 +110,67 @@ export class SessionMessageHistory extends EventEmitter {
   }
 
   /**
+   * Sweep an in-memory history array for `tool_start` entries that lack a
+   * matching `tool_result` and splice in a synthetic `tool_result` right
+   * after each one. Used during session restore (#4617) so that a session
+   * which was wedged on a tool when chroxy shut down does not zombify the
+   * dashboard's `activeTools` pill on the next history replay — the
+   * synthetic result rides the same handler path that normally clears
+   * activeTools (`handleToolResult.applyToActiveTools`).
+   *
+   * Returns a NEW array; the input is not mutated. The original ordering
+   * is preserved and the synthetic result is inserted immediately after
+   * its matching `tool_start`, with a timestamp one millisecond later so
+   * downstream consumers that sort by timestamp stay monotonic without
+   * pretending the tool completed "now".
+   *
+   * Safe to call on:
+   *   - empty / non-array input (returns the input unchanged)
+   *   - history with no tool_start entries (returns a shallow copy)
+   *   - history with all tool_starts already matched (returns a shallow copy)
+   *
+   * @param {Array} history
+   * @returns {Array}
+   */
+  static sweepUnresolvedToolStarts(history) {
+    if (!Array.isArray(history) || history.length === 0) return history
+    const resolved = new Set()
+    for (const entry of history) {
+      if (entry && entry.type === 'tool_result' && typeof entry.toolUseId === 'string') {
+        resolved.add(entry.toolUseId)
+      }
+    }
+    const out = []
+    for (const entry of history) {
+      out.push(entry)
+      if (
+        entry
+        && entry.type === 'tool_start'
+        && typeof entry.toolUseId === 'string'
+        && !resolved.has(entry.toolUseId)
+      ) {
+        const baseTs = typeof entry.timestamp === 'number' && Number.isFinite(entry.timestamp)
+          ? entry.timestamp
+          : Date.now()
+        out.push({
+          type: 'tool_result',
+          toolUseId: entry.toolUseId,
+          result: 'Tool was in flight when chroxy was last shut down. Tool may have continued or been cancelled — no record of outcome.',
+          interrupted: true,
+          isError: true,
+          synthetic: true,
+          reason: 'session_restored',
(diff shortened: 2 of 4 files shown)


blamechris:
## Code Review: PR #4619

### Summary

Adds a single static method, `SessionMessageHistory.sweepUnresolvedToolStarts(history)`, called from `SessionManager.restoreState()` between reading `saved.history` and seeding the in-memory ring buffer. Any `tool_start` without a matching `tool_result` gets a synthetic `tool_result` spliced in immediately after it, so the next `replayHistory()` chunk that fans out to dashboard subscribers carries the missing pair and the store-core `applyToActiveTools` filter clears the zombie entry naturally.

Cleanly scoped, surgical, and zero new wire-protocol surface.

### Strengths

- **Sweep placement is correct.** `session-manager.js:1344-1347` runs the sweep BEFORE `this._history.setHistory(sessionId, swept)`. The seeded array is the same one `getHistory()` returns (`session-manager.js:1467-1468` → `session-message-history.js:81-83`), which `replayHistory()` in `ws-history.js:368` consumes verbatim. No path reads `saved.history` directly after the sweep — the swept array is the only thing that ever reaches `_messageHistory`. End-to-end: load state → sweep → setHistory → getHistory → replayHistory → send to client → store-core `handleToolResult.applyToActiveTools` → entry filtered by toolUseId.
- **Production wire path.** The synthetic carries `type: 'tool_result'` + `toolUseId` matching the orphan tool_start. `applyToActiveTools` (`store-core/handlers/index.ts:3562-3565`) filters strictly on `toolUseId` and ignores everything else, so the synthetic clears the entry exactly like a real result. No new client code needed.
- **Defence-in-depth extras (beyond spec) are well-tested.** Date.now() fallback for missing timestamps and dedup on duplicate tool_starts both have dedicated unit tests (lines 491-503, 519-530 of the new test block).
- **Non-mutation is real.** `sweepUnresolvedToolStarts` builds a fresh `out` array and pushes each input entry by reference plus the synthetic. The input array's identity and contents are preserved (test at line 481 asserts deep-equality via JSON round-trip — see suggestion below).
- **Backward compat is clean.** Older `session-state.json` files with no unresolved tool_starts hit the no-op path (test at line 449). Older files WITH unresolved tool_starts heal automatically on first restore.
- **Healing the wedged session.** The user's current v0.9.18 zombie session-state.json (the AskUserQuestion stuck at 13:00 PDT showing \"Running X · 4h+\") will be cleaned on the next restart with this PR's code — the persisted unresolved tool_start gets a synthetic spliced in, the dashboard re-replay clears `activeTools`, and the footer pill drops.
- **CI is green** — all 13 server suites pass, 221 server tests pass locally.

### Issues Found

#### Critical (Must Fix)

None.

| `packages/server/src/session-message-history.js` | 158-163 | Add a code comment that `synthetic`, `interrupted`, `isError`, `reason` are **cosmetic/diagnostic only** — no client code consumes them today. The zod schema (`ServerToolResultSchema` in `packages/protocol/src/schemas/server.ts:148-153`) only declares `type`/`toolUseId`/`result`/`truncated`, so any client that runs the parser will strip the extras silently. The clearing behavior is entirely driven by `applyToActiveTools` matching on `toolUseId`. | Without this note, a future reader (or future #4618-style work) might assume the extras drive behavior and add code that depends on them. The fields are still worth keeping for log/grep visibility on disk and for future renderer work, but the comment makes the contract explicit. |
| `packages/server/tests/session-message-history.test.js` | 478-486 | The non-mutation test uses `JSON.parse(JSON.stringify(input))` + `deepStrictEqual` rather than asserting object identity. That catches *value* mutation but not *array-reference* mutation (e.g. `Array.prototype.splice` on the input that happened to leave values identical). Belt-and-braces: also assert `assert.strictEqual(out !== input, true)` and that `input.length` is unchanged. | The implementation is correct — it builds a fresh `out` — but the test could miss a regression where someone refactors to `history.splice(...)` and the splice happens to be a no-op for the test fixture. Tiny add. |
| `packages/server/src/session-message-history.js` | 152-154 | Edge case worth noting (not fixing): if two consecutive tool_starts share an identical `timestamp` (rare but possible — e.g. parallel tool dispatch in the same tick), the second's synthetic timestamp (`ts + 1`) could collide with the first synthetic's. Not a correctness bug — consumers don't require strict global monotonicity, only per-pair monotonicity — but worth a one-line comment if you want it documented. | Cosmetic / defensive. Skip if you'd rather keep the diff minimal. |

#### Nitpicks (Optional)

- Doc comment at line 130 says \"history with no tool_start entries (returns a shallow copy)\" but the implementation actually returns a NEW array containing the same entries by reference — that is a shallow copy, so the doc is accurate; just confirming.
- The integration test (`session-manager.test.js:906-953`) verifies the load → sweep → setHistory → getHistory chain, but stops short of asserting that the WS replay path emits the synthetic. That's fine — the unit test on the sweep covers the splice and the existing replay machinery is unchanged — but if you want full end-to-end coverage in one test, you could subscribe to the SessionManager's event emitter or stub `replayHistory`. Out of scope for this PR.

None — every suggestion above is either a comment add or a test polish; both fit cleanly in this PR if you want to land them. No follow-up issues created.

### Architecture Notes

The sweep belongs to `SessionMessageHistory` as a static helper (correct — it operates on a plain history array, no instance state needed) and is invoked exactly once from `SessionManager.restoreState`. This avoids the trap of running on every `setHistory` call, which would also have to handle truncation / streaming — restore is the only place a half-pair can legitimately appear.

Note that the sweep handles only the SessionManager restore path. If `JsonlSubprocessSession` or another resume path ever rehydrates history from a different source (e.g. the JSONL file via `getFullHistoryAsync`), it would also need the same treatment. Today `getFullHistoryAsync` reads `.jsonl` files which have their own structure and don't traffic in chroxy's `tool_start`/`tool_result` envelope types, so this is theoretical — but worth keeping in mind for future #4604-class fixes.

### Verdict

**Approve — Ready to merge.**

Sweep is correctly placed, non-mutating, well-tested, and heals the user's current production wedge on next restart. The two suggestions above (code comment about cosmetic fields, identity check in non-mutation test) are nice-to-haves, not blockers.

PR #4674, opened 2026-06-01
fix(claude-hook): use $CLAUDE_PROJECT_DIR for require-review hook path

blamechris:
## Summary

The PreToolUse Bash hook at `.claude/settings.json:9` was registered with a **relative path**:

```json
"command": "bash scripts/require-review-before-merge.sh"
```

That only resolves when the Bash invocation's cwd is the chroxy project root. Any time an agent or workflow ran a command from a subdirectory (e.g. `cd packages/dashboard && npm test`), the hook fired against `packages/dashboard/scripts/require-review-before-merge.sh` → script not found → non-blocking error spam in the Claude Code output AND the merge gate silently disabled for that invocation.

## Fix

Switch to `$CLAUDE_PROJECT_DIR` — Claude Code injects this env var on every hook invocation and it's always the project root regardless of the Bash command's cwd. One-line change:

```diff
- "command": "bash scripts/require-review-before-merge.sh"
+ "command": "bash \"$CLAUDE_PROJECT_DIR/scripts/require-review-before-merge.sh\""
```

The quoting protects against path components with spaces (no current risk, but cheap belt-and-braces).

## Test plan

- [ ] In a fresh Claude Code session, `cd packages/dashboard && ls` should no longer produce the `PreToolUse:Bash hook error` line
- [ ] Attempting `gh pr merge <unreviewed-PR>` from any cwd still blocks (hook still gates)
- [ ] Attempting `gh pr merge <reviewed-PR>` from `packages/dashboard/` still succeeds

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -6,7 +6,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "bash scripts/require-review-before-merge.sh"
+            "command": "bash \"$CLAUDE_PROJECT_DIR/scripts/require-review-before-merge.sh\""
           }
         ]
       }


blamechris:
## Agent Review

**Verdict: Approve**

One-line correctness fix to `.claude/settings.json` that switches the PreToolUse Bash hook from a cwd-relative path to a `$CLAUDE_PROJECT_DIR`-anchored absolute path. The diagnosis in the PR description matches the hook contract.

### Strengths

- **Correct root cause.** Claude Code injects `CLAUDE_PROJECT_DIR` into every hook invocation (changelog: *"Hooks: Added CLAUDE_PROJECT_DIR env var for hook commands"*), and it's the documented mechanism for cwd-independent hook paths. The bare relative form `bash scripts/require-review-before-merge.sh` is a known footgun called out in the official hook-development linter (`Hardcoded paths detected — Use $CLAUDE_PROJECT_DIR or $CLAUDE_PLUGIN_ROOT`).
- **Failure mode was the dangerous one.** A missing-script error from `bash` is non-fatal at the shell level and Claude Code does not propagate the hook's stderr as a block — so the merge gate was *silently disabled* whenever the hook fired from a subdirectory. This is exactly the security regression the gate exists to prevent (cf. memory note `feedback_review_hook.md`). Fixing it is a no-brainer.
- **Belt-and-braces quoting.** `"$CLAUDE_PROJECT_DIR/scripts/..."` survives a project path containing spaces. Current chroxy path has none, but a chroxy install on a path like `/Users/foo bar/Projects/chroxy` would have broken the unquoted form. Cheap, no downside.
- **Single env-var expansion, no command substitution.** Quoting placement is safe — there's no shell-injection surface introduced.
- **Scope discipline.** One line in one file, no drive-by changes.

### Issues

| Severity | Notes |
|----------|-------|
| Critical | None |
| Suggestions | None — would-be improvements (e.g. asserting `$CLAUDE_PROJECT_DIR` is set inside the hook script and emitting a louder diagnostic if not) are out of scope for a one-line settings.json fix |
| Nitpicks | Claude Code docs use `${CLAUDE_PROJECT_DIR}` (with braces) by convention; `$CLAUDE_PROJECT_DIR` here is identical to Bash and works the same. No change needed |

### Edge cases considered

- **Env var unset.** If Claude Code ever fails to inject `CLAUDE_PROJECT_DIR`, the expansion becomes `bash "/scripts/..."` which fails with "No such file or directory" — same non-blocking failure mode as before, just at a different path. Per the documented hook contract this should never happen; not a regression.
- **Platform.** Bash behaviour is identical on macOS and Linux; no Windows hook story for this repo.
- **Other hooks in repo.** Grep confirms `.claude/settings.json` has only this one hook entry, so no parallel paths need the same treatment.
- **Existing from-review issues.** None open with the `from-review` label that overlap this change.

### Test plan validity

The PR's three manual checks (no error from subdir, gate still blocks from any cwd, reviewed PR merges from subdir) directly exercise the fix and the regression contract. Adequate for a settings.json one-liner — no unit test surface here.

### Verdict

Approve. Ship it.

blamechris:
## Check-PR Summary

| — | _(no inline review comments)_ | — | Copilot review left only a summary, no findings |

- Reviews on PR: 1 (Copilot, `COMMENTED`, summary only, zero inline comments)
- Inline comments processed: 0
- Fixes committed: 0
- False positives: 0
- Follow-up issues created: 0
- Unresolved review threads: 0

Nothing to fix, reply to, or resolve. PR ready for merge gate.

PR #4712, opened 2026-06-01
fix(store-core): honor wire timestamp in handleUserQuestion (#4613)

### LINKED ISSUE #4613: fix(store-core): honour wire timestamp in handleUserQuestion (sibling of #4607)
## Context

Identified during review of PR #4612.

## Description

`handleUserQuestion` in `packages/store-core/src/handlers/index.ts:3091-3097` has the same `timestamp: Date.now()` pattern that #4607 fixed in `handleToolStart`. When a question prompt arrives via history replay (server's `replayHistory()` forwards the full ring buffer including question events, with `timestamp` stamped at original append time), the dashboard rebuilds the `prompt` ChatMessage with a fresh `Date.now()` instead of honouring the wire timestamp.

**Impact:** lower than #4607 — affects only the bubble's displayed timestamp ("at 14:23" style), not the "Running X · Ns" timer pill. Still a correctness bug: a question prompt that originally fired at 10:00 will show as "just now" if you tab away and back.

**Fix shape:** mirror the #4607 pattern. Add a `wireTimestamp` const with the `Number.isFinite` guard and use it instead of bare `Date.now()`.

## Original Review Comment

> `handleUserQuestion` (`packages/store-core/src/handlers/index.ts:3091-3097`) also uses `Date.now()` for the chat-bubble timestamp — out of scope for #4607 (only affects bubble display, not timer), but worth checking whether reviewer wants a follow-up issue filed.

## Acceptance Criteria

- [ ] `handleUserQuestion` honours the wire `timestamp` field on the `question` payload when present (finite number)
- [ ] Falls back to `Date.now()` when the wire timestamp is missing, `NaN`, or non-finite (defensive parse identical to `handleToolStart`)
- [ ] Unit test pins wire-timestamp honouring (mirrors `honours wire timestamp on the tool_start payload (#4607)` test)
- [ ] Unit test pins fallback when wire timestamp is missing
- [ ] Audit `packages/store-core/src/handlers/index.ts` for any other handlers that build ChatMessages with `timestamp: Date.now()` and rebuild on replay (`handleUserInput` already uses stable `messageId` for dedup so likely fine, but worth a sweep)

blamechris:
## Summary

- Mirrors the #4607 fix (PR #4612) for `handleToolStart`, applied to `handleUserQuestion` in `packages/store-core/src/handlers/index.ts`.
- The server's history ring buffer stamps `timestamp: Date.now()` at append time and forwards it on replay. The prompt ChatMessage rebuilt during `history_replay` now honors that wire timestamp instead of stamping a new `Date.now()`, so a question that originally fired at 10:00 no longer shows as "just now" after tab-switch.
- Lower-impact than #4607 (affects only the bubble's displayed timestamp, not the "Running X" timer pill), but a correctness bug per the original review comment on PR #4612.

## Implementation

- Defensive `Number.isFinite` guard mirrors `handleToolStart` exactly, so NaN / Infinity / non-number wire payloads fall back cleanly to `Date.now()`.
- Live (non-replay) `user_question` broadcasts never carry `msg.timestamp` on the wire — the fallback branch covers them unchanged.

## Tests

Three new pinned tests in `handlers.test.ts > handleUserQuestion`:

- honors a finite wire `timestamp` (#4613)
- falls back to `Date.now()` when wire timestamp is missing
- ignores non-finite (`NaN`) wire timestamp and falls back

All 905 store-core tests pass.

## Test plan

- [x] `cd packages/store-core && npm test` — 905/905 pass
- [x] New "honors wire timestamp" test fails on the pre-fix code, passes post-fix (TDD)
- [x] No call-site changes required — return shape unchanged

Closes #4613

### DIFF
--- a/packages/store-core/src/handlers/handlers.test.ts
+++ b/packages/store-core/src/handlers/handlers.test.ts
@@ -4433,6 +4433,58 @@ describe('handleUserQuestion', () => {
     expect(typeof out!.chatMessage.timestamp).toBe('number')
   })
 
+  // #4613 — mirrors the #4607 fix for handleToolStart. The server's history
+  // ring buffer stamps `timestamp: Date.now()` at append time and forwards it
+  // on every replay (session-message-history.js:208-216). When the dashboard
+  // rebuilds the `prompt` ChatMessage during history_replay (the question
+  // event is part of the replayed ring buffer), it must honour that wire
+  // timestamp instead of stamping a new Date.now(). Without this, a question
+  // prompt that originally fired at 10:00 shows as "just now" if you tab
+  // away and back. Lower-impact than #4607 (affects bubble display only, not
+  // the timer pill), but still a correctness bug.
+  it('honours wire `timestamp` field on the user_question payload (#4613)', () => {
+    const wireTimestamp = 1_700_000_000_000
+    const out = handleUserQuestion(
+      {
+        questions: [{ question: 'Pick a colour' }],
+        timestamp: wireTimestamp,
+      },
+      'sess-active',
+    )
+    expect(out).not.toBeNull()
+    expect(out!.chatMessage.timestamp).toBe(wireTimestamp)
+  })
+
+  it('falls back to Date.now() when wire `timestamp` is missing (live user_question, #4613)', () => {
+    const before = Date.now()
+    const out = handleUserQuestion(
+      { questions: [{ question: 'Pick a colour' }] },
+      'sess-active',
+    )
+    const after = Date.now()
+    expect(out).not.toBeNull()
+    expect(out!.chatMessage.timestamp).toBeGreaterThanOrEqual(before)
+    expect(out!.chatMessage.timestamp).toBeLessThanOrEqual(after)
+  })
+
+  it('ignores non-finite wire `timestamp` and falls back to Date.now() (#4613)', () => {
+    // Defensive: a malformed wire payload (NaN, Infinity, string-coerced)
+    // must not poison the displayed bubble timestamp with NaN-driven
+    // arithmetic downstream.
+    const before = Date.now()
+    const out = handleUserQuestion(
+      {
+        questions: [{ question: 'Pick a colour' }],
+        timestamp: Number.NaN,
+      },
+      'sess-active',
+    )
+    const after = Date.now()
+    expect(out).not.toBeNull()
+    expect(out!.chatMessage.timestamp).toBeGreaterThanOrEqual(before)
+    expect(out!.chatMessage.timestamp).toBeLessThanOrEqual(after)
+  })
+
   it('uses msg.sessionId when present', () => {
     const out = handleUserQuestion(
       { questions: [{ question: 'Q?' }], sessionId: 'sess-9' },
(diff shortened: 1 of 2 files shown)


blamechris (commented):
## Agent review — PR #4712 (#4613 mirror of #4607)

**Verdict: APPROVE.** The fix is a faithful mirror of the merged #4612 (#4607) handleToolStart fix, the failure mode the comment describes is real, and the wire path actually delivers `timestamp` to the handler. No correctness or contract concerns.

### Mirror fidelity vs #4612

The `wireTimestamp` expression is byte-identical to the merged `handleToolStart` version:

```ts
const wireTimestamp =
  typeof msg.timestamp === 'number' && Number.isFinite(msg.timestamp)
    ? msg.timestamp
    : Date.now()
```

Same guard semantics (`typeof === 'number' && Number.isFinite`), same fallback contract (live broadcasts that lack `timestamp` get a fresh `Date.now()`). Same three pinned tests (finite wire / missing / NaN). Good consistency for future readers — anyone who already understands #4607's reasoning gets this fix for free.

### Wire path verified end-to-end

1. **Server records `timestamp` in the ring buffer:** `packages/server/src/session-message-history.js:310-316` (`case 'user_question'`) stamps `timestamp: Date.now()` at append time, matching the PR claim verbatim.
2. **Replay forwards the full entry verbatim:** `packages/server/src/ws-history.js:380` does `send(ws, { ...entry, sessionId })` — `entry` is the raw ring-buffer record, so `timestamp` rides through to the client without reshaping.
3. **Schema isn't a stripping bottleneck:** `ServerUserQuestionSchema` in `packages/protocol/src/schemas/server.ts:276-280` doesn't declare `timestamp`, but neither store-core nor the dashboard/app parses incoming messages through the protocol schemas (no `ServerUserQuestionSchema` references outside the protocol package). The raw msg with `timestamp` reaches `handleUserQuestion` intact. (If the protocol schema is ever wired into runtime decode, this field will need to be added to the schema — worth noting for #4604 follow-ups but not a blocker here.)

### Fallback contract

The PR body claim that "live (non-replay) `user_question` broadcasts never carry `msg.timestamp` on the wire" matches the emit sites I checked (`packages/server/src/sdk-session.js:270-272`, `packages/server/src/cli-session.js:776`, `packages/server/src/claude-tui-session.js`) — all forward the raw `{ toolUseId, questions }` payload from the permissions/TUI sink without timestamping. The `Date.now()` fallback is exactly right for these.

### Defensive `Number.isFinite` guard

Mirrors #4607's reasoning: if a malformed wire payload ever delivers `Infinity` / `NaN` / a string, the bubble's `timestamp` would feed NaN-driven arithmetic into downstream "X ago" renderers. Guarded fallback keeps the displayed bubble timestamp sane. The `ignores non-finite` test pins this in `handlers.test.ts`.

### Test coverage

Three tests added in the existing `describe('handleUserQuestion')` block, parallel to the #4607 set:
- honours wire `timestamp` field (finite number, the replay case)
- falls back to `Date.now()` when missing (the live case)
- ignores `NaN` (the defensive case)

I ran `npx vitest run` against store-core locally: 902/902 pass.

### Scope sanity

The fix only touches the `prompt` ChatMessage `timestamp` field (the displayed bubble time), not any timer / "Running X" pill (the #4607 fix did both because tool_start drives an ActiveTool). The PR body's "lower-impact than #4607" framing is accurate — there's no equivalent of `ActiveTool.startedAt` to propagate here, so the fix is correctly scoped to a single field.

### No nits

- Comment block correctly attributes the symptom and references both #4607 and the server-side line range.
- No new exports / no return-shape changes / no call-site updates required.

Good to merge once the user signs off.

PR #4747, opened 2026-06-01
fix(scripts): flush JSONL before exiting tui-form-recorder

blamechris:
## Summary

- `tui-form-recorder.mjs` called `recording.end()` and then `process.exit()` in the same tick, which does not wait for the underlying write stream to flush — the Ctrl+D path was the worst case because it also kills the child PTY with `SIGTERM` the same tick, racing the flush and silently truncating the JSONL we then pin as multi-question form fixtures (#4604 / #4648 / #4668 work).
- Extract a tiny `flushAndExit(stream, exitCode, { exitFn, fallbackMs })` helper into `scripts/flush-and-exit.mjs` that waits for the stream's `finish` event, with a 1s `.unref()`-ed fallback so a misbehaving stream cannot hang the recorder.
- Wire the helper into both exit paths (`term.onExit` and the Ctrl+D handler). Add a self-contained node test harness (`scripts/__tests__/flush-and-exit.test.mjs`) covering flush-to-disk, single-fire exit, fallback timeout, and the unref behaviour. Hook it into the existing `scripts-tests` CI job alongside `bump-version.test.sh`.

## Test plan

- [x] `node scripts/__tests__/flush-and-exit.test.mjs` — all 4 cases pass locally on Node 22
- [x] `node --check scripts/tui-form-recorder.mjs && node --check scripts/flush-and-exit.mjs`
- [ ] CI `scripts-tests` job runs both `bump-version.test.sh` and the new `flush-and-exit.test.mjs`
- [ ] Manual smoke (developer-only script): run `node scripts/tui-form-recorder.mjs`, press Ctrl+D, confirm the JSONL ends with the `<<CTRL-D EXIT>>` event and not a truncated mid-event line

Closes #4729

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -440,3 +440,6 @@ jobs:
 
       - name: Run bump-version.sh tests
         run: bash scripts/__tests__/bump-version.test.sh
+
+      - name: Run flush-and-exit.mjs tests
+        run: node scripts/__tests__/flush-and-exit.test.mjs
--- a/scripts/__tests__/flush-and-exit.test.mjs
+++ b/scripts/__tests__/flush-and-exit.test.mjs
@@ -133,7 +133,45 @@ await test('fallback timeout calls exit if finish event never fires', async () =
   }
 })
 
-// --- Test 4: fallback timer is unref'd so it doesn't keep node alive -----
+// --- Test 4: caller's "closed" guard prevents write-after-end -------------
+// This documents the pattern the recorder uses: a `closed` flag flipped
+// before invoking flushAndExit, so any late writes (e.g. a PTY chunk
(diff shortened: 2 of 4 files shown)


blamechris:
## Review Comments Addressed

|---|---|---|---|
| 1 | Copilot suppressed-low-confidence: late `log()` after `recording.end()` could throw `ERR_STREAM_WRITE_AFTER_END` | **FIX** | Commit `b5e1da023` — added `recordingClosed` flag + `closeRecording()` helper; covered by new test case `write-after-flushAndExit guarded by caller flag does not throw` |

**Total:** 1 processed, 1 fixed, 0 false positive, 0 deferred

### CI

All 13 checks green on the final HEAD.

Issue #4615, opened 2026-05-30
Dashboard: tailored UI for ASK_USER_QUESTION_STALL error (retry-button affordance)


blamechris:
## Context

Identified during review of PR #4614 (#4604 chunks A+C — AskUserQuestion observability + watchdog).

PR #4614 emits a structured \`code: 'ASK_USER_QUESTION_STALL'\` error from the server when claude TUI never acknowledges an AskUserQuestion answer (multi-question form wedge). The error currently flows through the generic \`session.on('error') → session_event\` proxy in \`session-manager.js:1580\` and surfaces in the dashboard as a generic error toast carrying the \`message\` field.

That's acceptable as a first pass, but the message itself prompts the user to \"retry from your last message\" — a tailored UI affordance (e.g. a retry button that re-fires the last user message) would be much friendlier than expecting the user to scroll up, find their last message, copy, and re-send.

## Description

When a \`session_event\` with \`event: 'error'\` and \`data.code === 'ASK_USER_QUESTION_STALL'\` arrives in the dashboard:

- Render a distinct error card (not the generic toast)
- Include a \"Retry\" button that re-fires the user's last input message
- Optionally suppress the multi-question form rendering on screen since the prompt is now dead

## Original Review Comment

> \`ASK_USER_QUESTION_STALL\` has no specific dashboard handler. Verified via grep — the code surfaces via the generic \`session_event\` → \`error\` proxy in \`session-manager.js:1580\` (PROXIED_EVENTS list). The \`message\` field will render in the dashboard's generic error toast, which is acceptable for now, but consider tracking a follow-up to give it a tailored \"retry your last message\" UI affordance once the dashboard supports per-code error rendering.

## Acceptance Criteria

- [ ] Dashboard recognises \`code === 'ASK_USER_QUESTION_STALL'\` in session_event errors
- [ ] Dedicated UI card rendered (not the generic toast)
- [ ] Retry button re-fires the user's last input message
- [ ] Pending QuestionPrompt UI is cleared/replaced when the stall fires

## Related

- #4604 — root-cause multi-question form fix (Chunk B)
- PR #4614 — this PR (Chunks A+C)

PR #4814, opened 2026-06-02
refactor(store-core,app): lift ChatView message pipeline into store-core (#4806)

blamechris:
## Summary

Extract the pure portions of dashboard's `useChatMessages` hook into a shared `buildChatViewMessages` pure function in `@chroxy/store-core`. Dashboard hook becomes a thin React-memo wrapper; mobile `ChatView` consumes the same function inline.

Closes #4806.

## Why

Builder-agent audit finding #5 (P1): #4776 extracted `useChatMessages` from `App.tsx` for the dashboard, but `packages/app/src/components/ChatView.tsx` re-implemented the same filter + group + overlay + tail-id pipeline inline. The mobile version was missing `stalledPromptIds` entirely (the #4615 invariant added to the dashboard), so an `ASK_USER_QUESTION_STALL` would leave the now-dead interactive QuestionPrompt visible on mobile. Submitting it would fire keystrokes into a session that already discarded the prompt context.

## What

- `packages/store-core/src/buildChatViewMessages.ts` — pure function returning `{ chatMessages, displayGroups, chatToolGroupPayloads, chatTailMessageId, storeMsgMap, stalledPromptIds }`. `displayGroups` is exposed alongside `chatMessages` so the mobile renderer (which renders groups directly with `ActivityGroup`) and the dashboard renderer (which renders flattened rows via `renderMessage`) share the same source pass.
- `packages/store-core/src/buildChatViewMessages.test.ts` — 19 cases covering empty input, system filtering, singleton tool_use passthrough, 2+ collapse, tail-id derivation (incl. synthetic group id), streaming overlay, payload-map population gating, `storeMsgMap` shape, all five `stalledPromptIds` scenarios, and `toChatViewMessage` mapping.
- `packages/dashboard/src/hooks/useChatMessages.ts` — thin `useMemo` wrapper around the shared call. Same public API; existing 16 hook tests still pass unchanged.
- `packages/app/src/components/ChatView.tsx` — consumes `buildChatViewMessages`. Stalled prompts (`prompt` + `options` + no `requestId` + in `stalledPromptIds`) now skip `MessageBubble` rendering, mirroring `App.tsx:1333` in the dashboard.

## Mobile parity side effect

Mobile `ChatView` now suppresses interactive QuestionPrompt for ASK_USER_QUESTION_STALL-invalidated prompts, closing the #4615 gap that was previously dashboard-only.

## Test plan

- [x] `cd packages/store-core && npm test` — 1002/1002 pass (19 new in `buildChatViewMessages.test.ts`)
- [x] `cd packages/dashboard && npm test` — 2516/2516 pass (16 `useChatMessages` hook tests unchanged)
- [x] `cd packages/app && npm test` — 1429/1429 pass (updated `ComponentRendering.test.ts` static-source assertion to match the new `buildChatViewMessages` import)
- [x] `tsc --noEmit` clean across store-core, dashboard, app

### DIFF
--- a/packages/app/__tests__/ComponentRendering.test.ts
+++ b/packages/app/__tests__/ComponentRendering.test.ts
@@ -147,9 +147,12 @@ describe('ChatView component', () => {
     expect(chatViewSrc).toMatch(/Starting Claude Code\.\.\./)
   })
 
-  test('imports groupMessages from the shared store-core selector', () => {
-    expect(chatViewSrc).toMatch(/import\s*\{[^}]*groupMessages[^}]*\}\s*from\s*'@chroxy\/store-core'/)
-    expect(chatViewSrc).toMatch(/import\s*\{[^}]*applyStreamingOverlay[^}]*\}\s*from\s*'@chroxy\/store-core'/)
+  test('imports buildChatViewMessages from the shared store-core pipeline (#4806)', () => {
+    // Previously imported groupMessages + applyStreamingOverlay directly;
+    // now delegates to the shared buildChatViewMessages pure function
+    // (single source of truth for filter + group + overlay + tail-id +
+    // stalledPromptIds across dashboard + mobile).
+    expect(chatViewSrc).toMatch(/import\s*\{[^}]*buildChatViewMessages[^}]*\}\s*from\s*'@chroxy\/store-core'/)
   })
 
   test('listens for reduce motion accessibility setting', () => {
--- a/packages/app/src/components/ChatView.tsx
+++ b/packages/app/src/components/ChatView.tsx
@@ -19,7 +19,7 @@ import { COLORS } from '../constants/colors';
 import { ActivityGroup } from './chat/ActivityGroup';
 import { ToolDetailModal } from './chat/ToolDetailModal';
 import { MessageBubble } from './chat/MessageBubble';
-import { groupMessages, applyStreamingOverlay } from '@chroxy/store-core';
+import { buildChatViewMessages } from '@chroxy/store-core';
 import { useConnectionStore } from '../store/connection';
 
 // -- Props --
@@ -194,25 +194,28 @@ export function ChatView({
     setToolDetail({ toolName, content, toolResult, toolResultTruncated, toolResultImages, serverName });
   };
 
-  // Structural grouping — only re-runs when messages change (not on streaming deltas)
-  const baseGroups = useMemo(() => groupMessages(messages), [messages]);
-
-  // Overlay streaming isActive flag — cheap to recompute on every delta
-  const displayGroups = useMemo(
-    () => applyStreamingOverlay(baseGroups, messages, streamingMessageId),
-    [baseGroups, streamingMessageId, messages],
+  // #4806: shared ChatView message pipeline (lifted from dashboard's
+  // `useChatMessages`). Single source of truth for filter + group +
+  // overlay + tail-id + stalledPromptIds — both dashboard and mobile
+  // derive identically. The mobile renderer below consumes
+  // `displayGroups` (groups, not flattened rows) so it can keep using
+  // ActivityGroup / MessageBubble; `chatTailMessageId` and
+  // `stalledPromptIds` (#4615) flow straight through.
+  const {
+    displayGroups,
+    chatTailMessageId,
+    stalledPromptIds,
+  } = useMemo(
+    () => buildChatViewMessages(messages, streamingMessageId),
+    [messages, streamingMessageId],
   );
 
-  // #4496: tail message id + last user input — used to gate the
-  // stream-stall chip's Retry button. Mirrors the dashboard's `isTail`
-  // check (only the most recent stall retry-button-wires; historical
-  // replayed stalls render chip text only). Walking messages once here
-  // is cheaper than recomputing per bubble.
+  // #4496: last user input — used to gate the stream-stall chip's Retry
+  // button. Mirrors the dashboard's `isTail` check (only the most recent
+  // stall retry-button-wires; historical replayed stalls render chip
+  // text only). Walking messages once here is cheaper than recomputing
+  // per bubble.
   const sendInput = useConnectionStore((s) => s.sendInput);
-  const chatTailMessageId = useMemo(
-    () => (messages.length > 0 ? messages[messages.length - 1].id : null),
-    [messages],
-  );
   const lastUserInputContent = useMemo(() => {
     for (let i = messages.length - 1; i >= 0; i--) {
       if (messages[i].type === 'user_input') return messages[i].content;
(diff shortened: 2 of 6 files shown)


blamechris (commented):
## /agent-review — Verdict: APPROVE

Clean lift of dashboard's `useChatMessages` pipeline into a pure store-core function. Closes the silent #4615 mobile gap and parallels dashboard semantics line-for-line. Six targeted observations follow; none are blockers.

---

### Purity audit — buildChatViewMessages

**PASS.** `packages/store-core/src/buildChatViewMessages.ts` imports only `groupMessages` / `applyStreamingOverlay` / `DisplayGroup` from `./group-messages` and `ChatMessage` from `./types`. No React, no DOM, no store reads, no I/O. The function:

- Reads `storeMessages` only, never mutates it (`.filter()` creates a new array; `Map` constructor copies refs).
- Returns plain values + fresh `Map` / `Set` instances.
- Preserves original `ChatMessage` references in `storeMsgMap` (verified by the "Map values are the original references" test) — important for the renderer side that compares by identity.
- Does not depend on `Date.now()`, randomness, or any global state.

Genuinely pure. Calling it twice with the same inputs yields equivalent (though not reference-equal) results. Memoisation correctly lives at the call site.

### Dashboard thin-wrapper behavior loss check

**No behavior lost.** I diffed the old hook (six chained `useMemo`s) against the new wrapper (one `useMemo` over the pure call) field-by-field:

| Field | Old derivation | New derivation | Parity |
|-------|----------------|----------------|--------|
| `chatMessages` | filter → group → overlay → flatten with singleton passthrough | identical inside `buildChatViewMessages` | ✓ |
| `chatToolGroupPayloads` | only `messages.length >= 2` runs | same gate (#3794 invariant preserved) | ✓ |
| `chatTailMessageId` | last `chatMessages` row id (possibly synthetic `activity-<id>`) | identical | ✓ |
| `storeMsgMap` | built from unfiltered `storeMessages` | identical | ✓ |
| `stalledPromptIds` | last-stall index walk + `!answered` gate | identical | ✓ |
| `toChatViewMessage` | inline export with `prompt → response` + optional `code` spread | re-exported from store-core, same body | ✓ |

One micro-optimisation lost (worth noting, not blocking): the old hook split memos so `chatToolGroupPayloads` only recomputed when `chatDisplayGroups` changed and `storeMsgMap` only when `storeMessages` changed. Now they recompute together whenever either `storeMessages` or `streamingMessageId` changes. In practice both deps change together on every stream tick, so the difference is zero. The 16 existing hook tests still pass (including the "same input reference → same output reference" memoisation test).

### Mobile parity side effect — chatTailMessageId semantic change

**Intentional and correct.** Old mobile: `chatTailMessageId = messages[messages.length - 1].id` (raw store last, including `system` events). New mobile: last id in the filtered chat-view list, which can be a synthetic `activity-<id>` group key.

The only consumer of `chatTailMessageId` in mobile is the `stream_stall` Retry button gate at line 350 (`msg.id === chatTailMessageId`). The semantic change makes mobile match dashboard exactly:

- **Old bug**: a `system` event arriving after a `stream_stall` error would silently drop the Retry button (`error.id !== systemEvent.id`).
- **New correct**: `system` filtered → stall stays the tail → Retry stays visible.
- **Synthetic group ids**: `error` bubbles are `single` groups, so `msg.id` is always the raw error id; comparison against `chatTailMessageId` works whether the *next* group is a tool group or not.

Confirmed safe. This is a free correctness win on the side; PR description should arguably call it out as "fixes-also" rather than just "side effect".

### Mobile stalled-prompt gate alignment

**Exact parity with dashboard `App.tsx:1326-1333`:**

```
mobile:    msg.type === 'prompt' && msg.options && !msg.requestId && stalledPromptIds.has(msg.id)
dashboard: storeMsg.type === 'prompt' && storeMsg.options && !storeMsg.requestId
           && stalledPromptIds.has(storeMsg.id)
```

Both gates render `null` when satisfied. Both rely on the stall-chip rendered below the suppressed prompt to carry the retry affordance. Closing the #4615 gap on mobile is correct.

### Test coverage assessment

**Covers actual pipeline logic, not just signature.** The new store-core test file has 18 `it()` blocks (PR description says 19 — minor count error). Breakdown:

- 11 mirror the 16-case dashboard hook test, ported to the pure-function form (some merged because they don't need separate `renderHook` rerenders).
- 4 new `stalledPromptIds` cases (all five branches: no error, multiple stalled, answered-skipped, last-stall boundary, wrong code).
- 3 new `toChatViewMessage` cases (prompt→response mapping, `code` propagation, `code` omission when undefined).
- 1 new `displayGroups` assertion piggybacked on the streaming-overlay test.

What's tested:
- Empty input → empty/null derivations across all 6 fields.
- System filter applied to both `chatMessages` AND `displayGroups`.
- Singleton tool_use passthrough preserves `type: 'tool_use'`.
- 2+ run collapses to synthetic `tool_group` row with `activity-<firstId>` id.
- `chatTailMessageId` reflects synthetic group id when tail is a tool group.
- Streaming overlay flips trailing group `isActive: true`, AND the `displayGroups` field reflects it.
- `chatToolGroupPayloads` excludes singletons.
- `storeMsgMap` is built from unfiltered messages (includes system events) and preserves original refs.

What's NOT tested (intentional, not blocking):
- Memoisation/reference-equality — pure functions return fresh objects every call by design; this property only matters at the hook layer and is covered there.
- Interaction with `applyStreamingOverlay` when `streamingMessageId` matches a non-tail message — covered indirectly by the group-messages tests.

Coverage is real pipeline coverage. Not a signature stub.

### Minor suggestions (non-blocking, document-only)

1. **Hook return statement allocates a fresh wrapper each render** (`packages/dashboard/src/hooks/useChatMessages.ts:90-104`). The destructure-then-rewrap pattern means the hook's return object is never reference-stable even though its fields are. Existing test only asserts on `result.current.chatMessages` reference, so this is invisible. If a future caller does `const result = useChatMessages(...)` and depends on it in another memo, they'll get re-runs. Trivially fixable by `return useMemo(() => ({ chatMessages, chatToolGroupPayloads, ... }), [result])` or by returning `result` directly. Not for this PR.

2. **PR description test count** ("19 new in `buildChatViewMessages.test.ts`") should be 18 — minor accuracy nit.

3. **JsonlSubprocessSession middle-layer pattern** (project memory `feedback_jsonl_subprocess_middle_layer.md`) doesn't apply here, but the same lesson does: the `_AssertCompatible` type-level assertion between dashboard's local `ChatViewMessage` and the new store-core `ChatViewMessage` is the right tool — it'll fail typecheck if either drifts. Good defensive pattern, worth replicating elsewhere.

---

**Findings summary:**
- Critical: 0
- Suggestions: 3 (all non-blocking — return-reference stability, doc count, lesson note)
- Test coverage: real
- Behavior parity: confirmed
- Purity: confirmed

**Recommended action:** approve and merge once Copilot's pass clears.

Issue #2205, opened 2026-03-14
fix(ws): add backpressure monitoring and slow-client eviction


blamechris:
## Priority: P2 | Effort: S-M (1-2 days)

**Source**: Production readiness audit (Builder, Guardian)

## Problem
No monitoring of `ws.bufferedAmount`. Slow clients cause unbounded message buffering. No logging when backpressure events occur.

## Fix
- Check `ws.bufferedAmount` before large sends; log warning if > 64KB
- Close client if `bufferedAmount` > 1MB (stale client)
- Log backpressure events with client ID for monitoring
- Add metric: max bufferedAmount seen across clients

PR #4817, opened 2026-06-02
refactor(server): unify ws-client-sender backpressure with ws-broadcaster (#4804)

blamechris:
## Summary

Closes #4804. Related to #4775 (the partial unification that left this gap).

`ws-broadcaster.js _sendOneWithBackpressure` (pre-send, #4775) and
`ws-client-sender.js` (post-send, #2205) are **two distinct backpressure
systems** with different thresholds, different trigger conditions, and —
until this PR — different observability:

| | pre-send (broadcaster) | post-send (client-sender) |
|---|---|---|
| Threshold | 1MB (pre-write) | 64KB warn / 1MB evict (post-write) |
| Drop counter | yes (10-deep) | n/a (instant evict at 1MB) |
| `backpressure.drops` | yes | n/a |
| `backpressure.disconnects` | yes | **no** (silent eviction) |

## Why keep the post-send path

The audit's first option ("delete it, rely on pre-send") loses coverage
for every single-recipient `WsServer._send` call site — `pong`,
`token_rotated`, `auth_fail`, `rate_limited`, `server_status`, `error`,
`server_error`. Those don't iterate through `_sendOneWithBackpressure`,
so deleting the post-send path would leave them unprotected against slow
clients. Audit option B (route the eviction through the metric) is
strictly additive and closes the observability gap without losing
coverage.

## Changes

- `packages/server/src/ws-client-sender.js`: call
  `metrics.inc('backpressure.disconnects')` from the post-send eviction
  branch; add JSDoc explaining why both backpressure paths coexist so
  the next reader doesn't repeat this audit.
- `packages/server/tests/ws-backpressure-monitoring.test.js`: new
  `observability metrics (#4804)` describe block — asserts the metric
  fires on eviction, doesn't fire on warn-only / no-client / healthy
  sends.

## Test plan

- [x] `node --test tests/ws-backpressure-monitoring.test.js` — new
  observability suite passes (4/4)
- [x] `node --test tests/ws-client-sender.test.js tests/ws-broadcaster.test.js tests/ws-server-backpressure.test.js tests/ws-server-broadcast.test.js tests/ws-server-log-broadcast.test.js` — 103/103
- [x] Full server suite: only pre-existing env-dependent failures
  (service port test, TunnelManager Integration "requires cloudflared",
  WebTaskManager feature detection) — confirmed identical on `main`

### DIFF
--- a/packages/server/src/ws-client-sender.js
+++ b/packages/server/src/ws-client-sender.js
@@ -1,4 +1,5 @@
 import { encrypt, DIRECTION_SERVER } from '@chroxy/store-core/crypto'
+import { metrics } from './metrics.js'
 
 /** Backpressure thresholds (bytes) */
 const WARN_THRESHOLD = 64 * 1024    // 64KB — log warning
@@ -12,6 +13,16 @@ const WARN_THROTTLE_MS = 30_000
  * Handles JSON serialization, optional encryption, sequence numbering,
  * post-auth queue buffering, flush-overflow buffering, and post-send
  * backpressure monitoring.
+ *
+ * Post-send backpressure is intentionally distinct from the pre-send
+ * check in `WsBroadcaster._sendOneWithBackpressure` (#4775). The
+ * broadcaster only protects multi-recipient broadcast paths; single-
+ * recipient sends from `WsServer._send` (pong, token_rotated,
+ * auth_fail, rate_limited, server_status, error, etc.) bypass the
+ * broadcaster entirely, so this post-send path is the only thing that
+ * evicts a slow client on those code paths. Per #4804 both paths must
+ * emit `backpressure.disconnects` so alerting catches either eviction
+ * source — see `metrics.inc('backpressure.disconnects')` below.
  *
  * @param {object} log - Logger with .error() and .warn() methods
  * @param {object} [opts] - Optional overrides (for testing)
@@ -56,6 +67,10 @@ export function createClientSender(log, opts = {}) {
       const buffered = ws.bufferedAmount
       if (buffered > evictThreshold && client) {
         log.warn(`Backpressure: evicting client ${client.id} — bufferedAmount ${buffered} exceeds ${evictThreshold} bytes`)
+        // #4804: unify observability with WsBroadcaster._sendOneWithBackpressure
+        // so both backpressure systems feed the same metric. Without this the
+        // single-recipient eviction path silently bypasses alerting.
+        metrics.inc('backpressure.disconnects')
         ws.close(4008, 'Backpressure: slow client evicted')
       } else if (buffered > warnThreshold && client) {
         const now = Date.now()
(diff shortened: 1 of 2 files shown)


blamechris (commented):
## /agent-review — APPROVE (with one optional polish)

### Verdict
LGTM. Option B is the correct fix; the audit's Option A (delete) would have removed the only backpressure safeguard on every single-recipient `WsServer._send` call site. The PR has done the right thing and documented why.

### Audit-recommendation re-check

Walked the source paths to verify the PR's "deleting would leave them unprotected" claim:

- `WsServer._send` (ws-server.js:1570) → `this._clientSend(ws, client, message)` → `ws-client-sender.js`. **No reachability into `WsBroadcaster._sendOneWithBackpressure`.**
- `_sendOneWithBackpressure` (ws-broadcaster.js:73) is only invoked from `_broadcast` / `_broadcastToSession` / `_broadcastClientJoined`. Those are the *only* three callers.
- `_send` call sites from ws-server.js confirmed: lines 797, 800, 1070, 1135, 1320, 1331, 1337, 1345, 1355, 1372 — every one of those (pong, token_rotated, auth_fail, rate_limited, server_status, error, server_error) bypasses the broadcaster entirely.

So yes — the audit's Option A was genuinely wrong, not just inconvenient. Deleting the post-send path would silently regress those 7+ message types to zero backpressure protection. The PR description's defense holds up.

### Eviction-path coverage check

Are there any OTHER eviction paths in ws-client-sender that the new metric call would miss? Reviewed: there is exactly one `ws.close(4008, ...)` call in ws-client-sender.js (line 59), and the new `metrics.inc('backpressure.disconnects')` sits one line above it inside the same `if (buffered > evictThreshold && client)` branch. No drift possible.

### Test coverage check

The 4 new tests in `observability metrics (#4804)` exercise:
1. Eviction → metric fires (positive)
2. Warn-only (above warn, below evict) → metric does NOT fire (negative)
3. `client === undefined` → metric does NOT fire (the eviction branch's `&& client` guard works)
4. Healthy send below warn → metric does NOT fire

`metrics.reset()` in `beforeEach` correctly isolates the singleton between runs. All 15 tests in the file pass locally on Node 22 (4/4 in the new block).

That's complete coverage of the post-send branch's three observable outcomes (evict / warn-only / silent). Both the eviction *and* the warn-only paths are exercised by the metric assertions.

### Optional polish (non-blocking)

The JSDoc explains *why* both paths coexist but does not flag the semantic asymmetry between them: the broadcaster path evicts only after `_backpressureDrops >= maxDrops` (10 consecutive drops), while ws-client-sender evicts instantly on a single threshold breach. This is fine — they serve different traffic patterns — but a reader chasing a `backpressure.disconnects` spike will see the metric increment from two semantically different conditions. A one-line addendum like `// Note: this path evicts on a single breach; the broadcaster path evicts after N consecutive drops. The same metric covers both.` would save the next person a source dive. Not blocking.

### CI / mergeability

- All 13 checks green (Server Tests, Server Lint, etc.)
- mergeable: MERGEABLE, mergeStateStatus: UNSTABLE (likely due to a non-required check; not a blocker)
- +71 / -0 LOC; cleanly scoped to one src file + one test file

PR #5508, opened 2026-06-11
feat(control-room): per-repo Open-session button on the Project status tab

blamechris:
## What

Adds a per-repo **Open session** row action to every repo row in the Control Room's **Project status** tab (`ControlRoomSection`). Previously the only session entry point from the table was the Investigate verdict tag, which only appears for actionable verdicts and carries investigate-specific semantics. Now any repo can be opened directly.

Clicking **Open session** opens the existing `CreateSessionModal` pre-filled with that repo's path (`initialCwd`). The modal's existing logic suggests a session name from the repo basename and dedupes it against existing names; the user picks provider / model / permission mode / worktree as usual and hits Create. No new prompt UI — the modal *is* the basic prompt.

## Wiring

- **`ControlRoomSection.tsx`** — new `onOpenSession?: (req: { cwd: string; name: string }) => void` prop, threaded through `RepoRowsProps` → `RepoRows` → `RowActions` exactly like the existing `onInvestigate`. The button renders in the row-actions cell (next to View PRs / Copy path) on **every** row when the handler is wired — not gated on a verdict. A per-row `aria-label` gives screen readers the repo name; live repos get a title hint that the new session runs alongside the existing one (multi-session per repo is supported).
- **`ControlRoomView.tsx`** — forwards `onOpenSession` to the repo section alongside `onInvestigate`.
- **`App.tsx`** — `handleOpenSession` routes through the shared `openCreateSession({ cwd })` opener, **minus** the investigate composer seeding (no `seed`/reason note). The single-owner opener clears any stale seed, so no investigate reason can leak into an Open-session launch.

## Behaviour preserved

- Investigate is byte-identical — its `VerdictTag` button, `RepoInvestigateRequest`, and `handleInvestigate` paths are untouched.
- No protocol or server changes; dashboard-only wiring into the existing create-session flow.

## Tests

- `ControlRoomSection.test.tsx` — new `open-session action (#5507)` block: no action without a handler, action on every row regardless of verdict, per-row accessible name, callback payload `{ cwd, name }`, and the live-repo case still actionable.
- `ControlRoomView.test.tsx` — `onOpenSession` forwarding to the repo section (mirrors the `onInvestigate` forwarding test).
- Full dashboard suite green (3381 tests) and `tsc --noEmit` clean.

Closes #5507.

Related to #5159.

### DIFF
--- a/packages/dashboard/src/App.tsx
+++ b/packages/dashboard/src/App.tsx
@@ -90,7 +90,7 @@ import { ConsolePage } from './components/ConsolePage'
 import { EnvironmentPanel } from './components/EnvironmentPanel'
 import { SnapshotsPanel } from './components/SnapshotsPanel'
 import { PoolStatsPanel } from './components/PoolStatsPanel'
-import { type RepoInvestigateRequest } from './components/ControlRoomSection'
+import { type RepoInvestigateRequest, type RepoOpenSessionRequest } from './components/ControlRoomSection'
 import { ControlRoomView } from './components/ControlRoomView'
 
 /** Server-injected config from <meta name="chroxy-config"> tag */
@@ -1462,6 +1462,15 @@ export function App() {
     openCreateSession({ cwd: req.cwd, seed: req.reason })
   }, [openCreateSession])
 
+  // #5507 — open the create-session picker pre-filled for an "Open session"
+  // row action: cwd = the repo path. Same plumbing as Investigate minus the
+  // composer seeding (no reason note) — the modal suggests + dedupes the
+  // session name from the repo's basename, and the user picks
+  // model/provider/permission/worktree before creating.
+  const handleOpenSession = useCallback((req: RepoOpenSessionRequest) => {
+    openCreateSession({ cwd: req.cwd })
+  }, [openCreateSession])
+
   // #4695 / #4942 — bridge the macOS menu bar items to App-state
   // handlers. See the `useTauriMenuEvents` call below `handleShowQr`
   // (further down in this file) for the actual wiring — we can't
@@ -2448,7 +2457,7 @@ export function App() {
             double-rendering with the disconnected/startup screens. */}
         {controlRoomActive && (
           <div className="main-content" data-testid="control-room-main">
-            <ControlRoomView onInvestigate={handleInvestigate} />
+            <ControlRoomView onInvestigate={handleInvestigate} onOpenSession={handleOpenSession} />
           </div>
         )}
 
--- a/packages/dashboard/src/components/ControlRoomSection.test.tsx
+++ b/packages/dashboard/src/components/ControlRoomSection.test.tsx
@@ -870,3 +870,79 @@ describe('ControlRoomSection investigate action (#5202)', () => {
     expect(screen.getByTestId('cr-verdict-onboarded').tagName).toBe('SPAN')
   })
 })
+
+describe('ControlRoomSection open-session action (#5507)', () => {
+  it('does not render an Open session action when no handler is wired', () => {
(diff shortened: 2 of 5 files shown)


blamechris (commented):
## Code Review: PR #5508

### Summary

Adds a per-repo **Open session** button to the Control Room "Project status" tab. The `onOpenSession` prop is threaded `ControlRoomSection` → `RepoRows` → `RowActions`, mirroring the existing `onInvestigate` plumbing. `ControlRoomView` forwards it, and `App.tsx`'s `handleOpenSession` calls the shared `openCreateSession({ cwd })` opener (no composer seed). The diff is tightly scoped to 5 dashboard source/test files — no `node_modules`, no `package-lock` entries.

Verified against the six review focuses:

**1. Investigate regression — none.** `VerdictTag`, `RepoInvestigateRequest`, `ACTIONABLE_VERDICTS`, and `handleInvestigate` (`openCreateSession({ cwd: req.cwd, seed: req.reason })`) are untouched. Diffing `origin/main..PR` for investigate/verdict/seed lines surfaces only (a) doc comments that reference investigate for contrast and (b) the `RepoRows` destructure line where `onOpenSession` was inserted adjacent to the unchanged `onInvestigate` token. Composer seeding is intact; the investigate flow is byte-identical.

**2. Wiring correctness — confirmed.** `openCreateSession({ cwd = null, seed = null } = {})` exists in `App.tsx` (line ~664) and sets `pendingCwd` → `initialCwd={pendingCwd}` on the modal. `CreateSessionModal.tsx` derives the suggested name from `initialCwd` on fresh-open via `generateDefaultName(cwdValue, existingNames)`, which takes `basename(cwdPath)` and dedupes collisions with a ` (n)` suffix. The issue's "sensible suggested name" acceptance criterion is met by existing modal logic — no new code needed there.

**3. The `name` field — used, not dead weight.** The host `handleOpenSession` only consumes `cwd` (the modal independently re-derives the name from `initialCwd`), but `name` is the documented contract/suggested-label hint and is asserted in two tests (`toHaveBeenCalledWith({ cwd, name })`). The interface doc and the PR notes both state this explicitly, so it is not misleading.

**4. A11y + render cost — clean.** The button is keyed indirectly via `RepoRows`' stable `key={repo.path}`. The single inline `onClick={() => onOpenSession({ cwd, name })}` closure per row matches the established `VerdictTag` `onInvestigate!(...)` pattern (line ~410) — no new churn. `aria-label` includes both `${repo.name}` and `${repo.path}`, so it is unique per row and distinct from the identical visible "Open session" text. `title` adapts for live repos ("Open another session…").

**5. Tests — all green, regression-sensitive.** Full dashboard suite: **3381 passed / 171 files**. `tsc --noEmit`: clean. The 5 new tests assert button presence per verdict (live/investigate/onboarded), absence when unwired, the exact `{cwd, name}` payload, and an aria-label containing the repo name — each would fail if the button or payload regressed.

**6. XSS/injection — covered by React escaping.** Repo names flow into JSX text children and `aria-label`/`title`/`data-testid` attribute values, all of which React escapes. No `dangerouslySetInnerHTML` in the file. Matches the existing `cr-action-copy-${repo.name}` / `cr-action-prs-${repo.name}` pattern.

### Strengths

- Minimal, surgical diff that reuses the proven investigate plumbing rather than inventing a parallel path.
- Thorough inline documentation tying every prop hop back to #5507 and contrasting it with the investigate gating.
- Tests cover the unwired (no-button) case and every verdict row, plus the live-repo multi-session case.
- Correctly leans on the modal's existing name derivation/dedup instead of duplicating it in the host handler.

### Issues

None blocking. Minor observation: `RowActions` now also creates the `onOpenSession` closure per row, but this is consistent with the codebase's existing per-row action pattern and the table is operator-scale (repo count), so it is not a real concern.

### Verdict

- [x] **Approve**
- [ ] Request Changes

PR #5531, opened 2026-06-11
feat: typeable short pairing code — host display + dashboard entry

blamechris:
## Typeable short pairing code for camera-less devices

The TV-app pattern (epic #5509): the host displays a short, human-typable code; a device that can't scan a QR types it. The code is read off the host's **own screen** — physical presence — so pairing needs **no extra approval**, matching today's QR trust semantics. Codes delivered via any other channel must use the #5510 approval primitive instead (noted in the `pairing.js` header).

### One mechanism, not two
All pairing ids now use a typeable alphabet, so the QR carries the same id a user can type. No parallel code path.

- **Alphabet**: uppercase, ambiguity-free (`ABCDEFGHJKMNPQRSTUVWXYZ23456789` — no `0/O/1/I/L`), 8 chars, crypto-random with rejection sampling (unbiased).
- **Entry tolerance**: case-insensitive, spaces/dashes stripped. Inputs are normalized before lookup on both ends (`normalizePairingCode` server + dashboard), so existing QR/paste flows are byte-for-byte unaffected.

### Host display surfaces
- **CLI** — `chroxy pair-code` prints `<code>  (expires in NNs)  host: <ws url>` once and exits (scriptable; `--json` supported). Does not loop.
- **Dashboard** — the QR modal fetches `GET /pairing-code` alongside the QR and shows the code under it ("Or type this code:"). Per-session "Share" QRs keep no code (they issue a session-bound token).
- **New endpoint** — `GET /pairing-code` returns `{ code, url, expiresAtMs, expiresInSeconds }`, bearer-authed + grace-extending + CORS-restricted exactly like `/qr`.

### Client entry
ServerPicker gains a **"Have a code?"** panel: host + code fields. `parsePairingCodeEntry` normalizes the code and synthesizes the same `chroxy://host?pair=<code>` request the QR/paste path builds, then reuses `pairServer()` — identical handshake. A bad/expired code is rejected server-side and surfaces as the existing `pair_fail` alert.

### Out of scope / follow-up
- **Tray (Rust)** display is deliberately left as a follow-up per the issue.

### Protocol
No protocol/schema changes — the `pair` message already carries the id.

### Tests
- Server (`node --test --test-force-exit`): pairing + http-routes + pair-code + ws-auth suites — **279 passed, 0 failed** (pairing/http/cli: 124; ws-auth: 155). Custom lints (`lint-tests-state-file-path.sh`, `lint-session-opt-forwarding.sh`) pass.
- Dashboard (`vitest run`): pairing util + QrModal + ServerPicker + connection-pairing/request-pairing — **96 passed, 0 failed**. `tsc --noEmit` clean.

Closes #5512.

Related to #5509.

### DIFF
--- a/packages/dashboard/src/App.tsx
+++ b/packages/dashboard/src/App.tsx
@@ -606,6 +606,10 @@ export function App() {
   const [qrSvg, setQrSvg] = useState<string | null>(null)
   const [qrLoading, setQrLoading] = useState(false)
   const [qrError, setQrError] = useState<string | null>(null)
+  // #5512 — the typeable short pairing code shown beside the linking-mode QR so
+  // camera-less devices can type it instead of scanning. The QR encodes the same
+  // id. Null for per-session "Share" QRs (those carry a session-bound token).
+  const [qrPairingCode, setQrPairingCode] = useState<string | null>(null)
   const [settingsOpen, setSettingsOpen] = useState(() => {
     const params = new URLSearchParams(window.location.search)
     return params.get('settings') === '1'
@@ -1534,24 +1538,36 @@ export function App() {
     setQrLoading(true)
     setQrError(null)
     setQrSvg(null)
+    setQrPairingCode(null)
     const token = getAuthToken()
     if (!token) {
       setQrLoading(false)
       setQrError('No auth token available')
       return
     }
+    // The typeable code (#5512) only applies to the linking-mode QR — per-session
+    // "Share" QRs (/qr/session/…) issue a session-bound token with no displayed
+    // code. Fetch the code in parallel with the QR for the linking-mode path.
+    const isLinkingQr = path === '/qr'
     try {
-      const res = await fetch(path, {
-        headers: { Authorization: `Bearer ${token}` },
-      })
-      if (!res.ok) {
-        const body = await res.json().catch(() => ({ error: 'Request failed' }))
-        setQrError(body.error || `HTTP ${res.status}`)
+      const [qrRes, codeRes] = await Promise.all([
+        fetch(path, { headers: { Authorization: `Bearer ${token}` } }),
+        isLinkingQr
+          ? fetch('/pairing-code', { headers: { Authorization: `Bearer ${token}` } }).catch(() => null)
+          : Promise.resolve(null),
+      ])
+      if (!qrRes.ok) {
+        const body = await qrRes.json().catch(() => ({ error: 'Request failed' }))
+        setQrError(body.error || `HTTP ${qrRes.status}`)
         setQrSvg(null)
       } else {
-        const svg = await res.text()
+        const svg = await qrRes.text()
         setQrSvg(svg)
         setQrError(null)
+      }
+      if (codeRes && codeRes.ok) {
+        const body = await codeRes.json().catch(() => null)
+        if (body?.code) setQrPairingCode(String(body.code))
       }
     } catch (err) {
       setQrError(err instanceof Error ? err.message : 'Failed to fetch QR code')
@@ -2816,6 +2832,7 @@ export function App() {
             ? 'Scan to chat into this session only — the scanner cannot list, switch, or destroy other sessions.'
             : 'Scan with Chroxy app to pair your phone'
         }
+        pairingCode={qrShareMode === 'share' ? null : qrPairingCode}
       />
 
       {/* #3209: SkillsPanel — popover for manual-skill toggles + #3205 metadata */}
--- a/packages/dashboard/src/components/QrModal.test.tsx
+++ b/packages/dashboard/src/components/QrModal.test.tsx
@@ -55,6 +55,19 @@ describe('QrModal', () => {
     expect(screen.getByText(/Scan with Chroxy app/)).toBeInTheDocument()
   })
 
+  it('shows the typeable pairing code beside the QR (#5512)', () => {
+    const svg = '<svg><rect/></svg>'
+    render(<QrModal open={true} onClose={vi.fn()} qrSvg={svg} loading={false} pairingCode="ABCD2345" />)
+    expect(screen.getByTestId('qr-pairing-code')).toBeInTheDocument()
+    expect(screen.getByTestId('qr-pairing-code-value')).toHaveTextContent('ABCD2345')
+  })
+
+  it('omits the pairing code when none is provided (#5512)', () => {
+    const svg = '<svg><rect/></svg>'
+    render(<QrModal open={true} onClose={vi.fn()} qrSvg={svg} loading={false} />)
+    expect(screen.queryByTestId('qr-pairing-code')).not.toBeInTheDocument()
+  })
+
   it('closes on Escape key (#1549)', () => {
     const onClose = vi.fn()
     render(<QrModal open={true} onClose={onClose} qrSvg={null} loading={false} />)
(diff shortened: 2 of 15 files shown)


blamechris (commented):
## Code Review: PR #5531

### Summary

Switches all pairing ids (linking-mode `_current` + `generateBoundPairing` per-session) from `randomBytes(12).base64url` (~96 bits) to a rejection-sampled 8-char code over a 31-symbol ambiguity-free alphabet (`ABCDEFGHJKMNPQRSTUVWXYZ23456789`, ~39.6 bits). Adds `normalizePairingCode` (uppercase + strip whitespace/dashes) applied on lookup, a bearer-authed `GET /pairing-code` endpoint, a `chroxy pair-code` CLI, dashboard QR-modal code display, and a ServerPicker "Have a code?" panel. No protocol changes. I reviewed this security-first, since it touches pairing auth and deliberately reduces entropy.

### The entropy reduction — brute-force analysis (the headline risk)

Keyspace: 31^8 = 852,891,037,441 ≈ 8.53×10^11 (≈ 39.6 bits). The 56-bit drop from base64url(12) is only acceptable because the online-guessing legs hold. I verified each:

**(a) Rate limit per source.** A wrong guess returns `invalid_pairing_id`, which `handlePairMessage` (ws-auth.js) classifies as a brute-force signal: it increments the per-`rateLimitKey` `authFailures` bucket with exponential backoff `min(1000·2^(count-1), 60000)` ms *and closes the WS*. Keyed off `client.rateLimitKey` (CF-Connecting-IP behind the tunnel), so one attacker cannot share/poison another's bucket. Backoff sequence per source: 1s, 2s, 4s, 8s, 16s, 32s, 60s… Cumulative block after 6 wrong guesses is 1+2+4+8+16+32 = 63s, already exceeding the 60s code TTL.

> **Per-code window odds:** ≤7 guesses per source per 60s TTL window. P(hit a specific live code) ≈ 7 / 8.53×10^11 ≈ **8.2×10⁻¹² per window per source.** Even sustaining a fresh window every minute, that's ~4.3×10⁻⁶ per source-year. Parallel source IPs don't compound the threat: the code is single-use and rotates every 60s, so every IP must guess *the same current* value while each is independently throttled. Not online-brute-forceable.

**(b) Single-use on failure.** `validatePairing` sets `entry.used = true` only on the *success* path, after all validity checks. A wrong guess hits the `!entry` branch and never touches a live code, so an attacker cannot burn through a victim's code by guessing against it. (Expiry deletes the entry, but that's TTL, not attacker-driven.) Verified.

**(c) Grace window.** `/pairing-code` calls `extendCurrentId()` (max(existing, now+3min)) exactly like `/qr`. This extends the *display* lifetime of the current code, but the code stays single-use and the per-source guess throttle is unchanged, so grace does not materially widen the online-guessing odds (still ≤7 guesses/source before a multi-window lockout). The extension is bearer-gated, so only an authed host viewer can trigger it.

**(d) Bound pairings.** `generateBoundPairing` emits the identical `generateTypeableCode()` format, the same `this._ttlMs` (60s), `used:false` single-use, and validates through the same rate-limited `handlePairMessage` path. Same analysis holds. Verified.

Verdict on entropy: acceptable **given** the 60s single-use TTL + per-source exponential backoff. All four legs hold.

### Rejection sampling — correctness

`generateTypeableCode` computes `max = floor(256/31)*31 = 248` and rejects bytes `>= 248`, mapping the rest with `b % 31`. This is textbook modulo-bias-free rejection sampling: only the uniform sub-range [0,247] (8 full cycles of 31) is mapped, so each symbol is equally likely. Source is `crypto.randomBytes`. Correct. The 6-digit `_generateVerifyCode` (unchanged here) intentionally accepts negligible 2^32-mod-1e6 bias, documented and fine for a 120s human-comparison code.

### Normalization — collision / confusable analysis

`normalizePairingCode` uppercases and strips `[\s-]+` only. Applied on **lookup only**; stored keys are already canonical (uppercase, separator-free), so it's a no-op for the issued id. No collision: the function is idempotent and the only transforms (case-fold, separator-strip) can't map two distinct *alphabet* codes to one (the alphabet is all-uppercase and contains no whitespace/dashes). The confusable set `0/O/1/I/L` is *excluded from generation* but **not mapped on input** — a user who reads `O` and types `0` produces an off-alphabet char that simply fails lookup (counts as `invalid_pairing_id`). That's the "reject" choice rather than "map", which is the safe direction (no two display codes can ever collide). See minor UX note in Issues.

### `GET /pairing-code` endpoint

Mirrors `/qr` exactly: `_validateBearerAuth` (→ `_isTokenValid`), `extendCurrentId()`, `matchAllowedOrigin` CORS, `Cache-Control: no-store`, added to the OPTIONS `isRestricted` set. Returns only `{code,url,expiresAtMs,expiresInSeconds}` — no token material. The code itself is the credential and is logged nowhere (grep confirms no `log.*pairing-code`/code logging on this path). Polling keeps the *same* code alive (grace pushes rotation back) but it remains single-use, identical to `/qr`'s existing behavior — no regression.

### Back-compat

`parsePairingUrl` (dashboard) and the mobile app consume `?pair=<id>` verbatim with **no length/format/regex assumptions** — grep across `packages/app/src` and `packages/dashboard/src` found zero pairing-id length or charset checks. The shorter id flows through unchanged. `parsePairingCodeEntry` synthesizes `chroxy://host?pair=<code>` and reuses `parsePairingUrl` + `pairServer()`, so the "Have a code?" path drives the identical handshake. QR-scan and `chroxy://?pair=` paste flows are unaffected.

### Strengths

- One mechanism, not two: the QR encodes the same typeable id, so there is no parallel code path to keep in sync.
- Rejection sampling is correct and well-commented; crypto source is `randomBytes`.
- Normalization is lookup-only and idempotent; stored form stays canonical.
- `/pairing-code` is a faithful clone of `/qr`'s auth + grace + CORS contract.
- CLI prints once and exits, no loop, no token material; `--json` shape is sane and omits the apiToken (used only as a transient Bearer header).
- Test coverage is solid: alphabet/shape, case+separator tolerance, unknown-code rejection, snapshot helper, dashboard `parsePairingCodeEntry` host-form variants. 210 server tests + 86 dashboard tests pass locally; custom lints + tsc clean.

### Issues

**Minor / non-blocking**

1. **Confusable-input UX (not a security issue).** Because `0/O/1/I/L` are excluded from the alphabet but normalization doesn't fold a typed `0→O` / `1→I`, a user who mis-reads a glyph gets a silent `invalid_pairing_id` (and burns one of their ~7 guesses before backoff). Security-correct (no collision risk), but consider folding common confusables on *input* (`0→O`, `1→I`, `L→I` etc.) for friendlier entry. No test pins the reject-vs-map choice — adding one would lock the decision. Follow-up, not a blocker.

2. **`/pairing-code` is reachable by any pairing-issued session token, not only the primary/host token** — because `_validateBearerAuth → _isTokenValid` accepts `isSessionTokenValid` tokens. This is **identical to the existing `/qr` endpoint** (same helper), so it's not a regression from this PR, and a session-bound token is already host-trusted. Noting for the record: if you later want `/pairing-code` (and `/qr`) restricted to strictly primary/host class, that's a separate hardening across both routes.

No correctness or security blockers found.

### Verdict

- [x] **Approve** — entropy reduction is sound under the verified 60s single-use TTL + per-source exponential-backoff legs (≈8.2×10⁻¹² hit probability per window per source); rejection sampling is bias-free; normalization is collision-free and lookup-only; `/pairing-code` faithfully mirrors `/qr`; back-compat holds (no client-side id format assumptions). The two issues above are minor/non-blocking UX and a pre-existing cross-endpoint note.
- [ ] Request changes
- [ ] Comment only

PR #5546, opened 2026-06-11
fix(dashboard): auto-fetch Control Room tabs on open with a staleness guard

blamechris:
## Problem

Opening the Control Room (or switching tabs within it) showed empty/stale sections until the user clicked Refresh. All three tabs (Project status, Self-hosted runners, Integrations) are pull-model request/snapshot — nothing fired the request on mount.

## Fix

`ControlRoomView` now auto-requests on tab activation with a staleness guard:

- **Triggers** when the Control Room opens with a tab already active (including a localStorage-persisted tab) and on each tab switch — a `useEffect` keyed on the active tab + connection-ready state + the snapshot/loading slices.
- **Staleness window** (`CONTROL_ROOM_STALENESS_MS = 60_000`, a named constant): snapshot `null` → fetch; `generatedAt` unparseable → fetch; older than 60s → fetch; within the window → skip.
- **In-flight guard:** skips while the tab's loading flag is set, so it never double-fires or trips the server's per-client in-flight guard.
- **Disconnected-safe:** gated on `connectionPhase === 'connected'`; the request actions already no-op when the socket is closed, so opening while disconnected queues nothing and throws nothing.
- **Manual Refresh** still forces a re-fetch regardless of staleness.
- **No interval polling** — the runner/integration surveys shell out to git/gh per repo, so fetch-on-activation + manual refresh is the right cost model.

## Tests

New `ControlRoomView` auto-fetch suite (10 tests): null snapshot fires exactly one request for the active tab; persisted tab fetches its own survey (not repos); fresh snapshot (< window) does not re-fetch; stale snapshot (> window) and unparseable `generatedAt` re-fetch; in-flight loading flag prevents fetch; disconnected fires nothing; tab switch fetches the newly-activated tab; inactive tab is never fetched on open. Tests mock `../store/connection` per the codebase convention (App.test/ControlRoomSection.test). The App.test store mock gains the three snapshot/loading slices + request actions the new effect reads.

Dashboard vitest: 3448 passed. tsc --noEmit: clean.

Closes #5543.

Related to #5159.

### DIFF
--- a/packages/dashboard/src/App.test.tsx
+++ b/packages/dashboard/src/App.test.tsx
@@ -223,6 +223,19 @@ vi.mock('./store/connection', () => {
     switchServer: vi.fn(),
     connectToServer: vi.fn(),
     updateServer: vi.fn(),
+    // #5543 — Control Room auto-fetch: ControlRoomView's mount effect reads the
+    // three snapshot/loading slices and calls the matching request action when
+    // the active tab is stale and the WS is connected. Mirror the real store so
+    // the effect doesn't crash on a missing action when CR tests run connected.
+    hostStatus: null,
+    runnerStatus: null,
+    integrationStatus: null,
+    hostStatusLoading: false,
+    runnerStatusLoading: false,
+    integrationStatusLoading: false,
+    requestHostStatus: vi.fn(),
+    requestRunnerStatus: vi.fn(),
+    requestIntegrationStatus: vi.fn(),
   }
   const useConnectionStore = (
     selector?: (s: typeof baseState) => unknown,
--- a/packages/dashboard/src/components/ControlRoomView.test.tsx
+++ b/packages/dashboard/src/components/ControlRoomView.test.tsx
@@ -12,6 +12,40 @@
 import { describe, it, expect, vi, afterEach, beforeEach } from 'vitest'
 import { render, screen, fireEvent, cleanup } from '@testing-library/react'
 
+// #5543: the auto-fetch effect reads the connection store. Mock it (matching
+// the codebase convention in App.test/ControlRoomSection.test) so these tests
+// drive a plain state object + spy actions rather than the real store — which
+// also avoids the dual-React-instance hazard of rendering real store hooks
+// under testing-library.
+const requestHostStatusMock = vi.fn(() => true)
+const requestRunnerStatusMock = vi.fn(() => true)
+const requestIntegrationStatusMock = vi.fn(() => true)
+let storeState: Record<string, unknown> = {}
+
+function resetStore(over: Record<string, unknown> = {}) {
+  requestHostStatusMock.mockClear()
+  requestRunnerStatusMock.mockClear()
+  requestIntegrationStatusMock.mockClear()
+  storeState = {
+    connectionPhase: 'connected',
+    hostStatus: null,
+    runnerStatus: null,
+    integrationStatus: null,
+    hostStatusLoading: false,
+    runnerStatusLoading: false,
+    integrationStatusLoading: false,
+    requestHostStatus: requestHostStatusMock,
+    requestRunnerStatus: requestRunnerStatusMock,
+    requestIntegrationStatus: requestIntegrationStatusMock,
+    ...over,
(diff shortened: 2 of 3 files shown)


blamechris (commented):
## Code Review: PR #5546

### Summary

`ControlRoomView` gains a `useEffect` that auto-fetches the active tab's survey on Control Room open and on each tab switch, guarded by: WS connected, tab not already loading, and snapshot null-or-stale (`isStale(generatedAt)`, `CONTROL_ROOM_STALENESS_MS = 60_000`). Manual Refresh untouched; no interval polling. Diff is exactly 3 dashboard files (source + two test files) — verified no `package-lock.json`/`node_modules` noise. 10 new tests + additive App.test store-mock slices. Full dashboard suite: 3448 pass; `tsc --noEmit` clean.

### Adversarial analysis — effect loop / limp-server case

The key risk is a request→snapshot→effect-rerun cycle, especially against an *erroring* server. Traced both paths and **both are safe**:

1. **Valid error snapshot.** The server stamps every error reply with a fresh timestamp — `buildSurveyErrorSnapshot` in `packages/server/src/handlers/control-room-handlers.js:72` sets `generatedAt: new Date().toISOString()`, and all three builders (`errorSnapshot`/`runnerErrorSnapshot`/`integrationErrorSnapshot`) route through it. The schema is a plain `z.object` (not `.strict()`), so the additive `error`/`requestId` keys are stripped and the snapshot parses cleanly (`packages/protocol/src/schemas/server.ts:929`). The store handler then sets `{ hostStatus: parsed.data, hostStatusLoading: false }` (`message-handler.ts:2077`). On the effect re-run: `loading=false`, but `isStale(fresh generatedAt) === false` → **skip**. The 60s window gives the breathing room. No hot loop against an erroring server.

2. **Malformed snapshot (fallback net).** If a payload fails Zod parse, the handler returns early *without* clearing the loading flag — documented contract at `message-handler.ts:2070-2072` ("`hostStatusLoading` is cleared even on a successful parse only — a malformed payload leaves the spinner up"). The effect's in-flight guard (`if (loading) return`) then permanently blocks re-fire. So even the malformed case can't loop.

Two independent safety nets cover the adversarial case.

### Reconnect behavior

`connected` (`connectionPhase === 'connected'`) is in the dep array, so a reconnect while the Control Room is open re-fires the effect — desirable (refreshes the active tab after a drop). The effect resolves `snapshot`/`loading`/`request` off the *active* `tab` only, so a reconnect fires at most one request (the active tab), not a 3-tab burst. Confirmed.

### Tests

10 tests cover: null snapshot → one request for the active tab; persisted (non-default) tab fetches itself; fresh snapshot (< window) skips; stale (> window) fetches; unparseable `generatedAt` → fetch; in-flight loading guard; disconnected fires nothing; tab-switch fetches the new tab; inactive tab not fetched. The "fresh snapshot is not re-fetched" test exercises the *same* `isStale=false` code path that protects against the error-snapshot loop, so the loop-safety is covered transitively even without a dedicated error-snapshot test. Mocking `../store/connection` matches the `ControlRoomSection.test.tsx` precedent (verified). App.test additions are additive-only — new mock slices, no removed or weakened assertions.

### Strengths

- Genuine two-layer loop defense (fresh-timestamp staleness + leave-loading-up-on-malformed) — the adversarial error-loop is closed on the server *and* the store side, not just the view.
- Active-tab-only resolution keeps reconnect/open to a single request.
- Named, documented constant; manual Refresh path untouched; no polling added.
- Thorough test matrix with explicit count assertions (`toHaveBeenCalledTimes(1)`) that would catch a double-fire regression.

### Issues

- **(nit, non-blocking)** No test directly asserts "a fresh-timestamped error snapshot does not re-fire". The behavior is safe and covered transitively by the fresh-snapshot test, but a one-line test seeding `{ hostStatus: { ...snapshotAt(0), error: {...} } }` and asserting no second request would document the limp-server invariant explicitly. Optional follow-up, not a merge blocker.

### Verdict

- [x] Approve
- [ ] Approve with minor changes
- [ ] Request changes

PR #5587, opened 2026-06-11
fix(desktop): theme the summon hotkey input and buttons

blamechris:
## Summary

The summon hotkey field in Settings → Desktop (and its Save/Clear buttons) had no CSS beyond flex layout, so they rendered with browser-default chrome — a bright white input whose placeholder was nearly unreadable against the dark theme.

## Changes

- `.summon-hotkey-row input` now follows the `.settings-field select` / `.settings-textarea` convention: `--bg-input` background, `--text-primary` text, `--border-primary` border, accent-blue focus ring, `--text-disabled` placeholder.
- `.summon-hotkey-row button` (Save/Clear) gets the `.server-btn` secondary palette sized to align with the input, with hover and disabled states.

CSS-only; no markup or behavior changes. Dashboard builds clean.

### DIFF
--- a/packages/dashboard/src/theme/components.css
+++ b/packages/dashboard/src/theme/components.css
@@ -6711,10 +6711,41 @@ button.sidebar-token-view-session-row:hover {
 .summon-hotkey-row input {
   flex: 1 1 auto;
   min-width: 0;
+  padding: 8px 10px;
+  background: var(--bg-input);
+  color: var(--text-primary);
+  border: 1px solid var(--border-primary);
+  border-radius: 6px;
+  font-size: 14px;
+}
+
+.summon-hotkey-row input::placeholder {
+  color: var(--text-disabled);
+}
+
+.summon-hotkey-row input:focus {
+  outline: none;
+  border-color: var(--accent-blue);
 }
 
 .summon-hotkey-row button {
   flex: 0 0 auto;
+  padding: 8px 12px;
+  background: var(--bg-secondary);
+  color: var(--text-primary);
+  border: 1px solid var(--border-primary);
+  border-radius: 6px;
+  font-size: 13px;
+  cursor: pointer;
+}
+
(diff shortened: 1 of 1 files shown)


blamechris (commented):
## Code Review

CSS-only change that themes the Settings → Desktop "Summon hotkey" input and its Save/Clear buttons so they stop rendering as browser-default white chrome on the dark theme.

### Verification
- **Variables** — all referenced vars resolve in `theme/theme.css` (`--bg-input`, `--text-primary`, `--border-primary`, `--text-disabled`, `--bg-secondary`, `--accent-blue`). `--bg-hover` is not defined in the theme, but that is a pre-existing codebase convention: `.server-btn:hover` and ~9 other rules already reference `var(--bg-hover)` the same way, so the button hover here is consistent with (and exactly mirrors) the rest of the dashboard. Non-blocking — a separate cleanup could define `--bg-hover` once and fix all of them.
- **Selectors** — `.summon-hotkey-row input` / `.summon-hotkey-row button` match the markup in `SettingsPanel.tsx` (one input, Save + Clear buttons all inside the row). No stray elements affected.
- **Conventions** — input mirrors `.settings-field select` (same `8px 10px` padding, `--bg-input`/`--text-primary`, `1px solid --border-primary`, `6px` radius, `14px`, identical `:focus` border swap). Buttons mirror `.server-btn` (same bg/text/border vars, `--bg-hover` on hover, `opacity: 0.5; cursor: not-allowed` when disabled). Both claims accurate.
- **No conflicts** — this is the only block styling these elements; no competing specificity elsewhere in `components.css`.
- **States** — `:hover:not(:disabled)` and `:disabled` are correctly scoped; placeholder uses `--text-disabled`.

Dashboard Type Check is green. Clean, minimal, matches existing patterns.

### Verdict
- [x] Approve
- [ ] Request changes
- [ ] Comment only

Issue #3743, opened 2026-05-10
Mobile chat duplicates assistant turns and scrambles order after each reconnect


blamechris:
## Symptoms

After reconnecting an existing client (mobile app or dashboard) to a session that already had content, the chat view shows:

- The first assistant response duplicated **2 or 3 times** at different positions in the list
- User/assistant turns appearing in the wrong order
- Subsequent reconnects compound the problem (each reconnect adds another set of duplicates)

Reproduced live during the v0.7.16 dogfood smoke-test. The Tauri desktop stayed continuously connected (no reconnect) and showed the correct order; Android went through three reconnects in one session and visibly accumulated duplicates with each.

## Root cause

`replayHistory` in `packages/server/src/ws-history.js:284` is the path every reconnecting client takes to rehydrate session state from the in-memory ring buffer. It emits:

1. `history_replay_start` — **without** `fullHistory: true`
2. Each ring-buffer entry as-is, in chunks of 20
3. `history_replay_end`

On the client side (e.g., `packages/app/src/store/message-handler.ts:1289`), the `messages: []` clear only fires when `fullHistory: true`. So every reconnect-replay is treated as **additive** — entries get appended on top of whatever the client already had.

`isReplayDuplicate` is supposed to catch this, but when the live broadcast and the ring-buffer entry carry different messageIds (live: stable per-process IDs like `msg-cb310f-N`; replayed `message` events: server omits `messageId` so the client falls back to `nextMessageId(msgType)`), the dedup misses and the duplicate slides through.

The explicit `request_full_history` path ("Sync Full History" button) already sends `fullHistory: true` (`packages/server/src/handlers/conversation-handlers.js:121`) and forces the correct clear-then-replay behaviour. The auto-replay on reconnect is equally authoritative (same source: ring buffer ↔ disk), so it should mark the replay the same way.

## Fix

One-line server-side change in `replayHistory` — add `fullHistory: true` to the emitted `history_replay_start`. No client-side change required; both the mobile app and the dashboard already honor the flag correctly when it's set.

## Acceptance criteria

- [ ] Reconnect to an existing chat session on mobile → no duplicated assistant turns; order matches `request_full_history` output
- [ ] Repeat the reconnect a few times → still no duplicates accumulating
- [ ] Tauri dashboard cold-launch into an existing session → same correct order
- [ ] No regression on the explicit "Sync Full History" path (it already sets `fullHistory: true`, behaviour should be unchanged)
[...]

PR #5593, opened 2026-06-12
feat(ws): lastSeq delta replay + no-blank-flash reconcile (#5555.3/.4)

blamechris:
Implements sub-items **3 and 4** of epic #5555 (cold-connect first-paint). Builds on the merged #5590/#5592/#5591/#5588 work. Two coupled changes in one PR; protocol-additive, backward-compatible both directions, no flag day.

## 1. `lastSeq` delta replay (#5555.3)

Today every connect replays the **entire** ring buffer (ws-history had no cursor). Now the client sends a per-session history cursor and the server replays only what's newer.

### Cursor design choice — per-session map, advanced on the replay path
- Wire shape: `auth.historyCursors: { [sessionId]: lastSeq }` (additive optional, Zod-validated, server caps honoured keys at `MAX_HISTORY_CURSORS = 64`). A **map** (not just the active session) is future-proof for multi-session delta replay and costs nothing today; v1 only actually replays the bound/active session on connect.
- Each history entry gets a monotonic per-session `_seq` (server-internal; **stripped from the persisted state file**, reassigned `1..N` on restore). The wire exposes it as `historySeq` on replayed entries; `latestSeq` rides the `history_replay_start`/`_end` frames.
- The client tracks the highest applied `historySeq` per session and **advances the cursor only once a replay slice is fully delivered** (finalised from `latestSeq` at `history_replay_end`). Seeding at start would let a mid-replay socket drop claim un-applied entries → silent gap. Live-streamed messages carry no seq, so an active session that streams after its replay keeps the cursor pinned at the last replayed seq; the next reconnect backfills everything recorded since (bounded by activity-during-disconnect). Correct and append-only.

### Fallback semantics (server can't honour the cursor → full replay, flagged)
`resolveReplayPlan` falls back to a **full replay** with `fullHistory: true` (so the client rebuilds rather than appends) when:
- no cursor / cursor ≤ 0 (old client, first connect),
- the oldest **retained** entry's seq is `> lastSeq + 1` — the entry just after the cursor was trimmed off the ring buffer (or seqs reset on a server restart) → a delta would skip entries.

Cursor is **connect-only**. Session switch / forced re-home / new-subscribe pass `forceFull: true` (a background session accumulates live broadcasts in the client's per-session list while viewed elsewhere, so its cursor lags and a delta would re-send/duplicate them). `request_full_history` was already its own forced-full path — unchanged.

### Backpressure machinery: kept, not deleted
The epic predicted "most of the replay backpressure machinery becomes dead code." Audited and **kept all of it** (`scheduleAfterDrain`, the bufferedAmount chunk-entry + mid-chunk gates, the `BACKPRESSURE_MAX_WAIT_MS` cap): a cursor replay's tail is bounded by *activity-during-disconnect*, which for fat tool_result payloads (200KB+) can still blow past the 1MB eviction line over a long blip. The machinery now operates on the resolved slice (`startOffset`) instead of always `0`. Genuinely-dead code: **none** — the prediction assumed cursor replays are always tiny, which isn't true for long disconnects.

## 2. Replay UX — no blank flash + preserve-then-reconcile (#5555.4)

Shared in `store-core/replay-reconcile.ts`; both clients route through it.
- **No wipe on replay start.** Delta replay is naturally append-only. **Full rebuild** records the pre-replay message count as a baseline, keeps the existing messages **visible**, appends the authoritative replayed set after the baseline, and **swaps atomically** to the appended tail in one store update at `history_replay_end`. The old prefix vanishes only at the swap → no blank flash, scroll position preserved (array identity changes once).
- **Dedup scoped to the tail during a rebuild** (`replayDedupCache`) so a replayed entry isn't suppressed by an id in the about-to-be-discarded prefix and then lost in the swap. Respects the stream_id collision invariant (handlers still check existing message type before reusing a messageId) and the #3743 authoritative-replace contract.

### Reconcile × delta-flusher race (pinned by tests)
A forced flush landing **during** a full rebuild appends into the tail in array order → survives the swap, neither dropped nor reordered. A flush mutating a prefix message (below baseline) is harmless — that message is sliced off and the authoritative copy comes from replay. A flush **after** `history_replay_end` appends normally (no second swap). Pinned in `replay-reconcile.test.ts`.

## Protocol
Additive optional `historyCursors` on `AuthSchema`; tracked `dist/` regenerated. Old client → omits field → full replay. New client → old server ignores field → full replay. New client → new server with no prior cursor → full replay. No flag day.

## RTT / payload savings
- **Quick reconnect during idle** (tab refresh, tunnel blip): cursor == latest → empty delta slice → `start`+`end` with **zero** entries. Replaces a full ring-buffer dump (up to 1000 entries / many chunked RTTs of buffered frames) with ~1 frame. Near-zero replay, as the epic targeted.
- **Active-session reconnect**: replays only entries recorded during the disconnect window (bounded by activity), not the whole buffer.
- **First connect / trim-fallback**: unchanged full replay — but now without the blank-flash wipe.

## Tests
- Server: cursor exact boundary (entry AT cursor not resent, next is), trim-gap + unknown-session + old-client full-replay-unchanged fallbacks, `forceFull` override, seq ordering across the replay→live boundary, seq stamping/restore/cleanup, `auth.historyCursors` stash + cap. Temp `stateFilePath` throughout; `--test-force-exit`.
- store-core: cursor tracking, no-blank-flash invariant (messages never empty mid-replay), atomic swap on full rebuild, tail-scoped dedup, replay×flusher race ordering, `latestSeq` parsing.
- All 4 server custom lints green; eslint (server src) clean; full suites green — server 9281 pass (only the 2 documented `service.test.js` network flakes, which reproduce on clean main), dashboard 3538, store-core 1305, app message-handler 174 + connection 153, protocol 282. Typechecks: protocol/store-core/dashboard/app all green.

UI-layer note (not unit-testable in store): scroll anchoring is preserved because the full-rebuild swap changes the `messages` array identity exactly once (at end) instead of twice (wipe + rebuild); delta replay never changes identity destructively.

Related to #5555 (sub-items 5-7 remain open).

### DIFF
--- a/packages/app/src/store/connection.ts
+++ b/packages/app/src/store/connection.ts
@@ -144,6 +144,10 @@ import {
   // to `useSpeechRecognition({ mode })`. Now gated by the same guard
   // the dashboard uses (#4853).
   isVoiceInputMode,
+  // #5555.3 — per-session history cursors for delta replay, sent in `auth`.
+  getHistoryCursors,
+  // #5555.4 — hard-reset the replay reconcile state on explicit disconnect.
+  resetReplayReconcile,
 } from '@chroxy/store-core';
 import type { InputSettings } from '@chroxy/store-core';
 import { setCallback as setImperativeCallback, getCallback, clearAllCallbacks } from './imperative-callbacks';
@@ -969,6 +973,10 @@ export const useConnectionStore = create<ConnectionState>((set, get) => ({
             // stashed keypair. Generating eagerly is cheap and harmless even
             // when the server ends up not requiring encryption.
             const eager = prepareEagerKeyExchange();
+            // #5555.3 — send per-session history cursors so the server replays
+            // only entries newer than what we've applied. Empty on first
+            // connect → omitted → full replay (old-client shape).
+            const historyCursors = getHistoryCursors();
             socket.send(JSON.stringify({
               type: 'auth',
               token,
@@ -977,6 +985,7 @@ export const useConnectionStore = create<ConnectionState>((set, get) => ({
               capabilities: CLIENT_CAPABILITIES.mobile,
               eagerPublicKey: eager.publicKey,
               eagerSalt: eager.salt,
+              ...(Object.keys(historyCursors).length > 0 ? { historyCursors } : {}),
             }));
           }
         }
@@ -1119,6 +1128,11 @@ export const useConnectionStore = create<ConnectionState>((set, get) => ({
     clearInactivityWarningsAcrossSessions(set, get);
     // Reset replay flags in case disconnect happened mid-replay
     resetReplayFlags();
+    // #5555.3/.4 — explicit disconnect is a hard reset: drop the replay
+    // baseline AND the history cursors so a later connect (possibly to a
+    // different server) starts from a full replay. Tunnel-blip RECONNECTS keep
+    // cursors (they don't run disconnect(); auth_ok clears only the baseline).
+    resetReplayReconcile({ clearCursors: true });
     // Flush and clear any pending delta buffer
     clearDeltaBuffers();
     // Clear permission boundary split tracking
--- a/packages/app/src/store/message-handler.ts
+++ b/packages/app/src/store/message-handler.ts
@@ -69,6 +69,12 @@ import {
   handleConversationsList as sharedConversationsList,
   handleHistoryReplayStart as sharedHistoryReplayStart,
   handleHistoryReplayEnd as sharedHistoryReplayEnd,
+  // #5555.3 / #5555.4 — lastSeq cursor + no-blank-flash reconcile.
+  recordHistorySeq,
+  reconcileReplayStart,
+  reconcileReplayEnd,
+  replayDedupCache,
+  resetReplayReconcile,
   handlePermissionRequest as sharedPermissionRequest,
   handlePermissionResolved as sharedPermissionResolved,
   handlePermissionExpired as sharedPermissionExpired,
@@ -1242,6 +1248,10 @@ export function handleMessage(raw: unknown, ctxOverride?: ConnectionContext): vo
       // per-session replaying set so a reconnect doesn't leave stale ids
       // gating future activity bumps).
       _ctx.replayingSessions.clear();
+      // #5555.4 — clear any in-progress replay BASELINE (a rebuild that never
+      // saw history_replay_end before the socket dropped). Cursors are RETAINED
+      // so this reconnect's replay can be incremental.
+      resetReplayReconcile();
       _ctx.isSessionSwitchReplay = false;
       _ctx.pendingSwitchSessionId = null;
       if (!ctx.isReconnect) hapticSuccess();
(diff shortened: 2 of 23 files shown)


blamechris (commented):
## Code Review

Adversarial review of #5593 (epic #5555.3/.4: lastSeq cursor delta replay + no-blank-flash reconcile). Branch was cut from current `origin/main` — not stale. Traced the highest-risk correctness cases by hand through the code.

### Findings

**1. Cursor ahead of server latest seq — REAL BUG, fixed (`16346b0`).** `_seq` is server-internal and reassigned 1..N on state restore, so a client reconnecting after a server restart can hold a cursor numerically AHEAD of the freshly-reassigned latest (e.g. cursor=500, restored latest=300). The trim-gap fallback (`oldest > lastSeq+1`) does NOT catch this (oldest=1). `resolveReplayPlan` honoured the cursor, produced an EMPTY delta slice, and the client kept stale messages and never rebuilt — with a cursor that could never recover past latest (every future reconnect re-sent 500, got nothing). `resolveReplayPlan` now forces `fullHistory: true` when `lastSeq > latestSeq`; `latestSeq` is computed in `replayHistory` before plan resolution and passed in. Regression tests added (restart-reassignment + cursor-one-above-latest boundary + e2e). Copilot independently flagged the same case.

**2. Exact boundary — sound.** `_seq == cursor` not resent (`<=`), `cursor+1` is first sent. Pinned by existing test.

**3. Replay × live race — not a regression.** `getHistory` returns the live ring-buffer array; the chunked loop reads `history[i]`/`.length` lazily. An entry appended mid-replay is both replayed and broadcast live, but they share a messageId and the existing replay-dedup catches it. Pre-dates this PR; cursor monotonicity holds (live entries carry no `historySeq`).

**4. Replay × delta-flusher race — pinned.** `replay-reconcile.test.ts` exercises the racy interleaving (`old, r-1, f-1, r-2`); the swap only ever slices the pre-baseline prefix, so a racing flush is neither dropped nor reordered.

**5. Mid-replay disconnect — sound.** Cursor finalised only at `reconcileReplayEnd` (start-frame `latestSeq` deliberately not applied; test pins it). `auth_ok` clears only the baseline (cursors retained); the OLD cursor is re-sent, server re-fetches. No half-applied state.

**6. forceFull paths — confirmed.** Session switch / destroy-rehome / new-subscribe all pass `forceFull: true`; connect is the only cursor path.

**7. No-blank-flash — sound (dashboard + app).** `updateSession` syncs the flat `messages` and per-session state in ONE `setState` on the swap; array identity changes exactly once. No transient empty.

**8. Persisted-state hygiene — sound.** `truncateEntry` deletes `_seq` on every persisted entry; `setHistory` reassigns 1..N on restore, so old (no-seq) state files restore fine.

**9. Backpressure — sound.** `startOffset` math intact; chunk gates unchanged.

### Review threads (3, all resolved)
- `ws-history.js` cursor-ahead — FIX (finding 1).
- `replay-reconcile.ts` unbounded cursor map — FIX (`649b74a`): client now LRU-caps at 64 (mirrors server `MAX_HISTORY_CURSORS`), re-inserts the touched key so the active session is never evicted.
- `ws-auth.test.js` misleading test — FIX (`649b74a`): invalid cursor values reject auth at the schema (not silently filtered); test rewritten to pin the real behaviour.

### Verification
All touched suites green locally (server ws-history/ws-auth/session-message-history: 276 pass; store-core replay-reconcile: 19 pass; dashboard message-handler: 212 pass), all 4 server lints pass, server eslint clean, store-core/dashboard/app/protocol typechecks pass. All 16 CI checks green on the final head `649b74a8e` (Windows skipped).

- [x] Approve

### Verdict

PR #5624, opened 2026-06-12
fix(dashboard): dogfooding UX fixes — session modal, sidebar pairing, control room affordance

### LINKED ISSUE #5607: dashboard: sidebar SERVERS pairing actions are cut off after Discover on LAN, no horizontal scroll
## Problem

In the dashboard sidebar (v0.9.46), after pressing **Discover on LAN**, the pairing action row in the bottom-left is wider than the sidebar: "Request to pair" is visible but "Add with token…" is clipped at the right edge (renders as "Add with toke"). There is no way to horizontally scroll to reach the cut-off control.

## Expected

When the action row overflows the sidebar width, it should be horizontally scrollable (or the buttons should wrap/stack) so every control remains reachable. Content that can be cut off must never be unreachable.

## Where

Sidebar → SERVERS section → discovered-server pairing actions (dashboard package).

_Found while dogfooding v0.9.46._

blamechris:
Three independent dashboard dogfooding bugs found in v0.9.46. All are dashboard-local and small, so they ship as **one cohesive PR**. The #5608 behavior change stayed small (it reuses the existing per-row actions cell and an already-present create-session affordance), so no split was warranted.

## #5606 — New Session modal Advanced layout

The Advanced disclosure laid out each field as a 3-child flex row (`label | control | hint`), so the permission-mode help text was squeezed into a tall, narrow far-right sliver (~1-2 words/line) and the rows had an awkward empty gap.

**Fix:** Advanced fields now match the Provider row above — `label + control` on one row, help text breaking to a **full-width** secondary line below the control (`flex-wrap` + `flex-basis: 100%` on `.form-hint`, the same pattern the Provider billing note uses). The worktree checkbox keeps its column layout so its label stays inline with the box. Reuses existing `.form-field` / `.form-hint` classes — no new ones invented.

## #5607 — Sidebar SERVERS pairing actions clipped after Discover on LAN

The discovered-server row was a horizontal `info | actions` flex with the action group `flex-shrink: 0` and `nowrap` buttons, so in the narrow sidebar "Add with token…" overflowed the right edge and was unreachable.

**Fix:** stack the row vertically (info on top, actions below at full sidebar width) and let the action row `flex-wrap` so both "Request to pair" and "Add with token…" stay reachable. Chose wrap/stack over horizontal scroll — cleaner in a narrow sidebar, as the issue prefers.

## #5608 — Control Room: inconsistent badge affordance

Only the `investigate` verdict badge opened the New Session modal (#5202); the other badges did nothing, so it was undiscoverable that one badge was special.

**Affordance choice (reporter's preferred option 1):** a **dedicated row-action button**, not a clickable project name. Rationale: the row already has an actions cell with "Open session" / "Copy path" / gear, so a new "Investigate" button there is consistent with the established affordance pattern and keeps the project-name cell free for its existing drill-down disclosure toggle. Verdict badges are now pure status spans on every row. The button preserves the investigation-reason composer seed that distinguishes Investigate from plain Open session, and is shown only for actionable verdicts (table-driven via `ACTIONABLE_VERDICTS`).

## Verification

- `tsc --noEmit` clean
- `npm run build` (vite) green
- `vitest run` — 3576 passed (188 files); updated the #5202 badge tests to assert the new dedicated-button affordance, and the App.test seed-leak helper to click the new button.

Closes #5606
Closes #5607
Closes #5608

### DIFF
--- a/packages/dashboard/src/App.test.tsx
+++ b/packages/dashboard/src/App.test.tsx
@@ -2436,12 +2436,13 @@ describe('App', () => {
       }
     }
 
-    // Open the Control Room and click the repo's actionable Investigate verdict
-    // tag — this stashes the repo's note as the pending seed and opens the
+    // Open the Control Room and click the repo's dedicated Investigate row
+    // action (#5608 — moved off the verdict badge onto an explicit button) —
+    // this stashes the repo's note as the pending seed and opens the
     // create-session modal.
     function stashSeedViaInvestigate() {
       fireEvent.click(screen.getByTestId('sidebar-panel-slot-launcher-control-room'))
-      fireEvent.click(screen.getByTestId('cr-verdict-investigate'))
+      fireEvent.click(screen.getByTestId('cr-action-investigate-alpha'))
       // Modal opened (mock renders its node only when open=true).
       expect(screen.getByTestId('create-session-modal-mock')).toBeInTheDocument()
     }
--- a/packages/dashboard/src/components/ControlRoomSection.test.tsx
+++ b/packages/dashboard/src/components/ControlRoomSection.test.tsx
@@ -766,14 +766,31 @@ describe('ControlRoomSection control actions (#5272)', () => {
   })
 })
 
-describe('ControlRoomSection investigate action (#5202)', () => {
-  it('renders the investigate verdict as a non-interactive span when no handler is wired', () => {
+describe('ControlRoomSection investigate action (#5202 / #5608)', () => {
+  // #5608 — the verdict badge is now a pure status indicator on EVERY row;
+  // the session-creating action moved to an explicit "Investigate" button in
+  // the row's actions cell (shown only for actionable verdicts).
+  it('renders every verdict badge as a non-interactive span, even with a handler', () => {
+    render(
+      <ControlRoomSection
+        snapshot={makeSnapshot()}
+        loading={false}
+        onRefresh={() => {}}
+        now={() => NOW}
+        onInvestigate={() => {}}
+      />,
+    )
+    expect(screen.getByTestId('cr-verdict-investigate').tagName).toBe('SPAN')
+    expect(screen.getByTestId('cr-verdict-live').tagName).toBe('SPAN')
+    expect(screen.getByTestId('cr-verdict-onboarded').tagName).toBe('SPAN')
+  })
+
+  it('does not render an Investigate action when no handler is wired', () => {
     render(<ControlRoomSection snapshot={makeSnapshot()} loading={false} onRefresh={() => {}} now={() => NOW} />)
-    const tag = screen.getByTestId('cr-verdict-investigate')
-    expect(tag.tagName).toBe('SPAN')
+    expect(screen.queryByTestId('cr-action-investigate-medlens')).toBeNull()
   })
 
-  it('renders the investigate verdict as a button when onInvestigate is provided', () => {
+  it('renders the Investigate action button only for actionable verdicts', () => {
     render(
       <ControlRoomSection
         snapshot={makeSnapshot()}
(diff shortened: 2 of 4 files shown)


blamechris (commented):
## Code Review

Reviewed the full diff (`ControlRoomSection.tsx` + `.test`, `App.test.tsx`, `components.css`). Built and tested locally on the PR head (`d928284`).

### #5606 — Advanced modal layout (CSS)

The fix is correctly **scoped to `.advanced-section`** — `grep` confirms `.advanced-section` only renders inside `CreateSessionModal`, so this does **not** touch the non-advanced/global `.form-field` instances. The pattern (`flex-wrap: wrap` on the row + `flex-basis: 100%; margin-left: 0` on the direct-child `.form-hint`) reproduces the Provider-row layout: `label + control` on row 1, hint full-width on row 2. The added `.advanced-section select { min-width: 0 }` is the correct companion so the select can shrink in the flex row rather than forcing overflow.

Worktree checkbox: it carries both `form-field` and `form-field--checkbox`, and the latter's own `flex-direction: column` rule still wins for layout, so the label stays inline with the box (as the PR claims). The new `> .form-hint` rule also applies to the checkbox hint (and the skip-permissions / environment hints), but `flex-basis: 100%` in a `column` container is a benign preferred-main-size that shrinks back to content height; the only visible effect is the hint de-indenting from 22px to align with the label — consistent, not a regression.

### #5607 — Sidebar pairing actions clipped (CSS)

`.server-discover-item` (vertical stack, `align-items: stretch`) and `.server-discover-actions` (`flex-wrap`, dropped `flex-shrink: 0`) are used **only** in `ServerPicker.tsx` — scope-clean. Stacking info-over-actions at full sidebar width plus letting the action group wrap makes "Add with token…" reachable at the narrow sidebar width without horizontal overflow. The extra vertical height is acceptable for a transient discovered-server row.

### #5608 — Control Room Investigate affordance (behavior)

Verified end-to-end:
- `VerdictTag` no longer accepts/uses `onInvestigate`; every badge is now a pure `<span>` status chip. The old `.cr-tag-action` CSS and all references are fully removed (no orphans).
- The new `cr-action-investigate-<name>` button lives in `RowActions`, gated on `ACTIONABLE_VERDICTS[repo.verdict] === true && !!onInvestigate` — shown only for actionable verdicts, matching the prior badge gating.
- **Seed context preserved:** the button fires `onInvestigate({ cwd, name, reason: repo.note ?? '' })` — byte-identical to the old badge payload. Traced through `App.tsx` `handleInvestigate` → `openCreateSession({ cwd, seed: req.reason })` → `pendingSeedPromptRef` → composer injection. The `reason` seed is what still distinguishes Investigate from plain Open session (`handleOpenSession` passes no seed). No other code path depended on the badge being clickable.

### Test adaptations (not weakened)

The `#5202`/`#5608` test updates are equivalent-or-stronger: badges are now asserted as `SPAN` on every verdict **even with a handler wired** (previously investigate was a `BUTTON`); the new button is asserted present only for actionable verdicts and absent for `live`/`onboarded`/no-handler; the click→payload assertion (including the `reason: ''` noteless case) and the per-row `aria-label` assertion are preserved on the new button. `App.test.tsx`'s seed-leak helper correctly clicks `cr-action-investigate-alpha`.

### Verification

- `tsc --noEmit`: clean
- `npm run build` (vite): green
- `vitest run`: 3576 passed / 188 files
- CI: all Dashboard checks green (Dashboard Tests, Dashboard Type Check); App/Store/Protocol/Server Lint green; only unrelated Server Tests pending
- Base is current `origin/main` (`f690d28`) — not stale; no overlap with the other in-flight dashboard lanes (different selectors/components)
- Copilot: reviewed all 4 files, no comments. Zero open review threads.

CSS regression risk — the primary concern — is contained: all three changed rule groups are scoped to component-specific selectors, verified by usage grep.

### Verdict

- [x] Approve — ship it

PR #5625, opened 2026-06-12
feat(dashboard): push-to-talk — hold Space to dictate in the chat input

blamechris:
## Summary

Adds push-to-talk to the dashboard chat input (Claude Code parity, #5610): hold **Space** while focused in the message input to start voice transcription, release to stop and insert the transcript at the cursor. Reuses the existing voice pipeline (`useVoiceInput` — macOS Swift helper / Web Speech) behind the mic button next to Evaluate/Send.

## Space disambiguation (the whole trick)

The hard part is letting a quick Space still type a normal space while a sustained hold triggers capture, without flooding the field or dropping characters. Approach — **suppress-then-reinsert**:

- **First Space keydown** → `preventDefault()` (suppress the native character), record the caret as the dictation anchor, arm a `PTT_HOLD_MS` (300ms) timer.
- **Released before the timer fires** → it was a tap → re-insert the single suppressed space at the caret. Quick taps still type a space.
- **Timer fires while still held** → start capture; the space stays suppressed, and subsequent **key-repeat** keydowns (the browser fires keydown repeatedly while a key is held) are also `preventDefault`'d so the field isn't flooded with spaces while recording.
- **Any other key during the arm window** → cancel PTT and re-insert the suppressed space (the user was typing, not holding to talk).
- Arming/recording state lives on a **ref** so key-repeat can't re-arm or restart the timer.

Suppress-then-reinsert (rather than insert-then-delete) avoids any race where a character could be dropped or doubled under React's controlled value.

### Threshold choice

300ms — matches Claude Code's affordance; long enough to feel deliberate, short enough not to lag a fast typist. Exported as `PTT_HOLD_MS`.

## Recording indicator

PTT calls the same `voiceInput.start()` / `stop()` as the mic button, so `isRecording` flips and the existing mic-button recording state/UI lights up automatically — no new indicator.

## Focus / cleanup handling

- **Release while recording** → `stop()`; the existing transcript-merge effect splices the transcript at the dictation anchor (cursor at hold-start), not appended at the end.
- **Blur while arming** → cancel the timer (we deliberately do *not* re-insert a space on blur — silently appending to a draft the user left is more surprising than dropping the suppressed tap).
- **Blur while recording** → stop the mic so it never stays open.
- **Unmount** → clear any pending timer and stop the mic.
- PTT only arms on plain Space (no Cmd/Ctrl/Alt) and only when `voiceInput.isAvailable`.

## Tests

New `InputBar push-to-talk (#5610)` block with the voice pipeline mocked (no real mic):
- tap-Space types a space and does **not** record
- hold-past-threshold starts recording and suppresses the space
- key-repeat while recording doesn't flood the field
- other-key-during-arm cancels and re-inserts the space
- release while recording stops capture
- transcript splices at the caret anchor (not appended)
- blur-while-recording cleans up the open mic
- voice-unavailable and Space-with-modifier don't arm

`tsc --noEmit`, `vite build`, and the full dashboard suite (**3585 tests**) are green.

## Deferred

Mobile app push-to-talk (long-press mic is the more natural affordance there, per the issue) is left for a follow-up — this PR is dashboard-only.

Closes #5610

### DIFF
--- a/packages/dashboard/src/components/InputBar.test.tsx
+++ b/packages/dashboard/src/components/InputBar.test.tsx
@@ -2558,8 +2558,9 @@ describe('InputBar push-to-talk (#5610)', () => {
       />,
     )
 
-    // The transcript is spliced at the anchor (index 3), not appended at the end.
-    expect(captured).toBe('foo hello world')
+    // The transcript is spliced *in place* at the anchor (index 3): the prefix
+    // "foo" and the suffix "bar" both survive, with the transcript in between.
+    expect(captured).toBe('foo hello world bar')
   })
 
   it('cleans up the open mic when the input loses focus mid-recording', () => {
--- a/packages/dashboard/src/components/InputBar.test.tsx
+++ b/packages/dashboard/src/components/InputBar.test.tsx
@@ -2577,6 +2577,45 @@ describe('InputBar push-to-talk (#5610)', () => {
     expect(voice.stop).toHaveBeenCalledTimes(1)
   })
 
+  it('stops the mic on unmount mid-recording using the latest stop (not a stale closure)', () => {
+    vi.useFakeTimers()
+    // The engine — and therefore the memoised stop — is selected after mount,
+    // so the unmount cleanup must call the *current* stop, not the one captured
+    // on the first render. Re-render with a fresh stop before unmounting and
+    // assert that one is invoked.
+    const firstStop = vi.fn()
+    const start = vi.fn()
+    const { rerender, unmount } = render(
+      <InputBar
+        onSend={vi.fn()}
+        onInterrupt={vi.fn()}
+        controlledValue=""
+        onValueChange={() => {}}
+        voiceInput={{ isRecording: false, isAvailable: true, transcript: '', start, stop: firstStop }}
+      />,
+    )
+    const textarea = screen.getByRole('textbox') as HTMLTextAreaElement
+    fireEvent.keyDown(textarea, { key: ' ' })
+    act(() => { vi.advanceTimersByTime(300) })
+    expect(start).toHaveBeenCalledTimes(1)
+
+    // Engine selected → a new memoised stop replaces the original.
+    const latestStop = vi.fn()
+    rerender(
+      <InputBar
+        onSend={vi.fn()}
+        onInterrupt={vi.fn()}
+        controlledValue=""
+        onValueChange={() => {}}
+        voiceInput={{ isRecording: true, isAvailable: true, transcript: '', start, stop: latestStop }}
+      />,
+    )
+
+    unmount()
+    expect(firstStop).not.toHaveBeenCalled()
+    expect(latestStop).toHaveBeenCalledTimes(1)
+  })
+
   it('does not arm PTT when voice is unavailable (Space types normally)', () => {
     vi.useFakeTimers()
     const voice = makeVoice({ isAvailable: false })
--- a/packages/dashboard/src/components/InputBar.test.tsx
+++ b/packages/dashboard/src/components/InputBar.test.tsx
@@ -2,7 +2,8 @@
  * InputBar tests (#1162)
  */
 import { describe, it, expect, vi, afterEach } from 'vitest'
-import { render, screen, fireEvent, cleanup, waitFor } from '@testing-library/react'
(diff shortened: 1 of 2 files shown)


blamechris (commented):
## Code Review

Push-to-talk on the dashboard chat input (#5610): hold Space to dictate, release to splice the transcript at the caret; tap-Space still types a space. Suppress-then-reinsert disambiguation on a ref-tracked `idle → arming → recording` state machine. Reviewed the state machine against every path; ran the full InputBar suite, dashboard tsc, and build; triaged all review threads.

### State-machine trace (tap vs hold)

- **Tap (< 300ms)** — keydown suppresses the native space (`preventDefault`) and arms; keyup before the timer re-inserts exactly one space via `insertSpaceAtCaret`, which splices through `setValue` (the controlled onChange path — no direct DOM mutation React would clobber) and restores the caret in a `requestAnimationFrame` after the value commits. Normal fast typing is untouched: each space taps cleanly, no drop/double. Correct.
- **Hold ≥ 300ms** — timer fires, `recording`, `start()` called; the held space and all key-repeat keydowns are swallowed by the `pttStateRef !== 'idle'` early-return — no space flood. Correct.
- **Other key during arm** — `cancelPttArmOnOtherKey` clears the timer, returns to idle, and re-inserts the suppressed space; recording never starts. The user's `"a[space]b"` keeps its space. Correct.
- **Release while recording** — `stop()`; the transcript-merge effect splices at the captured anchor.
- **Blur** — arming: timer cleared, no space re-inserted (deliberate; documented). recording: `stop()` via a fresh closure. Correct.
- **Modifiers / unavailable** — Cmd/Ctrl/Alt+Space and `!isAvailable`/`disabled` return early before any `preventDefault`, so Spotlight (Cmd+Space) isn't hijacked and voice-off types normally. Correct.

The 300ms threshold is a reasonable tap/hold split (a 320ms deliberate-but-slow space could mis-arm, but that's accepted per the issue) — noted, not blocking.

### Issues found and fixed on the branch

1. **Suffix truncation — data loss (fixed, 618717c5c).** The shared transcript-merge effect rebuilt the value as `prefix + separator + transcript`, dropping everything after the dictation anchor. The mic button always anchored at `value.length` (empty suffix) so it never showed; PTT anchors at the caret, so mid-draft dictation silently deleted the tail — directly contradicting the "insert at cursor" intent. Now the post-anchor suffix is captured when capture begins and spliced as `prefix + separator + transcript + trailing + suffix` (empty for the mic button — unchanged). Updated the anchor-splice test to assert the tail survives.
2. **Leaked mic on unmount-while-recording (fixed, 1df13dfba).** `useVoiceInput` re-memoises `start`/`stop` on `[engine]`, and `engine` is selected after mount, so the empty-deps unmount cleanup captured the first-render `engine==='none'` no-op `stop` and would leave the mic open. Mirrored the live `stop` into a ref read by the cleanup, preserving empty-deps unmount semantics. Added a regression test (verified failing without the ref). Blur/keyup paths were already fresh closures — unaffected.

### Lower-priority (follow-up, not blocking)

- **IME composition** — a Space keydown during CJK candidate-confirm is `preventDefault`-ed and arms PTT (no `isComposing` guard). This matches the existing repo pattern (Enter-to-send has no IME guard either) and only bites voice-available CJK users; reasonable to defer.
- **Mic button + PTT shared state** — holding Space while already mic-recording re-anchors and stops on release. Single shared recording state (no double mic), narrow edge case; acceptable per the issue's "reuse the mic pipeline" design.

### Tests / build

- InputBar suite: 159 passed (8 original PTT tests + 1 added unmount-leak regression + the corrected splice assertion). Fake timers (`vi.useFakeTimers` + `advanceTimersByTime`) — no real 300ms sleeps (#5588 respected).
- `tsc --noEmit` clean; `vite build` succeeds (pre-existing chunk-size warning only).
- CI on the final commit (1df13dfba): Dashboard Tests + Dashboard Type Check green, all other jobs green; only Server Tests pending (server untouched by this dashboard-only PR).

### Threads

All 3 Copilot threads (suffix truncation x2 + stale-stop) were the issues fixed above — replied and resolved.

### Verdict

- [x] Approve — issues found were fixed on the branch; behavior is correct and tested

Issue #3735, opened 2026-05-10
Add cli-session test for mid-turn auto-mode panic-button (kill+respawn)


blamechris:
## Context

Identified during review of PR #3730.

## Description

PR #3730 changed \`base-session.setPermissionMode\` so that \`'auto'\` overrides the \`_isBusy\` guard. For SDK sessions this is well covered: \`tests/sdk-session.test.js\` has explicit tests for the panic-button path including auto-resolving pending prompts.

For \`cli-session\` the same change cascades through differently — \`setPermissionMode('auto')\` mid-turn will now succeed and call \`_killAndRespawn()\`, killing the in-flight \`claude -p\` process. The PR notes this is intentional ("destructive but workable"), but there's no explicit regression test for the cli-session path.

## Original Review Comment

> CLI session mid-turn auto flip kills the in-flight process. The PR description acknowledges this. Worth a comment in cli-session.js#setPermissionMode documenting the destructive behavior, plus an explicit test that flipping to auto mid-turn triggers _killAndRespawn.

## Acceptance Criteria

- [ ] Add a test in \`tests/cli-session.test.js\` (or an integration test) asserting that \`setPermissionMode('auto')\` while \`_isBusy=true\` triggers \`_killAndRespawn()\`
- [ ] Add a one-line comment in \`packages/server/src/cli-session.js\` near \`setPermissionMode\` clarifying that auto mode mid-turn IS destructive and that the in-flight turn dies — this is by design (#3729 panic-button semantics)

Issue #3729, opened 2026-05-10
Auto-bypass permission mode does not suppress existing allows or new prompts


blamechris:
## Description

When the permission mode is set to auto-bypass (the "dangerously accept all permissions" mode), it does not behave as expected:

1. **Existing allow rules are not auto-approved** — entries that were previously allowed still appear (or are otherwise still acted on as separate rules) instead of being implicitly accepted by the bypass mode.
2. **New permission prompts still surface** — switching into auto-bypass should suppress all permission prompts going forward, but new prompts continue to be raised to the user.

Expected behaviour: while in auto-bypass mode, **every** permission decision (existing allows + any new request) should be automatically approved without prompting the user. The mode should be a true "accept everything" override, not a per-rule toggle.

## Repro

1. Open a session with one or more existing allow rules configured.
2. Switch the permission mode to auto-bypass / "dangerously accept all".
3. Trigger a tool call that would normally hit a permission prompt.

**Actual:** Existing allow rules still listed/applied individually; new permission prompt appears.
**Expected:** No prompts. All requests auto-approved while bypass mode is active.

## Acceptance Criteria

- [ ] When auto-bypass mode is enabled, new permission prompts are auto-approved (no UI prompt shown to the user)
- [ ] When auto-bypass mode is enabled, existing allow rules are short-circuited by the bypass — the bypass takes precedence
- [ ] Switching out of auto-bypass restores normal prompt/allow-rule behaviour
- [ ] Add a regression test covering: bypass mode + existing allow rule + new permission request → auto-approved

PR #5626, opened 2026-06-12
fix(dashboard): surface CLI auto-switch turn interrupt in confirm copy (#5609)

blamechris:
## Problem

Switching the dashboard permission dropdown **Approve → Auto** while a CLI session is mid-response silently kills the in-flight turn. `CliSession._onPermissionModeChanged()` unconditionally respawns the `claude -p` subprocess (the #3729 "panic-button", pinned by #3735) — dropping the running response. But **SDK and TUI apply the exact same switch in-place** (SDK clears rules + auto-resolves pending prompts at the next tool check; TUI rewrites a sidecar with no PTY restart). The confirm dialog said only *"Tools will run without asking for permission."* — it never mentioned the kill, and the per-provider asymmetry was invisible from the UI (#5609).

## Chosen fix: accurate, provider-aware warning (preserve the panic button)

The brief offered *defer-until-idle* as a possible ceiling, but the panic-button exists specifically to **unstick a wedged session** — and a wedged turn never goes idle, so deferring would defeat the one use case the kill is for. The right fix is therefore to **keep the kill available but stop it being silent**: word the confirm so the user knows exactly what will happen on *their* provider, *right now*.

- **New capability `interruptsTurnOnAutoSwitch`** — `true` only on CLI (subprocess respawn), `false` on SDK/TUI (in-place). Declared on all three providers so the capability matrix stays uniform; flows through `listProviders()` to every client.
- **Server** (`settings-handlers.js`): the `confirm_permission_mode` warning now names the consequence — *"This session is mid-response. Switching to Auto will INTERRUPT the running turn and restart the session…"* — **only** when the active provider interrupts (`interruptsTurnOnAutoSwitch`) **and** the session is busy (`_isBusy`). SDK/TUI and idle CLI keep the plain bypass warning. The **mobile app renders this warning verbatim** (`SettingsBar.tsx`), so both clients are fixed by one server change.
- **Dashboard** (`connection.ts`): `setPermissionMode` builds its `window.confirm` copy via a new pure, unit-tested helper (`lib/auto-mode-confirm.ts`) off the active session's provider capability + `streamingMessageId`.

## Per-provider behavior this surfaces

| Provider | Mid-turn → Auto | Confirm copy (busy) |
|---|---|---|
| CLI | Interrupts + respawns (#3729) | Names the interrupt + restart |
| SDK | In-place, turn survives | Plain bypass warning |
| TUI | Sidecar rewrite, turn survives | Plain bypass warning |

## Panic-button preserved

This is **purely additive gating around** the #3729/#3735 mechanism — `CliSession._onPermissionModeChanged()` and `BaseSession.setPermissionMode`'s `'auto'`-bypasses-busy-guard are untouched. The user can still confirm and interrupt a wedged CLI turn. The pinned panic-button regression tests (`CliSession setPermissionMode panic-button mid-turn (#3735)`) still pass unchanged.

## Tests

- `auto-mode-confirm.test.ts` (dashboard) — 4 cases: interrupt copy only when interrupting-provider + streaming; plain copy otherwise (idle CLI / SDK / undefined flag).
- `settings-handlers.test.js` (server) — `confirm warning copy (#5609)`: interrupt copy for busy CLI; plain for idle CLI; plain for busy SDK.
- `providers.test.js` (server) — asserts `interruptsTurnOnAutoSwitch` is `true` only for `claude-cli`, `false` for SDK/TUI.
- Full dashboard suite (3580) green; server providers/settings-handlers/cli-session/sdk/base/ws-permissions suites green; #3735 panic-button tests green; dashboard + app typecheck clean; server eslint (0 errors) + all 4 custom shell lints pass.

Closes #5609

### DIFF
--- a/packages/app/src/store/types.ts
+++ b/packages/app/src/store/types.ts
@@ -169,6 +169,12 @@ export interface ProviderCapabilities {
   inProcessPermissions?: boolean;
   modelSwitch?: boolean;
   permissionModeSwitch?: boolean;
+  // #5609: true when switching to 'auto' mid-turn interrupts the running turn
+  // (CLI respawns its `claude -p` subprocess — the #3729 panic-button). The
+  // app renders the server-provided `confirm_permission_mode` warning verbatim,
+  // so this flag is informational on the app side; the server derives the
+  // warning copy from it.
+  interruptsTurnOnAutoSwitch?: boolean;
   planMode?: boolean;
   resume?: boolean;
   terminal?: boolean;
--- a/packages/dashboard/src/lib/auto-mode-confirm.test.ts
+++ b/packages/dashboard/src/lib/auto-mode-confirm.test.ts
@@ -1,0 +1,30 @@
+import { describe, it, expect } from 'vitest'
+import { buildAutoModeConfirmMessage } from './auto-mode-confirm'
+
+describe('buildAutoModeConfirmMessage (#5609)', () => {
+  it('warns about interrupting the turn when the provider interrupts AND a turn is streaming', () => {
+    const msg = buildAutoModeConfirmMessage({ interruptsTurn: true, isStreaming: true })
+    expect(msg).toMatch(/INTERRUPT/)
+    expect(msg).toMatch(/restart the session/)
+    // still explains the bypass consequence
+    expect(msg).toMatch(/without asking for permission/)
+  })
+
+  it('uses the plain copy when the provider interrupts but no turn is in flight', () => {
+    const msg = buildAutoModeConfirmMessage({ interruptsTurn: true, isStreaming: false })
+    expect(msg).not.toMatch(/INTERRUPT/)
+    expect(msg).toMatch(/Tools will run without asking for permission/)
+  })
+
+  it('uses the plain copy for non-interrupting providers (SDK/TUI) even mid-turn', () => {
+    const msg = buildAutoModeConfirmMessage({ interruptsTurn: false, isStreaming: true })
+    expect(msg).not.toMatch(/INTERRUPT/)
+    expect(msg).toMatch(/Tools will run without asking for permission/)
+  })
+
+  it('treats an undefined capability flag as non-interrupting', () => {
+    const msg = buildAutoModeConfirmMessage({ interruptsTurn: undefined, isStreaming: true })
+    expect(msg).not.toMatch(/INTERRUPT/)
+    expect(msg).toMatch(/Tools will run without asking for permission/)
+  })
+})
--- a/packages/dashboard/src/lib/auto-mode-confirm.ts
+++ b/packages/dashboard/src/lib/auto-mode-confirm.ts
@@ -1,0 +1,55 @@
+/**
+ * auto-mode-confirm — build the confirm-dialog copy for switching the active
+ * session's permission mode to "Auto" (bypass-permissions).
+ *
+ * #5609: switching to Auto mid-turn is destructive ONLY on the CLI provider,
+ * where it respawns the `claude -p` subprocess and drops the in-flight turn
+ * (the #3729 "panic-button"). SDK and TUI apply the same switch in-place and
+ * leave the running turn alone. Before this helper the dashboard showed one
+ * generic confirm ("Tools will run without asking for permission.") that never
+ * mentioned the kill — a silent footgun where flipping to Auto to stop being
+ * prompted would instead destroy the agent's running response, and only on
+ * CLI. This centralizes the wording so the copy reflects the ACTUAL
+ * per-provider consequence + whether a turn is currently in flight.
+ *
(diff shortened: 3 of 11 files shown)


blamechris (commented):
## Code Review

**PR #5626 — surface CLI auto-switch turn interrupt in confirm copy (#5609)**

Approve→Auto mid-turn silently killed the in-flight CLI response. This PR keeps the #3729/#3735 panic-button intact and makes its consequence explicit + provider-accurate, instead of deferring (a wedged turn never goes idle, so defer would defeat the kill's one use case). Reviewed all 11 files adversarially.

### Panic-button preserved (load-bearing invariant) — CONFIRMED
- `CliSession._onPermissionModeChanged()` and `BaseSession.setPermissionMode`'s `'auto'`-bypasses-`_isBusy` guard are **untouched** — the diff adds only the `interruptsTurnOnAutoSwitch` capability line inside each provider's `capabilities` object. Gating is purely additive around the mechanism.
- `cli-session.test.js` is **not modified** by this PR (0 lines). The pinned `#3735` panic-button suite passes unchanged: "mid-turn flip to auto kills the in-flight process and respawns" (3/3), full file 72/72. A wedged CLI turn remains killable.

### Capability matrix — CORRECT across every provider
Traced each provider's actual mid-turn-auto behavior against the declared flag:
- **CLI** `true` — `_onPermissionModeChanged` respawns `claude -p` (drops the turn). Correct.
- **SDK** `false` — in-process (clears rules + auto-resolves pending). Correct.
- **TUI** `false` — sidecar rewrite, no PTY restart. Correct.
- **BYOK** (absent → falsy) — overrides `_onPermissionModeChanged` but only calls `autoAllowPending()` ("Mirrors sdk-session.js's behaviour"); no respawn/interrupt. Correct (no false interrupt warning).
- **Codex/Gemini (JSONL-subprocess)** (absent → falsy) — `permissionModeSwitch: false`, mode change is a no-op; the absent flag yields the plain copy, no false interrupt alarm. Correct.
- **claude-channel** `permissionModeSwitch: false`; **Docker* subclasses** inherit via `...Parent.capabilities` (DockerSession→CLI `true` and it does respawn; DockerSdk→SDK `false`; DockerByok→BYOK falsy). All correct.

### Warning-copy accuracy ({CLI,SDK,TUI,JSONL} × {busy,idle}) — CORRECT
Server emits the interrupt copy only when `interruptsTurnOnAutoSwitch && _isBusy`:
- idle CLI → plain bypass warning (nothing to interrupt). ✓ (`settings-handlers.test.js` asserts this)
- busy SDK/TUI → plain (no interrupt). ✓
- busy CLI → interrupt + restart copy. ✓
- JSONL/BYOK any state → plain. ✓

### Server vs dashboard consistency — MATCHES
Both gate on the same predicate: server `interruptsTurn && session._isBusy`; dashboard helper `input.interruptsTurn && input.isStreaming`. Same branch logic, same provider source (`capabilities.interruptsTurnOnAutoSwitch`). The app renders the server `confirm_permission_mode` warning **verbatim** (`SettingsBar.tsx` Alert body = `pendingPermissionConfirm.warning`, no hardcoded string) — both clients fixed by the one server change. `buildAutoModeConfirmMessage` is a **pure** function (no side effects); its 4 tests cover the matrix (interrupt+streaming / interrupt+idle / non-interrupt+streaming / undefined-flag).

### Stale streamingMessageId — LOW RISK, acceptable
The dashboard keys "busy" off the per-session `streamingMessageId` (via `getActiveSessionState()`), an established signal cleared to `null` at every turn-end / interrupt / done handler (and on disconnect-reset). A stale non-null would also break the existing typing indicator, so the failure mode is shared, not new. Note: server uses authoritative `_isBusy` while the dashboard uses the client-mirrored `streamingMessageId` — these can momentarily diverge, but the dashboard `window.confirm` is advisory copy only (the actual switch + authoritative warning go through the server), so any mismatch is cosmetic, not a correctness bug.

### Tests / lints — all green
- Server: providers + settings-handlers (171 pass), cli-session (72, incl. #3735 panic-button 3/3 unchanged), sdk+base (295). All 4 custom shell lints pass; eslint 0 errors.
- Dashboard: helper test (4 pass), full suite (3580 pass, 189 files), tsc clean.
- App: tsc clean.
- CI on PR head a48ee759e: every check green (Server Lint, Server/Dashboard/App tests + typechecks, Store Core, Protocol, Hooks, Desktop Rust); Windows skipped by path filter.

### Threads / base
- No review threads. Copilot reviewed all 11 files and generated **no comments**.
- Not stale: merge-base == origin/main (`f690d286d`); #5624/#5625 not yet in main, no overlap conflict; `mergeable: MERGEABLE`.

### Verdict
- [x] Approve — panic-button preserved, capability matrix correct for every provider, server/dashboard copy consistent, all tests + CI green, no open threads.

PR #5757, opened 2026-06-13
fix(dashboard): price + correctly label current-gen Claude models (#5631 quick-win)

blamechris:
The quick-win slice of #5631. The dashboard's local pricing table (`model-pricing.ts`) was stale (3.x/3.5/3.7 era, no 4.x rows) and `claude-sonnet-4-5` was mislabeled **"Claude 3.7 Sonnet"**. This corrects the data.

## Change
- Add **`claude-opus-4-7`** (0.015/0.075), **`claude-sonnet-4-6`** (0.003/0.015), **`claude-haiku-4-5`** (0.001/0.005). Rates **mirror the server's authoritative `CLAUDE_PRICING_USD_PER_MTOK` table** (`packages/server/src/models.js`), converted USD/Mtok → USD/1k — verified exact by review, not invented.
- Correct `claude-sonnet-4-5` label → **"Claude Sonnet 4.5"**.
- Tests pin both rates + the label.

## Honest scope (corrected after review)
Adversarial review found this change is **currently inert in the UI**, and the original description overstated it:
- `MODEL_PRICING` is only consumed by `calculateCost` behind `CLIENT_ESTIMATED_COST_PROVIDERS = {codex, gemini}` — **Claude is not in that set**, so these Claude rows aren't on any live cost path (Claude cost stays server-authoritative).
- The UI model **label comes from the server `availableModels` list**, not `MODEL_PRICING.label`, so the label fix has no rendered effect today.

So this is **correctness / table-hygiene / future-proofing** (removes objectively-wrong data; makes the table right for any future consumer or if Claude is ever added to the estimate set), **not** a visible papercut fix. The genuinely *felt* part of #5631 (model labels in the UI) lives on the **server** `FALLBACK_MODELS` / `humanizeModelId` path and the structural wire-through — both remain tracked on #5631 (annotated).

## Deliberately omitted
`opus-4-8` + `fable-5` — the server doesn't price them yet either (#5630 owns opus-4-8); the `calculateCost` drift-warn surfaces them. Full dashboard suite green: 192 files / 3681 tests + tsc clean.

### DIFF
--- a/packages/dashboard/src/lib/model-pricing.test.ts
+++ b/packages/dashboard/src/lib/model-pricing.test.ts
@@ -200,7 +200,9 @@ describe('MODEL_PRICING table integrity', () => {
     // Rates mirror the server's authoritative table (USD/Mtok → USD/1k).
     expect(MODEL_PRICING['claude-opus-4-7']!.inputPer1k).toBe(0.015)
     expect(MODEL_PRICING['claude-opus-4-7']!.outputPer1k).toBe(0.075)
+    expect(MODEL_PRICING['claude-sonnet-4-6']!.inputPer1k).toBe(0.003)
     expect(MODEL_PRICING['claude-sonnet-4-6']!.outputPer1k).toBe(0.015)
+    expect(MODEL_PRICING['claude-haiku-4-5']!.inputPer1k).toBe(0.001)
     expect(MODEL_PRICING['claude-haiku-4-5']!.outputPer1k).toBe(0.005)
   })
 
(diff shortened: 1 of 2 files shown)


blamechris:
Addressed all three review comments — they were all the same valid point my own adversarial review also flagged: the comments overstated the runtime effect. Fixed in 4e33421d4:
- Corrected the `model-pricing.ts` + test comments to state plainly that this table feeds `calculateCost` only behind `CLIENT_ESTIMATED_COST_PROVIDERS = {codex, gemini}`, so the Claude rows are **not on a live cost path** (Claude cost is server-authoritative); they exist for table-correctness/future-proofing.
- Dropped the inaccurate "#5630 owns opus-4-8" reference (#5630 is closed and about BYOK cost-labeling); opus-4-8/fable-5 pricing is tracked under #5631.

PR #5760, opened 2026-06-13
feat(app): assertively announce a new permission prompt to screen readers (#5750 item 2)

### LINKED ISSUE #5750: mobile: per-tab pending-permission indicator + assertive prompt announcement (residual parity after #5674)
Follow-up to the now-closed #5674 (which shipped the prompt *label* + session-scoped routing). The swarm audit (2026-06-13, docs/reports/audit-results/) found two residual gaps on mobile vs the dashboard:

1. **No per-tab 'blocked on you' signal.** The session-tab pill (`packages/app/src/components/SessionPicker.tsx:102-108`) renders crash/busy/notification/shells dots but **no pending-permission indicator** — a background session blocked on a permission looks identical to one just thinking. The dashboard has `derivePendingPermissionSessions`/`selectNextPendingSession`; mobile doesn't. Port the selector + add a tab dot.

2. **No assertive a11y announcement of a new prompt (mobile mirror of #5733).** `PermissionDetail.tsx` has control labels + a polite countdown (:399) but no assertive live-region announcing that a new permission-prompt bubble appeared, so a screen-reader user won't know a (esp. background-session) prompt is waiting until they navigate to it.

Class: parity / a11y. Effort: M.

blamechris:
Closes #5750. The second and final item of the mobile permission-attribution parity — item 1 (the per-tab indicator) shipped in #5758.

## The gap
The mobile permission prompt **auto-denies on timeout** but announced nothing to VoiceOver/TalkBack when it appeared, so a screen-reader user could lose an action they'd have allowed without ever knowing it was asked. The dashboard got an assertive `alertdialog`/`aria-live` treatment in #5733; this is the mobile mirror.

## Change
- **`usePermissionAnnouncer(messages)`** — watches the active session's messages for the first *live* permission prompt (`type:'prompt'` + `requestId` + future `expiresAt` + unanswered — the requestId+expiresAt pair excludes AskUserQuestion) and pushes `Permission requested: <summary>` through `AccessibilityInfo.announceForAccessibility` (the same one-shot primitive `useConnectionAnnouncer`/`CheckInChip`/`ThinkingIndicator` use).
- **Seed-at-mount semantics** (mirroring `useConnectionAnnouncer`): a prompt already present when you arrive at a session stays silent; only genuinely *new* arrivals announce, once per `requestId`. After a prompt resolves, the next one announces.
- Mounted in **`ChatView`** (the active session's chat).
- Tests: the `firstLivePermissionPrompt` predicate + the announce / seed / resolve lifecycle (7 cases).

Full app suite green: 143 suites / 1908 tests + tsc clean.

### DIFF
--- a/packages/app/src/__tests__/hooks/usePermissionAnnouncer.test.tsx
+++ b/packages/app/src/__tests__/hooks/usePermissionAnnouncer.test.tsx
@@ -38,22 +38,22 @@ function permPrompt(over: Partial<ChatMessage> = {}): ChatMessage {
   } as ChatMessage;
 }
 
-function Harness({ messages }: { messages: ChatMessage[] }) {
-  usePermissionAnnouncer(messages);
+function Harness({ messages, sessionKey }: { messages: ChatMessage[]; sessionKey: string }) {
+  usePermissionAnnouncer(messages, sessionKey);
   return null;
 }
 
-function render(messages: ChatMessage[]) {
+function render(messages: ChatMessage[], sessionKey = 'A') {
   let tree!: renderer.ReactTestRenderer;
   act(() => {
-    tree = renderer.create(<Harness messages={messages} />);
+    tree = renderer.create(<Harness messages={messages} sessionKey={sessionKey} />);
   });
   return tree;
 }
 
-function update(tree: renderer.ReactTestRenderer, messages: ChatMessage[]) {
+function update(tree: renderer.ReactTestRenderer, messages: ChatMessage[], sessionKey = 'A') {
   act(() => {
-    tree.update(<Harness messages={messages} />);
+    tree.update(<Harness messages={messages} sessionKey={sessionKey} />);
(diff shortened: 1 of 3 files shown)


blamechris:
Addressed both:
1. **Multiple concurrent live prompts** — valid; keying on the first-by-position live prompt swallowed a second concurrent one. Switched to a **Set of announced requestIds** (pruned to still-live ids each run, so it stays bounded), announcing the first un-announced live prompt. Added a concurrent-prompt test. The 'once per requestId' contract now holds literally.
2. **Session-awareness** — already fixed in 671196837 (the commit before your review): the hook now takes `sessionKey` (the active session id, passed from ChatView) and re-seeds whenever it changes, so switching to a session with a pre-existing prompt stays silent. There's a cross-session regression test covering exactly that. The thread reviewed the pre-fix commit.

PR #5812, opened 2026-06-14
fix(app): compare allowSingleMultiSelect in MessageBubble memo comparator

blamechris:
## Summary
Fixes a correctness regression surfaced by the marathon convergence accuracy-audit (adversarially verified). #5791 threaded the new `allowSingleMultiSelect` gating prop into the app's `MessageBubble`, but its custom `React.memo` comparator (added in #5532) was never updated to compare it.

Unlike `allowMultiQuestion` (which depends only on the provider name), `allowSingleMultiSelect` also depends on the server-advertised `multiSelectReinject` capability — which can flip **under a live prompt** when `availableProviders` updates. With the prop uncompared, `React.memo` skipped the re-render and the bubble kept **stale gating** (the single-multiSelect checkbox form never appeared, or persisted after it should hide). The dashboard has no such trap — `useMessageRenderer` recomputes with `activeSessionCaps` in its deps — so this was an app-only divergence introduced by the convergence work.

## Fix
One line added to the comparator (`packages/app/src/components/chat/MessageBubble.tsx`), matching the existing scalar-boolean pattern, + a regression test that re-renders the **same message ref** with `allowSingleMultiSelect` flipped true and asserts the form appears.

## Verification
- app `tsc --noEmit` clean
- `MessageBubble.multiQuestion` jest: **10/10** incl. the new live-flip regression test (fails without the comparator change)

Found-by: marathon convergence accuracy swarm-audit (2026-06-14).

### DIFF
--- a/packages/app/src/components/__tests__/MessageBubble.multiQuestion.test.tsx
+++ b/packages/app/src/components/__tests__/MessageBubble.multiQuestion.test.tsx
@@ -281,4 +281,46 @@ describe('MessageBubble single-question multiSelect (#5776)', () => {
     expect(present(tree, 'approval-button-Cheese')).toBe(true);
     expect(present(tree, 'approval-button-Onion')).toBe(true);
   });
+
+  // #5791 review: the React.memo comparator must compare allowSingleMultiSelect,
+  // or the bubble keeps stale gating when the prop flips under a LIVE prompt
+  // (same message ref) — e.g. claude-tui's multiSelectReinject capability
+  // arriving in availableProviders after the prompt mounted. Render with it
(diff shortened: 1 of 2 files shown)


blamechris (commented):
## Agent review — clean (0 Copilot threads)
One-line fix for the convergence accuracy-audit's only verified correctness regression: the app MessageBubble React.memo comparator now compares allowSingleMultiSelect (was omitted by #5791), so the single-multiSelect form re-gates when the provider's multiSelectReinject capability flips under a live prompt — matching dashboard behavior. + a regression test (live-flip re-render). app tsc clean; MessageBubble jest 10/10. Self-merging per Unattended Merge Authority.

PR #5824, opened 2026-06-14
fix(protocol): single-source DEFAULT_PROVIDER; fix badge suppression (#5823)

### LINKED ISSUE #5823: Provider badge/label suppression still keys on 'claude-sdk' after default flip to claude-tui (#5819)
## Context

Follow-up from review of PR #5822 (flip default provider `claude-sdk` → `claude-tui`, #5819).

PR #5822 flips the zero-config default provider to `claude-tui`, but the provider-badge / provider-label **suppression** logic in the UIs still treats `claude-sdk` as the "common default, hide the badge" case:

- `packages/dashboard/src/components/Sidebar.tsx:869` — `session.provider && session.provider !== 'claude-sdk'`
- `packages/app/src/components/SessionPicker.tsx:95` and `:240` — same `!== 'claude-sdk'` guard (alert title + pill badge)

## Problem

After the default flip, the semantics are inverted:
- New **default** `claude-tui` sessions now render a provider badge/label (they're no longer the suppressed case).
- `claude-sdk` sessions — now an explicit non-default opt-in — render **no** badge, despite being the variant.

This is cosmetic (no functional break), which is why it's deferred rather than blocking the merge.

## Note

There are tests pinning the `!== 'claude-sdk'` string literal, so this isn't a one-line change:
- `packages/app/src/__tests__/components/SessionPickerProviderLabel.test.ts`
- `packages/app/src/__tests__/components/SessionPickerProviderBadge.test.ts`

## Suggested direction

Decide the intended UX (e.g. suppress the badge for the *current* default, or always show the provider badge now that there are several first-class providers), update the guard in dashboard + app, and update the pinned tests accordingly. Consider sourcing "the default" from a single shared constant rather than a hardcoded literal so the next default flip doesn't reintroduce the drift.

blamechris:
## Summary

Follow-up to PR #5822 (default flip `claude-sdk` → `claude-tui`). That flip inverted the provider-badge / label **suppression** semantics in the UIs, which still keyed on a hardcoded `'claude-sdk'`:
- New **default** `claude-tui` sessions rendered a provider badge (no longer the suppressed case).
- `claude-sdk` sessions — now an explicit non-default opt-in — rendered **no** badge.

This fixes it at the root by giving the default a single source of truth.

## Changes

- **`@chroxy/protocol`** gains `export const DEFAULT_PROVIDER = 'claude-tui'` — the one place server, dashboard, and app agree on the default. (dist rebuilt.)
- **Server** `providers.js` now imports + re-exports it; every server call site (`server-cli.js`, `doctor.js`, `session-manager.js`) is unchanged because they import from the registry.
- **Dashboard** `Sidebar.tsx` badge and **app** `SessionPicker.tsx` (pill chip + long-press alert) suppress for `DEFAULT_PROVIDER` instead of the literal.
- **Dashboard store** seeds `defaultProvider` from the constant.
- **Tests** — the source-text-pinning tests (app `SessionPickerProviderBadge`/`Label`, dashboard `ProviderPicker`) now assert the constant-based gate, so the next default flip can't reintroduce the drift. The misnamed dashboard "default provider is claude-sdk" test now asserts the modal sources its default from the store.

## Net effect

Badge now marks **non-default** provider sessions (codex, gemini, claude-sdk, docker-*, …) and stays clean for the current default — and the next default change is a one-line edit in `@chroxy/protocol`.

## Testing

- protocol: 289 pass · store-core: 1571 pass
- dashboard: typecheck clean, 961 pass (Sidebar + ProviderPicker + store)
- app: `tsc --noEmit` clean, SessionPicker provider tests 10 pass
- server: providers/banner/doctor/init/session-manager 356 pass; opt-forwarding + state-file-path lints clean

Closes #5823

### DIFF
--- a/packages/app/src/__tests__/components/SessionPickerProviderBadge.test.ts
+++ b/packages/app/src/__tests__/components/SessionPickerProviderBadge.test.ts
@@ -24,13 +24,15 @@ describe('SessionPicker pill chip — provider hint badge (#3940)', () => {
     expect(pillSection).toMatch(/getProviderInfo\(session\.provider\)/);
   });
 
-  it("only renders the provider hint when the session's provider is not the claude-sdk default", () => {
+  it("only renders the provider hint when the session's provider is not the default", () => {
     // Same gate as the long-press alert title from #3937 — gate on
-    // "session.provider && session.provider !== 'claude-sdk'" so the
-    // default SDK pill stays clean and only variant providers (claude-tui,
-    // codex, gemini, docker-cli, ...) get a badge.
+    // "session.provider && session.provider !== DEFAULT_PROVIDER" so the
+    // default pill stays clean and only non-default providers (claude-sdk,
+    // codex, gemini, docker-cli, ...) get a badge. Keyed on the shared
+    // DEFAULT_PROVIDER constant (#5823) so a default flip can't reintroduce
+    // the stale-literal drift.
     expect(pillSection).toMatch(
-      /session\.provider\s*&&\s*session\.provider\s*!==\s*['"]claude-sdk['"]/,
+      /session\.provider\s*&&\s*session\.provider\s*!==\s*DEFAULT_PROVIDER/,
     );
   });
 
--- a/packages/app/src/__tests__/components/SessionPickerProviderLabel.test.ts
+++ b/packages/app/src/__tests__/components/SessionPickerProviderLabel.test.ts
@@ -27,8 +27,10 @@ describe('SessionPicker long-press alert title — provider suffix (#3937)', () 
     expect(alertTitleSection).not.toMatch(/session\.provider\s*===\s*['"]claude-cli['"]/);
   });
 
-  it("only suffixes when the session's provider is not the claude-sdk default", () => {
-    expect(alertTitleSection).toMatch(/session\.provider\s*&&\s*session\.provider\s*!==\s*['"]claude-sdk['"]/);
+  it("only suffixes when the session's provider is not the default", () => {
+    // Keyed on the shared DEFAULT_PROVIDER constant (#5823), not a hardcoded
+    // 'claude-sdk', so a default flip can't reintroduce the stale-literal drift.
+    expect(alertTitleSection).toMatch(/session\.provider\s*&&\s*session\.provider\s*!==\s*DEFAULT_PROVIDER/);
   });
 
   it('feeds the computed providerLabel into the Alert.alert title argument', () => {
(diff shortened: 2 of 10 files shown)


blamechris:
## Chroxy Inspector — Review of #5824

### Summary
Single-sources the session default-provider into `@chroxy/protocol` (`export const DEFAULT_PROVIDER = 'claude-tui'`), has the server `providers.js` re-export it, and switches all three badge/label suppression sites (dashboard `Sidebar`, app `SessionPicker` pill chip + long-press alert) plus the dashboard store seed from the hardcoded `'claude-sdk'` literal to the shared constant. Source-text-pinning tests updated to assert the constant-based gate. Clean, well-scoped follow-up to the #5822 default flip.

### Strengths
- **Genuinely single-source.** Grep across all `src` packages for `!== 'claude-sdk'` (and `"claude-sdk"`) badge gates returns empty — every suppression site now references `DEFAULT_PROVIDER`. The only remaining `claude-sdk` literals in source are legitimate (the `if (name === 'claude-sdk')` provider lookup and a doc comment in `sdk-session.js`); the rest are `desktop/src-tauri/target/` build artifacts, not source.
- **Re-export works at runtime.** Verified `import { DEFAULT_PROVIDER } from './src/providers.js'` resolves to `claude-tui`. Dependency direction is correct (`server` → `@chroxy/protocol`; protocol has no server *import*, only comments) so there is no circular-import risk and no ESM named-export hazard.
- **dist artifacts are clean.** `git diff` of `dist/index.js` and `dist/index.d.ts` is purely additive (17/17 lines, **zero deletions**) — only the `DEFAULT_PROVIDER` doc + const block. `dist/index.js` exports `'claude-tui'`, `dist/index.d.ts` declares `"claude-tui"`, consistent with `src/index.ts`. No stray tsc-rebuild corruption (the known store-core dist hazard did not bite here).
- **Tests are meaningful, not vacuous.** All updated pinning regexes verified against real source: app `SessionPicker.tsx` L97/L242 match `!== DEFAULT_PROVIDER`; the renamed dashboard test reads `CreateSessionModal.tsx` (L185 `useConnectionStore(s => s.defaultProvider)`, L198 `useState(defaultProvider)`) and `connection.ts` L516 (`loadPersistedSetting('chroxy_default_provider', DEFAULT_PROVIDER)`). The old vacuous `toMatch(/claude-sdk/)` is correctly replaced with a store-sourcing assertion.
- All imports present and used; no unused imports.

### Issues
**Critical:** none.

**Suggestions:** none blocking.

**Nitpicks:**
- The shipped `desktop/src-tauri/target/**` build artifacts still carry the old `const DEFAULT_PROVIDER = 'claude-sdk'` in their bundled `doctor.js` copies. These are generated/gitignored build outputs (not edited here, and out of scope) — flagging only so they aren't mistaken for live source on a future grep.

### Verification
- protocol/server re-export: runtime resolves to `claude-tui`
- server `tests/providers.test.js`: 90 pass
- dashboard `ProviderPicker.test.tsx`: 24 pass
- app `SessionPickerProviderBadge`/`Label`: 10 pass

### Verdict
**Approve.** Correct, well-tested, no attribution issues, matches house style.

PR #6016, opened 2026-06-18
test(app): cover App-level view-mode effect transitions (#5998)

blamechris:
## Summary

Adds test coverage for the App-level view-mode effect in `packages/dashboard/src/App.tsx` (the effect that drives view-mode + the live terminal mirror for PTY / terminal-only providers). The effect was previously untested.

The `ViewSwitcher.test.tsx` AC lives in the dashboard package — the only `ViewSwitcher` in the repo, and the same effect/component PR #5996 (#5986) shipped. The effect's #5838 null-provider guard and `isPtyProvider = isTui || isUserShell` logic match the ACs exactly.

## What's tested (per AC)

**AC1 — user-shell forces terminal**
- A user-shell session sitting on the (now-hidden) Chat tab snaps `viewMode` to `terminal`.
- A user-shell session on a non-chat tab (System) is NOT snapped back (#5997).

**AC2 — non-PTY forces chat + #5838 null-provider guard**
- A non-PTY (chat) provider stranded on the Output tab is forced back to `chat` once the provider is known.
- During the load/reconnect window the provider is still `null` → the effect does NOT force-switch away from `terminal` (the #5838 guard that protects a persisted Output view for a claude-tui session).

**AC3 — PTY providers (claude-tui AND user-shell) subscribe to the mirror**
- Both claude-tui and user-shell subscribe to `subscribeTerminalMirror` on Output + connected.
- A non-PTY provider does not subscribe.
- No subscribe while disconnected; subscribes once the socket connects.
- Re-subscribes after a reconnect (unsubscribe on drop, fresh subscribe on reconnect — #5838).

**AC4 — Split + showChatTab=false**
- Dedicated `ViewSwitcher.test.tsx` case asserting the Split tab is hidden for a terminal-only provider (showChatTab=false, showTerminalTab=true).

## Test-support tweak
Added `subscribeTerminalMirror`, `unsubscribeTerminalMirror`, and `serverCapabilities` to the existing App.test store mock so the effect's selectors resolve.

## Verification
- `npx vitest run src/App.test.tsx src/components/ViewSwitcher.test.tsx` — 114 passed.
- Full dashboard suite: `npm test` — 3813 passed (202 files).
- `npm run typecheck` — clean.

Closes #5998

### DIFF
--- a/packages/dashboard/src/App.test.tsx
+++ b/packages/dashboard/src/App.test.tsx
@@ -236,6 +236,14 @@ vi.mock('./store/connection', () => {
     requestHostStatus: vi.fn(),
     requestRunnerStatus: vi.fn(),
     requestIntegrationStatus: vi.fn(),
+    // #5998 — the App-level view-mode effect opts a PTY-backed session into the
+    // live terminal mirror and forces the view-mode for terminal-only / non-PTY
+    // providers. The base mock exposes the two mirror actions + the
+    // serverCapabilities slice the effect (and the "New Shell" gate) read, so the
+    // effect's selectors don't crash on a missing action when these tests run.
+    subscribeTerminalMirror: vi.fn(),
+    unsubscribeTerminalMirror: vi.fn(),
+    serverCapabilities: null,
   }
   const useConnectionStore = (
     selector?: (s: typeof baseState) => unknown,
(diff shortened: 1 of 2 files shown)


blamechris:
## Code Review: PR #6016

### Summary
Test-only PR closing #5998. Adds 9 cases for the App-level view-mode effect in `packages/dashboard/src/App.tsx` (lines ~495-523) plus 1 `ViewSwitcher.test.tsx` case for the Split tab under `showChatTab=false`. The author's call to treat this as a **dashboard** (not mobile `packages/app`) task is correct: `ViewSwitcher` and the `isPtyProvider = isTui || isUserShell` effect live only in the dashboard; the mobile app has no equivalent surface. Confirmed genuinely test-only — both changed files are `*.test.tsx`, no production code touched.

### Strengths
- **Tests exercise the real effect, not a fabricated API.** Each case sets store state via `stateOverrides` and renders `<App />`; the effect's own branching runs unmocked. Only the store *actions* (`setViewMode`, `subscribeTerminalMirror`, `unsubscribeTerminalMirror`) are spies — the correct test boundary for "did the effect call the action."
- **Provider constants match ground truth.** `'claude-tui'` (DEFAULT_PROVIDER), `'user-shell'` (USER_SHELL_PROVIDER), `'claude-sdk'` (non-PTY) line up with `@chroxy/protocol` exactly, so `isTui`/`isUserShell`/`isPtyProvider` resolve as intended.
- **#5838 null-provider guard is genuinely covered** — the `sessions: []` case yields `activeSessionProvider == null`, and the test asserts `setViewMode` is NOT called, pinning the `activeSessionProvider != null` clause that protects a persisted Output view during the load/reconnect window. This is the load-bearing case and it's correct.
- **AC3 covers both PTY providers and the reconnect path.** Separate claude-tui and user-shell subscribe assertions; the disconnect→connect case pins the `connectionPhase === 'connected'` gate; the reconnect case (`connected`→`reconnecting`→`connected`) correctly asserts `unsubscribe` fires on the cleanup of the first run and `subscribe` is called twice total. React effect cleanup semantics are modeled accurately.
- **Negative cases included** (non-PTY does not subscribe; user-shell on System is not snapped back per #5997) — guards against over-broad future changes.
- Clean `beforeEach` reset (`stateOverrides = {}`) + `afterEach(cleanup)` already in the file; mirror-action spies created per-test, no cross-test leakage.

### Issues

#### Critical
_None._

#### Suggestions
_None blocking._

#### Nitpicks
| # | Note |
|---|------|
| 1 | AC4's new ViewSwitcher case (`showChatTab=false, showTerminalTab=true` → Split hidden) partially overlaps the existing #5997 "shows Split only when BOTH" test, which already pins the `showChatTab && showTerminalTab` gate. It's the explicit AC4 wording and adds the false-branch-with-terminal-present permutation, so it's defensible — just noting the mild redundancy. No change requested. |
| 2 | The store-mock comment says serverCapabilities is read by "the effect (and the 'New Shell' gate)"; the effect itself doesn't read `serverCapabilities` (it reads `isUserShell`, derived from the session's `provider`). Tests set `serverCapabilities: { userShell: true }` for realism but the redirect/subscribe behavior is driven purely by `provider`. Comment is slightly imprecise but harmless. |

### Verdict
**Approve.** Tests assert real behavior of the actual effect against verified source, the #5838 guard and AC3 reconnect path are correctly pinned, mocks are scoped to the right boundary without masking the logic under test, and the change is genuinely test-only. CI dashboard tests still finishing — assuming green, no concerns.

PR #6173, opened 2026-06-20
feat(server): WSL2 distro survey + start/terminate actions (#6138)

blamechris:
## Summary

#6138 (epic #5530), **server half** — the last sub of the device/runtime epic. Adds the WSL2 surface feeding the same Control Room "Device runtimes" tab as iOS/Android, mirroring the simulator/emulator pattern (does NOT fork it).

Related to #6138 (the dashboard WSL-panel slice is a fast-follow PR; the issue stays open until then).

## What changed

- **`control-room/wsl.js`** (new) — `surveyWsl` runs `wsl.exe -l -v` → `{name, state, version, isDefault}` per distro + `defaultDistro`. `runWslAction`: **start** runs `wsl.exe -d <d> -e true` (returns `running`), **terminate** runs `wsl.exe --terminate <d>` (returns `stopped`). Injectable `_execFile`/`_platform`/`_now` seams. The parser strips NUL bytes so wsl.exe's **UTF-16LE** output decodes whether it arrives clean or as NUL-interleaved bytes (NUL built via `String.fromCharCode(0)` — no control-char literal baked into source). **Windows-only**: off-`win32` (without even shelling out) OR no `wsl.exe` → `available:false`, a quietly-degraded first-class state.
- **`handlers/control-room-handlers.js`** — `wsl_status_request` + `wsl_action` with the full `docs/security/bearer-token-authority.md` checklist: host-authority gate, fresh-survey membership validation (client `distro` is a lookup key; **terminate never targets a non-enumerated distro**), state gate (start non-running / terminate running), per-distro in-flight + global cap that fast-rejects duplicates **before** the survey, exactly-once ack / `WSL_ACTION_FAILED`.
- **protocol** — `WslStatusRequestSchema`/`WslActionSchema` (client) + `ServerWslStatusSnapshotSchema`/`ServerWslActionAckSchema` (server) + dist rebuild; both documented in `ws-server.js`.
- **Coverage guards** — server-first split: `wsl_status_snapshot` + `wsl_action_ack` parked in both unhandled allowlists (the dashboard slice promotes them).

## Tests

- `control-room-wsl.test.js` — 13 cases: `parseWslList` (table parse, **UTF-16-as-bytes** decode, empty/garbage, null version), `surveyWsl` (win32 list + default, off-Windows quiet-degrade *without* exec, no-wsl degrade, null default), `runWslAction` (start/terminate command shape, bad args, failure propagation).
- `control-room-handlers.test.js` — survey + action handler cases (auth refusal, no-survey, unknown-distro, both state gates, missing distro, unavailable, concurrency fast-reject, failure relay, registry dispatch).

## Verification

- Server handler + wsl tests: **175 pass**; protocol full: **340 pass**
- Both coverage guards green; both server custom lints green; eslint clean; source verified 0 NUL bytes
- **Live validation deferred:** the `wsl.exe` round-trip needs a Windows host. The module + handlers are fully exercised here via injected exec seams + a simulated `win32` platform; on macOS/Linux the survey degrades to `available:false` (a tested first-class state).


blamechris:
**FIX** (d0d5b0045) — addressed the #6174 follow-up in-PR: WSL state gates are now default-CLOSED (start requires `Stopped`, terminate requires `Running`), aligning with the emulator sibling so a transitional-state distro can't slip the start gate. Added a transitional-state test. #6174 closed. (#6175 — live wsl.exe validation on a real Windows host — remains the only deferred step.)

Issue #5454, opened 2026-06-10
refactor(store-core): adopt shared handlers in the dashboard for the remaining duplicated message types


blamechris:
Scoped residue of #2661 (now closed as substantially complete — see the inventory in https://github.com/blamechris/chroxy/issues/2661#issuecomment-4668979301). store-core handles ~75 of 87 shared wire types; this issue tracks the remainder.

## Phase 1 — mechanical (dashboard adopts existing shared handlers)

The app already routes these through `@chroxy/store-core` handlers; the dashboard keeps local copies:

- `permission_resolved` (`message-handler.ts:1690`)
- `session_stopped` (`message-handler.ts:2594`)
- `checkpoint_restored` (`message-handler.ts:808`)
- `conversations_list` (`message-handler.ts:832`)

Plus extracting the small both-sides duplicates into shared handlers: `notification_prefs`, `pair_fail`, `session_cost_threshold_crossed`, `token_rotated`, and the trivial `pong`/`raw`/`raw_background`.

## Phase 2 — the permission pair (small but review-worthy)

- `permission_request` (`message-handler.ts:1591`, ~99 lines, includes a near line-for-line copy of the app's #554 stream-split block) — unify carefully; this is the highest-value and riskiest item.

## Also flagged

- Dashboard has no `permission_timeout` handler while the app does — confirm intentional or fix in Phase 1.

8 app-only and 32 dashboard-only types are legitimately platform-specific and stay put.

PR #6240, opened 2026-06-22
refactor(store-core): migrate checkpoint_created/checkpoint_list onto the shared dispatch table (#5618 Batch 6)

blamechris:
## Summary

Continues the dispatch-table migration (epic #5618): moves **`checkpoint_created`** and **`checkpoint_list`** off the twin platform-local switches onto the shared table (`createDispatchTable` + `runDispatch`). Both were byte-identical between the app and dashboard except the app's secondary mirror into its conversation store.

Related to #5618 (epic stays open).

## Approach (the #5595 playbook)

- **`dispatch-table.ts`**: `dispatchCheckpointCreated` / `dispatchCheckpointList` + registration + `DISPATCH_TABLE_TYPES` + `DispatchMessageMap` shapes.
  - `checkpoint_list` is a pure flat write (like `slash_commands`).
  - `checkpoint_created` is a **read-modify-write** (append to the prior list), so the adapter gains a **`getCheckpoints()`** read accessor (mirrors the existing `getSessions()`). This preserves each client's prior `if (next) { set(...) }` guard exactly — **no no-op state churn** (the reason I didn't reuse `updateState`).
- **App's conversation-store mirror** (`addCheckpoint` / `setCheckpoints`) rides on a **new optional `syncSecondaryCheckpoints?(op)` adapter hook** (`{kind:'append'|'replace'}`) — the same OUTSIDE-the-shared-contract pattern as `syncSecondaryInventory`. The **dashboard omits it** (flat write only).
- **Client handlers**: both cases removed; app wires `getCheckpoints` + `syncSecondaryCheckpoints`, dashboard wires only `getCheckpoints`. `checkpoint_restored` stays app-local (no dashboard case → still in the PENDING allowlist).

## Drift protection
- **6 contract fixtures** (`fixtures.ts`): write + non-active-session drop + malformed no-op, for each type.
- **10 dispatch-table unit tests**: ownership, empty + non-empty append (read-modify-write), gate drops, no-ops, and the **app-mirror vs dashboard-no-mirror** divergence.
- Removed `checkpoint_created` / `checkpoint_list` from the coverage-lint `PENDING_CONTRACT_TYPES` allowlist (it only shrinks).

## Verification (local)
- store-core full suite **1827 pass**
- dispatch-table + contract + coverage-lint **224 pass**
- dashboard `tsc --noEmit` clean · app `tsc --noEmit` clean
- app `message-handler.test.ts` **188 pass** (run `--runInBand --forceExit`)

### DIFF
--- a/packages/app/src/store/message-handler.ts
+++ b/packages/app/src/store/message-handler.ts
@@ -50,7 +50,9 @@ import {
   handleKeyExchangeOk as sharedKeyExchangeOk,
   handleServerMode as sharedServerMode,
   // checkpoint_created / checkpoint_list migrated to the shared dispatch table
-  // (#5618 Batch 6); only checkpoint_restored stays app-local.
+  // (#5618 Batch 6); checkpoint_restored is not migrated in this batch — it
+  // stays platform-local (here as a switch case; the dashboard via its
+  // HANDLERS map), so this shared import remains.
   handleCheckpointRestored as sharedCheckpointRestored,
   handleError as sharedError,
   handleSessionError as sharedSessionError,
--- a/packages/app/src/store/message-handler.ts
+++ b/packages/app/src/store/message-handler.ts
@@ -49,8 +49,8 @@ import {
   handleAuthFail as sharedAuthFail,
   handleKeyExchangeOk as sharedKeyExchangeOk,
   handleServerMode as sharedServerMode,
-  handleCheckpointCreated as sharedCheckpointCreated,
-  handleCheckpointList as sharedCheckpointList,
+  // checkpoint_created / checkpoint_list migrated to the shared dispatch table
+  // (#5618 Batch 6); only checkpoint_restored stays app-local.
   handleCheckpointRestored as sharedCheckpointRestored,
   handleError as sharedError,
   handleSessionError as sharedSessionError,
@@ -143,7 +143,6 @@ import { PROTOCOL_VERSION } from '@chroxy/protocol';
 import { hapticSuccess } from '../utils/haptics';
 import type {
   ChatMessage,
-  Checkpoint,
   ConnectionContext,
   ConnectionState,
   CustomAgent,
@@ -1163,6 +1162,8 @@ const _dispatchAdapter: ClientStoreAdapter<SessionState> = {
     ),
   addMessage: (m) => getStore().getState().addMessage(m),
   getSessions: () => getStore().getState().sessions,
+  // #5618 Batch 6 — checkpoint_created reads the prior flat list to append.
+  getCheckpoints: () => getStore().getState().checkpoints,
   // #5618 — user_question raises a background-session notification via the
   // app's own helper (which also mirrors the row into the mobile push store).
   pushSessionNotification: (sessionId, eventType, message) =>
@@ -1187,6 +1188,17 @@ const _dispatchAdapter: ClientStoreAdapter<SessionState> = {
       useConversationStore.getState().setSlashCommands(list as SlashCommand[]);
     } else {
       useConversationStore.getState().setCustomAgents(list as CustomAgent[]);
+    }
+  },
+  // #5618 Batch 6 — checkpoint_created / checkpoint_list also mirror into the
+  // app's secondary conversation store (its checkpoint timeline UI), exactly as
+  // the prior inline `useConversationStore.getState().addCheckpoint(...)` /
+  // `setCheckpoints(...)` did. The dashboard omits this hook (no secondary store).
+  syncSecondaryCheckpoints: (op) => {
+    if (op.kind === 'append') {
+      useConversationStore.getState().addCheckpoint(op.checkpoint);
+    } else {
+      useConversationStore.getState().setCheckpoints(op.checkpoints);
     }
   },
   // #5618 Batch 2 — the app tightens provider_list elements via mapProviderList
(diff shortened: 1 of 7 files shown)


blamechris (commented):
## Chroxy Inspector — review of #6240 (dispatch-table Batch 6: checkpoint_created / checkpoint_list)

**Verdict: Approve.** Behavior is provably preserved for both migrated cases across both clients.

### Summary
Migrates `checkpoint_created` + `checkpoint_list` off the twin platform-local switches onto the shared dispatch table, adding a required `getCheckpoints()` read accessor (for the read-modify-write append) and an optional `syncSecondaryCheckpoints?(op)` hook for the app's `useConversationStore` mirror (dashboard omits it). I compared the new `dispatchCheckpointCreated` / `dispatchCheckpointList` against the original `origin/main` switch cases in both clients, traced the shared helpers (unchanged), the adapter `setState`/`getCheckpoints` wiring, the contract/unit fixtures, and the coverage-lint invariants.

### Behavior-preservation verdict — per case

1. **checkpoint_created — preserved.** New handler reads the prior list via `getCheckpoints()`, passes the same three args (`msg`, prior list, `getActiveSessionId()`) to the unchanged `handleCheckpointCreated`, writes only on a non-null result (`if (!next) return`), then fires `syncSecondaryCheckpoints({kind:'append', checkpoint: msg.checkpoint as Checkpoint})` — `msg.checkpoint`, the same field the original `addCheckpoint(msg.checkpoint as Checkpoint)` used, gated on the same non-null guard. No `updateState`, so no no-op state churn — the reasoning holds.
2. **checkpoint_list — preserved.** Pure `setState({checkpoints: next})` on non-null + app `syncSecondaryCheckpoints({kind:'replace', checkpoints: next})`. The dashboard's `origin/main` case was `if (next) set({checkpoints: next})` with **no** conversation-store write, and the dashboard correctly omits the optional hook → flat write only. Matches.
3. **Active-session gate — preserved.** The gate + payload validation stay in the unchanged shared `handleCheckpointCreated`/`handleCheckpointList` helpers (`checkpoint.ts` untouched). The 3 `noop` contract fixtures (non-active-session drop + malformed payload, per type) run through the real `runDispatch` path for **both** `app` and `dashboard` adapters and assert `Object.keys(result.flat).toHaveLength(0)` + no mirror — a genuine proof that a non-active / malformed message stays a total no-op.
4. **checkpoint_restored — untouched / still both-client local.** Not migrated; the app `case 'checkpoint_restored'` and the dashboard's `checkpoint_restored: handleCheckpointRestored` HANDLERS entry are both still present on the PR branch. Correctly remains in the PENDING allowlist.
5. **Adapter wiring — correct.** App implements both `getCheckpoints` + `syncSecondaryCheckpoints`; dashboard implements only `getCheckpoints`; both read `getStore().getState().checkpoints`. The `setState` adapter on both is `(patch) => getStore().setState(patch)` — identical to the original inline `set({checkpoints})`.
6. **Drift guard — holds.** `checkpoint_created`/`list` are added to `DISPATCH_TABLE_TYPES` (no dups; both registered in `createDispatchTable` and present in `DispatchMessageMap`/`DispatchMessageType`), so the coverage-lint subtracts them from the both-clients universe; `checkpoint_restored` stays PENDING. Universe shrinks by 2 — still inside the lint's 42–54 band. 6 fixtures + 10 unit tests cover ownership, empty + non-empty append (read-modify-write via `getCheckpoints`), gate drops, no-ops, and the app-mirror-vs-dashboard-no-mirror divergence. No remaining references to `sharedCheckpointCreated`/`sharedCheckpointList` anywhere; the app cleanly dropped the now-unused `Checkpoint` type import (0 remaining references → tsc-safe).

### Issues

**[Low / doc-only] Misleading "no dashboard case" comments for checkpoint_restored.** Two comments assert the dashboard has no `checkpoint_restored` handler:
- `packages/dashboard/src/store/message-handler.ts` ~line 46: *"the dashboard has no checkpoint_restored case."*
- `packages/store-core/src/contract-fixtures/coverage-lint.test.ts` (the Batch 6 note): *"checkpoint_restored stays app-only (no dashboard case), so it remains pending."*

This is factually wrong — the dashboard registers `checkpoint_restored: handleCheckpointRestored` in its `HANDLERS` map (message-handler.ts line ~3112; added in #5454, which made the dashboard auto-switch to the restored session to match the app). **No behavior or test impact**: `checkpoint_restored` is a genuine both-clients handled type with no DISPATCH/SWITCH fixture, so keeping it in `PENDING_CONTRACT_TYPES` is exactly right and the lint stays green either way. The *outcome* is correct; only the stated *rationale* is misleading. Suggest correcting both comments to say something like "checkpoint_restored is still handled by both clients' own switch/HANDLERS (not migrated this batch) and has no contract fixture, so it stays pending." Non-blocking — fine to fix in this PR or fold into the next batch.

### Style
- The `as Record<string, Checkpoint[]>` / `as Checkpoint` casts match the established batch convention (e.g. the `availablePermissionModes` / `sessions` setState casts); no `any` abuse.
- The `syncSecondaryCheckpoints` discriminated-union param (`{kind:'append', checkpoint}` | `{kind:'replace', checkpoints}`) is clean and mirrors the `syncSecondaryInventory` precedent.
- Fixture adapter `getCheckpoints: () => (flat.checkpoints as Checkpoint[] | undefined) ?? []` reads the live `flat` ref, so the contract fixtures observe identical results — correct.

No other callers of the removed cases; no place relied on the inline behavior beyond the two switches.

**Verdict: Approve** — zero behavior change confirmed; the only finding is a non-blocking inaccurate comment.

PR #6242, opened 2026-06-22
test(dashboard): cover failed-only re-invoke in useTrayBadgeSync (#6227)

blamechris:
Closes #6227. From the #6226 review: the dedup key is `${blocked}:${failed}`, so a change to the **failed** slice alone (blocked constant) should re-invoke `update_tray_badge` — but the suite only exercised re-invoke via a `blocked` change.

Adds a test that renders with a fixed `blocked=1`, flips a session `running`→`failed` in the activity rollup between renders, and asserts a second `update_tray_badge` call with `failed: 1`. Test-only; the runtime path was already correct (7 pass in the file).

### DIFF
--- a/packages/dashboard/src/hooks/useTrayBadgeSync.test.tsx
+++ b/packages/dashboard/src/hooks/useTrayBadgeSync.test.tsx
@@ -103,4 +103,19 @@ describe('useTrayBadgeSync (#6184)', () => {
     expect(invokeSpy).toHaveBeenCalledTimes(2)
     expect(invokeSpy).toHaveBeenLastCalledWith('update_tray_badge', { blocked: 0, failed: 0 })
   })
+
+  // #6227 — the dedup key is `${blocked}:${failed}`, so a change to the FAILED
+  // slice alone (blocked held constant) must also re-invoke. The other re-invoke
+  // test only varies `blocked`; this covers the failed-only edge.
+  it('re-invokes when only the failed count changes (blocked constant)', () => {
+    setStore({ activity: activityOf({ s1: 'running' }), sessions: [{ sessionId: 's1', cwd: '/r' }] })
+    const { rerender } = render(<Harness blocked={1} />)
+    expect(invokeSpy).toHaveBeenCalledTimes(1)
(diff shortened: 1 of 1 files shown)


blamechris (commented):
## Review — test(dashboard): cover failed-only re-invoke in useTrayBadgeSync (#6227)

**Verdict: Approve** (posted as a comment — GitHub blocks self-approval).

**Summary**

Adds one test asserting the failed-only re-invoke path of `useTrayBadgeSync`. Verified against the hook + the existing file:

- **Exercises the right path.** The hook reads `failed` via `useConnectionStore((s) => selectCrossSessionActivity(s.activity, s.sessions).total.failed)`. The mock's `useConnectionStore` is `(selector) => selector(storeState)`, and `setStore` swaps `storeState`. On `rerender`, the component (and thus the hook + selector) re-runs against the **new** store, so `failed` moves 0→1 while `blocked` stays 1. The effect's `[blocked, failed]` deps change on `failed`, triggering the second invoke. This is genuinely the failed-only edge, not a blocked-driven one.
- **Assertion is correct.** `activityOf({ s1: 'failed' })` produces a `bySession` entry with `status: 'failed'` + `endedAt: 2`, which `selectCrossSessionActivity` rolls up to `total.failed: 1`. First render → `{blocked:1, failed:0}` (call 1); after the flip → `{blocked:1, failed:1}` (call 2). Matches `toHaveBeenCalledTimes(2)` and the `toHaveBeenLastCalledWith` payload.
- **Guards the dedup key, no false-pass.** If the key dropped `failed` (`${blocked}` only), it would be `"1"` on both renders and the second invoke would be suppressed → `toHaveBeenCalledTimes(2)` would fail. So the test does protect the `${blocked}:${failed}` contract.
- **Matches existing patterns.** Same `activityOf` / `setStore` / `Harness` / `render`+`rerender` shape as the sibling "re-invokes when blocked changes" test. No flakiness risk (synchronous, single subject session).
- **Test-only.** Diff is +15/-0 in `useTrayBadgeSync.test.tsx`; no production code changed.

Correct, non-flaky test that closes the #6227 coverage gap and would catch a regression in the dedup key.

PR #6322, opened 2026-06-25
test(store-core): soften protocol-type-coverage-lint to a floor-only guard

blamechris:
## Summary

Closes #6274. The `protocol-type-coverage-lint` extraction-sanity guard used a hardcoded `[80,120]` band on the server→client `z.literal` universe. With the control-room surface restored (#6273) the count is ~108 — only ~12 headroom before a legitimate new type or another tab-split trips the **ceiling** and fails the lint spuriously.

The ceiling has no protective value: the regression this guard exists to catch (a parser fault that stops matching `z.literal`, e.g. the #6272 barrel-split trap) *empties* the universe — it doesn't inflate it. So the ceiling only policed legitimate growth.

## Change

Replace the band with a generous **floor-only** guard (`>= 70`, well below the real ~108). It still trips loudly on a collapse-to-near-zero, but new server→client types and tab-splits no longer need a paired bound-bump. Rationale documented inline so the next surface-growth PR knows the policy.

## Verification

Full store-core suite: 1838 passed. Typecheck clean.

Closes #6274

### DIFF
--- a/packages/store-core/src/contract-fixtures/protocol-type-coverage-lint.test.ts
+++ b/packages/store-core/src/contract-fixtures/protocol-type-coverage-lint.test.ts
@@ -227,14 +227,17 @@ describe('protocol server→client type coverage lint (#6033)', () => {
   const { inApp, inDash } = clientCoverage()
 
   it('derives a non-trivial server→client schema-type union (extraction sanity)', () => {
-    // Pin to a band around the real count (~92) so a parser regression (e.g. the
-    // schema being reformatted in a way the literal regex stops matching) fails
-    // LOUDLY rather than silently shrinking the universe to nothing and passing
-    // vacuously. A real new server→client schema type that pushes past the
-    // ceiling is a DELIBERATE bump — raise both bounds in the same PR that adds
-    // the type (and its both-client coverage or allowlist entry).
-    expect(types.length).toBeGreaterThanOrEqual(80)
-    expect(types.length).toBeLessThanOrEqual(120)
+    // Floor-only guard (#6274). The only regression this protects against is a
+    // PARSER fault that silently empties the universe — e.g. the schema being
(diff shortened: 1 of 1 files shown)


blamechris:
## Inspector Review

**Summary**

Replaces the `[80,120]` band on the server→client `z.literal` universe with a floor-only guard (`>= 70`). Verified against the source:

- **Floor still catches the real regression.** The guard exists to catch a parser fault (reformat / barrel-split à la #6272) that stops the `z.literal` regex from matching and silently *empties* the `types` Set. With an empty universe, the downstream "handled by both clients" assertion passes vacuously (`[] === []`), so this floor is the sole tripwire — and `>= 70` against an empty/near-zero set still fails loudly. Protection intact.
- **`70` is a sensible floor.** Real count is **106** today (PR says ~108 — matches within normal churn). 70 sits ~34% below that: generous headroom for legitimate add/remove churn, comfortably above any collapse-to-near-zero.
- **Nothing of value lost by dropping the ceiling.** A parser regression can only *shrink* the matched-literal count, never inflate it. The only path past 120 is deliberately adding >12 new server→client types — healthy growth, not a bug. The old ceiling had ~12 headroom and would have spuriously failed on the next type or tab-split, forcing a paired bound-bump for no safety gain. Pure brittleness; correct to remove.
- **Inline rationale is accurate and clear.** Correctly states the failure mode ("empties the universe, it does not inflate it"), cites #6272/#6274, and tells the next surface-growth PR the policy. Good for the next maintainer.
- **No collateral.** Grepped the file: the band is referenced only by the removed/replaced lines. The other four assertions filter `types` for membership and are independent of the count bound.

**Verdict: Approve.** Tight, well-reasoned, and correctly narrows the guard to the only failure mode it can actually detect.

PR #6384, opened 2026-06-26
feat: extend the models.json overlay to non-Claude provider registries (#6377)

blamechris:
## What & why

Third item from the **model-serving freshness audit**. The hot-reloadable `~/.chroxy/models.json` overlay — the lever for adding/relabeling a model at runtime with no release — previously reached **only the default Claude registry** (`getRegistryForProvider` built non-Claude registries with no overlay). So an operator couldn't surface or relabel a Gemini/Codex/DeepSeek/Ollama model without a code release.

This routes overlay entries to the right registry via an optional per-entry **`provider`** field.

## How

```json
{
  "gemini-3.0-pro": { "provider": "gemini", "label": "Gemini 3.0 Pro", "contextWindow": 2000000 }
}
```

- **`loadModelsOverlayResult`** now returns `{ ok, overlay, byProvider }`. An entry with a `provider` field routes to `byProvider[name]`; an untagged entry stays in the default (Claude) overlay — **fully backward-compatible** (existing flat files unchanged).
- **Routing is by field _presence_, not name classification.** The overlay loads at module-init, before providers may have registered, so an `isClaudeProvider()` lookup there would be unreliable — presence routing is deterministic and boot-safe. (A Claude provider name, or omitting the field, lands on the Claude registry.)
- **`getRegistryForProvider`** threads `providerOverlays.get(name)` into each non-Claude registry — the model seeds the picker, resolves both ways, and **lands in the allowlist**, exactly like the Claude registry already does.
- **Hot-reload** (`reloadModelsOverlay`) re-folds the per-provider slices into already-built cached registries; `_resetModelsOverlayForTests` resets the provider state too.

## Scope note (deliberate)

**`pricing` for a tagged entry is NOT applied here.** Non-Claude cost flows through `ProviderClass._getPricing`, a separate seam from the registry — wiring overlay pricing into every provider class is its own change. Labels, context windows, and new-model seeding work for **every** provider; per-provider overlay *pricing* is a documented follow-up. (This PR is faithful to the issue title — "extend the overlay to non-Claude provider **registries**.")

## Tests

5 new (`tests/models-overlay-per-provider.test.js`): tagged entry seeds the provider registry (picker + allowlist) and does NOT leak to Claude · untagged entry stays on Claude · **hot-reload re-folds an already-built registry in place** · a dropped entry removes the overlay-only row · gemini↔codex isolation. eslint + 4 custom lints clean; **198 pass** across all models suites (no regression — existing overlay/per-provider behaviour intact). Docs: `model-overlay.md` gains the `provider` field + a "Targeting a specific provider" section.

Closes #6377.

### DIFF
--- a/docs/guides/model-overlay.md
+++ b/docs/guides/model-overlay.md
@@ -2,7 +2,7 @@
 
 Add, relabel, or re-price a model **at runtime, with no Chroxy release** — by dropping a small JSON file next to your config. The overlay is Chroxy's lightest "serve a model the build doesn't know about yet" lever: edit the file and the change is picked up live (hot-reload), no restart.
 
-> **Scope today:** the overlay applies to the **default Claude registry** only (`claude-sdk` / `claude-cli` / `claude-tui` / docker-wrapped Claude). Extending it to the other providers' registries (Gemini, Codex, DeepSeek, Ollama, config-driven endpoints) is tracked in [#6377](https://github.com/blamechris/chroxy/issues/6377). For those providers today, see [`providers.allowAnyModel`](../providers.md#serving-a-new-model-without-a-release-providersallowanymodel) (serve any API-valid id) and the config-driven [`models` / `modelDiscovery`](../providers.md#anthropic-compatible-endpoints-config-driven) blocks.
+> **Scope:** by default an entry applies to the **Claude registry** (`claude-sdk` / `claude-cli` / `claude-tui` / docker-wrapped Claude). Since [#6377](https://github.com/blamechris/chroxy/issues/6377) an entry can carry a **`provider` field** to target any other provider's registry instead (Gemini, Codex, DeepSeek, Ollama, config-driven endpoints) — see [Targeting a specific provider](#targeting-a-specific-provider-non-claude). One caveat: **`pricing` in a `provider`-tagged entry is not yet applied** — non-Claude cost flows through the provider class, not the registry, so overlay pricing currently only takes effect for Claude models (labels / context windows / new-model seeding work for every provider).
 
 ## Location
 
@@ -38,7 +38,8 @@ A JSON **object keyed by the model's full id**. Every field except the key is op
 | `shortId` | Optional alias (e.g. `opus-5`) the picker and `set_model` accept. Defaults to a derived short id. |
 | `label` | Optional display name in the picker. Defaults to a humanized id. |
 | `contextWindow` | Optional token window (positive number) for the context meter. |
-| `pricing` | Optional USD-per-MTok rates: `{ input, output, cacheRead, cacheWrite }`. Absent → cost reads `null` (not `$0`). |
+| `pricing` | Optional USD-per-MTok rates: `{ input, output, cacheRead, cacheWrite }`. Absent → cost reads `null` (not `$0`). (Claude registry only — see the scope note.) |
+| `provider` | Optional provider name (e.g. `"gemini"`, `"codex"`, `"deepseek"`) to target that provider's registry instead of Claude's. Omit for Claude models. |
 
 ### What an entry does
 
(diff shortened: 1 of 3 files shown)


blamechris:
## Chroxy Inspector — review of #6384 (per-provider models.json overlay, #6377)

### Summary
Extends the hot-reloadable `~/.chroxy/models.json` overlay to non-Claude provider registries via an optional per-entry `provider` field. `loadModelsOverlayResult` now returns `{ ok, overlay, byProvider }`; module state gains `providerOverlays`; `getRegistryForProvider` threads each provider's slice into its registry; `reloadModelsOverlay` / `_resetModelsOverlayForTests` re-fold cached registries. The implementation is tight, reuses the proven #5932 `applyOverlay` re-fold seam, and the scope note is honest. **Verdict: APPROVE.**

### Backward compatibility (focus #1) — VERIFIED
The default path is unchanged. An entry with no `provider` (or `provider` empty/non-string) hits the `else` branch at `models.js:274-276` and lands in `overlay` exactly as before — same `entry` object, same `overlay.set(fullId, entry)`. The boot binding `defaultOverlay = _bootOverlayResult.overlay` (291-292) is the same Map a flat file always produced; `defaultRegistry` (1072), `getModelPricing` (70-71), and `reloadModelsOverlay`'s default re-fold (1095-1096) are byte-for-byte equivalent for untagged files. The 99 tests across `models.test.js` + `models-overlay-reload.test.js` + the new suite pass locally with no regression.

### Hot-reload re-fold (focus #3) — VERIFIED correct on all three sub-points
- **(a) Provider that lost all entries → no stale leak.** `result.byProvider` simply omits a provider with zero tagged entries, so `providerOverlays.get(name) ?? new Map()` (1104) hands the cached registry an **empty** Map. `applyOverlay(empty)` recomputes `fallbackModels` from base only, dropping overlay-only rows. The "dropped entry" test asserts exactly this (row + allowlist entry both gone). Correct.
- **(b) Not-yet-built provider picks up its slice lazily.** Confirmed — `getRegistryForProvider` reads `providerOverlays.get(providerName)` at creation time (1347), and `reloadModelsOverlay` reassigns `providerOverlays` (1102) *before* the re-fold loop, so a later first-build sees the fresh slice. The "already-built registry" test covers the cached side; the lazy side is the create path.
- **(c) No SDK/cache re-merge interaction bug.** The per-provider re-fold calls the **same** `applyOverlay` the Claude registry uses, which correctly branches on `lastSdkModels` / `lastCacheModels` / bare-fallback (669-691). Non-Claude registries are typically in the cache/fallback state, and the overlay-only-row-removal semantics (preserve from cache, not `activeModels`) already handle the "operator removed a row" case. No new seam, no new bug surface.

### Other focus points
- **Routing + boot-safety (#2):** Sound. `loadModelsOverlayResult()` *is* called at module-init (`_bootOverlayResult`, 291) before `providers.js` populates `nameToProviderClass`, so an `isClaudeProvider()` classification there would see an empty map and misroute. Presence-routing is deterministic and order-independent — the right call. The consequence (a Claude-named `provider` tag is a no-op because `getRegistryForProvider` returns `defaultRegistry` for Claude names, and that registry never reads `providerOverlays`) is real and documented in both the code comment (264-267) and the guide. Acceptable.
- **TDZ / module-init ordering (#5):** Safe. `reloadModelsOverlay` (1089) and `_resetModelsOverlayForTests` (1182) reference `providerRegistryCache` (declared 1204) but are only ever invoked at runtime — the sole internal caller is `watchModelsOverlay`'s `fire()` closure (1140), which runs on a watch event, never at import. By call time the `const` is initialized.
- **State-leak / test isolation (#4):** Reset is complete. `_resetModelsOverlayForTests` now reassigns `providerOverlays` *and* re-folds every cached registry with its (possibly empty) slice. `providerOverlays` is `let`, read consistently at every site (create 1347, reload 1102-1104, reset 1187-1189). No leak path.
- **Scope honesty (#6) — VERIFIED true.** `getModelPricing` (71) resolves pricing from `defaultOverlay` **only**; `providerOverlays` is never consulted by any pricing path. A tagged entry's `pricing` block is collected into the entry (258-260) but inert — non-Claude cost flows through each `ProviderClass._getPricing` (deepseek/ollama/anthropic-compatible). The PR/doc claim is accurate, not a silent gap.
- **Docs + Zero Attribution (#7):** Doc is accurate and the caveats match the code. No attribution markers anywhere in the diff. Style conforms (no semicolons, single quotes, ES modules).

### Issues
**Critical:** none.

**Suggestions:**
1. *(minor, optional)* A tagged entry's `pricing` block is silently dropped from the effective cost path — accurate per the scope note, but an operator who writes `{ "gemini-x": { "provider": "gemini", "pricing": {...} } }` gets no signal it's ignored. Consider a one-time `log.debug` when a `provider`-tagged entry carries a `pricing` block, so the inertness is observable rather than purely doc-tribal. Out of scope for this PR; filing as a follow-up.

**Nitpicks:**
2. The JSDoc `@returns` on `loadModelsOverlayResult` (217) still reads `{{ ok: boolean, overlay: Map<...> }}` — it now also returns `byProvider`. Worth updating the annotation to match the new shape.

### Verdict
**APPROVE.** Backward compatibility is airtight, the hot-reload re-fold is correct on drop/lazy/re-merge, the boot-safety reasoning holds, and the scope note is faithful. Two non-blocking polish items above.

PR #6404, opened 2026-06-26
feat(dashboard): always-visible Enter-mode keyhint (chat redesign Phase 1)

blamechris:
Slice 5 of **Phase 1** (#6391) — the always-visible Enter-mode keyhint.

The composer's send key is no longer a guess: a muted hint sits inline in the actions row — **send on Enter / Shift+Enter for a newline** (or **Cmd/Ctrl+Enter to send** when Enter inserts a newline), driven by the `sendOnEnter` setting. Placed before the buttons so there's **no extra row -> no composer height churn**; `aria-hidden` since the existing `sr-only` shortcuts span already carries it for screen readers. Tokens only (ratchet-clean).

**Scope note:** the *send/stop visual morph* (the other half of planned slice 5) is **deferred**. During streaming the composer deliberately renders **both** a follow-up Send **and** Stop (the queue-while-streaming affordance) — a literal single-button morph would break that, so it needs a careful restyle that keeps both buttons. Tracked for a follow-on.

Verified: `tsc`, build, full dashboard suite (4089, +1 keyhint test) green.

Part of #6391 · epic #6389.

### DIFF
--- a/packages/dashboard/src/components/InputBar.test.tsx
+++ b/packages/dashboard/src/components/InputBar.test.tsx
@@ -25,6 +25,17 @@ describe('InputBar', () => {
     expect(screen.getByTestId('input-bar')).toHaveAttribute('data-activity-state', 'waiting')
   })
 
+  it('shows an always-visible Enter-mode keyhint (chat redesign #6391)', () => {
+    const { rerender } = render(<InputBar onSend={vi.fn()} onInterrupt={vi.fn()} sendOnEnter />)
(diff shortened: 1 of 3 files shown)


blamechris:
Self-review: muted always-visible Enter-mode keyhint inline in the actions row (no extra row). aria-hidden (sr-only hint already present); token-only styling. tsc + build + full suite (4089) green. Send/stop visual morph deferred (must keep both buttons during streaming).

PR #6467, opened 2026-06-27
test(server): mock-endpoint fidelity suite for the OpenAI-compatible provider (#6251)

blamechris:
Part of #6251 — an autonomous sub-task. Shrinks the dogfood ask from open-ended ('validate real-wire fidelity against LM Studio / OpenRouter / vLLM') to 'confirm a real server matches these pinned shapes'.

## Gap this closes
The sibling tests (`anthropic-openai-translate` / `-shim`) drive the translator + shim with a **mocked `openai` package** — they never exercise the real SDK's HTTP + SSE-parsing path. So a real-endpoint quirk (SSE framing, abort, the SDK's chunk handling) wouldn't be caught until a live spike.

## What
A real in-process **mock chat-completions endpoint** (127.0.0.1, `text/event-stream`), with the shim's **actual `openai` client** pointed at it — driving the full round-trip: `shim → openai SDK → HTTP → SSE → translator → Anthropic events`.

Pins 9 fidelity checks:
- **streaming text** reassembles in order → `content_block_*` + final `{type:'text'}`;
- **usage** maps faithfully (incl. `prompt_tokens_details.cached_tokens → cache_read_input_tokens`);
- **finish_reason → stop_reason**: `stop→end_turn`, `length→max_tokens`, `tool_calls→tool_use`, plus the *omitted-but-tool_calls* inference;
- **streamed tool_calls** with id/name/args split across chunks → a `tool_use` block with parsed `input`;
- the **forwarded request** is correctly translated (`stream:true` + `stream_options.include_usage`);
- **AbortSignal** cancellation reaches the network (upstream connection closes, stream stops early).

## Verified
9/9 pass; the openai-area suite stays green (40, 0 fail); custom lints pass. Purely additive (a new test file — no src change).

### DIFF
--- a/packages/server/tests/openai-compatible-fidelity.test.js
+++ b/packages/server/tests/openai-compatible-fidelity.test.js
@@ -1,0 +1,209 @@
+/**
+ * Mock-endpoint fidelity suite for the OpenAI-compatible provider (#6251).
+ *
+ * The sibling tests (anthropic-openai-translate / -shim) exercise the translator
+ * + shim with a MOCKED `openai` package — they never touch the real SDK's HTTP
+ * + SSE-parsing path. This suite closes that gap WITHOUT a live endpoint: it
+ * stands up a real in-process mock chat-completions server, points the shim's
+ * actual `openai` client at it, and drives the full wire round-trip
+ *   shim → real openai SDK → HTTP → SSE → translator → Anthropic events.
+ *
+ * It pins the fidelity dimensions a live-endpoint spike (LM Studio / OpenRouter /
+ * vLLM, #6251) would check — streaming text, usage, finish_reason → stop_reason,
+ * streamed tool_calls, the forwarded request shape, and AbortSignal cancellation
+ * — so the manual spike only has to confirm a real server matches these shapes.
+ *
+ * No network beyond 127.0.0.1; no live endpoint required.
+ */
+import { describe, it } from 'node:test'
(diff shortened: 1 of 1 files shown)


blamechris:
## Code Review: PR #6467 — OpenAI-compatible mock-endpoint fidelity suite
**✅ Approve, no critical.** Verified: (1) it genuinely drives the **real** `openai` v6.44.0 SDK against a real in-process HTTP/SSE mock (`createAnthropicShimClient` imported from the actual shim; SSE framing `data: {json}\n\n` + `[DONE]` is exactly what the SDK parser consumes) — the genuine gap vs the sibling tests' mocked package; (2) every fidelity assertion matches the real translator — `mapFinishReason` (stop→end_turn / length→max_tokens / tool_calls→tool_use), `mapUsage` (cached→cache_read), the `tool_use` block + `input_json_delta` streaming, id/model from chunks, the omitted-finish_reason→tool_use inference — all cross-checked (re-ran mapUsage, got the identical object); (3) **robust** — abort fires on the first delta with ~1.5s margin before the drip completes, streaming tests are fixed scripts, mocks close in `finally`; passes 5/5 + the full openai-area suite (40, 0 fail) with clean exit-0 (no `--test-force-exit` needed). Three non-blocking nitpicks (abort ~4s; block-start position; usage path for length/tool_calls covered separately).

PR #6517, opened 2026-07-02
feat(dashboard): Markdown syntax highlighting in the IDE file viewer (#6508)

blamechris:
## Summary

The opt-in IDE file viewer (#6469) rendered `.md`/`.markdown` files with **no syntax highlighting** — every line one `plain` colour — because those extensions were neither a `LANGUAGES` key nor an `ALIASES` entry in the shared tokenizer (`packages/store-core/src/syntax.ts`). Markdown is the **4th-most-common extension in this repo (113 files)** — `CLAUDE.md`, docs, guides — so it's one of the most-viewed file types in the viewer.

Same **bug class** as #6507 (`.mjs/.cjs/.mts/.cts`), but Markdown needs a real grammar, not a one-line alias.

## What changed

- **`packages/store-core/src/syntax.ts`** — a lightweight, line-oriented Markdown `LanguageDef`, registered as the `markdown` key + `md`/`mdx` aliases (`.markdown` hits the key directly; the server sends the bare extension as the language id).

The viewer tokenizes **each line independently** (`FileBrowserPanel` — `lines.map(line => tokenize(line, lang))`), so every rule is single-line:
- **Block constructs** (ATX headings, list/quote markers, fenced-code delimiter lines, thematic breaks) are `^`-anchored so they fire only at column 0.
- **Inline spans** (inline code, links, images, bare URLs, bold/italic/strikethrough) match anywhere on the line.

### Colour mapping
| Construct | Token type | Colour |
|-----------|-----------|--------|
| Heading | `keyword` | purple |
| Inline code + fence delimiter | `string` | green |
| Link / image / bare URL | `function` | blue |
| Bold / italic / strikethrough | `number` | warm |
| Blockquote / thematic break | `comment` | grey |
| List marker | `operator` | — |

### Two deliberate scope cuts (documented inline)
1. **`_`/`__` emphasis is omitted** — this repo's docs are full of `snake_case` identifiers and file names (`feedback_test_state_contamination.md`), so underscore emphasis would mangle far more than it highlights.
2. **A fenced block's *content* isn't tracked across lines** (the tokenizer is stateless per-line), so only the ` ``` ` delimiter line is coloured, not the code inside it.

Bold/italic/strikethrough share the warm `number` colour because output is colour-only spans (weight/slant can't be shown) — unifying emphasis to one hue keeps it distinct from body text. No new token types or CSS: it reuses the existing `.syn-*` classes every other language uses.

## Testing

- **7 new cases** in `packages/dashboard/src/lib/syntax.test.ts` (mirroring the #6507 pattern): `md`/`markdown`/`mdx` resolve; a heading is one non-plain token; a mid-line/spaceless `#` is **not** a heading; a mixed inline line yields code+link+bold token types; list markers colour but text stays plain (and `1.2.3` is not a marker); `snake_case` stays all-plain; bare URLs + blockquote markers colour.
- Full **store-core** (1919) and **dashboard** (4176) suites green.
- **Live-verified** against the running daemon: a `README.md` renders headings purple (`syn-keyword`), inline code green (`syn-string`), links blue (`syn-function`), and `_underscored_` prose stays `syn-plain`.

Closes #6508

### DIFF
--- a/packages/dashboard/src/lib/syntax.test.ts
+++ b/packages/dashboard/src/lib/syntax.test.ts
@@ -118,6 +118,59 @@ describe('tokenize', () => {
   })
 })
 
+describe('markdown highlighting (#6508)', () => {
+  it('resolves md / markdown / mdx (server sends the bare extension as the id)', () => {
+    // reader.js sends `extname` verbatim — `.md` → 'md', `.markdown` → 'markdown'.
+    // Without a Markdown LanguageDef these fall through to all-`plain` (no colours).
+    expect(getSyntaxRules('md')).not.toBeNull()
+    expect(getSyntaxRules('markdown')).not.toBeNull()
+    expect(getSyntaxRules('mdx')).not.toBeNull()
+  })
+
+  it('colours an ATX heading line as a single non-plain token', () => {
+    const tokens = tokenize('## Getting started', 'md')
+    expect(tokens).toEqual([{ text: '## Getting started', type: 'keyword' }])
+  })
+
+  it('does NOT treat a mid-line or spaceless # as a heading', () => {
+    // `#6508` and `C#` are not headings — an ATX heading needs a space/EOL after
+    // the hashes, so these must stay plain (no keyword token).
+    expect(tokenize('see issue #6508 for details', 'md').some(t => t.type === 'keyword')).toBe(false)
+    expect(tokenize('#hashtag', 'md').some(t => t.type === 'keyword')).toBe(false)
+  })
+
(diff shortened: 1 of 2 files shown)


blamechris:
Folded the M1 minor finding into this PR (commit d8d86a197) instead of deferring: the five `^`-anchored block rules now carry the `m` flag, so whole-file callers (mobile `FileBrowser.tsx`) highlight block constructs on every line, not just line 1. No-op for the dashboard's per-line path. Added a whole-file multi-line test. Closed #6518.

PR #6602, opened 2026-07-05
fix(server): detect claude-tui readiness via PTY output quiescence (#6601)

blamechris:
## Problem

Current claude (2.1.186/187) writes **no** `~/.claude/sessions/<pid>.json` for its **interactive** TUI (only `-p`/SDK mode and the Claude Desktop app do). chroxy's readiness probe keys on that file, so for every claude-tui session it never resolves — the probe burns the full `SPAWN_WARMUP_MAX_MS` (15s) on cold start and `TURN_PROMPT_WAIT_MAX_MS` (5s) on **every** turn, then falls through to "write anyway" (degraded). It works in a trusted cwd but wastes ~13s/session + ~5s/message.

Root-caused live for #6601 (see the issue for the full investigation — including why string-scraping the TUI is too fragile).

## Fix

Fall back to **PTY output quiescence** when no session file resolves: once claude has rendered on this spawn (`_sawFirstOutput`) and output has been quiet for `READY_QUIESCENCE_MS` (400ms), the composer's render burst has settled and it's ready. This is a robust, version-independent signal (no glyph/text scraping) that works for both cold-start and between-turns (bridges the post-Stop-hook redraw).

Surgical: only `checkReady`'s **no-file** branch changes (`return false` → the quiescence check). The session-file path is untouched for `-p` / claude-desktop / any version that writes a file, the warmup/turn ceilings still backstop a genuinely-stuck TUI, and the auth-failure scan still runs first in the poll loop so a logged-out banner can't read as ready.

## Validated live (daemon on this branch)

| Metric | Before (main) | After |
|---|---|---|
| Cold-start `claude_ready` | ~15.1s | **~1.2s** |
| Per-turn `waitForPrompt` | ~5.0s | **~0ms** |
| First message end-to-end | ~15s+ | **~2.3s** |

Daemon log confirms the quiescence path firing: `waitForPrompt (elapsedMs=1137 sawStatus=false ready=true)` — ready via quiescence (no file), and the "readiness gating disabled / may stall" warnings are gone.

## Tests

All 311 claude-tui-session tests pass, incl. 4 new ones pinning the fallback + its guards: quiescent→ready, still-flowing→not-ready, **no-output-yet→not-ready** (the `now - 0` quiet-by-default trap the `_sawFirstOutput` guard closes), and busy-file-overrides-quiescence.

## Scope / follow-up

An interstitial (trust/onboarding) is also quiescent, so this doesn't add interstitial *detection* — but it doesn't make it worse: chroxy already pre-trusts the cwd (so the trust dialog is pre-empted), and current claude's notices render inline (non-blocking). A dedicated interstitial guard can be a follow-up if a real case surfaces; adding a fragile token matcher now would risk its own false-negative wedges.

Closes #6601

### DIFF
--- a/packages/server/src/claude-tui-session.js
+++ b/packages/server/src/claude-tui-session.js
@@ -434,6 +434,13 @@ export class ClaudeTuiSession extends BaseSession {
     // UTF-8 strings already decoded, but the relevant control bytes
     // are 7-bit ASCII and survive the decode unchanged.
     this._outputTailRaw = Buffer.alloc(0)
+    // #6601: PTY output-quiescence readiness signal. `_lastOutputMs` is the
+    // monotonic time of the most recent onData chunk; `_sawFirstOutput` gates the
+    // signal until claude has actually rendered something on THIS spawn (so the
+    // `now - 0` at init doesn't read as "already quiescent"). Together (checkReady)
+    // they detect a settled composer when no ~/.claude/sessions file exists.
+    this._lastOutputMs = 0
+    this._sawFirstOutput = false
     // #5835 Phase 1: live remote-viewer mirror. PTY onData fires very frequently
     // during a TUI redraw, so coalesce bytes into a buffer and flush one
     // `terminal_output` event per tick (MIRROR_FLUSH_MS) — bounding the broadcast
@@ -1207,6 +1214,16 @@ export class ClaudeTuiSession extends BaseSession {
   // transitions are sub-second once the session is up.
   static get SPAWN_WARMUP_MAX_MS() { return 15_000 }
   static get TURN_PROMPT_WAIT_MAX_MS() { return 5_000 }
+  // #6601: how long PTY output must be quiet before the composer counts as ready
+  // when no session file resolves (current claude's INTERACTIVE TUI writes none,
+  // so the file probe never resolves and the caller used to burn the full
+  // warmup/turn ceiling then "write anyway"). The composer's render burst settles
+  // well within this window; the idle TUI then stays silent for seconds. Validated
+  // live: cold composer ready ~1.1s (vs the 15s warmup ceiling), between-turn
+  // redraw sub-second (vs the 5s per-turn ceiling). Comfortably above the observed
+  // intra-render gaps (~270ms) so it can't false-trigger mid-render; the
+  // warmup/turn ceilings still backstop a genuinely-stuck TUI.
+  static get READY_QUIESCENCE_MS() { return 400 }
   // #5317 (WP-2.3) — grace window between destroy()'s SIGTERM and the SIGKILL
   // escalation. Long enough for claude to flush its Stop hook + reap its own
   // tool children on a clean SIGTERM, short enough that a hung claude (or a
@@ -1881,6 +1898,11 @@ export class ClaudeTuiSession extends BaseSession {
     // the first spawn; this covers every subsequent _respawnPty.
     this._outputTail = ''
     this._outputTailRaw = Buffer.alloc(0)
+    // #6601: re-evaluate output-quiescence readiness for THIS spawn — require
+    // fresh output before trusting a quiet stretch, so a leftover _lastOutputMs
+    // from the prior process can't read as "ready" the instant we respawn.
+    this._sawFirstOutput = false
+    this._lastOutputMs = this._nowMonotonic()
 
     // #5794: a fresh PTY can swallow the first submit again, so re-arm the
     // first-turn submit nudge for the first message on THIS spawn. Reset after
@@ -2061,7 +2083,22 @@ export class ClaudeTuiSession extends BaseSession {
         sessFile = ClaudeTuiSession.resolveSessionFile(this._sessionId, pid, { allowDirScan })
         this._resolvedSessionFile = sessFile
       }
-      if (!sessFile) return false
+      if (!sessFile) {
+        // #6601: no session file resolved. Current claude's INTERACTIVE TUI
+        // writes no ~/.claude/sessions/<pid>.json at all, so the file probe
+        // never resolves and the caller used to burn the full warmup/turn
+        // ceiling then "write anyway" (the degraded path). Fall back to PTY
+        // OUTPUT QUIESCENCE: once claude has rendered on this spawn and output
+        // has been quiet for READY_QUIESCENCE_MS, the composer's render burst has
+        // settled and it is ready for input. Validated live (cold ready ~1.1s,
+        // between-turn redraw sub-second). The file path below is UNCHANGED for
+        // -p / claude-desktop / any version that DOES write a file; the
+        // warmup/turn ceilings still backstop a genuinely-stuck TUI, and the
+        // auth-failure scan (poll loop, above this) still runs first so a
+        // logged-out banner can't be mistaken for a settled composer.
+        return this._sawFirstOutput &&
+          (this._nowMonotonic() - this._lastOutputMs) >= ClaudeTuiSession.READY_QUIESCENCE_MS
+      }
       // A matching file was resolved → we have this session's readiness signal.
       sawStatus = true
       const status = ClaudeTuiSession.readSessionStatus(sessFile)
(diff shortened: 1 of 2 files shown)


blamechris:
## Code Review: PR #6602

### Summary
Replaces `claude-tui-session.js`'s no-session-file readiness branch (`return false`) with a **PTY output-quiescence** fallback: once claude has rendered on the current spawn (`_sawFirstOutput`) and output has been quiet for `READY_QUIESCENCE_MS` (400ms), the composer is treated as ready. This targets the fully-degraded probe on current claude (2.1.186/187), whose interactive TUI writes no `~/.claude/sessions/<pid>.json` — so the file probe never resolved and every session/turn burned the full warmup/turn ceiling before "writing anyway". The change is genuinely surgical: only the `!sessFile` branch inside `checkReady` changes; the session-file path (idle/busy/null-status, #6578 sessionId dir-scan) is byte-for-byte unchanged, and a resolved `busy` file still wins over quiescence.

### Strengths
- **Correctly closes the `now - 0` trap.** `_lastOutputMs` inits to `0`, so `now - 0` looks "quiescent" from the start — but the `_sawFirstOutput` guard (default `false`, flipped only in `_appendToOutputTail`) gates the whole expression, so a pre-render session can never read ready. The dedicated "no output yet" test pins exactly this.
- **Respawn reset is complete.** Every respawn path (`_scheduleRespawn` -> `_respawnPty` -> `_spawnPty`, and the initial `start()` -> `_spawnPty`) funnels through the one reset block (L1899-1905), which sets `_sawFirstOutput = false` + `_lastOutputMs = now` **before** `onData` is wired (L1917) — so a leftover stamp from the prior process can't read as ready the instant we respawn, and no chunk can land before the reset. There is no orphan tail-reset site (only the constructor and `_spawnPty` clear `_outputTail`, and both now clear the new fields alongside).
- **File path provably untouched.** `sawStatus`, `readSessionStatus`, and the `status === null ? true : status !== 'busy'` logic are unchanged; the quiescence branch is only reached when `!sessFile`.
- **Layered safety net.** The #5794 first-turn nudge (re-sends a bare `\r` if the first submit produces no output) still arms on every spawn before `_waitForPrompt` runs — so even if quiescence fires a hair early into a composer that swallows the submit, the nudge recovers it. Faster readiness doesn't disarm any existing wedge guard.
- **Tests are non-vacuous.** I traced the resolver: in these unit tests `_sessionId` is `null` (no `resumeSessionId`, `start()` not called), so `resolveSessionFile`'s `!sessionId` fast-path accepts the pid-named file verbatim — meaning the busy-file test's file genuinely resolves and returns via the file path (not quiescence), correctly pinning precedence. The quiescent/still-flowing/no-output tests each exercise a distinct guard.

### Issues Found

#### Critical (Must Fix)
_None._ The change is sound: no false-ready path I could construct survives the `_sawFirstOutput` guard + the poll-loop auth scan + the per-spawn reset, and the file path is untouched.

- **Auth-banner-as-quiescence on the per-turn path is defended only by timing, not structurally.** A logged-out banner is itself output — it sets `_sawFirstOutput = true` and stamps `_lastOutputMs`, then goes silent. The claimed defense ("auth scan runs first in the poll loop") holds for **warmup** (`detectAuthFailure: true`, in-loop scan at L2120 before `checkReady` at L2125, plus the post-return scan at L1956). But the **per-turn** call (`sendMessage`, L2547) passes `detectAuthFailure` = default `false`, so the quiescence branch runs with **no** auth scan ahead of it. This is benign today (you're authenticated by turn 2, and a mid-session logout is not the modeled failure), and the old behavior (`return false` -> write anyway) was arguably worse — but the PR body's blanket "the auth-failure scan still runs first" is only true for warmup. Worth a one-line comment at the per-turn call clarifying the scan is warmup-only, so a future reader doesn't lean on a guarantee that isn't there per-turn.
- **400ms leaves only ~130ms of margin over the observed ~270ms intra-render gap.** The comment cites ~270ms max intra-render gap; 400ms is comfortable for the *observed* renders but has no headroom for a slower cold start (MCP server enumeration, a slow skills-dir read, a model-catalog fetch, a loaded host). If a genuine mid-warmup pause exceeds 400ms, quiescence fires early and the throttled write lands on a not-yet-ready composer — recovered by the #5794 nudge, but a brief wedge. Since the *only* signal in the no-file case is this timer, consider making `READY_QUIESCENCE_MS` env-overridable (mirroring the other tunables) so a flaky host can widen it without a rebuild, and/or note the nudge is the backstop in the constant's comment.
- **Final-fall-through returns the new quiescence value without an auth re-scan.** After the while-loop, `const ready = checkReady()` (L2131) can now return `true` (previously always `false` for no-file). During warmup an auth banner appearing in the last <100ms before timeout would be read as ready here — the in-loop scan (every 100ms) and the post-return scan at L1956 cover all but that sub-100ms sliver, so this is negligible, but it's a real widening of what the fall-through can return. No action needed beyond awareness; flagging for completeness.

#### Nitpicks (Optional)
- The `READY_QUIESCENCE_MS` doc-comment is excellent but long; the "~270ms observed" figure is the load-bearing justification and would be easy to lose — fine as-is, just keep that number if the margin is ever revisited.
- No test asserts the **respawn reset** specifically (that a stale `_lastOutputMs` from a prior process can't read ready right after `_spawnPty`). The constructor-default case is covered by the "no output yet" test, but a targeted test that sets `_sawFirstOutput=true` + an old `_lastOutputMs`, calls the reset block's effect, and asserts not-ready would lock in the respawn guard the PR body calls out.

### Architecture Notes
The quiescence signal is a good fit for this file's design philosophy: it's decoupled from TUI glyph/text rendering (the lesson from the failed #4014/#4031/#4035 screen-scrape era) and version-independent — it keys on the *timing* of the render burst, not its content. It slots cleanly beneath the existing file-probe precedence (busy file > quiescence) and above the warmup/turn ceilings (which still backstop a genuinely stuck TUI). The two hot-path field writes in `_appendToOutputTail` are cheap and correctly placed. The interstitial caveat in the PR body is honest: quiescence can't distinguish a settled composer from a settled trust/onboarding dialog, but chroxy pre-trusts the cwd and current claude renders notices inline, so this doesn't regress the #5777 wedge — and a fragile token matcher would risk its own false-negatives. Deferring that guard is the right call.

### Verdict
**APPROVE** — This is a well-scoped, well-tested fix for a real, quantified degradation (~15s cold + ~5s/turn -> ~1.2s + ~0ms). The false-positive surface is genuinely closed at the guard I most worried about (the `now - 0` trap), the file path is untouched, respawn reset is complete, and the #5794 nudge backstops an early-fire. My suggestions are hardening/clarity, not blockers: the per-turn auth-scan gap is pre-existing-benign, and the 400ms margin is defensible for the observed workload with the nudge as a net. Ship it; consider the env-override + the per-turn-scan comment as fast-follows.

---
**Deferred-suggestion issues filed:**
- #6603 — make `READY_QUIESCENCE_MS` env-overridable + document the #5794 nudge backstop.
- #6604 — add a test pinning the per-spawn quiescence reset (stale `_lastOutputMs` must not read ready).

blamechris:
## Review addressed

**Verdict: APPROVE** (agent-review) + Copilot (overview only, no actionable comments). 0 unresolved threads.

- **Accuracy nit (auth scan "runs first"):** correct — that holds for the **warmup** `_waitForPrompt` call (`detectAuthFailure: true`). The **per-turn** call passes `detectAuthFailure: false`, so quiescence runs there with no auth scan ahead. Benign: a turn only reaches the per-turn probe after turn 1 already returned a response (you're authed), and the prior file-based per-turn check likewise ran no auth scan — so this changes nothing about mid-session logout handling. Not fixing in this PR.
- **#6604 (per-spawn reset test):** kept deferred. The reset was verified correct by code inspection; pinning it in a unit test means driving the real `_spawnPty` past its 15s warmup wait, which is awkward/flaky — tracked separately rather than forced here.
- **#6603 (env-overridable `READY_QUIESCENCE_MS` + doc):** deferred enhancement, tracked.

No code changes needed; CI stays green on the reviewed commit.

Issue #7178, opened 2026-08-15
chore: Renovate is configured but the GitHub App was never installed — 50 open advisories


blamechris:
## Description

`renovate.json` is committed at the repo root and is a real, carefully-written config — it extends `config:recommended`, sets `automerge: false`, and carries a bespoke `regexManager` for the `CLAUDE_CODE_VERSION` ARG in the sidecar Dockerfile plus a throttling `packageRule` (weekly window + 3-day `minimumReleaseAge`) to dodge yanked releases.

It has never run. The GitHub App was never installed on the repo:

```
$ gh pr list --state all --search "author:app/renovate" --limit 5
(empty — zero Renovate-authored PRs, ever)
```

So the dependency-update automation the repo believes it has does not exist. Current state:

```
$ npm audit --json | ...
{'info': 0, 'low': 4, 'moderate': 21, 'high': 22, 'critical': 3, 'total': 50}
```

**50 advisories — 3 critical, 22 high.**

The most consequential drift is the product's core dependency:

```
$ npm outdated
Package                          Current   Wanted   Latest
@anthropic-ai/claude-agent-sdk   0.2.114  0.2.141  0.3.233
@anthropic-ai/sdk                 0.81.0   0.81.0  0.117.1
```

`packages/server/package.json` declares `"@anthropic-ai/claude-agent-sdk": "^0.2.0"`, so the caret range **cannot reach 0.3.x** — the daemon is a full minor line behind upstream and no automated process will ever propose the bump.

This is a config that produces the *appearance* of dependency hygiene while providing none — the same failure shape as #7156 (a workflow that looks like a gate but is not one).

## Acceptance Criteria

- [ ] Either the Renovate GitHub App is installed and confirmed opening PRs, **or** `renovate.json` is deleted and a manual dependency-update cadence is documented
- [ ] The 3 critical advisories are triaged (fixed, or recorded as not-exploitable with reasoning)
- [ ] The `@anthropic-ai/claude-agent-sdk` 0.2 → 0.3 gap has an explicit decision (bump or documented pin), in its own PR with a full suite run — it is the daemon's core dependency
- [ ] Whichever path is chosen, the repo does not retain a config that silently does nothing

blamechris:
## Ordering note from PR #7711 — enable Dependabot alerts *before* the App goes live

[...]

PR #7191, opened 2026-08-15
chore(ci): migrate renovate.json off the deprecated manager syntax

blamechris:
## Description

Part of #7178.

`renovate.json` has never run — the GitHub App was never installed — so it has been sitting on syntax that Renovate now auto-migrates. The official validator on the current file:

```
$ npx --package renovate -- renovate-config-validator renovate.json
 WARN: Config migration necessary
```

Two renames:

| old | new |
|---|---|
| `regexManagers` | `customManagers` (+ `"customType": "regex"`) |
| `fileMatch` | `managerFilePatterns` (delimited regex) |

Renovate would still honour the old spelling, but the **first** thing it does on install is open a Config Migration PR. Landing this now means it starts on dependency work instead of housekeeping.

## Verification

Validator is clean after the change:

```
$ npx --package renovate -- renovate-config-validator renovate.json
 INFO: Config validated successfully
```

The custom manager still does its job — both halves checked against the real file rather than assumed:

```
packages/server/sidecar/Dockerfile:10:ARG CLAUDE_CODE_VERSION=2.1.128

✓ matchString captures currentValue = 2.1.128
✓ managerFilePatterns regex matches packages/server/sidecar/Dockerfile
```

## What this does not do

This does not install the app, which is the actual blocker in #7178 and needs repo-owner authorisation. It only ensures that when the app is installed, the config runs clean.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -2,10 +2,11 @@
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": ["config:recommended"],
   "automerge": false,
-  "regexManagers": [
+  "customManagers": [
     {
+      "customType": "regex",
       "description": "Track @anthropic-ai/claude-code version pinned in the sidecar Dockerfile ARG",
-      "fileMatch": ["^packages/server/sidecar/Dockerfile$"],
+      "managerFilePatterns": ["/^packages/server/sidecar/Dockerfile$/"],
       "matchStrings": [
         "ARG CLAUDE_CODE_VERSION=(?<currentValue>[^\\s]+)"
       ],


blamechris:
## Code Review: PR #7191

### Summary

Migrates `renovate.json` off deprecated Renovate config syntax: `regexManagers` → `customManagers` (+ `"customType": "regex"`), and `fileMatch` → `managerFilePatterns` as a delimited regex. Single file, +3/-2. Renovate has never run on this repo (app not installed, tracked in #7178), so this just ensures the config is clean for when it is. Verified independently rather than trusting the PR description — details below, everything reproduced with command output.

### What I verified

**1. Validator is clean on the new file; the claimed migration warning on the old file is real.**

Old file (`origin/main:renovate.json`) against `renovate@latest` (44.30.3):
```
WARN: Config migration diff:
  -   "regexManagers": [
  +   "customManagers": [
          {
  +           "customType": "regex",
              ...
  -       "fileMatch": [ "^packages/server/sidecar/Dockerfile$" ],
  +       "managerFilePatterns": [ "/^packages/server/sidecar/Dockerfile$/" ],
INFO: Config validated successfully against 1 file(s)
```
New file (this PR) against the same version:
```
INFO: Validating renovate.json as global config
INFO: Config validated successfully against 1 file(s)
```
No errors, no warnings, no migration diff. Confirms both halves of the PR's claim.

Caveat, not a blocker on this PR: running the *exact* bare command from the description (`npx --yes --package renovate -- renovate-config-validator renovate.json`, no version pin) resolved a stale cached `renovate@37.440.7` from this machine's npx cache, which predates `managerFilePatterns` and fails the new (correct) file:
```
ERROR: Found errors in configuration
  "message": "Custom Manager contains disallowed fields: managerFilePatterns"
  "message": "Invalid configuration option: customManagers[0].managerFilePatterns"
```
Pinning the version (`renovate@latest`) fixes it. Filed as #7196 rather than blocking here, since it's about how the check is reproduced, not the config itself.

**2. Delimited regex matches correctly; migration semantics preserved; no unintended paths.**

Loaded Renovate's own regex/glob matcher straight out of the installed package (`util/string-match.js`, `getRegexPredicate`/`isRegexMatch`) and ran the new pattern against the real target plus 8 near-miss candidates:
```
Pattern: /^packages/server/sidecar/Dockerfile$/
isRegexMatch: true
"packages/server/sidecar/Dockerfile"        -> true   (intended target)
"Dockerfile"                                -> false  (repo-root Dockerfile — the OTHER Dockerfile in this repo)
"packages/server/sidecar/Dockerfile.dev"    -> false
"packages/server/sidecar/subdir/Dockerfile" -> false
"xpackages/server/sidecar/Dockerfile"       -> false
"packages/server/sidecar/Dockerfile "       -> false  (trailing space)
"PACKAGES/SERVER/SIDECAR/DOCKERFILE"        -> false  (case)
"other/packages/server/sidecar/Dockerfile"  -> false
"packages/server/sidecar/dockerfile"        -> false  (case)
```
Repo has exactly two Dockerfiles (`git ls-files | grep -i dockerfile` → `Dockerfile`, `packages/server/sidecar/Dockerfile`); confirmed the pattern matches only the sidecar one.

Also checked Renovate's own auto-migration source (`config/migrations/custom/file-match-migration.js`, `FileMatchMigration.run`) — it unconditionally rewrites every `fileMatch` entry as `` `/${match}/` ``, i.e. exactly what this PR did by hand. Old `fileMatch` under regex/custom managers was always full-regex semantics (anchors and all), so wrapping in delimiters is a syntax change only — matching behavior is identical before and after.

**3. `matchStrings` still captures the version from the real file.**

`packages/server/sidecar/Dockerfile:10`:
```
ARG CLAUDE_CODE_VERSION=2.1.128
```
Ran the manager's own regex (`ARG CLAUDE_CODE_VERSION=(?<currentValue>[^\s]+)`) against the file content directly:
```
Match found: true
Captured currentValue: 2.1.128
```
Unaffected by this PR's changes (the `matchStrings` field itself is untouched), and still correct.

**4. No other deprecated syntax in the file.**

The full-file validator run above (latest Renovate) produced zero warnings on `extends`, `automerge`, `packageRules`, `schedule`, `prCreation`, or `minimumReleaseAge` — if any of those had a pending migration, the same validator that caught `regexManagers`/`fileMatch` would have caught it too.

Cross-checked directly against Renovate's option schema (`config/options/index.js`) by grepping every `deprecationMsg` in the file (~10 total across the whole schema). None touch any field this config uses — the only deprecated flags in the current schema are unrelated (`binarySource=docker`, direct `branchName`/`commitMessage`/`prTitle` editing, `hostRules.dnsCache`). `matchPackageNames`, `prCreation: "not-pending"`, and `minimumReleaseAge` are all current, non-deprecated option names/values. `config:recommended` is confirmed as a live preset (not the older `config:base`).

**5. First-install PR flood — bounded by Renovate defaults, but worth tightening given the backlog.**

Renovate's platform defaults (from the option schema): `prConcurrentLimit: 10`, `prHourlyLimit: 2`. Neither is set in this config, so it's not unbounded — but nothing here lowers it either, and `config:recommended`'s dashboard/grouping doesn't gate PR creation, it just also tracks things on an issue. Given #7178's own numbers (141 open issues, 50 advisories across 8 workspace packages), hitting the default 10-concurrent ceiling within the first several hours of app install is plausible. Not a defect in this PR — it correctly doesn't touch install policy — so filed as #7194 to decide alongside #7178 rather than blocking here.

### Issues Found

**Critical:** none.

**Suggestions:**
- Tune `prConcurrentLimit`/`prHourlyLimit` (or `dependencyDashboardApproval`) at app-install time given the existing backlog — filed as #7194.
- Add a CI check that runs `renovate-config-validator` (pinned version) against `renovate.json` on change, so this file has *some* automated guard instead of none — filed as #7196.

**Nitpicks:**
- The `EBADENGINE` warning when installing `renovate@latest` locally (wants Node `^24.11.0`, this machine has 26.3.0) is a local-verification artifact only — the hosted GitHub App runs its own infra, not this repo's Node version, so it has no bearing on actual behavior once installed.

### Verdict

**Approve.** Both renames are correct, verified against the real target file and Renovate's own parsing/migration code rather than assumed; no other deprecated syntax found; CI is green. The two suggestions above are follow-on/install-time decisions, not blockers on this PR.

Issue #2, opened 2026-02-06
Improve prompt detection for Claude Code permission dialogs


blamechris:
## Context
Claude Code presents interactive prompts (permission requests, selections) that need to be surfaced in the chat view as tappable options. The infrastructure exists (`prompt` message type with `options` array, amber "Action Required" cards with buttons) but the detection is unreliable.

## Problems
1. ANSI stripping loses spacing — "Yes, I trust this folder" becomes "Yes,Itrustthisfolder"
2. Numbered options (1. Yes / 2. No) may not be detected if formatting varies
3. Permission prompts (Allow/Deny/Always allow) need testing with real Claude Code output
4. The prompt accumulation timer (500ms) may be too short or too long

## Current Implementation
- `_detectPrompt()` in `output-parser.js` looks for `^\d+\.\s*(.+)` and permission keywords
- `_resetPromptFlush()` accumulates options over 500ms then emits
- Startup prompts (trust dialog) are suppressed — server auto-handles them
- Only prompts after `claudeReady` are forwarded to clients

## Next Steps
- Test with real permission prompts (file write, bash execution, etc.)
- Consider a more robust detection approach (maybe pattern match the full prompt block)
- Handle edge cases: multi-line options, prompts with descriptions

Issue #1, opened 2026-02-06
System message styling: distinguish alerts from Claude responses


blamechris:
## Status

**Partially addressed:** Error message styling was added in PR #22 (red-tinted bubble, red border, "Error" sender label). What remains is system message support.

## Remaining Work

System messages from Claude Code (usage limits, sub-agent info, background tasks) appear as regular "Claude" response bubbles. Need:

- New message type `system` with distinct styling (muted gray/neutral, not green/Claude or blue/You)
- Usage limit warnings, sub-agent spawn/completion notifications
- Differentiate system info from Claude's actual responses

## Current State

The output parser emits all non-prompt, non-tool content as `response` type. The app's `MessageBubble` now handles: `response`, `user_input`, `tool_use`, `thinking`, `prompt`, `error`. System messages would be a new type.

## Files

- `packages/server/src/cli-session.js` — emit `system` message type
- `packages/app/src/store/connection.ts` — add `system` to ChatMessage type
- `packages/app/src/screens/SessionScreen.tsx` — system bubble styling

Issue #7185, opened 2026-08-15
app: diff-comment E2E — submit-comments bar never appears after saving an inline comment


blamechris:
## Description

`packages/app/.maestro/diff-comment.yaml` fails deterministically in the Maestro nightly (both attempts of the runner's retry), and has since the flow landed with #6945 on 2026-07-23. It is one of the two flows keeping the nightly red (#7113).

The flow gets all the way through composing a comment and saving it, then the submit bar never appears:

```
Assert that id: diff-comment-input is visible...   COMPLETED
Tap on id: diff-comment-input...                   COMPLETED
Tap on id: diff-comment-save...                    COMPLETED
Assert that id: diff-submit-comments is visible... FAILED
Assertion is false: id: diff-submit-comments is visible
```

Run: https://github.com/blamechris/chroxy/actions/runs/31876250329

## Not testID drift

`diff-submit-comments` exists in the component and is asserted by passing jest tests:

```
packages/app/src/components/DiffViewer.tsx:514:  testID="diff-submit-comments"
packages/app/src/components/__tests__/DiffViewer.test.tsx:117
packages/app/src/components/__tests__/DiffViewer.test.tsx:174
```

So the element renders under jest but does not reach the screen in the running app after `diff-comment-save` is tapped. The likely gap is between the save handler and the pending-comments state that gates the submit bar — jest drives the component directly, the flow drives it through the real app.

## Acceptance Criteria

- [ ] Root cause identified: whether `diff-comment-save` fails to commit the comment, or the submit bar's visibility condition is wrong
- [ ] `diff-comment.yaml` passes on the nightly runner
- [ ] If the jest coverage passed while the real path was broken, add coverage at whatever layer would have caught it

Issue #10, opened 2026-02-06
Desktop/mobile session handoff


blamechris:
## Summary
Allow users to seamlessly swap between desktop and mobile for the same Claude Code session. Since the session lives in tmux, the underlying terminal is persistent — we just need to handle multiple clients viewing/controlling the same session.

## Current Behavior
- Only one client connects at a time via WebSocket
- Reconnecting from a different device works but there's no handoff UX

## Desired Behavior
- User can open Chroxy on desktop browser OR mobile app
- Both can view the session (read-only for one, or full duplex)
- Clean handoff: "Another device connected" notification
- Session state (chat history, terminal buffer) syncs on connect

## Technical Considerations
- tmux already supports multiple clients (`tmux attach`)
- WsServer currently supports multiple connections — need to decide on control semantics
- May need a lightweight web client for desktop access
- Related to #3 (multi-tab support)

## Related
- Depends on better output parsing or a structured API from Claude Code (see investigation notes)

PR #7204, opened 2026-08-16
fix(app): keep taps alive over the diff comment editor's scroll views

### LINKED ISSUE #7203: app: DiffViewer comment editor swallows the first tap after typing (keyboardShouldPersistTaps default)
## Context

Surfaced during review of PR #7202 (which works around this at the Maestro test layer for #7185).

The inline comment editor's Save/Cancel buttons (`diff-comment-save` / `diff-comment-cancel`, `packages/app/src/components/DiffViewer.tsx:211-221`) sit inside two nested `ScrollView`s:

```
packages/app/src/components/DiffViewer.tsx:257  <ScrollView style={styles.diffScroll} ...>
packages/app/src/components/DiffViewer.tsx:258    <ScrollView horizontal ...>
```

Neither sets `keyboardShouldPersistTaps`. React Native's default is `'never'`, documented as: "Tapping outside of the focused text input when the keyboard is up dismisses the keyboard. When this happens, children won't receive the tap." (https://reactnative.dev/docs/scrollview#keyboardshouldpersisttaps)

That matches the symptom PR #7202 root-caused on-device for the Maestro flow: after typing into `diff-comment-input` (which autofocuses and raises the keyboard), the first tap on `diff-comment-save` is consumed by the ScrollView's keyboard-dismiss handling rather than reaching the button's `onPress`. A second tap in the same spot then lands normally, because the keyboard is already down.

This isn't only a test-harness artifact — a real user tapping "Add comment" once, right after typing, would hit the same swallowed tap and see nothing happen.

## Suggested fix

Set `keyboardShouldPersistTaps="handled"` (or `"always"`) on the relevant `ScrollView`(s) in `DiffViewer.tsx` so the tap reaches `diff-comment-save`/`diff-comment-cancel` on the first try, keyboard up or not.

## Acceptance criteria

- [ ] First tap on `diff-comment-save` (or `diff-comment-cancel`) reaches the button while the keyboard is still up, no second tap needed
- [ ] No regression to normal scroll-to-dismiss behavior elsewhere in the diff viewer
- [ ] Once fixed, consider whether `packages/app/.maestro/diff-comment.yaml`'s retry loop (PR #7202) can be simplified back to a single tap — low priority, the retry is harmless to keep either way

blamechris:
## Description

Closes #7203.

The inline comment editor's Save / Cancel buttons sit inside two nested `ScrollView`s (`DiffViewer.tsx:257-258`), neither of which sets `keyboardShouldPersistTaps`. React Native's default is `'never'`, [documented](https://reactnative.dev/docs/scrollview#keyboardshouldpersisttaps) as: tapping outside the focused text input while the keyboard is up dismisses the keyboard, and **the children do not receive that tap**.

The comment input autofocuses, so a user who types and taps "Add comment" **once** has that tap eaten and sees nothing happen. A second tap works. That is what #7185's flow had been reporting for a month, and #7202 worked around it at the test layer.

## Change

`keyboardShouldPersistTaps="handled"` on **both** scroll views. The prop is per-`ScrollView`, so setting only the outer one leaves the inner one swallowing the tap. `"handled"` rather than `"always"` keeps tap-to-dismiss working everywhere else in the diff.

This matches the established idiom — `ConnectScreen.tsx:480`, `InputBar.tsx:287/352/397`, and `CreateSessionModal.tsx:219` all already use `"handled"`.

## ⚠️ I could not demonstrate this on-device — read before merging

The swallowed tap **did** reproduce reliably earlier in this session (single tap failed; a second tap on the same state saved the comment). It no longer does, and the reason is not that the bug went away:

```
$ maestro --device 0AD31F77… hierarchy   # editor open, input focused
  Keyboard  : 0
  key-      : 0
  shift     : 0
  space     : 0
```

**There is no software keyboard on screen.** The simulator is running with a hardware keyboard attached, so the keyboard-dismiss path this prop governs is never taken.

A clean A/B is therefore inconclusive — with a proper 25s settle between edits, both arms pass twice:

| | run 1 | run 2 |
|---|---|---|
| **without** the fix | 0 failures | 0 failures |
| **with** the fix | 0 failures | 0 failures |

Neither arm exercises the code path, so that table proves nothing about the fix. I am reporting it rather than quoting the passing runs as verification.

**What this rests on instead:** the RN documentation, the five existing uses of `"handled"` in this repo, and the earlier live reproduction of the exact symptom. Real users always have a software keyboard, so the path they hit is precisely the one this simulator cannot exercise.

**Recommendation:** keep #7202's retry loop until this is confirmed on real hardware or on a simulator with the software keyboard forced on (Cmd+K / `ConnectHardwareKeyboard = false`). The retry is harmless if this fix is correct, and still necessary if it is not.

## Tests

App suite: **2328 pass**. No behavioural unit coverage is added — the prop's effect only manifests in the RN gesture responder with a live keyboard, which is exactly what could not be reproduced here.

### DIFF
--- a/packages/app/src/components/DiffViewer.tsx
+++ b/packages/app/src/components/DiffViewer.tsx
@@ -254,8 +254,21 @@ function FileDiffView({
           </Text>
         </View>
       </View>
-      <ScrollView style={styles.diffScroll} contentContainerStyle={styles.diffScrollContent}>
-        <ScrollView horizontal showsHorizontalScrollIndicator>
+      {/* keyboardShouldPersistTaps on BOTH: the inline comment editor's Save /
+          Cancel buttons live inside these scroll views, and the comment input
+          autofocuses — so with RN's default of 'never', the first tap after
+          typing is consumed by the keyboard dismiss and never reaches the
+          button. The user taps "Add comment", nothing happens, and only a
+          second tap works (#7203; surfaced by the Maestro flow in #7185).
+          The prop is per-ScrollView, so setting it only on the outer one
+          leaves the inner one still swallowing the tap. 'handled' rather than
+          'always' keeps tap-to-dismiss working everywhere else in the diff. */}
+      <ScrollView
+        style={styles.diffScroll}
+        contentContainerStyle={styles.diffScrollContent}
+        keyboardShouldPersistTaps="handled"
+      >
+        <ScrollView horizontal showsHorizontalScrollIndicator keyboardShouldPersistTaps="handled">
           <View>
             {file.hunks.map((hunk, i) => (
               <DiffHunkView key={i} hunk={hunk} filePath={file.path} hunkIndex={i} commentApi={commentApi} />


blamechris:
## Code Review: PR #7204

### Summary

The fix is correct, minimal, and I was able to produce the on-device A/B the PR itself said it couldn't get: reverted the prop in a backed-up worktree copy, ran a single-tap test twice per arm, and got a clean result both times — **without** the fix the tap is swallowed (editor stays open, draft preserved, nothing saved); **with** the fix the tap succeeds (editor closes, comment saved, submit bar appears). Both nested `ScrollView`s are genuinely required — fixing only the outer one still swallows the tap.

The one thing I'd flag before merge isn't in the code: the PR's stated reason its own A/B was inconclusive ("There is no software keyboard on screen") does not hold up under my testing. I dumped the hierarchy on the same simulator UDID, via the same Maestro-driven method, four separate times — every time, a full, real, non-zero-bounds software keyboard was present. That doesn't call the fix into question (if anything, it's what let me actually prove the fix works), but the PR's verification section currently states something as fact that isn't reliably true, and I'd rather that get corrected than propagate.

### What I verified

**1. The "no software keyboard" claim — refuted, with the same tooling the PR used.**

```
$ maestro --device 0AD31F77-82DF-4E28-99E6-CAE0A8BB5E54 hierarchy   # editor open, input focused
  ... 26 letter keys, "shift", "space", "delete", "return", "Emoji", "Dictate", "Typing Predictions" ...
```
Bounds check confirmed these aren't phantom/off-screen nodes:
```
space -> [100,752][301,798] enabled=true      # screen is 402x874pt; keyboard occupies the bottom ~336pt
```
Screenshot at that exact moment shows the same thing visually (keyboard fully rendered, "Add comment"/"Cancel" sitting above it). I repeated this — fresh `clearState: true` app launch each time — 4 separate times over ~15 minutes (once standalone, then once as the pre-tap check in each of 3 A/B trials below). All 4 showed a real keyboard. I never touched `ConnectHardwareKeyboard` or any other simulator setting to get this — `defaults find ConnectHardwareKeyboard` returns "no domain, key, nor value" anywhere on the system, so nothing was forced. I can't explain why the author's session differed; I'm reporting the discrepancy, not a diagnosed root cause for it.

**2. A real, repeated A/B — the main deliverable of this review.**

Metro on :8081 turned out to be bound to `~/Projects/chroxy` (the pristine main checkout, off-limits to edit per my instructions), not this worktree, so it was serving pre-fix code regardless of what I did here. I started a second Expo/Metro instance rooted at this worktree's `packages/app` on port 8082 so I had a server actually reflecting my edits, and pointed the dev client at it via the deep-link URL.

Backed up the fixed file, then edited the worktree copy to remove both `keyboardShouldPersistTaps` props — confirmed byte-identical to pristine `main`'s `DiffViewer.tsx` via `diff` (exit 0) before testing. Ran 2 trials per arm plus 1 outer-only-fix trial; for 3 of the 5 I split setup and tap into separate flows so I could dump the hierarchy *immediately before* the single tap, not just after:

| Trial | Pre-tap keyboard (directly dumped) | Post single-tap | Result |
|---|---|---|---|
| without-fix #1 | not separately captured (combined flow) | `diff-comment-editor-2` still mounted, draft text unchanged, no `diff-submit-comments` | **swallowed** |
| without-fix #2 | present — 4 key rows confirmed | same as above | **swallowed** |
| with-fix #1 | not separately captured (combined flow) | `diff-comment-editor-2` gone, `diff-comment-note-2` + `diff-submit-comments` rendered | **saved** |
| with-fix #2 | present — 4 key rows confirmed | same as above | **saved** |
| outer-ScrollView-only (inner reverted) | present — 4 key rows confirmed | `diff-comment-editor-2` still mounted, draft unchanged | **swallowed** |

For the two "not separately captured" rows: without-fix #1's swallowed result is itself indirect evidence the keyboard was up at tap-time — per the source in finding #4 below, the capture-phase interception this bug depends on requires `_keyboardIsDismissible()` to be true, so a swallowed tap couldn't happen with no keyboard up. With-fix #1's success doesn't carry the same inference (it would succeed either way), so I don't count it as keyboard evidence — only as a fix-works data point, corroborated by with-fix #2's directly-confirmed keyboard-up trial.

Every state transition was captured via `maestro hierarchy` (parsed for the relevant `resource-id`s) plus an `xcrun simctl io screenshot`. Restored the fix afterward and confirmed `diff --stat` / `git status --porcelain` clean against the PR's actual commit before finishing.

**3. Per-ScrollView, not inherited — confirmed from this repo's actual RN 0.81.5 source, not just docs**, and confirmed empirically by the outer-only trial above:
```
node_modules/react-native/Libraries/Components/ScrollView/ScrollView.js:1444
  this.props.keyboardShouldPersistTaps === 'handled' && ...
```
Read directly off `this.props` inside each `ScrollView` instance's own responder handlers (`_handleStartShouldSetResponderCapture`, `_handleResponderRelease`, `_handleTouchEnd`) — no context, no propagation from a parent. The PR's claim is accurate.

**4. Why the keyboard-dependence exists at all, at the source level** — `_keyboardIsDismissible()` (`ScrollView.js:1517-1536`) gates the entire tap-eating mechanism on `this._keyboardMetrics != null`, which is only populated inside the native `keyboardWillShow`/`keyboardDidShow` handlers. If no software-keyboard-shown event ever fires, this PR's whole fix is provably inert regardless of the prop's value — which is exactly the mechanism the PR describes, even though (per #1 above) the specific "no keyboard" premise didn't hold in my testing.

**5. `"handled"` vs `"always"`, for the Save/Cancel buttons specifically**: functionally identical for the buttons themselves (neither value lets the capture-phase intercept a tap on a Touchable child). They differ on a **blank-area** tap: `_handleResponderRelease`'s keyboard-dismiss guard explicitly excludes `'always'`/`true` but not `'handled'` (`ScrollView.js:1368-1369`), so `'always'` would have silently broken "tap empty space in the diff to dismiss the keyboard" everywhere else in the viewer. `"handled"` is the correct choice.

**6. Precedent citations — all 5 verified exact**, plus 2 more the PR didn't cite (`ChatView.tsx:592`, `SessionOverview.tsx:318`) — the existing-idiom argument is if anything stronger than claimed.

**7. Existing `DiffViewer.test.tsx` (210 lines, 8 `it()` blocks)** has no ScrollView/prop assertions and drives interactions via RTL `fireEvent.press`, which calls the handler directly and never touches the native gesture-responder system this bug lives in — confirms unit tests structurally can't cover this, not just that none happen to.

**8. `PreWriteDiffReview.tsx`** (the "read-only viewers" the doc-comment references) really is unaffected: it calls `DiffHunkView` without `commentApi`, so no `TextInput` ever mounts there, and the component has no `ScrollView`/`FlatList` wrapper at all.

**9. Full app jest suite**: `180 passed, 180 total` / `2328 passed, 2328 total` — matches the PR's claim exactly. `npx tsc --noEmit` also clean.

**10. Regression check — a real (pre-existing) risk this fix makes easier to hit.** `handleOpenComment` (`DiffViewer.tsx:364-371`) unconditionally discards `draft` and switches `editing` with no confirmation whenever any other diff-line/comment-note is tapped. Verified live: typed an unsaved draft on line 2, tapped line 3 once with the keyboard still up — the line-2 draft vanished silently, editor jumped straight to line 3's empty state. Under the old default this took 2 taps on the other line (first swallowed as a keyboard-dismiss, second triggers the switch); this fix makes it 1. The underlying no-confirm-on-switch design predates this PR, so it's not this PR's bug to fix, but it's now more reachable by accident.

**11. Maestro retry loop (`diff-comment.yaml`'s `repeat: times: 3`) — agree it should stay**, and for a slightly different reason than the PR gives. The PR frames it as a hedge "until this is confirmed" — my A/B above is that confirmation. But my testing also directly demonstrated that keyboard-visibility on this exact Simulator/Maestro combination is not stable across otherwise-identical runs (same UDID, same driving method, author saw none / I saw one consistently). That instability is about the test environment, not this fix, and it's exactly the kind of thing the retry loop protects against regardless of which way the product code is proven. Keep it.

### What I could NOT verify

- **Why** the author's session showed no keyboard while mine consistently did. I have no access to their session state, and I didn't need to change anything to get a real keyboard in mine — so I can't reconcile the two beyond reporting both observations.
- The "switching editors loses 2 taps' worth of draft under the old code too" half of finding #10 is reasoned from the same RN source citations (`_keyboardIsDismissible()` requires focus + metrics; blur happens on the swallowed tap) plus the exact `handleOpenComment` code, not separately reproduced on pre-fix code on-device — I judged the primary A/B a better use of the remaining device cycles.
- Android and physical hardware — simulator/iOS only, same as the PR itself.

### Issues Found

**Critical**
1. The PR description's stated diagnosis ("There is no software keyboard on screen [on this simulator]") is not reliably true — I refuted it 4/4 times with the same method the PR used. This does **not** block merge and does not implicate the code change (my own A/B, run because I found a real keyboard, is the strongest evidence yet that the fix is correct). It's a correction to the PR's evidence trail: I'd suggest editing the PR description (or leaving a follow-up comment) to point at this review's A/B instead of the inconclusive one, so anyone reading the merged history later isn't working from a false premise about Simulator keyboard behavior.

**Suggestions**
2. `handleOpenComment`'s no-confirmation draft-discard-on-switch (finding #10) predates this PR but is now 1 tap away from firing instead of 2. Worth a low-priority follow-up (confirm before switching, or persist the draft per-line) — out of scope for this PR.
3. No CI-enforced regression guard for either `keyboardShouldPersistTaps` prop going forward. `ChatView.virtualization.test.tsx:180-183` already has the pattern this needs — a source-regex assertion, because RTL can't exercise the real behavior:
   ```js
   expect(source).toMatch(/keyboardShouldPersistTaps=["']handled["']/);
   ```
   Two lines like that (one per `ScrollView`, or a `toMatch(/.../g)` count) in the existing `DiffViewer.test.tsx` would catch a future accidental removal in CI, which nothing currently does.
4. Filed as separate issues (same bug class as #7203, different files — confirmed via grep + JSX-nesting, **not** independently reproduced live the way DiffViewer.tsx was):
   - #7206 — `GitView.tsx`'s commit-message `TextInput` + "Commit" button share a `FlatList` (`GitView.tsx:534-540`) with zero `keyboardShouldPersistTaps` occurrences in the file.
   - #7207 — Settings' `QuietHoursEditor` time inputs + "Save Quiet Hours" button are mounted inside `SettingsScreen.tsx`'s single `ScrollView` (`SettingsScreen.tsx:303-667`), also zero occurrences.

**Nitpicks**

None of substance. The 9-line in-JSX doc comment (`DiffViewer.tsx:257-265`) is long for a comment but earns its length given how expensive this exact failure mode was to root-cause — I would not ask for it to be trimmed.

### Verdict

**Approve.** The code change is correct — verified with a real, repeated, on-device A/B rather than inferred from docs — and both `ScrollView`s are genuinely necessary. Nothing in this review implicates the diff itself; the one Critical item is about tightening the PR's own evidence trail, not about the fix. Recommend keeping the Maestro retry loop per reasoning above, and optionally tightening the PR description's verification section to reference this review's A/B.

blamechris:
## You refuted the central claim, and you were right — corrected in `9012b34b3`

This is the outcome I asked for and did not expect. My "there is no software keyboard on this simulator" conclusion was **wrong**, and the probe that produced it was invalid:

```
$ maestro --device 0AD31F77… hierarchy | grep -ci "Keyboard|key-|shift|space"
  0
```

I was grepping the **app's** accessibility hierarchy for keyboard node names. The system keyboard is not part of the app's hierarchy, so that search could only ever return zero — it would have "proved" no keyboard exists on any device, in any state. A negative result from a probe that cannot produce a positive is worthless, and I treated it as evidence.

Your dump found a full 26-key keyboard occupying ~336pt of an 874pt screen, four times over, with no simulator settings changed.

### What that changes

Everything, in the right direction. With a real keyboard-up A/B:

| arm | result |
|---|---|
| without the fix | tap swallowed, editor stays open, draft preserved (2/2) |
| with the fix | tap lands, comment saved (2/2) |
| **outer ScrollView only** | **still swallowed** (1/1) |

That third row is the one I could not have produced by reasoning, and it is the empirical answer to "are both really needed" — yes, and now the commit says so on evidence rather than on source-reading.

### The commit message is amended, not just this comment

This repo squashes with `squash_merge_commit_message: COMMIT_MESSAGES`, so the commit body lands in `main`'s permanent history. My incorrect "could not verify" note would have been enshrined there. `9012b34b3` replaces it with the real A/B results, and keeps a short record of the bad inference — because that same invalid probe would mislead the next person identically.

### On your suggestions

- **#7206 / #7207** — agreed, same root cause in `GitView` and `QuietHoursEditor`. Correctly filed as unverified-live rather than asserted; they deserve the same on-device A/B this one now has before anyone calls them fixed.
- **Stray-tap draft loss** — real, and this fix does make it easier to reach. Worth its own issue rather than expanding this PR.
- **CI regression coverage for the prop** — the concrete suggestion mirroring `ChatView.virtualization.test.tsx` is the right shape; queued as follow-up work.

Also noted your disclosure about the mistaken stray agent call and that you disregarded its output. Flagging that rather than quietly dropping it is the right instinct.

Issue #7183, opened 2026-08-15
bump-version.sh doesn't update CLAUDE.md/AGENTS.md version references


blamechris:
## Problem

`CLAUDE.md`'s Project Overview has a `**Current Status (vX.Y.Z):**` line (currently line 28) and a footer `*Version: X.Y.Z*` (currently line 462) that state the current release version in prose. `AGENTS.md` carries the same two lines (generated from `CLAUDE.md` per `node scripts/gen-agents-md.mjs`).

`scripts/bump-version.sh` never touches either file — confirmed by reading the full script (`ROOT/CLAUDE.md` / `ROOT/AGENTS.md` don't appear anywhere in it) and by grepping both files after PR #7180's 0.10.0 → 0.11.0 bump: both still read `**Current Status (v0.10.0):**`, and the footer on both still reads `*Version: 0.9.47*` — already two releases stale even before that bump, which shows this has been drifting silently across multiple releases, not just this one.

## Why it matters

These lines are the first thing a session (human or agent) reads about "what version is this" at the top of the project's own instructions file, and they're wrong immediately after every release cut.

## Suggested fix

Either:
1. Teach `scripts/bump-version.sh` to also rewrite the `**Current Status (vX.Y.Z):**` line and the `*Version: X.Y.Z*` footer line in `CLAUDE.md` (same regex-anchored, verify-after-write pattern the script already uses for `Cargo.toml`/`Cargo.lock`), then run `node scripts/gen-agents-md.mjs` in the same step so `AGENTS.md` stays in sync per the existing convention (see CLAUDE.md's "Editing CLAUDE.md → regen AGENTS.md" note), or
2. Add a lightweight CI/lint check (mirroring `packages/server/scripts/lint-*.sh`) that fails when `CLAUDE.md`'s footer version doesn't match `packages/server/package.json`'s version, so drift is caught even if the prose-editing step is a manual part of the release checklist rather than automated.

Option 1 is preferred since it closes the gap the same way every other version-bearing file in the repo is already handled.

## Scope

Docs-only; no behavior change. Filed from review of PR #7180 (release-cut PR), which surfaced the drift but intentionally left it out of scope (pure `scripts/bump-version.sh` run, no hand-edited prose).

Issue #7197, opened 2026-08-15
build-publish-dir.mjs self-check imports hardcoded paths, not the manifest it just wrote


blamechris:
## Context

Found during adversarial review of #7192 (the `verify-publish-artifacts.mjs` pre-publish gate).

`packages/store-core/scripts/build-publish-dir.mjs` assembles `@chroxy/store-core`'s publishable manifest and, at the end, self-checks both entry points before finishing:

```js
for (const [label, entry] of [['.', 'dist/index.js'], ['./crypto', 'dist/crypto.js']]) {
  try {
    await import(new URL(`publish/${entry}`, `file://${pkgDir}/`).href)
    console.log(`  ✓ "${label}" entry imports under Node`)
  } catch (err) {
    throw new Error(`published "${label}" entry (${entry}) fails to import: ${err.message}`)
  }
}
```

The `entry` values (`'dist/index.js'`, `'dist/crypto.js'`) are a **hardcoded literal list**, not read back from the `exports` object the script just wrote into `publish/package.json` a few lines earlier (`out.exports['.'].import`, `out.exports['./crypto'].import`). So if the generated manifest's `exports` field ever points somewhere wrong — a typo, a stale path after a refactor — this self-check still imports the *correct* file directly by its known-good literal path and prints `✓`, while the actual npm consumer (resolving through `package.json#exports`) would get `ERR_MODULE_NOT_FOUND`.

## Reproduction

Verified by mutating the manifest-building line in `build-publish-dir.mjs` (temporary, reverted after testing — not committed):

```diff
- '.': { types: './dist/index.d.ts', import: './dist/index.js' },
+ '.': { types: './dist/index.d.ts', import: './dist/index-typo.js' },
```

Result:
```
  ✓ "." entry imports under Node          <- self-check still passes
  ✓ "./crypto" entry imports under Node
built @chroxy/store-core@0.11.0 -> .../packages/store-core/publish
```
`build:publish` completes with both checkmarks despite the manifest pointing `.` at a nonexistent file.

`scripts/verify-publish-artifacts.mjs` (new in #7192) *does* independently catch this downstream, because its entry-point check imports via the package specifier (`import("@chroxy/store-core")`) through a real `npm install` in a from-scratch consumer project, which resolves through the (broken) `exports` field rather than a hardcoded path:

```
FAIL  import("@chroxy/store-core") threw: Error [ERR_MODULE_NOT_FOUND]: Cannot find module
[...]

PR #7208, opened 2026-08-16
fix(ci): realpath both sides of gen-agents-md's entry-point guard

### LINKED ISSUE #7198: gen-agents-md.mjs's CLI-guard silently no-ops under symlink-traversing paths (macOS /tmp, /var)
## Problem

`scripts/gen-agents-md.mjs`'s top-level "am I being run directly" guard is:

```js
if (process.argv[1] && resolve(fileURLToPath(import.meta.url)) === resolve(process.argv[1])) {
  // ... writeFileSync(AGENTS_PATH, expected) or --check ...
}
```

Node's ESM loader realpaths the entry module's `import.meta.url` (resolving symlinks), but `process.argv[1]` stays exactly as passed on the command line, unresolved. On macOS, `/tmp` and `/var` are symlinks to `/private/tmp` / `/private/var`, so invoking the script via an absolute path constructed through either prefix makes the two sides of that comparison diverge — the guard evaluates `false`, the whole body (write **and** `--check`) is skipped, and the process exits `0` having done nothing.

Reproduced directly:

```
$ node -e "
  import('node:path').then(({resolve}) => import('node:url').then(({fileURLToPath}) => {
    console.log('argv[1]           =', process.argv[1]);
    console.log('import.meta.url   =', import.meta.url); // n/a in -e, see below
  }))
"
```
(full isolated repro, run against a copy of gen-agents-md.mjs invoked as `node "$ROOT/scripts/gen-agents-md.mjs"` from a `mktemp -d` directory, i.e. under `/var/folders/...`):

```
process.argv[1]                         = /var/folders/6c/.../T/tmp.IpABlEfhIO/scripts/probe.mjs
fileURLToPath(import.meta.url)          = /private/var/folders/6c/.../T/tmp.IpABlEfhIO/scripts/probe.mjs
resolve(process.argv[1])                = /var/folders/6c/.../T/tmp.IpABlEfhIO/scripts/probe.mjs
resolve(fileURLToPath(import.meta.url)) = /private/var/folders/6c/.../T/tmp.IpABlEfhIO/scripts/probe.mjs
EQUAL?                                  = false
```

Confirmed against the real `gen-agents-md.mjs`: `node "$ROOT/scripts/gen-agents-md.mjs"` (the exact invocation `bump-version.sh` uses) exits `0` and writes nothing when `$ROOT` is a `/var/folders/...`-rooted temp dir; the identical invocation under `/private/tmp/...` (already-resolved) works correctly and prints "Wrote AGENTS.md from CLAUDE.md."

## Why it matters

`scripts/bump-version.sh` (#7183 / the CLAUDE.md-version-refs PR) added:

```bash
if [ -f "$ROOT/scripts/gen-agents-md.mjs" ]; then
  node "$ROOT/scripts/gen-agents-md.mjs" >/dev/null
  echo "  AGENTS.md regenerated from CLAUDE.md"
fi
```

This trusts node's exit code alone — no post-write content check — unlike every other write in that script (Cargo.toml, the CLAUDE.md rewrite itself, both lockfiles all verify what actually landed on disk). Combined with the guard bug above: if `bump-version.sh` is ever run from a checkout/worktree rooted under an unresolved `/tmp` or `/var` path on macOS (plausible for a throwaway scratch worktree), it prints the reassuring "AGENTS.md regenerated from CLAUDE.md" line and exits 0 while AGENTS.md is left completely untouched.

This is **not** a silent-merge risk in practice: the CI "Scripts Tests" job's drift gate (`node scripts/__tests__/gen-agents-md.test.mjs`) imports `renderAgentsMd`/`readClaudeMd`/`readAgentsMd` directly rather than subprocess-invoking the CLI guard, so it's immune to this specific bug and still fails the build on a stale AGENTS.md. That job also runs on the self-hosted **Linux** pool for same-repo events (`runner-target` routing in `.github/workflows/ci.yml`), where `/tmp` isn't a symlink, so CI re-running the same broken step wouldn't itself reproduce this. The exposure is local/dev-time: a confusing false-success message that wastes time chasing, on a class of platform-specific bug this repo has been bitten by before (see `bash32_perf_cliff_and_platform_blind_ci.md` / `set_u_macos_linux_divergence.md` in project memory — Linux CI can't see it).

Also worth noting: this exact call path has **zero test coverage** either way. `scripts/__tests__/bump-version.test.sh`'s `build_fake_repo` fixture never creates `scripts/gen-agents-md.mjs` in the fake tree, so the `if [ -f "$ROOT/scripts/gen-agents-md.mjs" ]` branch is never exercised by any existing test — neither the success path nor a failure path.

## Suggested fix

Either (or both):
1. Make the CLI-invocation guard in `gen-agents-md.mjs` robust to symlink-traversing paths — e.g. don't compare a realpath'd side against a non-realpath'd side (realpath both, or compare basenames + a weaker check, or use whatever direct-execution idiom is safest for the Node version this repo targets).
2. Have `bump-version.sh`'s new call site verify AGENTS.md's content after regeneration (mirroring the verify-after-write pattern already used for `Cargo.toml`/`CLAUDE.md` in the same script), so a no-op can't report success.
3. Add fixture coverage in `scripts/__tests__/bump-version.test.sh` for the AGENTS.md-regeneration call site (both a working `gen-agents-md.mjs` and a failing one), since it's currently entirely untested.

## Context

Found during adversarial review of PR #7195 (`chore(docs): bump CLAUDE.md's version references with the release`, closes #7183), which added the `bump-version.sh` call site above. `gen-agents-md.mjs` itself is unmodified by that PR — this is a pre-existing latent gap in the generator that the new call site newly exposes to the release-bump path.

blamechris:
## Description

Closes #7198.

`gen-agents-md.mjs` guards its CLI side-effects with:

```js
if (process.argv[1] && resolve(fileURLToPath(import.meta.url)) === resolve(process.argv[1])) {
```

Node's ESM loader **resolves symlinks** in `import.meta.url`, but leaves `process.argv[1]` exactly as the caller typed it — and `resolve()` does not follow symlinks. On macOS `/tmp` is a symlink to `/private/tmp`:

```
$ node -e "console.log(resolve('/tmp/x.mjs'))"
/tmp/x.mjs          # vs '/private/tmp/x.mjs' from import.meta.url  -> not equal
```

So the guard was false, the CLI branch never ran, and the script **exited 0 having done nothing.**

## Reproduced before fixing

Not inferred from the source — run against a real copy through `/tmp`, with `AGENTS.md` deliberately drifted:

```
$ node /tmp/garepro/scripts/gen-agents-md.mjs
  exit=0
  still drifted? True     <- silently regenerated nothing
```

## Why the silence matters

`bump-version.sh` calls this and trusts the exit code, so **a release could skip the AGENTS.md regeneration and still report success**. And `--check` — the CI drift gate — would pass on a stale file for exactly the same reason, since the branch that performs the check never executes.

That is the same shape as the other guards-that-quietly-no-op addressed this cycle (#7183's skipped pattern, #7197's hardcoded self-check paths): the check reports success by not running.

## Fix

`realpathSync` both sides, wrapped because `argv[1]` need not exist (an `-e` eval, a deleted script) — an unresolvable path simply is not this file.

The import path is deliberately unaffected: the guard must stay false when the module is imported by the test harness, and that is asserted (the suite imports it and the existing three assertions still pass).

## Verification

| scenario | before | after |
|---|---|---|
| run via symlinked `/tmp` path | exit 0, no regeneration | regenerates ✓ |
| `--check` via symlinked path, drifted | would pass silently | **exit 1** ✓ |
| run via real `/private` path | works | works ✓ |
| imported by tests | guard false | guard false ✓ |

Regression test added and **mutation-checked** — restoring the `resolve()`-only guard fails exactly that test and nothing else:

```
  ok   render includes the entire CLAUDE.md body unmodified
  ok   render prepends the auto-generated / do-not-edit header
  ok   committed AGENTS.md is in sync with CLAUDE.md (drift gate)
  FAIL runs through a symlinked invocation path (#7198): generator silently no-opped
  3 passed, 1 failed
```

Scripts suite: **18 pass**.

### DIFF
--- a/scripts/__tests__/gen-agents-md.test.mjs
+++ b/scripts/__tests__/gen-agents-md.test.mjs
@@ -11,7 +11,10 @@
  *   node scripts/__tests__/gen-agents-md.test.mjs
  */
 
-import { dirname, resolve } from 'node:path'
+import { dirname, join, resolve } from 'node:path'
+import { cpSync, mkdirSync, mkdtempSync, readFileSync, rmSync, symlinkSync, writeFileSync } from 'node:fs'
+import { tmpdir } from 'node:os'
+import { spawnSync } from 'node:child_process'
 import { fileURLToPath } from 'node:url'
 
 const __filename = fileURLToPath(import.meta.url)
@@ -65,6 +68,46 @@ await test('committed AGENTS.md is in sync with CLAUDE.md (drift gate)', async (
     committed === expected,
     'AGENTS.md is stale — run `node scripts/gen-agents-md.mjs` and commit AGENTS.md'
   )
+})
+
+
+// #7198 — the entry-point guard must survive a symlinked invocation path.
+//
+// Node's ESM loader resolves symlinks in `import.meta.url` but leaves
+// `process.argv[1]` as the caller typed it, and `resolve()` does not follow
+// symlinks. On macOS /tmp is a symlink to /private/tmp, so invoking
+// `node /tmp/…/gen-agents-md.mjs` compared '/private/tmp/…' against '/tmp/…',
+// the guard was false, and the script exited 0 having regenerated NOTHING.
+// bump-version.sh calls this and trusts the exit code, so a release could
+// silently skip the regeneration and still report success.
+await test('runs through a symlinked invocation path (#7198)', async () => {
+  const dir = mkdtempSync(join(tmpdir(), 'genagents-'))
+  try {
+    // Mirror the real layout: the generator resolves CLAUDE.md / AGENTS.md
+    // relative to its own parent directory, i.e. the repo root.
+    const root = join(dir, 'root')
+    const link = join(dir, 'link')
+    mkdirSync(join(root, 'scripts'), { recursive: true })
+    symlinkSync(root, link)
+
+    cpSync(scriptPath, join(root, 'scripts', 'gen-agents-md.mjs'))
+    cpSync(resolve(__dirname, '..', '..', 'CLAUDE.md'), join(root, 'CLAUDE.md'))
+    writeFileSync(join(root, 'AGENTS.md'), 'STALE\n')
+
+    const run = spawnSync(process.execPath, [join(link, 'scripts', 'gen-agents-md.mjs')], { encoding: 'utf8' })
+    assert(run.status === 0, `expected exit 0, got ${run.status}: ${run.stderr}`)
+    assert(
+      readFileSync(join(root, 'AGENTS.md'), 'utf8') !== 'STALE\n',
+      'generator silently no-opped when invoked through a symlinked path'
+    )
+
+    // --check must FAIL on drift through that same path, not pass silently.
+    writeFileSync(join(root, 'AGENTS.md'), 'STALE\n')
+    const checked = spawnSync(process.execPath, [join(link, 'scripts', 'gen-agents-md.mjs'), '--check'], { encoding: 'utf8' })
+    assert(checked.status === 1, `--check should exit 1 on drift, got ${checked.status}`)
+  } finally {
+    rmSync(dir, { recursive: true, force: true })
+  }
 })
 
 // --- summary --------------------------------------------------------------
--- a/scripts/gen-agents-md.mjs
+++ b/scripts/gen-agents-md.mjs
@@ -12,7 +12,7 @@
 // Keeping CLAUDE.md as the ONE source and generating AGENTS.md means the two can
 // never drift and a convention can never be silently dropped from the mirror.
 
-import { readFileSync, writeFileSync } from 'node:fs'
+import { readFileSync, writeFileSync, realpathSync } from 'node:fs'
 import { fileURLToPath } from 'node:url'
 import { dirname, join, resolve } from 'node:path'
 
@@ -62,9 +62,35 @@ export function readAgentsMd() {
 }
 
 // Only run the CLI side-effects when invoked directly (not when imported by tests).
-// Normalize both sides: process.argv[1] may be relative depending on how node was
-// invoked, while fileURLToPath(import.meta.url) is absolute.
-if (process.argv[1] && resolve(fileURLToPath(import.meta.url)) === resolve(process.argv[1])) {
+//
+// Both sides are realpath'd, not just `resolve`d. Node's ESM loader resolves
+// symlinks in `import.meta.url`, but `process.argv[1]` is left exactly as the
+// caller typed it, and `resolve()` does not follow symlinks — so on macOS,
+// where /tmp is a symlink to /private/tmp, invoking this as
+// `node /tmp/…/gen-agents-md.mjs` gave '/private/tmp/…' on one side and
+// '/tmp/…' on the other. The guard was false, the CLI branch never ran, and
+// the script exited 0 having done nothing (#7198).
+//
+// That silence is the dangerous part: bump-version.sh calls this and trusts
+// the exit code, so a release could regenerate nothing and still report
+// success — the same shape as every other guard-that-quietly-no-ops fixed
+// this cycle. realpathSync is wrapped because argv[1] need not exist (an
+// `-e` eval, a deleted script); an unresolvable path simply is not this file.
+const invokedDirectly = (() => {
+  if (!process.argv[1]) return false
+  const realpathOrNull = (p) => {
+    try {
+      return realpathSync(p)
+    } catch {
+      return null
+    }
+  }
+  const self = realpathOrNull(fileURLToPath(import.meta.url))
+  const argv = realpathOrNull(resolve(process.argv[1]))
+  return self !== null && self === argv
+})()
+
+if (invokedDirectly) {
   const expected = renderAgentsMd(readClaudeMd())
   if (process.argv.includes('--check')) {
     if (readAgentsMd() !== expected) {


blamechris:
## Code Review: PR #7208

### Summary

The fix is correct, well-scoped, and independently reproducible: I rebuilt the pre-fix guard from `main` in an isolated harness (not the PR's own test) and confirmed the exact failure mode described — silent exit 0, no regeneration, `--check` passing on a stale file — then confirmed the fix resolves all three. The mutation check also holds exactly as claimed. One important correction to the PR's stated rationale: **CI's actual drift gate was never exposed to this bug** (details below) — the code fix is still worth taking, just not for the CI-safety reason given. I also found a live sibling instance of the same bug class in a file this PR doesn't touch. Filed three follow-up issues; none block this PR.

### What I verified

**1. Reproduced the original bug independently, then confirmed the fix.**

Built my own symlink-based harness (not the PR's test file) against the pre-fix content pulled from `main` (`git show main:scripts/gen-agents-md.mjs` — confirmed byte-identical to the diff's "before" hunk):

```
############ PRE-FIX (main) VARIANT ############
===== plain run via symlinked path =====
exit=0
RESULT: AGENTS.md still STALE (no-op / bug present)

===== --check on STALE file via symlinked path =====
exit=0
RESULT: --check PASSED on a stale file (drift gate defeated -- bug present)

===== plain run via NON-symlinked (real) path, for comparison =====
Wrote AGENTS.md from CLAUDE.md.
exit=0
RESULT: AGENTS.md was regenerated
```

```
############ POST-FIX (this PR) VARIANT ############
===== plain run via symlinked path =====
Wrote AGENTS.md from CLAUDE.md.
exit=0
RESULT: AGENTS.md was regenerated (fix working)

===== --check on STALE file via symlinked path =====
::error::AGENTS.md is out of sync with CLAUDE.md. Run `node scripts/gen-agents-md.mjs` and commit AGENTS.md.
exit=1
RESULT: --check FAILED (exit 1) on a stale file (correct)
```

Matches the PR's claimed behavior exactly, confirmed from a clean-room repro rather than the PR's own transcript.

**2. The `--check` / "CI drift gate" claim — the code is fine, but the framing is wrong.**

The PR states: *"`--check` — the CI drift gate — would pass on a stale file for exactly the same reason, since the branch that performs the check never executes."*

This conflates two different things. CI's actual drift-check step (`.github/workflows/ci.yml:896`, "Check AGENTS.md is in sync with CLAUDE.md") does **not** invoke `gen-agents-md.mjs --check`. It runs `node scripts/__tests__/gen-agents-md.test.mjs`, whose Test 3 ("committed AGENTS.md is in sync with CLAUDE.md (drift gate)") calls the exported `readAgentsMd()` / `readClaudeMd()` / `renderAgentsMd()` **directly** and compares them in-process — it never touches the `invokedDirectly` guard or the `--check` CLI branch at all, by construction (the module is `import()`-ed, so `process.argv[1]` is the test file's own path, which can never equal `gen-agents-md.mjs`'s path regardless of whether the guard is buggy or fixed).

Proved this directly: restored the pre-fix (buggy) guard into the working file, deliberately staled the **committed** `AGENTS.md`, and ran the literal command CI runs:

```
$ node scripts/__tests__/gen-agents-md.test.mjs      # pre-fix guard, staled AGENTS.md
  ok render includes the entire CLAUDE.md body unmodified
  ok render prepends the auto-generated / do-not-edit header
  FAIL committed AGENTS.md is in sync with CLAUDE.md (drift gate): AGENTS.md is stale — run `node scripts/gen-agents-md.mjs` and commit AGENTS.md
  FAIL runs through a symlinked invocation path (#7198): generator silently no-opped
2 passed, 2 failed
```

The drift gate (Test 3) still correctly caught the staleness under the **pre-fix, buggy guard** — proof that CI's real gate was never exposed to this bug. (Test 4 failing is expected — that's the new regression test this very PR adds, not a pre-existing CI gate.) Separately: CI's `scripts-tests` job runs on the self-hosted Linux pool or `ubuntu-24.04` for fork PRs (`runner-target` job, `SELF_HOSTED_OR_HOSTED` routing) — never macOS — so the `/tmp` → `/private/tmp` symlink that triggers this bug doesn't even exist on those runners by default, a second, independent reason CI was never organically exposed.

What **is** accurate and verified: the plain (non-`--check`) CLI path is genuinely used in production by `scripts/bump-version.sh:275-276` (`node "$ROOT/scripts/gen-agents-md.mjs" >/dev/null`, trusting exit code with no verification the file changed) as part of preparing a release — and `bump-version.sh` is not invoked by any GitHub Actions workflow (checked `ci.yml`, `release.yml`, `auto-tag-on-release.yml`); it's run locally/by an agent, an environment where a `/tmp`-rooted worktree is entirely plausible. That's the real, live exposure this fix protects — it's just not "CI's drift gate." Worth softening that line in the PR description before merge so it doesn't read later as "CI was broken," which it wasn't.

**3. try/catch swallow risk — real but narrow, filed as a follow-up.**

`realpathOrNull` returns `null` on *any* thrown error, collapsing "argv[1] doesn't exist" (legitimately not-this-file) and "realpath failed for some other reason" (unknown) into the same signal. Since the final check is `self !== null && self === argv`, a `self`-side failure makes the guard false unconditionally — even if `argv` is textually identical to `self`. Demonstrated `realpathSync` throwing under ordinary conditions (`EACCES` on a permission-restricted directory), and confirmed the shipped closure's behavior on a forced-null `self`: `{ invokedDirectly: false, self: null, argv: null }` even when the two input strings were identical. Low likelihood in this repo's actual CI (fresh checkout, no permission weirdness), but the failure mode is exactly as silent as the bug being fixed. Filed as #7214 (not blocking).

**4. Import path stays side-effect-free.**

Mirrored the test harness's own invocation shape (a separate real file as `argv[1]`, dynamically importing the module — not `node -e`, which wouldn't set `argv[1]` realistically) and checked `AGENTS.md`'s mtime + checksum before/after:

```
=== BEFORE ===
1786854649
0a2ca174cf52baaefce373afa9778a41b4743530  AGENTS.md
imported exports: [ 'readAgentsMd', 'readClaudeMd', 'renderAgentsMd' ]
=== AFTER (unchanged) ===
1786854649
0a2ca174cf52baaefce373afa9778a41b4743530  AGENTS.md
```

No write, no console output from the CLI branch. Confirmed.

**5. Windows — real, documented risk, and currently untested either way. Filed as a follow-up.**

Two things compound here:
- `scripts-tests` (the only CI coverage for this file, including the new regression test) never runs on Windows — it's routed through `SELF_HOSTED_OR_HOSTED` (Linux/`ubuntu-24.04`), not `chroxy-win`/`windows-latest`. Pre-existing gap, not introduced by this PR, but relevant given the change is specifically about path normalization.
- The default (non-native) `fs.realpathSync` has documented Windows-specific gaps: [nodejs/node#50019](https://github.com/nodejs/node/issues/50019) — the same physical file reached via different mount points/junctions can silently realpath to *different* strings (no error) — which would reproduce this exact "guard false when it should be true" bug on Windows through a junction instead of a macOS symlink; and [nodejs/node#8897](https://github.com/nodejs/node/issues/8897) — `realpathSync` throwing on certain Windows container/UNC paths. Both are caught silently by the new try/catch (see #3 above).

I can't test this on real Windows from here, so this is a documented risk, not a proven regression — filed as #7215 with the citations, recommending `realpathSync.native` and/or actual Windows coverage for this test file.

**6. Test suite + mutation check.**

Ran every step of the `Scripts Tests` CI job locally and cross-checked against the actual CI log for this PR's run (`gh run view ... --job <id> --log`) — all match:

| file | result |
|---|---|
| `bump-version.test.sh` | 18 passed, 0 failed |
| `flush-and-exit.test.mjs` | 5 passed, 0 failed |
| `compile-skill-targets.test.mjs` | 18 passed, 0 failed |
| `lint-no-raw-color-literals.test.sh` | 4/4 cases |
| `lint-no-nul-bytes.test.sh` | 7/7 cases |
| `gen-agents-md.test.mjs` | **4 passed, 0 failed** |
| `check-release-pr-subject.test.mjs` | 23 passed, 0 failed |

Total across the job: **79 passed**, not the "Scripts suite: 18 pass" the PR description states (that number matches two *individual* files' counts by coincidence — `bump-version.test.sh` and `compile-skill-targets.test.mjs` — not the gen-agents-md file this PR touches, which has 4, and not the suite total, 79). Nitpick, not a blocker — worth a quick correction.

Mutation check: reverted only the guard (kept the real, committed `AGENTS.md`) and reran:

```
  ok render includes the entire CLAUDE.md body unmodified
  ok render prepends the auto-generated / do-not-edit header
  ok committed AGENTS.md is in sync with CLAUDE.md (drift gate)
  FAIL runs through a symlinked invocation path (#7198): generator silently no-opped when invoked through a symlinked path
3 passed, 1 failed
```

Matches the PR's claimed mutation-check output exactly. Restored the fix afterward; `node scripts/__tests__/gen-agents-md.test.mjs` is back to 4 passed, 0 failed.

**7. Bonus finding: the same bug class is live elsewhere in the repo.**

While checking for sibling instances of this exact guard pattern (`process.argv[1]` vs `fileURLToPath(import.meta.url)`), found `scripts/compile-skill-targets.mjs:350` using the identical vulnerable comparison:

```js
if (process.argv[1] && resolve(process.argv[1]) === fileURLToPath(import.meta.url)) {
  main()
}
```

Confirmed live with an isolated mirror of the exact shipped expression (stub `main()`, no repo side effects):

```
=== direct invocation ===
MAIN RAN
=== via symlinked path ===
GUARD FALSE -- main() skipped, exiting 0 having done nothing
```

This is the developer/agent-facing skill compiler (`node scripts/compile-skill-targets.mjs --name <name>`, documented in `CLAUDE.md` as a direct manual invocation) — the same "reports success, does nothing" shape as #7198, unfixed. Filed as #7213 (high priority — it's live, not hypothetical). Also checked three more files sharing the same guard family (`packages/server/sidecar/agent.js`, `packages/server/src/server-cli-child.js`, `packages/server/src/channels/chroxy-channel-server.js`) — none are currently exposed in their actual shipped invocation paths (details in #7213), but worth a defensive pass in the same sweep.

### Issues Found

**Critical:** None in this diff.

**Suggestions (all filed as follow-ups, none blocking):**
- #7213 — `scripts/compile-skill-targets.mjs` has the same symlink entry-point-guard bug this PR fixes elsewhere; confirmed live. (high priority)
- #7214 — `realpathOrNull`'s blanket catch conflates "not this file" with "can't tell," which can silently re-skip the CLI branch on a genuine invocation under fs-error conditions. (low priority)
- #7215 — Windows: no CI coverage for this file's test suite, plus documented `fs.realpathSync` Windows junction/mountpoint gaps (nodejs/node#50019, #8897) that could reintroduce the exact bug this PR fixes, on Windows, through junctions instead of symlinks. (low priority)
- Soften the PR description's "`--check` — the CI drift gate" framing before merge — CI's actual gate (the test file's direct function calls) was never exposed to this bug; see verification #2 above. The `bump-version.sh` release-path exposure is real and is the accurate motivating case.

**Nitpicks:**
- PR description's "Scripts suite: 18 pass" doesn't match the actual suite total (79) or the gen-agents-md file's own count (4) — looks like a stray copy from one of the other files' output. Worth a quick fix for accuracy.

### Verdict

**Approve.** The guard fix itself is correct, matches its own regression test, survives a mutation check, doesn't leak side effects into the import path, and is a real improvement for the `bump-version.sh` release path and any manual `--check` use. The one thing worth fixing before this settles into history is the PR description's overstated CI-exposure claim — the code doesn't need to change for that, just the narrative. Everything else is filed and tracked, none of it blocking.

blamechris:
## You corrected the central claim — amended in `eff644c50`

I asserted CI's AGENTS.md drift gate was silently defeatable. **It was not**, and I verified your correction before acting on it:

```
$ grep -A2 "AGENTS.md is in sync" .github/workflows/ci.yml
      - name: Check AGENTS.md is in sync with CLAUDE.md
        run: node scripts/__tests__/gen-agents-md.test.mjs

$ grep -rn "gen-agents-md.mjs" .github/workflows/
(no matches — nothing in CI invokes the script, let alone --check)
```

CI runs the **test file**, which imports the exported functions directly and never evaluates `invokedDirectly`. Your empirical proof — reverting the guard, staling AGENTS.md, and showing the drift gate still caught it — settles it. The Linux runners also lack the `/tmp` symlink that makes this reachable on macOS.

The real exposure is narrower and remains accurate: `scripts/bump-version.sh:276` calls this and trusts the exit code, and that call site has **no** test covering it. Which is a little pointed — I added that call in #7195 two hours ago.

I amended rather than only commenting because this repo squashes with `COMMIT_MESSAGES`, so the overstated claim would have gone into `main`'s history permanently.

### On your follow-ups

**#7213 is the valuable one** — the identical guard bug live in `scripts/compile-skill-targets.mjs`, confirmed by isolated repro. That is a real bug found by generalising from this one, which is exactly the sweep I should have done here and didn't. Taking it next.

**#7214** (`realpathOrNull` swallowing `EACCES` and reintroducing a silent no-op through a different door) — genuine, and it is the same failure mode as the bug being fixed, so it deserves the fix rather than sitting as a latent note.

**#7215** (Windows) — fair, and sharpened by your observation that "Scripts Tests" never runs on Windows CI, so nothing would catch a regression there.

**Nitpick accepted:** the "18 pass" figure in the description was the scripts suite total, not this file's — sloppy phrasing on my part.

Issue #7222, opened 2026-08-16
Consolidate scripts/'s two independent entry-point-guard copies, or migrate to Node 22.18+'s import.meta.main


blamechris:
## Context

PR #7217 (closing #7213) made `packages/server/src/utils/is-entry-point.js` the one implementation for the server package, and deliberately duplicated the same realpath-comparison pattern into `scripts/compile-skill-targets.mjs` because `scripts/` sits outside the server package's scope.

That reasoning holds for reaching into `packages/server/src/` (confirmed: zero existing precedent of `scripts/*.mjs` importing from `packages/*/src` — `grep -rn "from '.*packages/" scripts/` returns nothing), but it left a second-order duplication unaddressed: `scripts/gen-agents-md.mjs` already carries its own independent copy of this exact pattern (from #7208, predating #7217), with different local naming (`realpathOrNull`/`argv`) but identical logic. `scripts/` now has two hand-maintained copies of the same ~15 lines — the "one bug, N copies" shape #7213/#7217 exists to eliminate, recreated one directory level down. #7213 itself raised this ("worth deciding whether a single shared helper... is warranted") but the question wasn't resolved before #7217 shipped a second copy instead of consolidating.

## Two ways to close this

1. **Minimal**: factor `scripts/lib/is-entry-point.mjs` (no existing `scripts/lib/` convention, but trivial to add) shared by `gen-agents-md.mjs` and `compile-skill-targets.mjs`, mirroring the server package's helper.

2. **Better**: Node 22.18.0 shipped `import.meta.main` as a stable, unflagged, SEMVER-MINOR feature — a native replacement for the whole realpath-comparison pattern that resolves symlinks correctly by construction. Verified empirically on this machine (Node v22.22.3):

   ```
   $ node /private/tmp/imm-test/entry.mjs   # real path
   entry import.meta.main = true
   $ node /tmp/imm-test/entry-link.mjs      # same file via a symlink
   entry import.meta.main = true
   $ node /private/tmp/imm-test/lib-import-probe.mjs   # imported, not entry
   lib import.meta.main = false
   ```

   This would let all five guards (the shared helper, its two `packages/server` consumers, and the two `scripts/`/`sidecar/` duplicates) collapse to `if (import.meta.main) { ... }` — no hand-rolled comparison code anywhere, which eliminates the duplication question rather than managing it.

[...]

Issue #7215, opened 2026-08-16
gen-agents-md.mjs's realpath fix is unverified on Windows: Scripts Tests never runs there, and fs.realpathSync has known Windows junction gaps


blamechris:
## Context

PR #7208 (closes #7198) switched `scripts/gen-agents-md.mjs`'s entry-point guard from a plain `resolve()` comparison to `realpathSync` on both sides. This is a cross-platform path-normalization change, but it ships with zero Windows verification, for two independent reasons.

## 1. The only test coverage for this file never runs on Windows CI

`.github/workflows/ci.yml`'s `scripts-tests` job (which runs `scripts/__tests__/gen-agents-md.test.mjs`, including the new symlinked-invocation regression test) is routed via:

```yaml
runs-on: ${{ fromJSON(needs.runner-target.outputs.runner) }}  # SELF_HOSTED_OR_HOSTED
```

which resolves to `["self-hosted", "Linux", "chroxy-linux"]` for same-repo events or `ubuntu-24.04` for fork PRs — never the Windows pool (`chroxy-win` / `windows-latest`, only used by `server-tests-windows` and `desktop-tests-windows`). There is no Windows leg of Scripts Tests. This is pre-existing (not introduced by #7208), but it means a Windows-specific regression in this exact code would ship silently.

## 2. `fs.realpathSync` (default, non-native) has documented Windows gaps in exactly this scenario

- [nodejs/node#50019](https://github.com/nodejs/node/issues/50019): on Windows, `fs.realpathSync()` can return **different, non-canonical strings for the same physical file** when accessed via different mount points/junctions — silently, no error. If a direct invocation's path and `import.meta.url` reach the script via different junction routes, the two `realpathSync` calls could disagree even though it genuinely is the same file, reproducing the "guard false when it should be true → silent no-op" bug this PR fixes, specifically on Windows.
- [nodejs/node#8897](https://github.com/nodejs/node/issues/8897): `fs.realpathSync()` has thrown `ENOENT` on Windows container/UNC-mounted paths by misinterpreting the path, where the old plain `resolve()` (no filesystem access at all) would not have touched the filesystem and would not have thrown.
- `fs.realpathSync.native` is documented/known to behave differently (more completely) than the default JS implementation for some of these Windows cases (e.g. `os.tmpdir()` resolution).

The fix as shipped uses the plain (non-native) `realpathSync` import from `node:fs`.

## Suggested follow-up

[...]

PR #7224, opened 2026-08-16
fix(ci): decide the entry point before touching the filesystem

### LINKED ISSUE #7214: gen-agents-md.mjs's realpathOrNull collapses "not this file" and "can't tell" into the same false, risking a silent no-op via a different trigger
## Context

PR #7208 (closes #7198) fixed `scripts/gen-agents-md.mjs`'s entry-point guard by `realpathSync`-ing both `fileURLToPath(import.meta.url)` and `resolve(process.argv[1])` before comparing, wrapped in a `realpathOrNull` helper that returns `null` on any thrown error:

```js
const realpathOrNull = (p) => {
  try {
    return realpathSync(p)
  } catch {
    return null
  }
}
const self = realpathOrNull(fileURLToPath(import.meta.url))
const argv = realpathOrNull(resolve(process.argv[1]))
return self !== null && self === argv
```

## The gap

`realpathOrNull` treats every error identically, but there are two semantically different situations underneath:

1. `argv[1]` genuinely doesn't exist (e.g. `-e` eval, deleted script) → legitimately "not this file", guard should be false. This is the intended, documented case.
2. `realpathSync` throws for some other reason on a path that *is* this file (permission error on an intermediate directory, transient filesystem hiccup, exotic mount) → the guard still silently becomes false, and the CLI branch is skipped exactly as it was in the original #7198 bug — just triggered by an fs error instead of a symlink-normalization mismatch.

Demonstrated directly: `realpathSync` throws `EACCES` under ordinary conditions (a directory the current user can't traverse) — not a contrived scenario:

```
$ chmod 0000 locked && node -e "require('node:fs').realpathSync('locked/self.mjs')"
realpathSync THREW: EACCES - EACCES: permission denied, lstat '.../locked/self.mjs'
```

And feeding the exact shipped closure a case where `self` fails to realpath (even with `argv` textually identical to it) shows `invokedDirectly` evaluates to `false`:

```
case: self and argv are literally the same string, but self's realpath throws
-> { invokedDirectly: false, self: null, argv: null }
```

Because the script does nothing after the `if (invokedDirectly) { ... }` block, a real, legitimate direct invocation that hits this would exit 0 having done nothing — no error, no diagnostic — the identical silent-no-op shape PR #7208 exists to eliminate, just reachable through fs errors instead of path-string mismatches.

## Why this is low/not blocking

- Requires an unusual environment (restrictive permissions, exotic/network filesystem, a TOCTOU race in the narrow window between process start and the check) — not a routine trigger.
- Doesn't undo the fix for the reported #7198 bug (the common symlink case).
- CI's actual drift gate (`scripts/__tests__/gen-agents-md.test.mjs`'s Test 3) is unaffected either way, since it calls the exported functions directly and never exercises this guard.

## Suggested improvement

Distinguish the two cases instead of collapsing them:
- If `argv[1]` doesn't exist (`ENOENT`), that's a legitimate "not this file" signal — keep current behavior.
- If realpath fails for any other reason on either side, consider falling back to the plain `resolve()` string comparison (degrading to the pre-fix behavior only in this narrow case, rather than silently doing nothing), or at minimum emit a stderr diagnostic so a real invocation that hits this doesn't fail in total silence.

Filed from PR #7208's review.

blamechris:
`scripts/gen-agents-md.mjs`'s entry-point guard resolved `realpath` for both sides first and returned `false` whenever either call threw:

```js
const self = realpathOrNull(fileURLToPath(import.meta.url))
const argv = realpathOrNull(resolve(process.argv[1]))
return self !== null && self === argv
```

An `EACCES` on a parent directory, or a script unlinked after launch, therefore made a direct `node scripts/gen-agents-md.mjs` evaluate `false` — AGENTS.md never regenerated, `--check` never checked, exit 0. The same silent no-op as #7198, reached through a different trigger.

## The fix

The plain comparison is conclusive on its own: identical paths mean this **is** the entry point. Running it first settles the common case with no filesystem access, so no fs error can reach it.

```js
const self = fileURLToPath(import.meta.url)
const argv = resolve(process.argv[1])
if (self === argv) return true
// only now is realpath needed: differing paths can still be one file via a
// symlink, or via case normalisation on a case-insensitive filesystem
```

The #7198 symlink fix is untouched — realpath still runs for exactly the case it was added for. The residual is a direct invocation through a symlink whose realpath *also* fails, which is genuinely undecidable without the call that just failed.

## Coverage

The guard is pinned directly, in `scripts/__tests__/gen-agents-md.test.mjs`. A `module.register` load hook breaks the filesystem in the window between node reading the source and the module body evaluating — which is where the guard runs — under two triggers, `unlink` and `chmod 0000`.

Mutation-verified:

| variant | exit | result |
|---|---|---|
| this PR | 0 | 7 passed, 0 failed |
| realpath-first restored | 1 | both #7214 tests FAIL — `got 0 (0 = the silent no-op this guards)` |
| restored again | 0 | 7 passed, 0 failed |

The suite carries a **positive control**: without it, both tests pass for the wrong reason, since a hook that fails to break realpath leaves the guard working normally and `--check` exits 1 on the stale AGENTS.md either way. The control asserts `realpathSync` actually throws afterwards.

The EACCES case is **skipped as root** — `chmod 0000` does not deny root, so it would otherwise pass without exercising the path at all.

## Verification

Every step the `Scripts Tests` job runs, locally on Node 22:

```
bump-version.test.sh                 exit=0
flush-and-exit.test.mjs              exit=0
compile-skill-targets.test.mjs       exit=0
lint-no-raw-color-literals.test.sh   exit=0
lint-no-nul-bytes.test.sh            exit=0
gen-agents-md.test.mjs               exit=0
check-release-pr-subject.test.mjs    exit=0
```

Plus the guard end-to-end: `node scripts/gen-agents-md.mjs --check` and the same through a symlinked path both exit 0 and report in sync.

## Corrections to this PR's earlier claims

Two statements in the original body were wrong and are replaced above, both caught in review:

1. **"the identical logic is pinned in `packages/server/tests/is-entry-point.test.js`"** — that file covers the shared helper, a *different copy*. Reverting this file's reordering left the whole repo suite exiting 0. The change was unpinned; it is now pinned by the tests above.
2. **"the EACCES trigger is not reproducible in a subprocess"** — false, and the load-hook test is the counterexample. Node only has to *load* the script, not keep it readable through evaluation.

The original body also said "every test the Scripts Tests job runs" while listing five of seven; the full seven are above.

This is the same reordering applied to the other three guards in #7217.

Closes #7214.

### DIFF
--- a/scripts/__tests__/gen-agents-md.test.mjs
+++ b/scripts/__tests__/gen-agents-md.test.mjs
@@ -12,7 +12,7 @@
  */
 
 import { dirname, join, resolve } from 'node:path'
-import { cpSync, mkdirSync, mkdtempSync, readFileSync, rmSync, symlinkSync, writeFileSync } from 'node:fs'
+import { chmodSync, cpSync, mkdirSync, mkdtempSync, readFileSync, realpathSync, rmSync, symlinkSync, writeFileSync } from 'node:fs'
 import { tmpdir } from 'node:os'
 import { spawnSync } from 'node:child_process'
 import { fileURLToPath } from 'node:url'
@@ -109,6 +109,108 @@ await test('runs through a symlinked invocation path (#7198)', async () => {
     rmSync(dir, { recursive: true, force: true })
   }
 })
+
+// --- #7214: realpath failing must not silently skip the run -----------------
+//
+// The guard decides "am I the entry point?" from process.argv[1]. When it
+// consulted realpath FIRST, any throw was read as "not the entry point", so a
+// direct run exited 0 having done nothing — no regeneration, no --check, no
+// diagnostic. Same silent no-op as #7198, different trigger.
+//
+// Reproducing that needs the filesystem to break in the window AFTER node has
+// read the script and BEFORE the module body evaluates, which is where the
+// guard runs. A `module.register` load hook sits exactly in that window: node
+// only has to LOAD the file, not keep it readable afterwards.
+const HOOKS = `
+import { unlinkSync, chmodSync } from 'node:fs'
+import { fileURLToPath } from 'node:url'
+import { dirname } from 'node:path'
+export async function load (url, context, nextLoad) {
+  const result = await nextLoad(url, context)
+  if (url.endsWith('/gen-agents-md.mjs')) {
+    const p = fileURLToPath(url)
+    if (process.env.CHROXY_BREAK === 'unlink') unlinkSync(p)
+    else if (process.env.CHROXY_BREAK === 'eacces') chmodSync(dirname(p), 0o000)
+  }
+  return result
+}
+`
+const REGISTER = `
+import { register } from 'node:module'
+register('./hooks.mjs', import.meta.url)
+`
+
+const withBrokenRealpath = (breakMode, fn) => {
+  // realpathSync the tmpdir: on macOS os.tmpdir() is /var/..., a symlink to
+  // /private/var. Leaving it symlinked would ALSO trip #7198's non-canonical
+  // path case, and the two triggers together are the residual this fix does
+  // not claim to cover. Canonicalising isolates #7214.
+  const dir = mkdtempSync(join(realpathSync(tmpdir()), 'genagents-7214-'))
+  const scriptsDir = join(dir, 'root', 'scripts')
+  try {
+    mkdirSync(scriptsDir, { recursive: true })
+    cpSync(scriptPath, join(scriptsDir, 'gen-agents-md.mjs'))
+    cpSync(resolve(__dirname, '..', '..', 'CLAUDE.md'), join(dir, 'root', 'CLAUDE.md'))
+    writeFileSync(join(dir, 'root', 'AGENTS.md'), 'STALE\n')
+    writeFileSync(join(dir, 'hooks.mjs'), HOOKS)
+    writeFileSync(join(dir, 'register.mjs'), REGISTER)
+    fn(spawnSync(
+      process.execPath,
+      ['--import', join(dir, 'register.mjs'), join(scriptsDir, 'gen-agents-md.mjs'), '--check'],
+      { encoding: 'utf8', env: { ...process.env, CHROXY_BREAK: breakMode } },
+    ))
+  } finally {
+    chmodSync(scriptsDir, 0o755)
+    rmSync(dir, { recursive: true, force: true })
+  }
+}
+
+// Positive control for the two tests below. If the hook does NOT actually break
+// realpath, both assertions pass for the wrong reason: the guard works normally,
+// --check sees the stale AGENTS.md, and exits 1 either way. That is the vacuous
+// pass this whole file exists to distrust, so prove the trigger fires first.
+await test('the load hook genuinely breaks realpath (positive control)', async () => {
+  withBrokenRealpath('unlink', (run) => {
+    assert(run.status !== null, `child did not run at all: ${run.error && run.error.message}`)
+  })
+  const dir = mkdtempSync(join(realpathSync(tmpdir()), 'genagents-ctl-'))
+  try {
+    const probe = join(dir, 'probe.mjs')
+    writeFileSync(probe, 'process.exit(0)\n')
+    writeFileSync(join(dir, 'hooks.mjs'), HOOKS.replace(/gen-agents-md\.mjs/g, 'probe.mjs'))
+    writeFileSync(join(dir, 'register.mjs'), REGISTER)
+    spawnSync(process.execPath, ['--import', join(dir, 'register.mjs'), probe], {
+      encoding: 'utf8',
+      env: { ...process.env, CHROXY_BREAK: 'unlink' },
+    })
+    let threw = false
+    try {
+      realpathSync(probe)
+    } catch {
+      threw = true
+    }
+    assert(threw, 'the unlink hook did not make realpathSync throw — the tests below would pass vacuously')
+  } finally {
+    rmSync(dir, { recursive: true, force: true })
+  }
+})
+
+await test('--check detects drift when the script is unlinked after launch (#7214)', async () => {
+  withBrokenRealpath('unlink', (run) =>
+    assert(run.status === 1, `--check must exit 1 on drift; got ${run.status} (0 = the silent no-op this guards)`))
+})
+
+// chmod 0000 does not deny root, so as root realpath would succeed and this
+// would pass without exercising the EACCES path at all — vacuously. Skipped
+// rather than silently weakened.
+if (typeof process.getuid === 'function' && process.getuid() === 0) {
+  process.stdout.write('  skip --check detects drift when realpath fails with EACCES (#7214): running as root\n')
+} else {
+  await test('--check detects drift when realpath fails with EACCES (#7214)', async () => {
+    withBrokenRealpath('eacces', (run) =>
+      assert(run.status === 1, `--check must exit 1 on drift; got ${run.status} (0 = the silent no-op this guards)`))
+  })
+}
 
 // --- summary --------------------------------------------------------------
 process.stdout.write(`\n${pass} passed, ${fail} failed\n`)
--- a/scripts/gen-agents-md.mjs
+++ b/scripts/gen-agents-md.mjs
@@ -78,6 +78,16 @@ export function readAgentsMd() {
 // `-e` eval, a deleted script); an unresolvable path simply is not this file.
 const invokedDirectly = (() => {
   if (!process.argv[1]) return false
+  // The plain comparison runs first and is conclusive on its own: identical
+  // paths mean this IS the entry point, decided without touching the
+  // filesystem. Consulting realpath first and reading a failure as false let an
+  // EACCES on a parent directory (or a script unlinked after launch) turn a
+  // direct run into a silent exit-0 no-op — the same shape as #7198, reached
+  // through a different trigger (#7214). Realpath is only needed once the paths
+  // differ, where a symlink is the sole thing that can still make them one file.
+  const self = fileURLToPath(import.meta.url)
+  const argv = resolve(process.argv[1])
+  if (self === argv) return true
   const realpathOrNull = (p) => {
     try {
       return realpathSync(p)
@@ -85,9 +95,9 @@ const invokedDirectly = (() => {
       return null
     }
   }
-  const self = realpathOrNull(fileURLToPath(import.meta.url))
-  const argv = realpathOrNull(resolve(process.argv[1]))
-  return self !== null && self === argv
+  const realSelf = realpathOrNull(self)
+  const realArgv = realpathOrNull(argv)
+  return realSelf !== null && realSelf === realArgv
 })()
 
 if (invokedDirectly) {
--- a/scripts/gen-agents-md.mjs
+++ b/scripts/gen-agents-md.mjs
@@ -84,7 +84,8 @@ const invokedDirectly = (() => {
   // EACCES on a parent directory (or a script unlinked after launch) turn a
   // direct run into a silent exit-0 no-op — the same shape as #7198, reached
   // through a different trigger (#7214). Realpath is only needed once the paths
-  // differ, where a symlink is the sole thing that can still make them one file.
+  // differ — via a symlink, or via case normalisation on a case-insensitive
+  // filesystem (default APFS/NTFS), both of which it resolves.
   const self = fileURLToPath(import.meta.url)
   const argv = resolve(process.argv[1])
   if (self === argv) return true
--- a/scripts/gen-agents-md.mjs
+++ b/scripts/gen-agents-md.mjs
@@ -84,8 +84,9 @@ const invokedDirectly = (() => {
   // EACCES on a parent directory (or a script unlinked after launch) turn a
   // direct run into a silent exit-0 no-op — the same shape as #7198, reached
   // through a different trigger (#7214). Realpath is only needed once the paths
-  // differ — via a symlink, or via case normalisation on a case-insensitive
-  // filesystem (default APFS/NTFS), both of which it resolves.
+  // differ, where a symlink is the only thing that still makes them one file —
+  // the same wording as the canonical copy in
+  // packages/server/src/utils/is-entry-point.js, deliberately.
   const self = fileURLToPath(import.meta.url)
   const argv = resolve(process.argv[1])
   if (self === argv) return true


blamechris:
## Code Review: PR #7224

### Summary

Reorders the entry-point guard in `scripts/gen-agents-md.mjs` so the plain string comparison of `fileURLToPath(import.meta.url)` against `resolve(process.argv[1])` runs first and returns `true` immediately on a match, with `realpathSync` demoted to a fallback for the case the paths differ. Closes #7214.

**The code change is correct.** I verified the defect is real, the fix removes it, and the #7198 symlink fix survives — all with exit-code assertions and a positive control. Details below.

Two problems are in the PR body rather than the diff, and together they are the reason for the verdict: the change ships with **no regression coverage anywhere**, and the body's two-part justification for that — "not reproducible in a subprocess", and "is pinned in `packages/server/tests/is-entry-point.test.js`" — is disproven by construction and by `git cat-file` respectively.

### Verification performed

Three variants were built from this PR's file, differing only in the guard, each in a self-contained mini-repo with a **drifted `AGENTS.md`** so the signal is purely the child process's exit code (`1` = guard fired, `0` + no output = silent no-op):

- `V0` = pre-#7198 (plain compare only) — the **positive control**
- `V1` = current `main` (realpath first)
- `V2` = this PR (plain first, realpath fallback)

| Scenario | V0 (pre-#7198) | V1 (main) | V2 (this PR) |
|---|---|---|---|
| direct absolute path | FIRED | FIRED | FIRED |
| `/tmp` symlink prefix (#7198 trigger) | **SILENT-NOOP** | FIRED | FIRED |
| symlinked directory | **SILENT-NOOP** | FIRED | FIRED |
| symlinked file | **SILENT-NOOP** | FIRED | FIRED |
| relative `argv[1]` | FIRED | FIRED | FIRED |
| realpath fails — `EACCES` on parent dir | — | **SILENT-NOOP** | FIRED |
| realpath fails — script unlinked after launch | — | **SILENT-NOOP** | FIRED |

`V0` failing rows 2–4 is what proves the harness can actually see a broken guard; without that control the `V1`/`V2` greens would mean nothing. The two `V1` no-ops are #7214 reproduced, 8/8 runs, deterministic.

**Drift gate** (`--check` on the real repo): in-sync `exit=0`; `AGENTS.md` drifted `exit=1`; `gen-agents-md.test.mjs` under the same drift `exit=1`; regenerate then re-check `exit=0`, byte-identical to the committed file.

**All seven steps of the CI `Scripts Tests` job** (`ci.yml` L865-899), locally, Node 22: `bump-version.test.sh` 0 · `flush-and-exit.test.mjs` 0 · `compile-skill-targets.test.mjs` 0 · `lint-no-raw-color-literals.test.sh` 0 · `lint-no-nul-bytes.test.sh` 0 · `gen-agents-md.test.mjs` 0 · `check-release-pr-subject.test.mjs` 0.

**Reordering correctness** — I went looking for a case where the new code returns `true`, the old returned `false`, and `true` is wrong. Given `self === argv` as strings, the old code realpath'd the *same string* twice, so it can only have differed by that realpath failing — and "realpath failed on the path that equals this module's own resolved path" cannot mean the module is not at that path. The single constructible divergence is `process.argv[1]` deliberately reassigned to the module's own path while the module is *imported*, with realpath simultaneously broken: `V2` runs the side effect, `V1` does not. But the control run shows `V1` runs it too the moment realpath works — so the old code's "protection" there is accidental, not a property anything relies on. **No real regression.**

### Strengths

- The ordering argument is right and well stated: the cheap comparison is conclusive on its own, so no filesystem error can reach the common case.
- The residual case is named honestly rather than papered over — and it is real. My first attempt at a regression test tripped it by accident, because `os.tmpdir()` on macOS is `/var/...` (a symlink to `/private/var/...`); combined with a broken realpath, `V2` no-ops exactly as the body says it will. The test needs `realpathSync(tmpdir())` to isolate #7214 from #7198.
- The inline comment explains *why* the order is load-bearing rather than restating the code, which is what stops a future reader "tidying" it back.
- Matches #7217's `is-entry-point.js` reordering exactly — I diffed them; that claim holds.

### Issues Found

#### Critical (Must Fix)

|---|---|---|---|
| PR body | "On test coverage" | **States the logic "**is** pinned, in `packages/server/tests/is-entry-point.test.js` (added in #7217)".** #7217 is **OPEN**. That file, and `packages/server/src/utils/is-entry-point.js`, are absent from `origin/main` **and** from this PR's branch (`git cat-file -e` fails on both refs). Mutation-confirmed: revert the reordering and the **entire repo suite still exits 0** — nothing anywhere detects it. If #7217 is closed or reworked, this fix is unpinned permanently and silently. | Land the regression test in this PR (below), and correct the body to say the sibling coverage is *proposed* in an open PR. |
| PR body | "On test coverage" | **"the `EACCES` trigger is not reproducible in a subprocess, since node has to be able to load the very script whose path it cannot `lstat`" is false.** The premise is true and the conclusion does not follow: node only has to *load* it. A `module.register` `load` hook breaks the filesystem in the window between source read and module evaluation — which is precisely the #7214 window. Deterministic, 8/8, node builtins only. | Replace the claim, and land the test. |

|---|---|---|---|
| PR body | "Verification" | Lists five tests under "Every test the 'Scripts Tests' job runs". The job runs **seven**; `lint-no-raw-color-literals.test.sh` and `lint-no-nul-bytes.test.sh` are missing. Both pass, so nothing is hidden — but the two were added in #7122, well before this branch, so base drift does not explain it. | "Every test" should be true or not claimed. |
| `scripts/gen-agents-md.mjs` | 98-100 | A fourth copy of this guard now exists. Already tracked by #7222 — no action here, just noting the copy count went up while the consolidation issue is open. | Consolidation is the fix; duplicating once more is reasonable in the meantime. |

#### Nitpicks (Optional)

- The comment says realpath is needed "once the paths differ, where a symlink is the sole thing that can still make them one file" — on a case-insensitive filesystem (default APFS/NTFS) case normalisation also makes differing strings one file, and the realpath fallback is what covers it. Worth a clause, since it is a second reason the fallback must stay.

### The regression test, ready to land

This passes on your code and fails when the reordering is reverted. It follows the same shape as the existing #7198 test in `scripts/__tests__/gen-agents-md.test.mjs` — mini-repo in a tmpdir, stale `AGENTS.md`, assert on the child's exit code — and adds no dependency:

```js
// A load hook that breaks realpath() for the generator after node has read it.
const HOOKS = `
import { unlinkSync, chmodSync } from 'node:fs'
import { fileURLToPath } from 'node:url'
import { dirname } from 'node:path'
export async function load (url, context, nextLoad) {
  const result = await nextLoad(url, context)
  if (url.endsWith('/gen-agents-md.mjs')) {
    const p = fileURLToPath(url)
    if (process.env.CHROXY_BREAK === 'unlink') unlinkSync(p)
    else if (process.env.CHROXY_BREAK === 'eacces') chmodSync(dirname(p), 0o000)
  }
  return result
}
`
const REGISTER = `
import { register } from 'node:module'
register('./hooks.mjs', import.meta.url)
`

const withBrokenRealpath = (breakMode, fn) => {
  // realpathSync the tmpdir: on macOS os.tmpdir() is /var/... , a symlink to
  // /private/var. Leaving it symlinked would ALSO trip #7198's non-canonical
  // path case, and the two triggers together are the residual this fix does
  // not claim to cover. Canonicalising isolates #7214.
  const dir = mkdtempSync(join(realpathSync(tmpdir()), 'genagents-7214-'))
  const scriptsDir = join(dir, 'root', 'scripts')
  try {
    mkdirSync(scriptsDir, { recursive: true })
    cpSync(scriptPath, join(scriptsDir, 'gen-agents-md.mjs'))
    cpSync(resolve(__dirname, '..', '..', 'CLAUDE.md'), join(dir, 'root', 'CLAUDE.md'))
    writeFileSync(join(dir, 'root', 'AGENTS.md'), 'STALE\n')
    writeFileSync(join(dir, 'hooks.mjs'), HOOKS)
    writeFileSync(join(dir, 'register.mjs'), REGISTER)
    fn(spawnSync(
      process.execPath,
      ['--import', join(dir, 'register.mjs'), join(scriptsDir, 'gen-agents-md.mjs'), '--check'],
      { encoding: 'utf8', env: { ...process.env, CHROXY_BREAK: breakMode } }
    ))
  } finally {
    chmodSync(scriptsDir, 0o755)
    rmSync(dir, { recursive: true, force: true })
  }
}

// #7214 — realpath failing on the generator's own path must not turn a direct
// run into a silent exit-0 no-op. The entry point is decided by the plain
// comparison, before any filesystem access, so this must still catch drift.
await test('--check detects drift when realpath fails with EACCES (#7214)', async () => {
  withBrokenRealpath('eacces', (run) =>
    assert(run.status === 1, `--check must exit 1 on drift; got ${run.status} (0 = the silent no-op this guards)`))
})

await test('--check detects drift when the script is unlinked after launch (#7214)', async () => {
  withBrokenRealpath('unlink', (run) =>
    assert(run.status === 1, `--check must exit 1 on drift; got ${run.status} (0 = the silent no-op this guards)`))
})
```

Needs `chmodSync`, `realpathSync`, `mkdtempSync` added to the existing `node:fs` import and `tmpdir` from `node:os` (already imported).

Mutation result: **passes** on this branch, **fails both cases** with the guard reverted to realpath-first, while `gen-agents-md.test.mjs` stays green throughout — which is the gap.

|---|---|---|
| The undecidable residual (paths differ AND realpath fails) still returns a bare `false` with no output. Undecidable need not mean silent — warn on stderr and keep returning `false`. | [#7226](https://github.com/blamechris/chroxy/issues/7226) | Behaviour change beyond the reordering; wants its own test and should land in the shared helper once #7222 consolidates. |
| `bump-version.sh:275-278` prints "AGENTS.md regenerated from CLAUDE.md" unconditionally. `set -euo pipefail` catches only a *non-zero* exit, and this whole defect class is exit **zero** having done nothing — so in the exact failure mode it worries about, it prints a false success. Also `ROOT` uses logical `pwd`, retaining symlink components. | [#7227](https://github.com/blamechris/chroxy/issues/7227) | Different file, different blast radius (release tooling); the caller verifying its own postcondition is a separate layer from the guard. |

### Architecture Notes

The guard is a CI-integrity control, not product code: `--check` is the gate that keeps `AGENTS.md` an honest mirror of `CLAUDE.md`, and `bump-version.sh` regenerates through the same entry point during a release. Both consume it as an exit code, so the failure mode is not a crash but a **false green** — the class this repo treats as its most expensive. That is what makes the missing test disproportionate to the three-line diff: the change is correct today, and nothing in the repo will notice when it stops being.

The right end state is #7222 — one implementation, one test — at which point this file imports the helper and the fourth copy disappears. Until then the inline copy needs its own pin.

### Verdict

- [x] **Request Changes** - Issues must be addressed

The fix is right and I would happily approve the diff as-is. What blocks it is that the PR asserts coverage that does not exist, on the strength of a reproducibility claim that does not hold — and the mutation test confirms the practical consequence: revert the fix, everything stays green. The remedy is small and drafted above: land the test, and correct the two statements in the body.

blamechris:
Both blocking findings addressed in `48abee1ca`. Both were correct, and the second was a genuine refutation rather than a style point.

## 1. The coverage claim was false

Right — `packages/server/tests/is-entry-point.test.js` covers the shared helper, a **different copy**. Reverting this file's reordering left the whole repo suite exiting 0, exactly as the mutation showed. The change was unpinned and the body asserted otherwise.

## 2. "Not reproducible in a subprocess" was wrong

This is the useful one. The premise was true and the conclusion did not follow — node only has to *load* the script, not keep it readable through evaluation, and a `module.register` load hook sits precisely in that window. Landed as drafted, with two additions:

**A positive control.** Without it both tests pass for the wrong reason: if the hook fails to break realpath, the guard simply works, `--check` sees the stale `AGENTS.md`, and exits 1 anyway. The control asserts `realpathSync` actually throws afterwards, so the trigger is proven before anything is concluded from it.

**A root skip on the EACCES case.** `chmod 0000` does not deny root, so as root that test would pass without exercising the path at all — the vacuous pass this file exists to distrust. Skipped rather than quietly weakened.

Mutation result:

```
baseline   exit 0   7 passed, 0 failed
mutant     exit 1   FAIL --check detects drift when the script is unlinked after launch (#7214)
                    FAIL --check detects drift when realpath fails with EACCES (#7214)
                    both: "got 0 (0 = the silent no-op this guards)"
restored   exit 0   7 passed, 0 failed
```

Your note about `realpathSync(tmpdir())` was load-bearing and is kept with the comment explaining why: leaving `/var` symlinked trips #7198's case simultaneously, and the two triggers together are the residual this fix does not claim to cover.

## 3. Nitpick + body

Comment now says realpath covers "a symlink, or case normalisation on a case-insensitive filesystem (default APFS/NTFS)".

Body rewritten: the two false claims are replaced, both are recorded explicitly as corrections rather than quietly dropped, and the Scripts Tests list now has all **seven** steps (it listed five — the two lint golden tests were missing, and base drift doesn't explain it since #7122 predates this branch).

All seven verified locally on Node 22, exit 0 each.

#7226 and #7227 both look right to me as scoped. #7227 is done in #7229 — the caller now runs `--check` and fails loudly, mutation-verified against a generator patched to exit 0 without writing, and `ROOT` moved to `cd -P`/`pwd -P`.

blamechris:
## Re-verification of the remedy — **Approve**

Scoped to whether the remedy for the two blocking findings works, not a fresh review of the whole PR. Everything below was verified by mutation on `48abee1c`, not by reading. macOS, Node v22.22.3, uid 501.

**Both blocking findings are genuinely fixed.** One new claim introduced by the remedy does not hold up; it is comment-only and non-blocking, but it is worth a one-line fix (see N1).

---

### 1. The guard is now actually pinned ✅

Reverted the reordering in `scripts/gen-agents-md.mjs` to the pre-fix realpath-first form (`git show a0f75b9c9:` — diff confirmed to touch only the guard), then restored by `cp`:

| variant | exit | result |
|---|---|---|
| `48abee1c` as-is | 0 | 7 passed, 0 failed |
| realpath-first restored | **1** | both #7214 tests FAIL |
| restored | 0 | 7 passed, 0 failed |

Failure text is verbatim what the body claims:

```
FAIL --check detects drift when the script is unlinked after launch (#7214): --check must exit 1 on drift; got 0 (0 = the silent no-op this guards)
FAIL --check detects drift when realpath fails with EACCES (#7214): --check must exit 1 on drift; got 0 (0 = the silent no-op this guards)
```

The prior finding — "reverting the reordering left the whole repo suite exiting 0" — no longer holds. This file now carries the pin.

### 2. The positive control is real, and load-bearing ✅

Made the `load` hook a no-op (both break branches removed):

```
FAIL the load hook genuinely breaks realpath (positive control): the unlink hook did not make realpathSync throw — the tests below would pass vacuously
  ok --check detects drift when the script is unlinked after launch (#7214)
  ok --check detects drift when realpath fails with EACCES (#7214)
6 passed, 1 failed   exit=1
```

Exactly the predicted vacuity: both #7214 tests pass for the wrong reason, and only the control catches it. Without the control this suite would have exited 0 while testing nothing. This also settles the second blocking finding — the EACCES trigger *is* reproducible in a subprocess.

### 3. The root skip is correct, and not hypothetical ✅

- `chmod 0000` does not deny root — correct, and already established in this repo. `.github/workflows/ci.yml:109` says "it is root-in-container that bypasses them", and `POSIX_PERM_SKIP` (`packages/server/tests/test-helpers.js:117`) exists for the same reason (#6075). The predicate here matches its root arm exactly.
- **It will fire in CI.** `Scripts Tests` routes through `runner-target` to the same mixed self-hosted Linux pool documented at `ci.yml:98-102`: Mac Docker members run as **uid 0**, the winbox member as uid 1001. So the EACCES case skips on the Docker members and runs on winbox. The skip is load-bearing, not defensive decoration.
- Does not misfire when non-root: at uid 501 the test runs (`ok ...`). Forcing `process.getuid = () => 0` via `--import` (no file mutation) produces the skip path:
  ```
  skip --check detects drift when realpath fails with EACCES (#7214): running as root
  6 passed, 0 failed   exit=0
  ```
  Visible, same `  `-indent as `ok`/`FAIL`, and it does not inflate the pass count.

### 4. Filesystem hygiene ✅

Zero leaked temp dirs and zero `0000`-mode dirs after two consecutive green runs **and** after the two failing mutation runs (the `finally` runs on the throw path too).

The `chmodSync(scriptsDir, 0o755)` restore is load-bearing, not cosmetic — replicated the eacces layout standalone:

```
A (no chmod restore):   rmSync threw = EACCES scandir
B (with chmod restore): rmSync threw = no
```

Without it, cleanup fails and the tree leaks.

### 5. All seven `Scripts Tests` steps ✅

The list matches `ci.yml:935-957` exactly, and in order. Each run individually, exit code captured per command:

```
bash scripts/__tests__/bump-version.test.sh                exit=0   18 passed, 0 failed
node scripts/__tests__/flush-and-exit.test.mjs             exit=0    5 passed, 0 failed
node scripts/__tests__/compile-skill-targets.test.mjs      exit=0   18 passed, 0 failed
bash scripts/__tests__/lint-no-raw-color-literals.test.sh  exit=0   PASS — all 4 cases
bash scripts/__tests__/lint-no-nul-bytes.test.sh           exit=0   PASS — all 7 cases
node scripts/__tests__/gen-agents-md.test.mjs              exit=0    7 passed, 0 failed
node scripts/__tests__/check-release-pr-subject.test.mjs   exit=0   23 passed, 0 failed
```

End-to-end also confirmed: `node scripts/gen-agents-md.mjs --check` exit 0 "in sync", and the same through a symlinked path exit 0 "in sync". CI is green on `48abee1c` across all 21 reporting checks.

### 6. Body claims

Every assertion in the rewritten body checks out — mutation table (exact exit codes and failure text), the positive-control rationale, the root-skip rationale, the seven-step list, and the end-to-end runs. Both corrections are accurate: `packages/server/tests/is-entry-point.test.js` exists on `origin/main`, contains no reference to `gen-agents-md`, and covers the shared helper only.

---

## Non-blocking

**N1 — the new case-insensitive clause does not hold, and contradicts the canonical copy.** The added comment says realpath is needed when paths differ "via a symlink, **or via case normalisation on a case-insensitive filesystem (default APFS/NTFS), both of which it resolves**". On APFS (confirmed case-insensitive here) neither half reproduces:

- The two sides never diverge by case. Invoking `node <root>/SCRIPTS/probe.mjs` where the real dir is `scripts/` gives `import.meta.url` = `.../SCRIPTS/probe.mjs` and `resolve(argv[1])` = `.../SCRIPTS/probe.mjs` — **plain equal = true**. The fallback is never reached. A variant with the realpath fallback deleted entirely still fires the guard on a case-variant path (and correctly silently no-ops on a symlinked path, which is the positive control for that probe).
- `realpathSync` would not fix a case difference anyway. This file imports the **non-native** one (`gen-agents-md.mjs:15`), which preserves the case as given: `realpathSync(.../SCRIPTS/f.txt)` returns `SCRIPTS`, not `scripts`. Only `realpathSync.native` case-normalizes.

More to the point, `packages/server/src/utils/is-entry-point.js` on `origin/main` — the canonical implementation of this identical logic, landed by #7217 — documents the opposite: "the **sole** reason they can differ while still naming the same file is a symlink somewhere in the invocation path." Two copies of one piece of reasoning now disagree, and the one that disagrees is the copy that is wrong. Suggest just dropping the clause so this comment reads like the helper's.

Flagging it plainly because it is the third unverified claim in this PR's lineage, and the two before it were the blocking findings. One line, cheap to delete.

**N2 — the positive control only covers the `unlink` trigger.** Neutering *only* the eacces break (`chmod 0o755` instead of `0o000`) leaves all 7 green, exit 0. Low risk, since the unlink test carries the pin and is controlled, but the eacces test can silently go vacuous. A control inside `fn` — assert `realpathSync(join(scriptsDir, 'gen-agents-md.mjs'))` throws before the run returns — would cover both triggers directly and is a strictly better shape than the out-of-line `probe.mjs` indirection.

**N3 — the skip covers root but not Windows.** `POSIX_PERM_SKIP` guards both arms (`process.platform === 'win32'` *and* uid 0) because "Windows ACLs don't map to chmod bits". This copy has only the root arm. Moot for CI — `Scripts Tests` never routes to a Windows runner — but a local Windows run gets a vacuous pass on the EACCES case.

**N4 —** the control's first block (`withBrokenRealpath('unlink', run => assert(run.status !== null))`) is a redundant full spawn: the following tests already assert `run.status === 1`, which fails if the child never ran. The probe below it is the actual control.

**N5 —** if `mkdirSync(scriptsDir)` ever throws, the `finally`'s `chmodSync(scriptsDir, ...)` throws `ENOENT`, masks the original error, and skips `rmSync`. Cosmetic.

**N6 —** the branch is 4 behind `origin/main` and predates #7217, which is why the comment in N1 could drift from the helper it duplicates. Not blocking (`mergeStateStatus: CLEAN`).

---

**Verdict: Approve.** The remedy does what it claims — the guard is pinned, the pin is controlled against vacuity, the skip is justified and visible, and the suite is clean and repeatable. N1 is a one-line comment fix worth making before merge.

<sub>Verified in an isolated worktree; all mutations restored by `cp`, `git status` clean, no temp dirs leaked.</sub>

blamechris:
N1 taken — the clause is deleted in `e6cb2b3df`. You were right on both halves, and I verified before removing rather than taking it on report:

```
actual directory:  SCRIPTS
realpathSync('…/scripts/x.mjs') -> '…/scripts/x.mjs'      (not normalised)
```

`gen-agents-md.mjs:15` imports the non-native `realpathSync`, which does not case-normalise — only `realpathSync.native` does. And the two sides cannot diverge by case anyway, since `import.meta.url` preserves the path as typed, so the plain comparison already matches.

The contradiction with `packages/server/src/utils/is-entry-point.js` is the part that actually bothers me: two spellings of one rule disagreeing in comments is precisely how these four copies drifted into four different bugs, which is what #7222 exists to end. Reverted to main's wording with a note that the match is deliberate.

That clause came from a review nitpick I applied without checking — which makes it, as you say, the third unverified claim in this PR's lineage. The first two were caught by you, this one by you again. The pattern is consistent enough to be worth naming: every one of them was a *plausible mechanism asserted from a result*, never from an observation.

On your two other flags, both fair and left as-is deliberately:

- **The control covers only the `unlink` trigger.** True — neutering just the eacces break leaves all 7 green. The control's job is to prove the hook mechanism works at all, which is the vacuity that would silently void both tests; a per-trigger control would be stricter and I would take it in the consolidation work.
- **The skip has a root arm but not a Windows arm.** Correct, and moot for CI since `Scripts Tests` never routes to the Windows pool — which is #7215's actual subject. Worth carrying if that Windows leg ever lands.

Thank you for checking whether the skip will really fire: `Scripts Tests` routing to a pool whose Mac Docker members run as uid 0 is the detail that makes it load-bearing rather than defensive, and I had assumed it was the latter.

7/7 gen-agents-md tests still pass after the removal.

Issue #7239, opened 2026-08-17
Extend the config-dir guard beyond packages/server/src


blamechris:
## Context

Found during review of PR #7238 (#7052). The `lint-config-dir` gate scans only `packages/server/src`, but its CI step is named "Check ~/.chroxy resolves through config-dir.js" — which overclaims, since the guard structurally cannot see the surfaces below.

## Live resolvers outside the guarded scope

- **`packages/claude-hooks/src/config.js:36`** — `return env.CHROXY_CONFIG_DIR || join(homedir(), '.chroxy')`. Correct today, and it must stay hand-rolled because claude-hooks is deliberately zero-runtime-dep and cannot import the server's accessor. But it is exactly the "each copy is correct in isolation and each is a place the convention can drift" shape the lint's own docstring argues against. Pointing the lint at that directory flags it immediately, so the shape is detectable — only the scope excludes it.
- **`scripts/docker-entrypoint.sh:4`** — `CONFIG_DIR="$HOME/.chroxy"`, a shell copy that ignores the override outright. `docker run -e CHROXY_CONFIG_DIR=…` sends the daemon somewhere the entrypoint never writes, so its token-preservation block finds no `apiToken` and **mints a fresh token on every restart**, while `docker-compose.yml:39` keeps the `chroxy-data` volume mounted at a path the daemon no longer uses. Silent, and it looks like it worked. Fix is `CONFIG_DIR="${CHROXY_CONFIG_DIR:-$HOME/.chroxy}"`.

## Guard hardening (from adversarial fixture testing)

The rules are single-line regexes over a per-line window. These evade and are worth closing, roughly in reachability order:

- `const home = homedir()` on one line, `join(home, '.chroxy', …)` on another — **the two-step idiom already exists** at `ws-file-ops/browser.js:24` and `devcontainer-config.js:206` (neither is a chroxy path today, so nothing is broken — but the shape is live in the codebase).
- Module-scope capture inside a multi-line object literal (`supervisor.js:208-212` is this idiom one scope down), a lazy/memoized singleton, or a class static field.
- A template-literal path — `` `${homedir()}/.chroxy/x` `` has no quoted `.chroxy`, so `CHROXY_SEGMENT` misses it.
- `join(homedir(), '.chroxy/logs/x')` with combined segments; `join(process.env.HOME, …)` (`tests/smoke-test.mjs:27` does this).
[...]

Issue #7052, opened 2026-07-25
fix(server): CHROXY_CONFIG_DIR relocates only half the daemon's state (session-state family is home-rooted)


blamechris:
## Summary

`CHROXY_CONFIG_DIR` is documented and treated as "relocate all persistent state", and roughly ten
modules honour it. The **session-state family does not**. The two groups therefore disagree about
where `~/.chroxy` is whenever the variable is set to anything other than `$HOME/.chroxy`, and
nothing in the test suite can currently detect it.

Surfaced while triaging a review comment on #7013 (`chroxy schedule`), which proposed making the
CLI's `CONFIG_DIR` honour the variable. That change would have been actively harmful — see
"Why the obvious fix is wrong" below.

## Honours `CHROXY_CONFIG_DIR`

`models.js:166,178,315` · `connection-info.js:7` · `device-preferences.js:42` · `doctor.js:41`
`shell-approval-info.js:20` · `byok-compose-state.js:66` · `event-ingest.js:162`
`snapshots-store.js:65` · `checkpoint-manager.js:20` · `orchestration/git-ops.js:69`
`cli/tokens-cmd.js:20` · `ws-server.js:633` · `server-cli.js:1008` · `claude-hooks/src/config.js:36`

`byok-compose-state.js:62` even documents it as the convention: "Honors CHROXY_CONFIG_DIR like the
rest of…".

## Does NOT honour it (home-rooted)

| Path | Location | Notes |
|---|---|---|
| `~/.chroxy/session-state.json` | `session-manager.js:47` (`DEFAULT_STATE_FILE`) | `server-cli.js` never passes `stateFilePath` — the string does not appear in that file — so the daemon always uses this default |
| `~/.chroxy/scheduled-tasks.json` | `session-manager.js:685` via `defaultScheduledTasksPath(this._stateFilePath)` | inherits the above |
| `~/.chroxy/supervisor.pid` | `supervisor.js:37` (`DEFAULT_PID_FILE`) | no production caller passes `pidFilePath` |
| `~/.chroxy/known-good-ref` | `supervisor.js:66` | same |
| `~/.chroxy/config.json` | `server-cli-child.js:24` | the daemon child's own config read |
| `~/.chroxy/…` (all CLI commands) | `cli/shared.js:13-14` (`CONFIG_DIR`/`CONFIG_FILE`) | consumed by `schedule`, `session`, `deploy`, `init`, `tunnel`, `config`, `providers`, `worktree-gc` |

[...]

PR #7245, opened 2026-08-17
fix(desktop): honour CHROXY_CONFIG_DIR, and document it as the config root

### LINKED ISSUE #7241: Docs and desktop still pin ~/.chroxy as fixed
## Context

Found during review of PR #7238 (#7052). `CHROXY_CONFIG_DIR` went from a half-working accident to a CI-gated, fully-honoured feature, but the docs and the desktop app were not updated.

## Docs

- **`packages/server/CONFIG.md`** — the canonical reference. `CHROXY_CONFIG_DIR` appears in neither the "Env-only settings" table nor the "Direct reads" paragraph, while the file asserts `~/.chroxy/…` as *the* location in ~14 places. `packages/server/.env.example:5` likewise only names `~/.chroxy/config.json`.
- **`docs/security/bearer-token-authority.md`** §2/§7/§11 pin `~/.chroxy/config.json`, `~/.chroxy/session-tokens.json` and `~/.chroxy/pages/<slug>/` as fixed; only the ingest-secret row carries the "(`CHROXY_CONFIG_DIR` honored)" note, which is now true of every row.
- **`docs/design/orchestration/engine.md:4`** asserts "`~/.chroxy` is **NOT** in `FORBIDDEN_HOME_SUBDIRS`" — factually wrong (`handler-utils.js:217` has it), pre-existing, and directly load-bearing for the cwd deny-list finding PR #7238 fixed. It sits in a section labelled "key confirmations found while verifying".
- **`docs/design/orchestration/engine.md:49,380`** prescribe `createRunStore({ filePath: join(homedir(), '.chroxy', 'orchestration-runs.json') })` — a shape the new lint now bans. `run-store.js` does not exist, so this is unimplemented spec, but an implementer following it writes code CI rejects.
- **`docs/design/orchestration/verification-consistency.md:11`** F4 describes the engine/metering split as open; the config-dir half is now settled repo-wide.
- **`packages/server/src/worktree-gc.js:262,279`** and **`packages/claude-hooks/README.md:44,47`** still say `~/.chroxy/worktrees` in prose.

## Desktop (Tauri/Rust)

`config.rs:28`, `settings.rs:68`, `qrcode.rs:103` and `lib.rs:2307` each hardcode `~/.chroxy/…` and never read `CHROXY_CONFIG_DIR`. The autostart path is internally consistent (`server.rs` spawns the embedded server with a curated env that omits the var, so both halves stay home-rooted) — this is **not** a live split-brain. But the moment a user runs a relocated daemon separately and uses the tray app as a client, the QR code reads a stale `connection.json`, the settings pane reads the wrong port, and "Reveal in Finder" opens the wrong directory.

## Acceptance criteria

- [ ] `CHROXY_CONFIG_DIR` documented in CONFIG.md (env table + direct reads) and `.env.example`
- [ ] Security docs note the override applies to every listed path
- [ ] Correct the false `FORBIDDEN_HOME_SUBDIRS` claim in engine.md
- [ ] Orchestration design docs stop prescribing the now-banned shape
- [ ] Desktop: add a `CHROXY_CONFIG_DIR` fallback, or record an explicit decision that desktop does not support relocation

blamechris:
Closes #7241.

`CHROXY_CONFIG_DIR` became a CI-gated feature in #7052 / #7238 / #7239, but the docs and the desktop app still presented `~/.chroxy` as a fixed location.

## Docs

- **`packages/server/CONFIG.md`** gains a canonical **The config root** section — `~/.chroxy` is the *default*, the override relocates **all** of it, it must be absolute (a relative value is refused, not resolved), and relocating leaves state behind (`chroxy config-dir status` / `migrate --yes`). Added to the env-only table and the direct-reads paragraph, and to `.env.example`. Every other doc's `~/.chroxy/…` now reads correctly against this one statement rather than needing 170 edits.
- **Security docs** stop pinning their paths. `bearer-token-authority.md` carried the `(CHROXY_CONFIG_DIR honored)` note on the **ingest-secret row alone**, which read as though the other paths were *not* relocatable — the classic partially-pinned shape. The note is now document-level, and it records a wrinkle found while verifying rather than asserted: **mode enforcement is not uniform.** `session-tokens.json` fails closed (`session-token-store.js` refuses a non-0600 file and logs `devices will re-pair`); `ingest-secret` sets `0600` at exclusive create but **never re-checks the mode on read**. Notes also added where relocation changes a doc's own threat reasoning — `credentials-at-rest.md` (§3's "misconfigured sync tool" becomes the configuration if you relocate into a synced dir), `spawned-binary-provenance.md` (the ledger's integrity is reduced to *the directory's* ACL, so a shared mount makes the gate cosmetic), `encryption-threat-model.md` (a non-persistent root re-mints the identity key, so pinned clients refuse to connect — operator-triggered, not an attacker).

### One factual correction

`orchestration/engine.md:4` asserted **"`~/.chroxy` is NOT in `FORBIDDEN_HOME_SUBDIRS`"**. It is — [`handler-utils.js:218`](../blob/main/packages/server/src/handler-utils.js), since `d70954431` (2026-04-12, adversary finding A9), which is **three months before that document was written**. So this was wrong when written rather than drift, it inverted a live security control, and it sat in a list captioned *"key confirmations found while verifying"* — the strongest claim to a reader that someone checked. It was load-bearing for the cwd deny-list gap #7238 fixed.

`engine.md` §1.2/§12 also **prescribed `join(homedir(), '.chroxy', …)`** — the exact shape `lint-config-dir.mjs` now fails CI on — so an implementer following the spec wrote code the gate rejects. Paths corrected in place, and the section marked **SUPERSEDED** per `verification-consistency.md` F4's already-adjudicated resolution (RunLedger wins; `run-store.js` is not to be created). F4's config-dir half is marked closed; the rest is flagged still-open rather than silently folded in.

## Desktop

`config.rs`, `settings.rs`, `qrcode.rs` and the Finder-reveal action each resolved the root from `dirs::home_dir()` and never read the override. They now share **one** `config::config_dir()` resolver mirroring `config-dir.js` — read **per call**, because a `OnceCell`/`lazy_static` here would reproduce the #7052 freeze bug in Rust. `setup.rs` and the `allowAutoPermissionMode` commands inherit it via `config_path()`.

**The issue's premise about the autostart path was wrong, and it matters.** #7241 says `server.rs` spawns the embedded server with "a curated env that omits the var, so both halves stay home-rooted". There is no `env_clear()` anywhere in the desktop crate, so the child **inherits** `CHROXY_CONFIG_DIR`: the server read the relocated root while the tray read `~/.chroxy`. `server.rs:996` already `env_remove`s `CHROXY_HOST` for precisely this reason ("so an inherited CHROXY_HOST … can't silently pin the server"), so the codebase knew omitting ≠ clearing and applied it to one variable only. Silent for #7239's reason: `API_TOKEN` is passed explicitly in the spawn env, so the token survives and only the *other* state diverges. Honouring the var on both sides resolves this without needing an `env_remove`.

Sharpest user-visible symptom, now fixed: the QR code encoded a stale `connection.json` — silently, because a stale file still parses.

## Testing

7 new Rust tests. Semantics are driven through a **pure** `resolve_config_dir(raw, home)` so the suite never mutates process env — `cargo test` runs tests as threads in a *single* process, so a `set_var` would race every other path-resolving test.

The cross-site guard *does* set the variable, behind a lock, and that is deliberate: **with `CHROXY_CONFIG_DIR` unset, `config_dir()` IS `~/.chroxy`**, so a guard asserting against it would pass against the very mutant it exists to catch — false safety of exactly the kind `docs/false-safety-guards.md` describes. Proven by mutation, restoring from a scratchpad copy rather than `git checkout --`:

| Mutation (the literal pre-fix body) | Result |
|---|---|
| `settings.rs::path()` → `dirs::home_dir()…` | guard **FAILS**, all 229 others pass |
| `qrcode.rs::connection_info_path()` → `dirs::home_dir()…` | guard **FAILS** (exit 101) |

- `cargo test --locked`: **230 pass, 0 fail** (exit 0)
- Server tests (config-dir + worktree-gc, 7 files): **165 pass, 0 fail** (exit 0)
- All 6 custom server shell lints + `lint-config-dir.mjs`: **PASS**
- Every relative link added verified to resolve

Desktop Rust needed CI's own prep step (`mkdir -p server-bundle ../dist`, `ci.yml:838`) to build in a worktree; the `swift/speech-helper` binary the build re-signs was restored rather than committed.


blamechris:
Both review comments addressed in `1e5d9cfb3`, and the second was a real correctness bug rather than a wording nit — thanks.

**`config.rs` — warning promised a fallback it may not have.** Correct catch. The message hardcoded `Using the default ~/.chroxy instead`, but when no home resolves (possible in a GUI launch) `resolve_config_dir` returns `None` and there *is* no default, so the text would have sent an operator looking for state in a directory nothing wrote to. It now names the directory actually in use and only claims one when it exists:

```rust
match &fallback {
    Some(dir) => eprintln!("… not an absolute path. Using {} instead.", dir.display()),
    None      => eprintln!("… not an absolute path, and no home directory could be \
                            resolved — no config directory is available."),
}
```

This also brings it closer to the daemon, which interpolates `defaultConfigDir()` for the same reason. The return value was already correct and covered by `resolve_config_dir_no_home_no_override_is_none` (which asserts `None` for a relative value with no home) — only the operator-facing text was wrong.

**`settings.rs` — stray bracket in the doc comment.** Fixed by unnesting the intra-doc link from the parenthetical. I also checked whether it was a *broken* link rather than just an odd render: `cargo doc -W rustdoc::broken_intra_doc_links` is clean. While there I fixed the same rendering class one file over — `[packages/server/src/config-dir.js]` in `config.rs` was wrapped in brackets as though it were an intra-doc link, but a file path is not a Rust item so it rendered as literal brackets. Now plain backticks.

Verification: `cargo test --locked` → **232 pass, 0 fail**. The `#[cfg(windows)]` branch was re-checked by temporarily inverting its cfg so this host type-checks and runs it — cfg'd-out code is parsed but never type-checked, which is how the first commit on this PR shipped a Windows-only break from a green macOS run.

PR #7247, opened 2026-08-17
fix(ci): fail on a FOURTH hand-rolled entry-point guard, not just on drift

### LINKED ISSUE #7235: Nothing prevents a FOURTH hand-rolled entry-point guard appearing
## Context

Surfaced by the review of #7232 (which closed #7222).

#7232 added a drift gate keeping the three surviving copies of the entry-point guard identical: `scripts/lib/is-entry-point.mjs`, `packages/server/src/utils/is-entry-point.js`, `packages/server/sidecar/agent.js`. That closes the *current instance*.

It does not close the **failure mode**. `GUARD_COPIES` in `scripts/__tests__/is-entry-point.test.mjs` is a hardcoded three-element list, so the gate is silent when a *fourth* copy appears — which is exactly how #7198 happened ("the same guard hand-rolled in four files"). The list was verified complete at the time of review.

## Suggested fix

A grep lint in the style of the existing `packages/server/scripts/lint-*.sh`, failing on a hand-rolled `realpathSync` + `process.argv[1]` comparison anywhere outside the three sanctioned copies. The repo already has that idiom and wires those lints into **Server Lint**.

## Acceptance criteria

- [ ] A new hand-rolled guard outside the three known copies fails CI
- [ ] The three sanctioned copies and the tests are exempt
- [ ] Wired into a CI job that runs on PRs

blamechris:
Closes #7235.

## The gap

#7222 added a drift gate keeping the three surviving copies of the entry-point guard identical. It iterates `GUARD_COPIES`, a **hardcoded three-element list**, so it is silent when a *fourth* copy appears — which is exactly how #7198 happened (the same guard hand-rolled in four files, **all four wrong**, failing silently: the guard reads false, `main()` never runs, the process exits 0).

That closed the instance, not the failure mode. `docs/false-safety-guards.md` calls this "the list that stopped growing" and cites #7192 and #7197 for the same shape.

## What this adds

`packages/server/scripts/lint-entry-point-guard.mjs` — a **repo-wide** walk (~1900 JS/TS files, 0.4s) that fails on any entry-point determination outside the sanctioned set. Wired into **Server Lint**, which is a required check.

Deliberately repo-wide, not `packages/server`: two of the three sanctioned copies already live outside it, and so could a fourth. The step name says exactly what is covered (#7239's lesson about a gate overclaiming in its own title).

Four shapes are banned, matched on comment-stripped source with `\s*` spanning newlines:

| kind | why |
|---|---|
| `argv1-index` | the shape all four #7198 copies had |
| `argv1-at` | the same read spelled `.at(1)` |
| `argv1-destructure` | `const [, script] = process.argv` binds the slot without naming an index |
| `import-meta-main` | `undefined` on Node 22.0–22.17; the declared floor is `>=22`, so it is a falsy guard on a supported runtime — the same silent no-op, as a version skew no `node-version: 22` job would show |

Documented non-matches, each with a test and a positive control: `require.main === module` (CJS compares module objects, no symlink trap), ordinary `slice(2)` parsing, a rest element at position 1, `.sh` sources (`docker-entrypoint.sh` uses the same slot for a config path), and prose in comments.

## The two gates now read one list

Two hardcoded lists of the same three files is this bug one level up: add a fourth copy to the walk's allowlist, forget the drift gate, and the new copy is exempt from **both**. `scripts/lib/entry-point-guard-copies.mjs` is now the single list, imported by the drift gate and the walk.

The asymmetry that makes hardcoding safe *here*: the walk's list is the set **exempt**, so falling behind fails loudly. The drift gate's was the set **checked**, so falling behind was quiet. Same structure, opposite failure direction — sharing it means the safe direction covers the unsafe one. The allowlist is ratcheted both ways: an entry whose file no longer contains a guard fails as stale, and a sanctioned path that has vanished exits 2.

Exit 2 is distinct from 1 throughout — a bad flag, a missing sanctioned path, or a walk below its `--min-files` floor means the **guard** broke, and must not be reported as "clean" or as "dirty".

## Proven red before being trusted

Not "the tests pass" — the CI step's own shell body, run against real mutants in the real tree, checking **exit codes**:

| mutant | step exit |
|---|---|
| clean tree | 0 |
| 4th guard in `scripts/` (the #7198 `pathToFileURL` shape) | **1** |
| 4th guard in `packages/dashboard/` (the *fixed* realpath shape, copy-pasted) | **1** |
| `import.meta.main` in a `.ts` under `packages/store-core/` | **1** |
| a sanctioned copy renamed (lint exits 2) | **1** |

36 unit tests drive the same walk against fixture trees via `--repo-root`, so there is no second enumeration path that only tests exercise. Every exemption carries a positive control.

### One removal worth naming

A raw-text prefilter (skip files whose raw source has neither `argv` nor `import.meta.main`) was written and then removed. It is unsound in the one direction that matters: a comment *inside* the expression — `import./* */meta.main` — leaves the raw text without the substring while the stripped text still matches, so the file is skipped and the tree reports clean. That is "silently skipped an input", the fourth false-safety mode. It bought 260ms (0.14s vs 0.40s). The test for that case **fails against the prefilter and passes without it**, verified by re-adding it.

## Also in this PR, deliberately

- **`packages/server/scripts/lib/strip-comments.mjs`** — extracted verbatim from `lint-claude-family-explicit.mjs`, which now imports it. Writing a fourth copy of a comment stripper in the PR whose subject is "stop the fourth hand-rolled copy" seemed like poor form. The two other lints carry divergent versions (different signatures and string-awareness); migrating them is real work and is filed separately rather than smuggled in here.
- **`docs/false-safety-guards.md`** — its "Derive the set; never hardcode it" rule gains the complement this PR is an instance of (when the set cannot be derived, invert the list and walk everything). Its `is-entry-point.js` bullet was stale: it says #7222 is open and that the server's copy is the only unit-tested one. Both were true when written on 2026-08-16 and superseded by #7232 the same day.

## Verification

- `scripts/lint-entry-point-guard.sh` — OK, 1903 files walked
- all 7 custom server shell lints — pass
- `packages/server`: `lint-entry-point-guard` (36) + `lint-claude-family-explicit` + `lint-config-dir` + `is-entry-point` — 107 pass / 0 fail
- `scripts/__tests__/is-entry-point.test.mjs` (the drift gate, now importing the shared list) — 18 pass / 0 fail
- `npm run lint` (eslint) — 0 errors

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -15,3 +15,9 @@
 packages/server/node_modules
 packages/server/tests
 packages/server/coverage
+# Dev-only lints. They are never run in the image, and lint-entry-point-guard.mjs
+# imports repo-root scripts/lib/, which the image does not copy — so shipping
+# them would put a module with an unresolvable import inside the container
+# (inert, but pointless). fix-node-pty-helper.js is the one the build needs.
+packages/server/scripts/lint-*
+packages/server/scripts/lib
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -421,6 +421,40 @@ jobs:
             exit 1
           fi
 
+      - name: Check no hand-rolled entry-point guard exists (whole repo, JS/TS only)
+        run: |
+          # Issue #7235 — #7198 was one guard hand-rolled in four files, three
+          # of them wrong, and its failure mode is SILENCE: the guard reads
+          # false, main() never runs, the process exits 0. #7222 added a drift
+          # gate keeping the three surviving copies identical, but that gate
+          # iterates a hardcoded three-element list, so a FOURTH copy is
+          # invisible to it. This lint is the other half: it walks the repo and
+          # fails on any entry-point determination outside the sanctioned set.
+          #
+          # Deliberately REPO-WIDE and not packages/server: two of the three
+          # sanctioned copies live outside it, and so could a fourth. The step
+          # name says what is actually covered — .js/.mjs/.cjs/.ts/.tsx/.jsx.
+          # Shell is NOT walked (docker-entrypoint.sh's `node -e` uses the same
+          # argv slot for a config path), and neither is generated dist/.
+          #
+          # Exit 2 means the lint could not do its job (bad flags, a sanctioned
+          # path that vanished, or it walked fewer files than its floor) and is
+          # reported separately from exit 1 — "the guard broke" must never be
+          # shown as "the guard passed", nor as "the code is dirty".
+          set +e
+          scripts/lint-entry-point-guard.sh
+          rc=$?
+          set -e
+          if [ "$rc" -eq 0 ]; then
+            echo "OK — entry-point-ness is decided in the sanctioned files only"
+          elif [ "$rc" -eq 2 ]; then
+            echo "::error::lint-entry-point-guard could not run (bad flags, a stale allowlist entry, or it walked fewer files than --min-files). The GUARD is broken, not necessarily the code. See packages/server/scripts/lint-entry-point-guard.mjs."
+            exit 1
+          else
+            echo "::error::A file decides 'was this module run directly?' by hand, or a sanctioned-copy allowlist entry is stale. Import isEntryPoint instead. See packages/server/scripts/lint-entry-point-guard.sh."
+            exit 1
+          fi
+
   dashboard-tests:
     name: Dashboard Tests
     needs: runner-target
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -421,21 +421,24 @@ jobs:
             exit 1
           fi
 
-      - name: Check no hand-rolled entry-point guard exists (whole repo, JS/TS only)
+      - name: Check no hand-rolled entry-point guard exists (JS/TS sources, excluding generated dist/)
         run: |
-          # Issue #7235 — #7198 was one guard hand-rolled in four files, three
-          # of them wrong, and its failure mode is SILENCE: the guard reads
-          # false, main() never runs, the process exits 0. #7222 added a drift
-          # gate keeping the three surviving copies identical, but that gate
-          # iterates a hardcoded three-element list, so a FOURTH copy is
-          # invisible to it. This lint is the other half: it walks the repo and
-          # fails on any entry-point determination outside the sanctioned set.
+          # Issue #7235 — #7198 and the #7213 sweep found one guard hand-rolled
+          # in four files, and ALL FOUR were wrong. Its failure mode is SILENCE:
+          # the guard reads false, main() never runs, the process exits 0. #7222
+          # added a drift gate keeping the three surviving copies identical, but
+          # that gate iterates a hardcoded three-element list, so a FOURTH copy
+          # is invisible to it. This lint is the other half: it walks the tree
+          # and fails on any entry-point determination outside the sanctioned
+          # set.
           #
-          # Deliberately REPO-WIDE and not packages/server: two of the three
+          # Deliberately not scoped to packages/server: two of the three
           # sanctioned copies live outside it, and so could a fourth. The step
-          # name says what is actually covered — .js/.mjs/.cjs/.ts/.tsx/.jsx.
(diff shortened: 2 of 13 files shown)


blamechris:
## Review round: 3 agent reviewers + Copilot

Copilot: 9/9 files, 0 comments. Three agent reviewers (adversarial evasion, correctness/integration, conventions/claim-verification) ran against the real tree. Everything below was **reproduced on exit codes**, not read off the diff, and every fix is in `cb92fee99`.

### The headline: the lint was unsound in both directions, inside itself

Two reviewers converged on one root cause from opposite ends — the hand-written comment stripper could not tell a regex literal from a comment delimiter.

| shape | what the scanner did | consequence |
|---|---|---|
| `u.replace(/\/*$/, '')` | `\/` leaves a `/`, next char is `*` → read as **block-comment open**, blanking the rest of the file | a guard below it is **invisible** (silent false negative) |
| `/(["])/g` | the quote starts a **phantom string state**, so it stops blanking comments | ordinary prose reads as code (**spurious red** in a required check) |

Measured over the 1903 walked files: **83 differed from the truth — 40 hiding real code, 51 leaking comment text.** Reproduced live in both directions: a guard hidden behind a trailing-slash trim, and one added `// NOTE:` comment under `src/utils/win-spawn.js`'s `CMD_META` regex turning Server Lint red.

Telling regex-start from division needs the grammar, so the stripper now asks a parser — TypeScript's, because the walk covers `.ts`/`.tsx` and acorn cannot parse those. 0 parse failures over the same 1903 files, ~2.4s, offsets asserted rather than assumed. `typescript` is now a declared devDependency rather than inherited from another package's hoist, and the stripper loads dynamically so an absent one exits 2 rather than 1.

### The detection hole that was already live

Every rule required the literal `process` before `.argv`. But `import { argv } from 'node:process'` is Node's own documented idiom, and **two scripts in this repo already use it** — `scripts/merge-updater-feeds.mjs` and `packages/server/scripts/spike-mcp-elicitation-shim.mjs`, both executable and guard-less today. A guard added to either would have been spelled `argv[1]` and sailed past. Rules now anchor on the identifier, which also picks up optional chaining, `const { argv } = process`, `import { argv as x }`, and aliases.

### Also closed

- **Whole-file exemption.** A *second* guard could hide inside `sidecar/agent.js` — 1339 lines the drift gate only samples one body of. The three copies are now held to the same site count, with **no hardcoded number**: the drift gate already proves the bodies identical, so the ratchet maintains itself.
- **Ignore marker** was matched on the raw line, so the same text inside a template literal exempted the line below it.
- **Unreadable file** escaped as an uncaught EACCES → exit 1, which the CI step reported as a hand-rolled guard that did not exist. Now exit 2.
- `.mts`/`.cts` were not walked.
- **The wrapper-floor test was itself blind** — it hardcoded `1500` and never read the wrapper. Mutation proved it: setting the wrapper to `--min-files 99999` made the real step exit 2 while the test stayed green. It parses the value out now.

### Accuracy

`#7198` was described as *"three of the four wrong"* in three places, one of them live `console.error` output. [#7217](https://github.com/blamechris/chroxy/pull/7217) says **"There were four... all wrong the same way"**, and the pre-existing `docs/false-safety-guards.md` agrees. Corrected, along with the step title's "whole repo" (65 committed files under `dist/` are not walked), the sidecar Dockerfile claim (it also copies `package-lock.json`), and a stale file count.

Independently verified: `import.meta.main` did land in **Node 22.18.0** against a declared floor of `>=22`, so that rule's justification holds.

### Verification

19 fixture probes on exit codes — 11 evasions that previously passed, 8 shapes that must stay clean — plus all three findings re-tested **in situ on the real tree**. The new stripper tests fail against the old scanner and pass against the parser.

`78` tests across the two new suites; 149 pass across all affected server test files; drift gate 18/18; 7/7 custom server lints; eslint 0 errors; server lockfile fresh.

### Deferred

[#7248](https://github.com/blamechris/chroxy/issues/7248) — `lint-config-dir.mjs` and `lint-session-opt-forwarding.mjs` still carry their own strippers with the same regex blindness. Different signatures, so migrating them is its own change.

PR #7249, opened 2026-08-18
fix(ci): scope the entry-point walk to the repository, not the filesystem

blamechris:
Follow-up to #7247, which I found broken **after** merging it — by running the shipped gate against the main checkout rather than my worktree.

## The bug

#7247's lint walks the filesystem. That is not the same set as "the repository".

`packages/desktop/src-tauri/server-bundle/` is generated and gitignored (`.gitignore:60`). It contains a stale copy of the entire server, from before #7217 removed the hand-rolled guards — so it still has two of the original #7198 ones:

```
packages/desktop/src-tauri/server-bundle/src/server-cli-child.js:165
packages/desktop/src-tauri/server-bundle/src/channels/chroxy-channel-server.js:241
  const isEntryPoint = process.argv[1] && import.meta.url === pathToFileURL(process.argv[1]).href
```

A fresh CI checkout has no bundle, so all 22 checks passed on #7247 while the gate was **already failing in every working copy that had built the desktop app** — reporting guards the developer never wrote, in files they cannot fix.

### Why nothing caught it

Worth recording, because it is the interesting part. My worktree *has* `server-bundle`, but empty — so the lint passed there, the 36 fixture tests passed, and all three in-situ mutation runs passed. Every check I ran was against a tree that happened not to contain the failing input. It surfaced only when I ran the merged gate against `~/Projects/chroxy` as a post-merge verification.

## The fix

Adding `server-bundle` to `SKIP_DIRS` would have fixed this one directory and left the next generated tree to rediscover it — the hardcoded-list failure this lint exists to prevent. So the rule is the real one:

```
git ls-files --cached --others --exclude-standard
```

A guard has to be **committed** to matter. A tracked file still counts even if a broad ignore rule would match it (`--cached` lists it regardless), so `git add -f`'d output is not a hole.

**When git cannot answer** — not a repository, git absent — nothing is filtered. That direction is deliberate: no filter means scanning MORE files, never fewer, so an unavailable git can never hide a guard. It is also what lets the existing tests point `--repo-root` at a plain fixture tree and still exercise this exact code path; that fallback is now asserted rather than left implicit.

`--min-files` is counted after the filter, so it tracks what was actually checked.

## Verification

Against the main checkout, which is where the bug reproduces:

| | exit |
|---|---|
| shipped lint (#7247) on the main checkout | **1** ← the bug |
| this lint on the main checkout | **0** |
| this lint + a real fourth guard there | **1** |
| this lint + a guard planted inside the ignored bundle | **0** |

Four new tests build **real git repos** (`git init`, `gc.auto 0` against the known ENOTEMPTY teardown race) rather than faking ignore rules: ignored guard not reported; positive control with the ignore rule pointed elsewhere; a tracked-but-ignore-matching file still reported; and a non-repo root filtering nothing. The first fails against a filter-disabled mutant.

60 tests in the suite, 102 across all affected server test files, drift gate 18/18, 7/7 custom server lints, eslint clean.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -421,7 +421,7 @@ jobs:
             exit 1
           fi
 
-      - name: Check no hand-rolled entry-point guard exists (JS/TS sources, excluding generated dist/)
+      - name: Check no hand-rolled entry-point guard exists (JS/TS sources in the repository)
         run: |
           # Issue #7235 — #7198 and the #7213 sweep found one guard hand-rolled
           # in four files, and ALL FOUR were wrong. Its failure mode is SILENCE:
@@ -436,9 +436,12 @@ jobs:
           # sanctioned copies live outside it, and so could a fourth. The step
           # name says what is actually covered, because a gate that overclaims
           # in its own title is #7239 — walked are
-          # .js/.mjs/.cjs/.ts/.mts/.cts/.tsx/.jsx; NOT walked are shell
-          # (docker-entrypoint.sh's `node -e` uses the same argv slot for a
-          # config path) and the 65 committed files under generated dist/.
+          # .js/.mjs/.cjs/.ts/.mts/.cts/.tsx/.jsx that git considers part of the
+          # repository. NOT walked: shell (docker-entrypoint.sh's `node -e` uses
+          # the same argv slot for a config path), the 65 committed files under
+          # generated dist/, and anything gitignored — notably
+          # packages/desktop/src-tauri/server-bundle/, a generated copy of the
+          # server that lags its source and still holds pre-#7217 guards.
           #
           # Exit 2 means the lint could not do its job (bad flags, a sanctioned
           # path that vanished, or it walked fewer files than its floor) and is
--- a/packages/server/scripts/lint-entry-point-guard.mjs
+++ b/packages/server/scripts/lint-entry-point-guard.mjs
@@ -65,7 +65,9 @@
  *     entirely. Shell is not scanned and this file's name does not claim it is
  *     (#7239's lesson: a gate must not overclaim in its own title).
  *   - Anything git does not consider part of the repository — gitignored build
- *     output, local scratch files. A guard has to be COMMITTED to matter, and
+ *     output and ignored local scratch files. An untracked file that is NOT
+ *     ignored is still scanned, deliberately: it is on its way to being
+ *     committed. A guard has to be COMMITTED to matter, and
  *     `packages/desktop/src-tauri/server-bundle/` is a generated copy of the
  *     whole server that lags the source by however long since it was last built.
  *     See `gitKnownFiles`.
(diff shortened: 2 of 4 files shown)


blamechris:
## Review round on this PR: 12 agents, 3 dimensions + adversarial verify

19 findings → **5 refuted**, 10 nitpicks, **4 confirmed**. The confirmed ones collapse to three defects, all in code added by this PR. Fixed in `c7fc142c1`.

### 1. The filter dropped files whose path spelling differed — a silent false negative

The comparison was an exact string match, and the two sides come from different producers. macOS git defaults to `core.precomposeunicode=true` and emits **NFC**; `readdirSync` returns the on-disk bytes, which stay **NFD** for a file created that way. Measured here:

```
readdir:        "café.js"  isNFC: false
git ls-files:   "café.js"  isNFC: true
raw equal: false | NFC equal: true
```

So a **committed** file with a decomposed name was dropped from the scan and a hand-rolled guard inside it went unreported — the silently-skipped-an-input mode this lint's own header rejects the raw-text prefilter for, reintroduced one layer down. Both sides normalise to NFC now.

Case is deliberately **not** folded, and the reasoning is in the code: an unrecorded case-only rename still drops the file locally, but folding case would be wrong on the case-sensitive filesystems CI runs on, where two such paths are genuinely different files. CI catches that variant.

### 2. The fail-open was silent

When git cannot answer — the `safe.directory` "dubious ownership" refusal is the common one on self-hosted runners and in containers — the filter is inert and #7247's red comes back with **nothing pointing at git**. Failing open is still right (no filter means scanning *more* files, so it can never hide a guard); it just says so on stderr now.

### 3. `--min-files` counting the filtered set was unpinned

The header claimed it, the code did it, and nothing tested it: moving the filter below the floor check left all 60 tests green. It's load-bearing — a tree with lots of generated output could clear the floor on ignored files while the scanned set had collapsed, which is the exact failure the floor exists to prevent. Pinned.

### A vacuous test, caught

Worth recording since it's this repo's signature defect. My first NFD test **passed against its own mutant** — because the mutation only removed the normalisation from one side: the find/replace targeted `split('/')` where the code says `join('/')`, matched nothing, and exited fine. The test was "verified" against a mutant that was never applied. Caught by grepping for leftover occurrences instead of trusting the replace. Both new tests now fail against a fully-applied mutant and pass against the fix.

### Documentation corrections

- The header promised ignored "local scratch files" were skipped; untracked-but-not-ignored files are deliberately still scanned.
- The CI step title still read "excluding generated dist/" when the gate now excludes every ignored path — the file quotes #7239's don't-overclaim rule 350 lines above, so it should follow its own advice.
- The wrapper's floor comment explained a local-vs-CI file-count divergence that this PR is precisely what eliminates (both are 1903 now).

### Refuted (not acted on)

Five findings did not survive the skeptic pass. The surviving nitpicks — a nested untracked git repo having its files dropped, and case-only renames — are documented as known limits rather than fixed; both are latent (this repo has no submodules and no non-ASCII tracked paths) and CI is case-sensitive.

### Verification

104 tests across affected server suites, 62 in this file, drift gate 18/18, 7/7 custom server lints, eslint clean, `ci.yml` parses.

Issue #7213, opened 2026-08-16
scripts/compile-skill-targets.mjs has the same symlink entry-point-guard bug fixed in gen-agents-md.mjs (#7198)


blamechris:
## Summary

`scripts/compile-skill-targets.mjs:350` guards its CLI side-effects with the exact same vulnerable pattern that PR #7208 (fixing #7198) just removed from `scripts/gen-agents-md.mjs`:

```js
if (process.argv[1] && resolve(process.argv[1]) === fileURLToPath(import.meta.url)) {
  main()
}
```

Node's ESM loader resolves symlinks when computing `import.meta.url`, but `process.argv[1]` is left exactly as the caller typed it, and `resolve()` does not follow symlinks. On macOS, where `/tmp` is a symlink to `/private/tmp`, invoking this script through a `/tmp`-rooted path makes the guard false, `main()` silently never runs, and the process exits 0 having compiled nothing — no error, no non-zero exit, nothing to signal the skip.

## Reproduction

Confirmed directly against the shipped guard expression (isolated mirror, stub `main()`, no repo side effects):

```
$ node /private/tmp/.../root/mirror.mjs      # direct path
MAIN RAN

$ node /tmp/.../link/mirror.mjs              # same file, via a symlinked path
GUARD FALSE -- main() skipped, exiting 0 having done nothing
```

## Why it matters here specifically

`compile-skill-targets.mjs` is not CI-invoked — it's the tool developers/agents run directly and manually per `CLAUDE.md`'s documented workflow (`node scripts/compile-skill-targets.mjs --name <name>`), including from ad-hoc worktrees/scratch directories that are commonly rooted under `/tmp` on macOS. A silent no-op here means `/skill add`/`/skill update` can report success while leaving `.claude/skills/<name>/SKILL.md` (and the other compile targets) stale or missing — the same "reports success while doing nothing" failure shape as #7198, just in the skill-compiler instead of the AGENTS.md generator.

## Suggested fix

Apply the same fix PR #7208 used for `gen-agents-md.mjs`: `realpathSync` both sides (wrapped in try/catch, since `argv[1]` need not exist), and add a regression test mirroring `scripts/__tests__/gen-agents-md.test.mjs`'s new symlinked-invocation case.

## Related (checked, lower priority — not currently exposed but same guard family)

[...]

PR #7251, opened 2026-08-18
test(ci): cover compile-skill-targets.mjs's entry-point call site

### LINKED ISSUE #7236: compile-skill-targets.mjs's entry-point call site has no test coverage
## Context

Surfaced by the test-validity review of #7232.

`gen-agents-md.mjs`'s call site is covered — hardwiring its guard to `false` reddens four subprocess tests. `compile-skill-targets.mjs`'s equivalent is not:

```js
if (isEntryPoint(import.meta.url)) { main() }
```

Hardwiring that to `false` leaves `scripts/__tests__/compile-skill-targets.test.mjs` at **18/18 green**. The script would compile nothing and exit 0 — the same silent no-op class as #7198/#7214, on the other consumer of the shared helper.

Pre-existing rather than introduced by #7232, but #7232 moved both consumers onto the shared helper, which is what makes the asymmetry visible.

## Suggested fix

A subprocess test mirroring `scripts/__tests__/gen-agents-md.test.mjs`: run the script for real in a temp fixture and assert it actually emitted something, rather than only asserting the exit code.

## Acceptance criteria

- [ ] Hardwiring `compile-skill-targets.mjs`'s guard to `false` turns a test red
- [ ] The test asserts an observable side effect (a file written), not just exit status

blamechris:
## Problem

Every test in `scripts/__tests__/compile-skill-targets.test.mjs` imported the module and
called an exported function. That says nothing about whether `node
scripts/compile-skill-targets.mjs` still compiles anything.

Hardwiring the script's `if (isEntryPoint(import.meta.url))` to `false` left the file at
**18/18 green** while the script compiled nothing and exited 0 — the same silent no-op
class as #7198/#7214, on the other consumer of the shared guard. `gen-agents-md.mjs`'s
equivalent call site was already covered; this one was the asymmetry #7232 made visible.

## What changed

Four subprocess tests that run the script for real out of a temp fixture and assert on the
**artifact it wrote**. Exit status alone cannot tell "compiled everything" from "never
ran" — both are 0.

| test | what it pins |
|---|---|
| direct invocation | every selected target's artifact is written, so the emitter **loop** is covered, not one lucky write |
| symlinked invocation path | #7198's actual shape at this call site |
| `--dry-run` writes nothing | with the identical run **without** it as an inline positive control |
| no `.claude/commands` → exit 1 | a detector for the same mutation that does not depend on a file |

Staging is extracted to `scripts/__tests__/lib/stage-script.mjs` and shared with
`gen-agents-md.test.mjs`, which grew it first. A second hand-maintained copy is the shape
#7213/#7222 were filed to remove, recreated one directory down.

## Proof it goes red

Restores were `cp` from an absolute-path backup (never `git checkout --`), and every
mutant was grepped for after application to prove it landed — a find/replace that matches
nothing exits 0 and the test then "passes" against a mutant that never existed.

| mutation | new suite | which test caught it |
|---|---|---|
| **A** — guard → `if (false)` (the acceptance criterion) | **exit 1**, 18 passed / 4 failed | all four |
| **A** against the *pre-change* test file | exit 0, **18/18 green** | none — the issue reproduced exactly |
| **B** — `for (const t of targets)` → `targets.slice(0, 1)` | exit 1, 21/1 | direct-invocation (second target absent) |
| **C** — guard reverted to the naive pre-#7213 comparison | exit 1, 21/1 | **symlink only** — proving it is not a duplicate of the direct-run test |
| **D** — `if (!dryRun)` → `if (true)` | exit 1, 21/1 | dry-run |

Each mutation is caught by exactly its intended assertion, and mutation A's failure
messages name the real failure mode rather than a bare boolean.

## Controls validated, not just asserted

- **The injected `HOME` really reaches `homedir()`.** Differential probe: a temp home
  *with* `.codex` produced the "installed but not selected" hint, one *without* did not.
  So the child cannot see the developer's real home. The real `~/.codex/prompts` was
  confirmed untouched.
- **`stageScript` throws** when the sibling `lib/` is absent, rather than silently staging
  an incomplete fixture — verified by calling it with a lone script.
- **The fixture tmpdir is realpath'd.** On macOS `os.tmpdir()` is a symlink, so a fixture
  on the raw path would put every run through a symlinked prefix and quietly turn the
  direct-invocation test into a second copy of the symlink test, on macOS only.

## Safety

Every run passes `--targets` explicitly rather than trusting the profile fallback, and the
child gets a throwaway `HOME`. The `codex` and `pi` emitters write into `~/.codex` and
`~/.pi`, so a fixture that ever reached them would write to the developer's real home.

## Verification

- `compile-skill-targets.test.mjs` 22/22 (was 18/18), Node 22 and Node 26
- `gen-agents-md` 8/8, `is-entry-point` 18/18 (incl. the three-copy drift gate),
  `flush-and-exit` 5/5, `check-release-pr-subject` exit 0 — all on Node 22

Closes #7236

### DIFF
--- a/scripts/__tests__/compile-skill-targets.test.mjs
+++ b/scripts/__tests__/compile-skill-targets.test.mjs
@@ -45,7 +45,11 @@ const test = async (name, fn) => {
   }
 }
 
+// Counted, so withFixture below can prove its callback actually checked
+// something rather than trusting that it was called.
+let assertions = 0
 const assert = (cond, msg) => {
+  assertions++
   if (!cond) throw new Error(msg || 'assertion failed')
 }
 
@@ -270,6 +274,7 @@ await test('detectUncompiledAgents flags an installed-but-unselected pi (~/.pi)'
 
 import { existsSync, readFileSync, realpathSync, symlinkSync, writeFileSync } from 'node:fs'
 import { spawnSync } from 'node:child_process'
+import { pathToFileURL } from 'node:url'
 import { stageScript } from './lib/stage-script.mjs'
 
 const SKILL_BODY = '# Demo skill\n\nCompiles a demo skill for the entry-point fixture.\n'
@@ -285,6 +290,7 @@ const SKILL_BODY = '# Demo skill\n\nCompiles a demo skill for the entry-point fi
  */
 const withFixture = (fn) => {
   const dir = mkdtempSync(pjoin(realpathSync(tmpdir()), 'compile-skill-'))
+  const before = assertions
   try {
     const root = pjoin(dir, 'root')
     const home = pjoin(dir, 'home')
@@ -295,6 +301,22 @@ const withFixture = (fn) => {
     fn({ dir, root, home, script })
   } finally {
     rmSync(dir, { recursive: true, force: true })
+  }
+  // Proof of WORK, not proof of call. Every assertion in this section lives
+  // inside `fn`, and the harness counts a test as `ok` whenever it does not
+  // throw — so a `withFixture` that quietly failed to call `fn` would report
+  // every test here as passing having checked nothing. That is the same
+  // false-safety shape this whole section exists to catch, one level up in the
+  // scaffolding.
+  //
+  // A `ran = true` after the call does NOT close it: the obvious mutation makes
+  // the CALL conditional and still falls through to the flag, which is exactly
+  // how the first version of this line passed a mutant that skipped `fn`
+  // entirely. Counting assertions cannot be dodged that way, and it also
+  // catches the weaker case of a callback that runs but checks nothing.
+  // Unreachable when `fn` throws, since the exception propagates past it.
+  if (assertions === before) {
+    throw new Error('withFixture: the callback made no assertions — this test checked NOTHING')
   }
 }
 
@@ -310,16 +332,24 @@ const withFixture = (fn) => {
  * Determinism second: main()'s "installed but not selected" hint fires off
  * `homedir()`, so on a machine with ~/.codex present the child's stdout would
  * differ from CI's.
+ *
+ * `cwd` is pinned to the fixture for the same reason. Every call passes --repo
+ * explicitly, so it should never matter — but `repo` DEFAULTS to
+ * `process.cwd()`, and the suite is normally invoked from the repo root. If
+ * --repo handling ever regressed, an unpinned child would compile the real
+ * checkout and rewrite its committed artifacts instead of failing. Pointing it
+ * at a directory with no .claude/commands turns that regression red.
  */
-const runCompiler = (scriptPath, args, home) =>
+const runCompiler = (scriptPath, args, home, cwd) =>
   spawnSync(process.execPath, [scriptPath, ...args], {
     encoding: 'utf8',
+    cwd,
     env: { ...process.env, HOME: home, USERPROFILE: home },
   })
 
 await test('invoked directly, it compiles a real skill and writes every target artifact (#7236)', () => {
-  withFixture(({ root, home, script }) => {
-    const run = runCompiler(script, ['--repo', root, '--targets', 'claude,gemini'], home)
+  withFixture(({ dir, root, home, script }) => {
+    const run = runCompiler(script, ['--repo', root, '--targets', 'claude,gemini'], home, dir)
     assert(run.status === 0, `expected exit 0, got ${run.status}: ${run.stderr}`)
 
     const skillMd = pjoin(root, '.claude', 'skills', 'demo', 'SKILL.md')
@@ -355,6 +385,7 @@ await test('invoked through a symlinked path, it still compiles (#7198 at this c
       pjoin(link, 'scripts', 'compile-skill-targets.mjs'),
       ['--repo', root, '--targets', 'claude'],
       home,
+      dir,
     )
     assert(run.status === 0, `expected exit 0, got ${run.status}: ${run.stderr}`)
     assert(
@@ -365,10 +396,10 @@ await test('invoked through a symlinked path, it still compiles (#7198 at this c
 })
 
 await test('--dry-run emits nothing, and the identical run without it does (#7236)', () => {
-  withFixture(({ root, home, script }) => {
+  withFixture(({ dir, root, home, script }) => {
     const skillMd = pjoin(root, '.claude', 'skills', 'demo', 'SKILL.md')
 
-    const dry = runCompiler(script, ['--repo', root, '--targets', 'claude', '--dry-run'], home)
+    const dry = runCompiler(script, ['--repo', root, '--targets', 'claude', '--dry-run'], home, dir)
     assert(dry.status === 0, `--dry-run should exit 0, got ${dry.status}: ${dry.stderr}`)
     assert(!existsSync(skillMd), '--dry-run wrote an artifact')
 
@@ -376,7 +407,7 @@ await test('--dry-run emits nothing, and the identical run without it does (#723
     // file" is equally what a guard that never fired produces, so the assertion
     // above proves nothing on its own. Same fixture, same flags minus
     // --dry-run, must now write.
-    const wet = runCompiler(script, ['--repo', root, '--targets', 'claude'], home)
+    const wet = runCompiler(script, ['--repo', root, '--targets', 'claude'], home, dir)
     assert(wet.status === 0, `expected exit 0, got ${wet.status}: ${wet.stderr}`)
     assert(existsSync(skillMd), 'the control run emitted nothing either — the no-op above was vacuous')
   })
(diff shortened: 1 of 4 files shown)


blamechris:
## Adversarial review round — three defects found in the first commit, all fixed in `cb61c3314`

Ran a 22-agent adversarial pass (six lenses, refuters on every non-nitpick finding, completeness critic). It confirmed 12 findings; refuters downgraded all to minor/nitpick and refuted 3. **I reproduced each of the three below myself before fixing** — two of them changed the change.

### 1. The coverage was one-directional

Every test faced a guard stuck **false**. Stuck **true** was invisible: hardwiring the guard to `true` left the suite at 22/22 green, because `main()` then runs during this test file's own `await import(...)` — compiling whatever repo the suite was invoked from and rewriting its committed artifacts. Verified from a full checkout: `Compiling` appeared in the output while every test still passed.

I then hit this for real. My own `guard -> true` mutation run compiled 31 skills into the worktree and dirtied 11 tracked artifacts. Only the explicit-path staging rule kept them out of the commit. The new test spawns an *import* of the module and asserts it neither logs nor emits.

This is #7250's lesson recurring: testing only the direction the bug arrived from shipped a worse hole than the one it fixed.

### 2. The fixture had no proof of work — and my first fix for it did not work

Making `withFixture` skip its callback left all four tests reporting `ok` having asserted nothing — the same false-safety shape this section exists to catch, one level up in the scaffolding.

My first fix was a `ran = true` after the call. It **passed the mutant**: the obvious mutation makes the *call* conditional and still falls through to the flag. Caught only because I re-ran the battery rather than trusting the fix. Replaced with an assertion counter, which cannot be dodged that way and also catches a callback that runs but checks nothing.

### 3. `stageScript` threw before `mkdir`

Callers wrap it in a `try/finally` that chmods and removes the fixture, so throwing before the directory existed made their `chmodSync(scriptsDir)` fail with `ENOENT` — masking the diagnostic and skipping the `rmSync`. A regression the extraction introduced; the pre-refactor code mkdir'd first. Ordering restored.

Also pinned the spawned child's `cwd` to the fixture: every call passes `--repo`, but `repo` defaults to `process.cwd()` and the suite runs from the repo root, so a `--repo` regression would have compiled the real checkout instead of failing.

### Mutation battery, 23/23 baseline

| mutation | result |
|---|---|
| guard → `false` | 4 red |
| guard → `true` | 1 red — **was 0** |
| emitter loop → first target only | 1 red |
| naive pre-#7213 guard | 2 red |
| `--dry-run` writes anyway | 1 red |
| `--repo` ignored, falls back to cwd | 3 red |
| `withFixture` skips its callback | 5 red — **was 0** |

Restores were `cp` from absolute-path backups, every mutant grepped for after application, and the restore check compares against the backup rather than `HEAD` (the first version compared against `HEAD` and reported failure on every iteration, since the fixes were uncommitted — a broken control, caught and fixed).

### Filed rather than folded in

The panel found real gaps that are **not** this change's scope, so they are scoped issues instead:

- #7253 — compiled skill artifacts are stale on `main` and nothing gates them. Verified on pristine `origin/main`: **11 drifted**, including `catchup`, which has a source but no compiled artifact at all and so is not loadable.
- #7254 — the two *server* `isEntryPoint` call sites still have no call-site coverage, including `server-cli-child.js` (the daemon's child process — far more blast radius than this one).
- #7255 — `main()` branches with no coverage: the `--name` traversal guard, the emit-failure exit path, the Gemini skip, `emitCodex`, `targetsFromProfile`.

Critic verdict: **safe to merge**, both acceptance criteria met and independently re-verified.

blamechris:
## Code Review: PR #7251 — **Approve** (0 critical, 3 suggestions, 1 nitpick)

Three reviewers (style/conventions, correctness/testing, architecture/docs) over the final diff, then synthesis. The branch was verified independently in an isolated `git archive` copy under Node 22.22.3; the shared checkout was left clean and on `main` throughout.

**All three suggestions and the nitpick are now fixed in `0c810f6a2`.** I reproduced each before acting — two of them changed the change, and one contradicted my own earlier adversarial pass, which had *refuted* it.

### 1. The other-direction test did not reach the branch it named

It spawned `node -e "await import(...)"`, and under `-e` there is **no `argv[1]`** — so `isEntryPoint` returns at its first line (`if (!process.argv[1]) return false`) and never reaches the path comparison. That comparison is what actually decides the question when CI runs this suite, where `argv[1]` is a real, existing, *different* file.

Reproduced: inverting `self === invoked` to `!==` in the guard left the suite at **23/23 green** while the test file's own top-level import ran `main()` and wrote **12 files into the invoking checkout** — verbatim the failure the test's comment claimed to close. `is-entry-point.test.mjs` does catch that mutant, but at a *later* step in the same job, so the call-site step reports green having already rewritten the checkout on the way past.

The importer is now a staged file, so `argv[1]` is real and different.

### 2. `withFixture` was sync-only without saying so

Making a callback `async` and awaiting once left the suite green with the positive control **never executed**: the synchronous prefix satisfies the assertion counter, the returned promise is neither awaited nor returned, and the `finally` removes the fixture out from under the rest of the body. All 18 pre-existing tests use async callbacks the harness awaits, so async is this file's ambient habit and only these five callbacks are not. A thenable now throws.

Worth noting: my earlier 22-agent pass **refuted** this exact finding. It was real. Refuter verdicts are evidence, not verdicts.

### 3. The fixture body exercised none of the emitter transformations

`stripStamp`, `emitClaude` and `emitGemini` are **not exported**, so these subprocess tests are their only coverage anywhere in the repo — and three mutations stayed green:

- `stripStamp` reduced to a no-op — 28 of 31 real sources carry a `<!-- skill-templates: -->` stamp, so every artifact would ship install metadata as skill content
- `emitGemini`'s `$ARGUMENTS` → `{{args}}` substitution deleted — every compiled Gemini command silently loses argument passing
- `emitClaude` dropping its closing frontmatter `---` — a `startsWith` check passes on output Claude cannot parse

The fixture now carries a stamp and an arg token, and the frontmatter assertion pins the whole block.

### Nitpick: the "ONE implementation" claim was false

`bump-version.test.sh`'s `install_agents_generator()` is a third copy staging the same script from shell, and it **disagrees** on the missing-lib case — it skips where the new helper refuses, which is the "'cannot check this' treated as 'nothing to check'" shape from `docs/false-safety-guards.md`. Header scoped honestly; reconciliation filed as #7256.

Also moved to `scripts/__tests__/helpers/`: every non-`node_modules` `lib/` in this repo is production code, and `packages/server/tests/helpers/` is the precedent for shared test scaffolding.

### Mutation battery — 23/23 baseline, 12 mutations, all red

| mutation | before this round | now |
|---|---|---|
| `self === invoked` → `!==` | **green** | 1 red |
| callback made `async` | **green** | 1 red |
| `stripStamp` → no-op | **green** | 1 red |
| gemini drops `$ARGUMENTS` | **green** | 1 red |
| `emitClaude` drops closing `---` | **green** | 1 red |
| guard → `false` | 4 red | 4 red |
| guard → `true` | 1 red | 1 red |
| emitter loop → first target | 1 red | 1 red |
| naive pre-#7213 guard | 2 red | 2 red |
| `--dry-run` writes anyway | 1 red | 1 red |
| `--repo` ignored → cwd | 3 red | 3 red |
| `withFixture` skips callback | 5 red | 5 red |

Two mutations initially failed to land (zsh ate a backslash in one needle; the other matched twice) and were re-run with corrected needles rather than counted as passes — the `assert count == 1` refusal is what caught both.

### Architecture

Sharing `stageScript` is the right call: two copies of "stage a script plus its sibling `lib/`" is a hardcoded list beside a growing set — `scripts/lib/` gained `is-entry-point.mjs` (#7222) and `entry-point-guard-copies.mjs` (#7247), and each addition would have needed both copies updated.

The subject script and the guard are **unmodified** by this PR (`git diff --stat -- scripts/compile-skill-targets.mjs scripts/lib/` is empty). Repo-wide entry-point walk: 1904 files, 6 sanctioned guard sites, exit 0. All 7 custom server lints green.

### Deferred

| Item | Issue |
|---|---|
| Compiled skill artifacts stale on `main`, nothing gates them (11 drifted) | #7253 |
| Two server `isEntryPoint` call sites uncovered (+ the sidecar's third, see comment) | #7254 |
| `main()` branches uncovered: `--name` traversal, emit-failure exit, gemini skip, `emitCodex`, `targetsFromProfile` | #7255 |
| `bump-version.test.sh` staging copy disagrees on missing-lib semantics | #7256 |
| `docs/false-safety-guards.md` still describes the guard as held by "two gates" | #7257 |

### Verdict

**Approve.** Test-only change, subject unmodified, coverage where there was none, and every claim it makes about itself is now backed by a mutation that goes red.

Issue #7262, opened 2026-08-18
tests/_setup.mjs: named and namespace ESM fs imports bypass the write sandbox (41 modules unguarded)


blamechris:
## Summary

`packages/server/tests/_setup.mjs` is the sandbox that stops tests from clobbering the real `~/.chroxy` / `~/.claude` (#4633). Its header comment claims that production code doing `import { writeFileSync } from 'fs'` **sees the patched value**. It does not. Named *and* namespace ESM imports of `node:fs` read a link-time snapshot taken **before** the patch is applied, so they bypass the guard entirely.

**41 modules under `packages/server/src/` import a write-side `fs` function by name** (`supervisor.js`, `service.js`, `billing-budget.js`, `user-shell-registry.js`, `byok-mcp-trust.js`, …). For all of them, the sandbox is not armed.

This is the [`docs/false-safety-guards.md`](docs/false-safety-guards.md) class exactly: the guard reports success while checking less than it claims, and no test can see it because the guard's *output* is correct and its *coverage* is what is wrong.

## The claim

`packages/server/tests/_setup.mjs:57-65`:

> ``…any production code that does `import { writeFileSync } from 'fs'` then sees our patched value because the ESM named exports are derived from this same object at link time.``

## Reproduction

Under the **real** harness (`node --import ./tests/_setup.mjs`), a probe printing `.name` for each binding form:

```
CJS require        : patchedWriteFileSync
ESM default import : patchedWriteFileSync
ESM namespace      : writeFileSync      <-- UNPATCHED
ESM named import   : writeFileSync      <-- UNPATCHED
```

Only the live CJS exports object — reached via `require('node:fs')` or the ESM **default** import, which is that same object — sees the patch.

## Root cause (verified, with a paired control)

`_setup.mjs` **itself** does `import { mkdtempSync } from 'node:fs'` at line 51. ESM imports are evaluated before the module body, so that import links the `node:fs` synthetic module — snapshotting its named exports from the *unpatched* `module.exports` — before line 132 replaces `fs.writeFileSync`.

Two minimal setups, identical except for that import, isolate it:

| setup | ESM named | ESM namespace |
|---|---|---|
| imports `node:fs` by name (today's `_setup.mjs`) | `writeFileSync` (unpatched) | `writeFileSync` (unpatched) |
| takes `mkdtempSync` off the required CJS object | `patchedWriteFileSync` | `patchedWriteFileSync` |

[...]

Issue #4633, opened 2026-05-31
Bullet-proof server test isolation from user state (~/.chroxy, ~/.claude)


blamechris:
## Context

While iterating on the #4628 fix (2026-05-30), running the full server test suite repeatedly clobbered the user's live `~/.chroxy/session-state.json` with `{id:'s1', name:'Session 5', cwd:'/tmp'}` test fixture data — and corrupted `~/.chroxy/checkpoints/` with `sess-1`/`sess-a`/`sess-b` fixtures. The user lost the chroxy-side history for two actively-dogfooded sessions (No-it-all TUI: 124 → 51 events; Rah6 CLI: 188 → 51 events). Salvage backup recovered both at an older snapshot; claude TUI's own `~/.claude/projects/.../<uuid>.jsonl` survived untouched.

This is exactly the bug class PR #2314/#2315 tried to fix — "SessionManager tests MUST always pass `stateFilePath`" — but the rule isn't enforced and new tests + existing offenders still bypass it.

## Audit (current scope)

`grep -rln "new SessionManager(" packages/server/tests` returns 13 files. Of those, **3 use bare `new SessionManager()` with no `stateFilePath` option** at all:

- `packages/server/tests/auto-label.test.js`
- `packages/server/tests/session-manager-delta-limit.test.js`
- `packages/server/tests/session-manager-history-error.test.js`

(The remaining 10 either explicitly pass `stateFilePath` or pass options that include it via a helper — not yet audited line-by-line; the full pass should find more cases that pass partial options.)

## Why Docker isn't the right tool here

The user asked about Docker for isolation. Real talk on the trade-off:

| Approach | Setup | Per-test-run cost | Bulletproof? | Dev iteration cost |
|---|---|---|---|---|
| **Docker (`docker run --rm -v $PWD:/app node:22`)** | image + Dockerfile + volume mount | ~3-10s container startup + slower FS layer | YES (full sandbox) | painful (each iteration pays the startup tax + filesystem perf hit) |
| **HOME override (test setup file sets `process.env.HOME = mkdtempSync(...)`)** | one `--import` flag in `package.json` test script | zero (native test runs unchanged) | YES (any code reading `~/.chroxy/...` gets the tmp path) | zero |
| **Sandbox guard (test setup throws if `~/.chroxy/` or `~/.claude/` is written)** | one `--import` flag, `fs.writeFileSync` monkey-patch | trivial | YES — loud failure on violation | zero, plus surfaces the test that contaminated |
[...]

PR #7272, opened 2026-08-19
fix(server): guard the whole fs mutation surface in the test sandbox

### LINKED ISSUE #7267: tests/_setup.mjs: the write sandbox patches no delete or copy API — cpSync writes into the real ~/.chroxy
## Summary

`packages/server/tests/_setup.mjs` guards a **named list** of `fs` functions, not a category. Everything that deletes, copies, or links is outside it, so a test can still mutate the developer's real `~/.chroxy` / `~/.claude` while the guard reports nothing (#4633's whole point).

Found while fixing #7262. That PR made the guard visible to named/namespace importers for the first time, which is what makes this gap the entire remaining reach problem rather than one of two.

## Measured

Under the real harness (`node --import ./tests/_setup.mjs`), against a path under the real protected root:

| API | result |
|---|---|
| `writeFileSync`, `mkdirSync`, `openSync` | GUARDED |
| `truncateSync` | GUARDED — but only incidentally, it opens `r+` and trips `openSync` |
| `unlinkSync`, `rmSync`, `rmdirSync` | **unguarded** |
| `cpSync`, `copyFileSync` | **unguarded — and these CREATE** |
| `symlinkSync`, `linkSync`, `chmodSync` | **unguarded** |
| `promises.unlink`, `promises.rm`, `promises.cp`, `promises.copyFile` | **unguarded** |

`cpSync` and `copyFileSync` are not merely destructive-bypass: running the probe left a real directory and file at `~/.chroxy/__probe-<pid>/x` on the developer's machine, created straight through an armed sandbox. (Cleaned up; noted because it means this is a write breach, not only a delete one.)

## Not currently exploited

An extended guard covering the whole destructive/copy/link surface was armed and the **entire server suite** run under it: **13,391 tests, zero hits**. No existing test deletes, copies, links or chmods anything under the real protected roots. This is latent, not live — which is why it is filed rather than hot-fixed.

It is also **pre-existing and not worsened by #7266**: before that PR, named-import `unlinkSync` bypassed the guard *and* so did named-import `writeFileSync`. #7266 is a net reduction in reach gap.

## Why it still matters

`src/connection-info.js:37` `removeConnectionInfo()` is `unlinkSync(getConnectionInfoPath())`, called automatically from `readConnectionInfo()` whenever the recorded pid is dead. `configPath()` reads `CHROXY_CONFIG_DIR` per call, and fourteen test files `delete process.env.CHROXY_CONFIG_DIR`. A test that lands in that state and reads connection info would delete a running daemon's real `~/.chroxy/connection.json`, and the guard would not fire.

## Suggested fix

Patch the destructive/copy surface — `rmSync/rm`, `unlinkSync/unlink`, `rmdirSync/rmdir`, `cpSync/cp`, `copyFileSync/copyFile`, `symlinkSync/symlink`, `linkSync/link`, `chmodSync/chmod` — checking **both** paths for the two-path ones, as `renameSync` already does.

Better: drive the patch list from a single exported array so a missing method is a visible omission rather than an invisible one. `tests/setup-sandbox-binding-forms.test.js` currently hand-maintains a parallel list of methods to assert, which is `docs/false-safety-guards.md`'s own cause #1 (a hardcoded list beside a set that grows); deriving both from one array closes that at the same time.

## Acceptance criteria

- [ ] Every patched method is proven to fail: attempt each against a protected path and confirm `CHROXY_TEST_SANDBOX`
- [ ] `cpSync`/`copyFileSync` specifically verified to create nothing under the real home
- [ ] The patch list and the test's assertion list derive from one source
- [ ] The `_setup.mjs` header's coverage claim updated to match (it was narrowed in #7266 to name the gap and point here)
- [ ] Full server suite run afterwards; any newly-surfaced write is a genuine #4633 bug to triage, not a reason to weaken the guard

## Provenance

Adversarial review of #7266, confirmed independently by measurement.

### LINKED ISSUE #7268: claude-hooks test sandbox patches a narrower fs list than the server's, and its errors carry no code
## Summary

`packages/claude-hooks/tests/_setup.mjs` is the sibling of the server write sandbox. Its binding-form hole was fixed in #7266 (it had the identical #7262 defect), but two differences from the server's guard remain.

**1. Narrower patch list.** It patches `writeFileSync`, `mkdirSync`, `renameSync`, `rmSync`, `unlinkSync`, `createWriteStream`, `promises.writeFile`. Missing versus the server's: `appendFileSync`, `openSync`, `promises.appendFile`, `promises.rename`, `promises.mkdir`, `promises.open`.

`openSync` is the load-bearing omission: `installer.js`'s atomic-write path is `openSync -> writeSync -> fsyncSync -> renameSync`, so only the final `renameSync` is intercepted. A partial write that never reaches the rename is unguarded.

(It is *ahead* of the server on `rmSync`/`unlinkSync`, which the server does not patch at all — see #7267. The two guards should converge on one list.)

**2. Its errors have no `code`.** It throws a plain `Error` whose message begins `CHROXY_TEST_SANDBOX:`, while the server sets `err.code = 'CHROXY_TEST_SANDBOX'`. Any caller matching on `code` — including a probe written against the server's convention — reads a fired guard as an unrelated failure. This cost a false reading during the #7266 review: a guarded call was initially classified "unguarded(undefined)".

## Suggested fix

Converge the two sandboxes: one shared patch list (or a shared helper), `err.code` set on both, and the same method coverage. #7267 is the server-side half of the same convergence.

## Acceptance criteria

- [ ] Each newly patched method proven to fail against a protected path
- [ ] `err.code === 'CHROXY_TEST_SANDBOX'` on both packages' guards
- [ ] `installer.js`'s `openSync` path specifically covered
- [ ] claude-hooks suite green afterwards

## Provenance

Adversarial review of #7266.

blamechris:
## Summary

The `#4633` write sandbox patched a **named list** of eleven `fs` methods, not a category. Everything that deletes, copies, links or changes a mode reached the developer's real `~/.chroxy` while the guard went on reporting success — the "Checked a subset" mode in `docs/false-safety-guards.md` — a hardcoded list next to a set that grows.

Found inside the fix for `#7262`, which is this document's own thesis: making the guard *visible* to named importers left its *reach* as the only remaining problem, and the reach was the list.

## Measured, under the real harness, against the live `~/.chroxy`

| API | before |
|---|---|
| `writeFileSync` `mkdirSync` `openSync` | GUARDED |
| `truncateSync` | GUARDED — only incidentally, it opens `r+` and trips `openSync` |
| `unlinkSync` `rmSync` `rmdirSync` | **unguarded** — 34 `unlinkSync` call sites in `src/`, more than `openSync`'s 14 |
| `chmodSync` `symlinkSync` `linkSync` | **unguarded** |
| `cpSync` `copyFileSync` | **unguarded — and these CREATE** |
| `fs.writeFile(path, data, cb)` and the whole callback surface | **unguarded** |

`cpSync` is the one worth remembering. A probe aimed at a path whose **parent directory did not exist** — chosen precisely so a failing measurement could create nothing — left a real directory tree inside the live `~/.chroxy`, because `cpSync` creates the destination's parents before it fails. The safety argument was sound for every other API and wrong for that one, and nothing said a word.

## What changed

**One table, in `scripts/lib/test-fs-sandbox.mjs`.** It expands to 61 guards across the sync, callback and `promises` surfaces, and to the probes that prove each one fires. A row cannot be guarded without being proven, or proven without being guarded. That retires the hand-maintained list `setup-sandbox-binding-forms.test.js` carried as a `KNOWN GAP (#7267)` — deleted, not kept in step.

**The guard is now a category, enforced.** `FS_EXEMPTIONS` classifies every remaining `fs` function with a reason (`read`, `fd`, `class`, `internal`), and `GUARDED ∪ EXEMPT` must cover the live `fs` surface exactly. A Node release that adds a path-taking mutator turns the suite red instead of quietly widening the hole. On Node 22.22.3 that is 100 `fs` functions and 31 `fs.promises` functions, zero unclassified, zero overlap.

**Both packages install the same module** (closes `#7268`). They were separate hand-written lists that had drifted in *both* directions — claude-hooks patched `rmSync`/`unlinkSync` the server did not, the server patched `openSync` and four `promises` methods claude-hooks did not. The `openSync` gap was **latent, not reachable through `installHooks`** — issue #7268 claimed otherwise and was wrong, and I have corrected it there. `installer.js`'s atomic write is `mkdirSync → openSync → writeSync → fsyncSync → chmodSync → renameSync`, and `mkdirSync` *was* on the old list, so the sequence aborted at step one (measured against a fake protected home with exactly the old guard installed: one guard fired, nothing created). `openSync` was covered there the way `truncateSync` was covered on the server side — by accident, through a neighbour — which is the kind of coverage this change replaces with the deliberate kind. Its errors also carried no `code`, so a caller matching `err.code` read a **fired** guard as an unrelated failure — that cost a false reading during the `#7266` review.

**The callback guard throws synchronously**, on purpose: `fs.unlink(p, () => {})` is a common fire-and-forget idiom, and a callback-delivered breach would be silent — the exact failure the sandbox exists to prevent. Zero callback-form `fs` call sites exist under `packages/*/src`, so nothing in production code can be surprised by it.

**The `#7262` rule now covers the graph, not one file.** The guard lives one import away from `_setup.mjs`, so an ESM `node:fs` import in the shared module would disarm it identically. The structural check walks the whole import graph, and an import it *cannot* follow is a failure rather than a skip — "cannot check this" must never read as "nothing to check".

## The part that is not obvious

**A derived assertion cannot test the parameter it derives from.** Both lists expand from rows like `{ base: 'rename', paths: 2 }`, so narrowing one to `paths: 1` deletes the guard on the second argument *and* the probe that would have caught it. That mutation was run and it **passed every other assertion in the suite** — `rename(tmp, '~/.chroxy/session-state.json')` would have walked straight through. Which arguments are paths is therefore stated a second time, deliberately, as a specification with a reason per row. It is the only thing here written twice.

## Every guard proven to fail

Thirteen mutations, each restored from a `cp` backup rather than `git checkout --`, each turning the suites red:

| | mutation | caught by |
|---|---|---|
| M1 | the original `#7262` ESM `fs` import, re-applied to the fixed file | binding forms (51 failing) |
| M2 | an ESM `fs` import in the new shared module | binding forms + graph walk |
| M3 / M4 | drop the `unlink` / `cp` row (the `#7267` defect itself) | category completeness |
| M5b / M5c | `rename` / `symlink` narrowed to one path | two-path pin |
| M6 | `isWriteIntent` always false | the three `open` probes |
| M7 | `cpSync` silently skipped at install | per-method probe + probed→installed |
| M9 | the guard error loses its `code` (the `#7268` defect) | error-shape test |
| M10 | an unwalkable import edge in the bootstrap graph | graph walk |
| M11 | claude-hooks guarding the wrong root | all 84 hooks probes |
| M13–M25 | see the review-round comment | 13 more, added after the panel |
| M12 | a two-path row without a stated reason | the specification table |

M5 was a **green miss on the first run** — the two-path pin was added because of it.

## Verification

- **Full server suite: 13,570 tests, 4 failures — every one of them also fails on `main` in the same environment** (live daemon on `:8765`, real `~/.codex/auth.json`). `main` in fact fails 5. No new failures.
- **Zero sandbox breaches surfaced** across the whole suite, matching the issue's prediction that this is latent, not live.
- claude-hooks: 184/184. New sandbox suites: 133/133 server, 87/87 hooks.
- All 14 server custom lints, all 10 `Scripts Tests` steps, and the AGENTS.md gate pass locally.
- The real `~/.chroxy` is clean afterwards — asserted by the suites themselves, not just checked by hand.

Closes #7267
Closes #7268

---

## Review round (`14e9d3376`)

A six-lens dimension panel with three-lens adversarial verification found **three genuine holes in the change itself**, all now fixed and each with a mutation proving the guard fails without it:

- **The write-intent mask was Linux's literals.** `O_CREAT` is 64 on Linux but **512 on macOS**, and `O_TRUNC` was not in the mask at all — `openSync('<protected>/state.json', O_TRUNC)` emptied a real file through an armed sandbox. Now derived from `fs.constants`. The *allow* direction was unpinned too.
- **Three claude-hooks probes could destroy live config.** They targeted real-home paths whose parent exists, so a regression would rewrite and truncate the developer's `~/.claude/settings.json` — the same class as the `cpSync` discovery above, in a file written for this PR.
- **Platform-blind probes** — the CI failure. `lchmod` is macOS-only; plans are now partitioned by availability, with every dropped plan corroborated against the installer's own skip record.

Plus fail-closed path handling, `Uint8Array` paths, case-folding on darwin/win32, an independent mutator roster, `~/.claude` and `~/.claude.json` probes, dynamic-import detection in the structural walk, and corrections to several of my own claims — including that **#7268's `openSync` premise was wrong** (corrected on the issue). Full detail in the [review-round comment](https://github.com/blamechris/chroxy/pull/7272#issuecomment-5337839555).

**28 mutations, every one red.** Full server suite 13,586 tests / 4 failures, all four also failing on `main` in this environment. claude-hooks 185/185. Both suites also green under a simulated Linux fs surface with `lchmod` removed.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -311,7 +311,9 @@ Every test that constructs a `SessionManager` **must** pass `stateFilePath` poin
 
 Two layers of defence are wired up:
 
-1. **Sandbox guard** (`packages/server/tests/_setup.mjs`, loaded via `node --import`) — monkey-patches `fs.writeFileSync`/`promises.writeFile`/`renameSync`/`mkdirSync`/`createWriteStream`/`openSync(w*)` to throw `CHROXY_TEST_SANDBOX` if any test writes to the real `~/.chroxy/` or `~/.claude/` tree. The error includes the call site, so the next bare `new SessionManager()` fails loudly at the offending test.
+1. **Sandbox guard** (`packages/server/tests/_setup.mjs`, loaded via `node --import`) — monkey-patches every `fs` function that takes a path and mutates the filesystem, across the sync, callback and `promises` surfaces, so any test write to the real `~/.chroxy/` or `~/.claude/` tree throws `CHROXY_TEST_SANDBOX`. The error includes the call site, so the next bare `new SessionManager()` fails loudly at the offending test.
+
+   The method list is NOT written here, and not written in `_setup.mjs` either: it expands from the one table in `scripts/lib/test-fs-sandbox.mjs`, which `packages/claude-hooks/tests/_setup.mjs` installs from too (the two guards were separate hand-written lists that had drifted in both directions — #7267, #7268). Adding a row there installs the guard AND its proof; `tests/setup-sandbox-coverage.test.js` fails if `GUARDED ∪ EXEMPT` stops covering the live `fs` surface, so a Node upgrade that adds a path-taking mutator goes red rather than quietly widening the hole. Reads are deliberately untouched — tests legitimately read the developer's real config.
 2. **CI lint** (`packages/server/scripts/lint-tests-state-file-path.sh`) — fails the build if any `new SessionManager(...)` in `tests/` is missing `stateFilePath`. Run locally with `cd packages/server && ./scripts/lint-tests-state-file-path.sh`.
 
 If you need to write to the real home for a legitimate reason (no current test does), set `process.env.CHROXY_TEST_ALLOW_REAL_HOME_WRITES = '1'` scoped to the test and restore it after. The sandbox guard MUST stay enabled in `package.json`.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -292,7 +292,9 @@ Every test that constructs a `SessionManager` **must** pass `stateFilePath` poin
 
 Two layers of defence are wired up:
 
-1. **Sandbox guard** (`packages/server/tests/_setup.mjs`, loaded via `node --import`) — monkey-patches `fs.writeFileSync`/`promises.writeFile`/`renameSync`/`mkdirSync`/`createWriteStream`/`openSync(w*)` to throw `CHROXY_TEST_SANDBOX` if any test writes to the real `~/.chroxy/` or `~/.claude/` tree. The error includes the call site, so the next bare `new SessionManager()` fails loudly at the offending test.
+1. **Sandbox guard** (`packages/server/tests/_setup.mjs`, loaded via `node --import`) — monkey-patches every `fs` function that takes a path and mutates the filesystem, across the sync, callback and `promises` surfaces, so any test write to the real `~/.chroxy/` or `~/.claude/` tree throws `CHROXY_TEST_SANDBOX`. The error includes the call site, so the next bare `new SessionManager()` fails loudly at the offending test.
+
+   The method list is NOT written here, and not written in `_setup.mjs` either: it expands from the one table in `scripts/lib/test-fs-sandbox.mjs`, which `packages/claude-hooks/tests/_setup.mjs` installs from too (the two guards were separate hand-written lists that had drifted in both directions — #7267, #7268). Adding a row there installs the guard AND its proof; `tests/setup-sandbox-coverage.test.js` fails if `GUARDED ∪ EXEMPT` stops covering the live `fs` surface, so a Node upgrade that adds a path-taking mutator goes red rather than quietly widening the hole. Reads are deliberately untouched — tests legitimately read the developer's real config.
 2. **CI lint** (`packages/server/scripts/lint-tests-state-file-path.sh`) — fails the build if any `new SessionManager(...)` in `tests/` is missing `stateFilePath`. Run locally with `cd packages/server && ./scripts/lint-tests-state-file-path.sh`.
 
 If you need to write to the real home for a legitimate reason (no current test does), set `process.env.CHROXY_TEST_ALLOW_REAL_HOME_WRITES = '1'` scoped to the test and restore it after. The sandbox guard MUST stay enabled in `package.json`.
--- a/docs/false-safety-guards.md
+++ b/docs/false-safety-guards.md
@@ -133,6 +133,50 @@ is the control that proves the mechanism — a separate synthetic module,
 never ESM-imported by `_setup.mjs`, and all four of *its* binding forms were
 guarded the whole time.
 
+### 9. The sandbox that guarded a named list, not a category — `#7267`
+
+Found inside the fix for the eighth, which is how this document started.
+
+Making the guard visible to named importers (`#7262`) turned its *reach* into
+its only remaining problem — and its reach was a hand-written list of twelve
+method names. Everything that deletes, copies, links or changes a mode was
+outside it. Measured under the real harness, against the developer's live
+`~/.chroxy`:
+
+```
+writeFileSync mkdirSync openSync   GUARDED
+truncateSync                       GUARDED — but only because it opens 'r+'
+                                             and trips openSync
+unlinkSync rmSync rmdirSync        unguarded   (52 unlinkSync call sites in src/,
+chmodSync symlinkSync linkSync     unguarded    more than openSync has)
+cpSync copyFileSync                unguarded — and these CREATE
+fs.writeFile(path, data, cb)       unguarded — the whole callback surface
+```
+
+`cpSync` is the one worth remembering. A probe aimed at a path whose parent
+directory did **not** exist — chosen precisely so a failing measurement could
+create nothing — left a real directory tree inside the developer's live
+`~/.chroxy`, because `cpSync` creates the destination's parents before it fails.
+The safety argument was sound for every other API and wrong for that one, and
+nothing in the guard, the suite, lint or CI said a word.
+
+The fix is one exported array (`scripts/lib/test-fs-sandbox.mjs`) that expands
+to the patch list *and* to the probes asserting each patch fires, plus a
+complement (`FS_EXEMPTIONS`) that classifies every remaining `fs` function with
+a reason — so `GUARDED ∪ EXEMPT` must cover the live surface exactly, and a Node
+release that adds a path-taking mutator turns the suite red instead of widening
+the hole.
+
+**The part that is not obvious: a derived assertion cannot test the parameter it
+derives from.** Both lists expand from rows like `{ base: 'rename', paths: 2 }`,
+so narrowing one to `paths: 1` deletes the guard on the second argument *and*
+the probe that would have caught it — and the suite stays green while
+`rename(tmp, '~/.chroxy/session-state.json')` walks through. That mutation was
+run, and it passed. Which arguments are paths is therefore stated a second time,
+deliberately, as a specification with a reason per row; it is the only thing in
+the change written twice.
+
+
 What the reach was hiding: `tests/http-routes.test.js` renamed the developer's
 real `~/.chroxy/connection.json` aside in a `before` hook and moved it back in
 `after` — so a crash, a test timeout, or a SIGKILL between the two left a
--- a/docs/false-safety-guards.md
+++ b/docs/false-safety-guards.md
@@ -133,6 +133,21 @@ is the control that proves the mechanism — a separate synthetic module,
 never ESM-imported by `_setup.mjs`, and all four of *its* binding forms were
 guarded the whole time.
 
+What the reach was hiding: `tests/http-routes.test.js` renamed the developer's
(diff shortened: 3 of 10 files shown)


blamechris:
## Review round — all findings addressed

Six review lenses ran over this diff and every non-nitpick finding went to a three-lens refuter panel. **Three of the criticals were genuine holes in the change itself**, two of which I would have shipped. Fixed in `14e9d3376`; each one now has a probe and a mutation that turns the suite red.

### Critical

| | finding | what changed |
|---|---|---|
| C1 | `probePlans` expanded the whole table while `patch()` skips platform-absent methods, so Linux CI called `undefined` | Plans are partitioned by availability, and **every dropped plan must be corroborated by the installer's own `SANDBOX_SKIPPED` entry** — "cannot run this" must not become "nothing to check". Re-ran the CI failure under a simulated Linux fs surface: red without the fix, green with it. |
| C2 | `isWriteIntent`'s numeric branch tested Linux's literals (`O_CREAT` is 64 on Linux, **512 on macOS**, 256 on Windows) and never tested `O_TRUNC` | Mask now derived from this runtime's `fs.constants`. **Reproduced the hole first**: `openSync('<protected>/state.json', O_TRUNC)` emptied a real file with the sandbox armed and silent. The *allow* direction was also unpinned — `isWriteIntent -> true` left the suite fully green — so a read-open probe now pins it. |
| C3 | three claude-hooks probes targeted real-home paths whose **parent exists**, so a regression would rewrite and truncate the developer's live `~/.claude/settings.json` | All aim inside a missing-parent probe dir. The `~/.claude.json` probes (both packages) use `renameSync(<absent source>, target)` — that file's parent is `$HOME` and exists, so a write probe would clobber a real file, while a missing source fails before touching the destination. |

C3 is the same class as the `cpSync` discovery this PR is about, in a file I wrote for it.

### Suggestions — all taken

`fd` exemption rationale corrected (`fchmod`/`fchown`/`futimes` gate on **ownership**, not write access, and succeed on the read-only fd the guard allows — measured, mode `600 → 0`; reclassified as `fd-metadata-known-residual` and named rather than hidden) · independent roster pins which operations belong in the mutator table — moving one into `FS_EXEMPTIONS` deleted the guard, its probe, **and** the category check's objection in one edit, and `rmdirSync` then removed a real directory with the suite green · read-opens pinned · `~/.claude` and `~/.claude.json` probed · `URL`/`Buffer`/`Uint8Array` probed and unrecognised shapes now **fail closed** · the structural walk sees `import('node:fs')` · `PROBE_EXTRA_ARGS` actually refuses a recipe-less row (the comment claimed a check that did not exist) · the claude-hooks source scan is recursive, covers `bin/` and `node:fs/promises`, and fails on a default/namespace import it cannot enumerate · doc-claim fixes (S9, S10, S12).

Nitpicks N1–N8 are also fixed, including case-folding on darwin/win32 where `~/.Chroxy/x` **is** `~/.chroxy/x` — both platforms run this sandbox in CI.

### Corrections to my own claims

- **claude-hooks' `openSync` gap was not reachable through `installHooks`.** `writeSettingsAtomic` opens with `mkdirSync`, which the old guard *did* patch, so the sequence aborted at step one. I measured this against a fake protected home with exactly the old list installed: one guard fired, nothing was created. **Issue #7268 said otherwise and was wrong** — I have corrected it there too. The real defects are the missing `err.code` and the unguarded delete/copy/link/callback surface.
- `52 unlinkSync` was token *mentions*; call sites are **34** (openSync's are 14 — the comparison holds either way). `45 modules` is specific to the six-name pre-#7267 write set; against the surface now guarded it is **53**. The base commit patched **11** methods, not 12.

### Verification

**28 mutations, every one red** — the original #7262 import, the CI failure under a simulated Linux surface, a dropped mutator row, a narrowed path arity, a defeated write-intent gate, a dropped `O_TRUNC`, a silently skipped install, an over-dropping availability filter, a relocated exemption, a basename match, a case-sensitive compare, a dynamic import, a recipe-less row, and the dropped error `code`.

Full server suite: **13,586 tests, 4 failures — the same four that fail on `main` in this environment** (live daemon on `:8765`, real `~/.codex/auth.json`). Zero sandbox breaches. claude-hooks 185/185. Both suites also green under a simulated Linux surface with `lchmod` removed. All 14 server lints and all 10 `Scripts Tests` steps pass.

### Deferred

Nothing. The one item I would have deferred — fd-level bookkeeping to close `fchmod`/`fchown`/`futimes` on a read-only fd — is not fixable with a path-based guard, so it is **named** in the exemption table as a known residual rather than covered by a rationale the code does not support. No call site in this repo does it.

Issue #7273, opened 2026-08-20
Windows: project-directory containment rejects valid in-project paths (12 test files, 2 security-relevant)


blamechris:
## Summary

Measured on a real Windows host (Win 11, Node 22.23.1, the `chroxy-win` runner box) while inverting the Windows CI test list for #7270: **12 server test files fail on Windows because the project-directory containment check rejects paths that are legitimately inside the project**, or fails to reject ones that are outside it.

These are NOT POSIX-only tests. They assert cross-platform behaviour and they are wrong on Windows, so they are tracked as `windows-defect` debt in `WINDOWS_EXEMPT` (`packages/server/scripts/lib/windows-test-set.mjs`) rather than exempted permanently.

## Two of these are security-relevant

Both need investigation before they are dismissed as test-fixture problems:

- **`tests/handler-utils.test.js` — `rejects backslash traversal (..\)`** fails with `backslash traversal should be caught`. On POSIX `..\` is a literal filename; on Windows it is a real traversal separator. The assertion says it must be caught; on Windows it was not.
- **`tests/handlers/conversation-handlers.test.js` — `uses the authoritative recorded cwd, ignoring a spoofed in-scope cwd hint`** fails with `read path must derive from the recorded cwd, not the client-supplied hint` (expected `false`, actual `true`), alongside `Adversary A8: scopes results to bound session cwd`.

I have not confirmed either as an exploitable path — these are test failures observed on Windows, and the fixture itself may be the thing that does not translate. But they are the two worth looking at first, and the daemon does run on Windows (there is a `chroxy-win` runner and a Windows desktop build).

## The failing files

The dominant symptom is `Access denied: … restricted to the project directory` on a path inside the project, or a realpath/separator mismatch.

| file | failing | of | representative failure |
|---|---|---|---|
| `tests/ws-server-file-ops.test.js` | 18 | 53 | `'Directory not found'` vs `'Access denied: directory listing is r…'` |
| `tests/write-file.test.js` | 6 | 15 | `writes a file successfully` → `Access denied: file writing is restricted to the project directory` |
| `tests/handler-utils.test.js` | 5 | 142 | `rejects backslash traversal (..\)`; `blocks symlink that resolves outside project directory` |
| `tests/ws-file-ops-error-paths.test.js` | 5 | 12 | `browseFiles returns Not a directory for ENOTDIR` |
| `tests/ws-file-ops-common.test.js` | 4 | 8 | `resolves an existing path to its realpath` |
[...]

Issue #7287, opened 2026-08-20
Consolidate the remaining hand-rolled path-containment copies, and lint against new ones


blamechris:
## Summary

#7273 gave path containment one root-aware implementation
(`packages/server/src/utils/path-containment.js`) and routed **fifteen** call sites through it — the ws-file-ops, `file_ref` attachment, conversation-scope and IDE surfaces. It did **not** convert every copy in the repo, and it added no lint to stop a sixteenth being written.

This issue tracks the remainder. Each was verified as a genuine survivor during the #7286 review (the file is untouched by that PR's diff).

## The survivors

| site | spelling | assessment |
|---|---|---|
| `src/docker-byok-session.js:253` `remapToContainerPath` | `root + '/'` on a **host** path | **Confirmed broken on a Windows host.** Executed against the real exported function with `path` rebound to `path.win32`: `remapToContainerPath('C:\Users\dev\proj\src\a.js', 'C:\Users\dev\proj')` throws `EACCES "path outside workspace"`. Fail-closed, so not a hole — docker-byok simply cannot map any file on Windows. The comment directly above claims it normalises "so prefix-matching works on Windows hosts too"; it does not. Note lines 262/275 are **correct** as-is — those are container-internal POSIX paths. |
| `src/devcontainer-config.js:205-219` `validateMounts` | `absCwd.endsWith(sep) ? absCwd : absCwd + sep` | A fourth spelling, unlisted in the module header. Reproduced against the real exported function under a win32-bound `path`. |
| `src/pages-store.js:245-252` `resolveFile` | `root + sep`, twice (lexical + post-realpath) | Reviewed and found **not currently reachable** for a wrong answer — both operands derive from the same string and `resolve()` does no case folding. Still a second implementation of the predicate. |
| `src/permission-floor.js:229`, `src/permission-manager.js:236,242` | `relative()` + `isAbsolute(rel)` + `'..' + sep` | **Semantically correct** — a differential over 18 cases on both `path.posix` and `path.win32` (cross-drive, UNC, `\\?\`, drive root, `..hidden`, case variants) found **zero** divergence from `makeIsPathWithin`. Consolidate for single-implementation reasons only, and carefully: these sit on the protected-path floor (`docs/security/permission-floor.md`), which explicitly has one implementation by design. |
[...]

PR #7289, opened 2026-08-21
fix(server): stage the pathspec that was validated, not the client's string

### LINKED ISSUE #7281: security(server): gitStage forwards the raw client string to git, so a magic pathspec escapes the validated path
## Summary

`gitStage` / `gitUnstage` in `packages/server/src/ws-file-ops/git.js` validate a **resolved absolute path** but then hand git the **raw client string**:

```js
const absPath = normalize(resolve(cwdReal, file))
await validatePathWithinCwd(absPath, sessionCwd)   // validates the RESOLVED path
validatedFiles.push(file)                          // pushes the RAW client string
await execFileAsync(GIT, ['add', '--', ...validatedFiles], { cwd: cwdReal })
```

What is validated is not what is executed. `resolve()` treats the input as a filesystem path; `git add` treats it as a **pathspec**, which has its own magic syntax. A payload such as `:/` (pathspec magic for "from the repo root") resolves, as a plain path, to something harmless inside the cwd — and then means something entirely different to git.

This is **platform-independent** — it is not a Windows bug, and it is not fixed by #7273's containment work.

## Suggested fix

- Push the **validated, cwd-relative** path rather than the raw string, so what git receives is what was checked.
- Pass `--literal-pathspecs` (or the `GIT_LITERAL_PATHSPECS=1` env) so wildcards and magic prefixes can never be reinterpreted.
- Keep the existing `--` separator.

## Verification

Prove it red before fixing: craft a pathspec payload that passes `validatePathWithinCwd` and demonstrably stages something outside the intended set, then confirm the fix makes that case fail. A test that only stages ordinary relative paths will pass before and after and prove nothing.

## Provenance

Surfaced by the adversarial review panel run while root-causing #7273. Flagged there as low-confidence on the Windows framing; the reviewer's re-derivation confirmed the general thesis (validated ≠ executed) holds on **every** platform, which is why it is filed separately rather than folded into #7273.

blamechris:
## What

`gitStage`/`gitUnstage` validated a **resolved absolute path** and then handed git the
**raw client string**. git reads that string as a *pathspec*, not a path, and a pathspec
has its own magic syntax — so what was checked and what was executed were written in
different languages.

`:/` resolves, as a filesystem path, to a harmless `<cwd>/:` that passes containment. To
git it means "from the repo root".

## Proven, not argued

Driving the real handlers against a real repo whose session cwd is a subdirectory, with
the workspace root set to that subdirectory:

| payload | server replied | actually staged |
|---|---|---|
| `:/` `:(top)` `:/*` `:(top,glob)**` | `error: null` | a file **above the workspace root** |
| `:/OUTSIDE-THE-WORKSPACE.txt` | `error: null` | that exact file |
| `:!inside.txt` `:(exclude)nothing` | `error: null` | the whole repo |
| `../OUTSIDE-...` | `Access denied` ✅ | nothing |

Two distinct mechanisms: **root-anchoring** re-bases the pathspec on the repo root, and
**exclusion-only** pathspecs mean "everything except", resolved repo-wide. Any
authenticated client could do this in one WebSocket message. Platform-independent — not a
Windows defect, and not covered by #7273.

## The fix

Two halves, and only both together close it:

- **`--literal-pathspecs`** on the invocation, disabling all magic and glob expansion.
  This is the load-bearing half — re-deriving the path is *not* sufficient on its own,
  because `relative()` leaves `:/etc/shadow` byte-identical.
- **handing git the validated path** rather than the client's string.

The pathspec is derived from the lexically-resolved `absPath`, *not* the validator's
symlink-resolved `realPath`: for a symlink inside the cwd, `realPath` names its target, so
staging it would record a different object than the client asked for.

But `absPath` is not always inside the cwd either — containment is decided on `realPath`,
so it approves a path that merely *reaches* the cwd through a symlink or aliased prefix.
The second commit handles that, with `isPathWithin` (the same root-aware predicate from
#7273, not a second implementation), giving a stateable invariant: **the function never
returns a pathspec that leaves the cwd.**

## Bugs this fixes beyond the escape

Found while verifying, each measured on `main`:

- **Glob over-staging.** Selecting the single file `a[bc].txt` stages `a[bc].txt`,
  `ab.txt` **and** `ac.txt` — two files the client never selected, reported as success.
- **A symlink above the cwd pointing back into it** (`../lure.txt`) staged `lure.txt`,
  above the workspace root, replying `error: null`. Open on `main`; closed here.
- **A file whose name begins with `:` is un-stageable today** — git reads it as magic.
  Now stageable.
- **An absolute path through a symlinked directory** errored with "beyond a symbolic
  link"; now works.

## Verification

Red-then-green against `origin/main`, by restoring the original file and re-running:
**12 failures before, 0 after** (exit code checked, not output grepped). The suite carries
positive controls so it cannot pass by denying everything — the failure mode
`docs/false-safety-guards.md` entry 11 names *this exact test file* for:

- ordinary staging still succeeds;
- a `:`-prefixed filename **is** stageable, which fails if `--literal-pathspecs` is ever
  dropped — a positive proof the flag is armed, not merely that something was blocked.

Fixture geometry is what makes the bug observable: every pre-existing test in this file
uses the **repo root** as the session cwd, where `:/` has nothing to reach. That is why
none of them caught it.

184/184 green across the eleven affected suites; all 8 server lints pass. The
`EXPECTED_EMIT_SITES` drift guard in `ws-git-result-schemas.test.js` correctly went red on
the new rejection branch — the new branch is now driven by that file's fixture and the
counts updated, and I verified the bumped guard still fires on a 7th site.

## Not fixed here — filed separately

An adversarial audit of every client-controlled argv reaching a spawned process turned up
neighbours that are **not** this bug and are not fixed by `--literal-pathspecs`. They are
filed as their own issues rather than widened into this PR.

Closes #7281

### DIFF
--- a/docs/false-safety-guards.md
+++ b/docs/false-safety-guards.md
@@ -287,6 +287,69 @@ is also wrong, on both platforms: `'C:\Documents'.startsWith('C:\' + '\')` and
 a separator denies everything. Two of the six copies had already "fixed" it that
 way.
 
+### 12. The guard that validated one language and executed another — `#7281`
+
+`gitStage` resolved the client's string to an absolute path, asked
+`validatePathWithinCwd` whether that path was inside the project, and — having
+been told yes — handed **the client's original string** to `git add`.
+
+The check was correct. The path it checked was never the thing that ran.
+`git add` takes a *pathspec*, not a path, and a pathspec has its own grammar.
+`:/` resolves, as a filesystem path, to a harmless `<cwd>/:` that is plainly
+inside the project; to git it means "from the repository root". From a session
+confined to a subdirectory, one WebSocket message staged files above the
+workspace root, and the server replied `error: null`.
+
+Two grammars, two mechanisms: root-anchoring (`:/`, `:(top)`, `:/*`,
+`:(top,glob)**`) re-bases the pathspec on the repo root, and exclusion-only
+(`:!x`, `:(exclude)x`) means "everything except", resolved repo-wide.
+
+**Why no test caught it.** Every fixture in `ws-file-ops-git-paths.test.js` and
+`git-stage-commit.test.js` used the repo root as the session cwd — the one
+topology in which `:/` has nothing to reach. The suite tested the escape
+mechanism it knew about (`../`, correctly refused) and never asked about the
+one that worked. Fixture *geometry*, not payload, was what made the bug
+invisible; a reviewer scanning six adversarial payloads in a describe block
+named "path validation" would reasonably conclude the surface was covered.
+
+**The fix has two halves and needs both.** `--literal-pathspecs` disables the
+magic; handing git the validated path restores "what is checked is what runs".
+Re-deriving the path alone is *not* sufficient — `relative()` leaves
+`:/etc/shadow` byte-identical.
+
+**Two traps inside the fix**, both found by review after the first cut shipped:
+
+- Deriving the pathspec from the validator's `realPath` stages a symlink's
+  TARGET. Deriving it from the lexical `absPath` can point *out* of the cwd,
+  because containment is decided on `realPath` — so it says yes to a path that
+  merely reaches the cwd through a symlink or an aliased prefix (the ordinary
+  macOS `/var` → `/private/var` case, not an attack). Neither alone is right.
+- Falling back to `realPath` then re-introduces the original sin in miniature:
+  when the fallback yields an empty relative path, answering `'.'` turns a
+  one-path request into a whole-cwd stage. Measured: `['../aliasdir']` staged
+  every file in the cwd, `error: null`.
+
+**And the flag is silent on half the surface.** `git reset` does not error on a
+non-matching pathspec the way `git add` does, so deleting `--literal-pathspecs`
+from the unstage invocation left all 31 tests green. The guard needed its own
+positive control: a file whose name begins with `:` is unstageable *without* the
+flag and unstageable *with* it, so the assertion goes red the moment the flag is
+dropped. The stage half has the mirror control.
+
+**Environment can refuse the fix outright.** git rejects `--literal-pathspecs`
+when the environment also selects another global pathspec mode, so an inherited
+`GIT_GLOB_PATHSPECS=1` or `GIT_ICASE_PATHSPECS=1` would make *every* stage die
+with `fatal: global 'literal' pathspec setting is incompatible with all other
+global pathspec settings`. The daemon inherits the user's shell environment;
+those variables are dropped for our own invocations.
+
+The generalisation is worth more than the instance: **whenever a validated value
+is handed to something that parses it under a different grammar — a pathspec, a
+revision, a glob, a shell word, a URL — the validation proves nothing.** The
+sibling instances that audit turned up are `#7290` (a revision allowlist whose
+comment claims it blocks flags, while `-` is inside its character class) and
+`#7291` (client prompts in positional argv slots with no `--`).
+
 ## The one that got me while writing this
 
 A Maestro `repeat` was written with `maxRuns: 3`. `YamlRepeatCommand` has no
--- a/packages/server/src/ws-file-ops/git.js
+++ b/packages/server/src/ws-file-ops/git.js
@@ -1,4 +1,4 @@
-import { normalize, resolve, join } from 'path'
+import { normalize, resolve, join, relative, sep } from 'path'
 import { execFile as execFileCb } from 'child_process'
 import { promisify } from 'util'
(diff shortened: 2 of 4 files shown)


blamechris:
## Code Review: PR #7289

Dimension panel (5 independent lenses: security, test-quality, correctness-regression, Windows/cross-platform, style-architecture), with adversarial refuters on every critical and suggestion finding — 2 refuters per finding, distinct lenses, instructed to default to REFUTED. 23 agents, 0 errors.

### Verdict

**Request Changes → all findings now addressed** in `fb5f1dab2` and `0829d530b`. Every fix is proven by a targeted mutation that reverts exactly one guard and turns exactly one test red.

### Findings

| # | Severity | Finding | Refuters | Resolution |
|---|---|---|---|---|
| 1 | **critical** | `toLiteralPathspec` widened a **one-path request into a whole-cwd stage** when the realPath fallback yielded an empty relative path — a client naming something outside the cwd that *resolves* to it. Measured: `['../aliasdir']` staged all three files, `error: null`. The unstage direction discards the user's entire staged set. | 6/6 UPHELD | `fb5f1dab2` — `'.'` is returned only when the client's own path is lexically inside the cwd; the fallback rejects |
| 2 | suggestion | **`--literal-pathspecs` on the unstage half had no failing proof.** `git reset` doesn't error on a non-matching pathspec the way `git add` does, so deleting the flag from the reset argv left all 31 tests green — a guard that could not go red. | 4/4 UPHELD | `fb5f1dab2` — mirror positive control: a `:`-prefixed filename is unstageable without the flag, unstageable *with* it |
| 3 | suggestion | **The test suite could not see a corrupted derivation.** Every path it used was a single component, so truncating the pathspec to its last component left **89/89 green** across five suites. | 2/2 UPHELD | `0829d530b` — nested path + same-basename decoy at the cwd root, for stage and unstage. Same mutant now fails exactly those two tests |
| 4 | suggestion | No `docs/false-safety-guards.md` entry, against the repo's established practice for exactly this class. | 4/4 UPHELD | `fb5f1dab2` — entry 12 |
| 5 | nitpick | `--literal-pathspecs` is **refused by git** when the environment selects another global pathspec mode. A daemon inheriting `GIT_GLOB_PATHSPECS=1` or `GIT_ICASE_PATHSPECS=1` would fail *every* stage with `fatal: global 'literal' pathspec setting is incompatible...`. Not an attack — just a user with an exported variable. | verified directly | `fb5f1dab2` — the family is dropped for our own invocations, with a test per variable |
| 6 | nitpick | `'Invalid file path: empty'` was also emitted for a non-string element, where "empty" is untrue. | — | `fb5f1dab2` — the message names which |

### What the security lens could not defeat

30 payloads against the real handlers with the session cwd a strict subdirectory: `:/`, `:(top)`, `:/*`, `:(top,glob)**`, `:!x`, `:(exclude)x`, `:(attr:x)`, `:(icase)`, `::/`, `:::/`, `.git/config`, NUL-embedded strings, and traversal in several spellings. **Zero escapes.** Root-anchoring payloads reach git as inert literals; traversal is denied by containment; a NUL string throws inside Node and is caught fail-closed. Fail-fast ordering holds — a hostile *last* element aborts with nothing staged.

### Strengths noted by the panel

- `--literal-pathspecs` removes the entire magic grammar rather than blacklisting spellings, so there is no residue to enumerate.
- The `absPath`-over-`realPath` preference is a subtle call most fixes would get wrong, and it is pinned by a real mode-`120000` assertion.
- The suite is explicit about its own limits: why `:(glob)**` was excluded as "a test that cannot fail", why fixture *geometry* rather than payload is what hid the bug, and one case recorded-but-not-asserted with the reason it can't be a portable fixture.

### Deferred — filed, not folded in

| Finding | Issue |
|---|---|
| `getDiff`'s allowlist admits leading-dash git options, behind a comment claiming it blocks them | #7290 |
| Client prompts in positional argv slots with no `--` (web-task-manager, codex-session), plus a substring feature-gate that reports `--remote` available on a CLI that only has `--remote-control` | #7291 |
| `gitStatus` emits repo-root-relative, C-quoted paths that `gitStage` re-resolves cwd-relative — the stage round-trip is broken for subdirectory sessions | #7292 |
| `claude-tui` `respondToQuestion` test asserts a throttled PTY write after a fixed 100 ms sleep (flaked on this PR's CI, unrelated to it) | #7293 |

### Architecture

Reuses `isPathWithin` from #7273 rather than adding a second containment implementation — the rule `docs/security/permission-floor.md` and #7287 both insist on. Containment is still decided by `validatePathWithinCwd`; `toLiteralPathspec` only chooses which validated form to name.

Issue #7291, opened 2026-08-21
security(server): client prompt lands next to a flag with no `--` separator (web-task-manager, codex-session)


blamechris:
## Summary

Two spawn sites put a raw client string in a **positional** argv slot with no `--`
separator, so a value beginning with `-` is parsed as an option rather than as text.
Distinct class from #7281's pathspec magic — `--` is what would protect these, and it is
absent.

### 1. `web-task-manager.js:193`

```js
const child = execFile('claude', ['--remote', task.prompt], { cwd: task.cwd, timeout: 300_000 }, ...)
```

`task.prompt` is client-supplied. A prompt of `--print` or
`--dangerously-skip-permissions` is parsed as an option to a locally-spawned agent binary.

### 2. `web-task-manager.js:66-68` — the gate that was supposed to prevent this is broken

```js
const stdout = await exec('claude', ['--help'])
this._remoteAvailable = stdout.includes('--remote')
```

Substring match. The current claude CLI has **no `--remote` flag at all** — its help text
contains only `--remote-control` and `--remote-control-session-name-prefix`. So
`includes('--remote')` is `true` on a CLI that does not support the flag, and the
availability gate that was meant to keep client prompts out of that argv is open. This is
the repo's signature defect class: a check that reports success without checking
(`docs/false-safety-guards.md`).

### 3. `codex-session.js:228-230`

```js
const args = threadId
  ? ['exec', '--sandbox', sandbox, 'resume', threadId, text, '--json', '--skip-git-repo-check']
  : ['exec', ...]
```

`text` is the raw user prompt in a positional slot — the highest-frequency client input in
the product (every chat message on a codex-exec session).

## Suggested fix

- `detectFeatures`: word-boundary match, `/--remote(?![-\w])/`. Needs a negative control
  proving it goes red on a help text that contains only `--remote-control`.
- Both spawn sites: terminate option parsing with `--` where the CLI supports it, or
  reject a leading `-` explicitly.

## Verification

For the gate, feed a captured `--help` text containing only `--remote-control` and assert
`_remoteAvailable === false`. For the argv sites, assert the spawned argv contains a `--`
before the client string, or that a leading-dash prompt is refused.

## Provenance

Adversarial argv audit run while fixing #7281 (#7289). The `--remote` substring false
positive was verified by piping the installed CLI's real `--help` through the check.
Filed separately: different injection class, different files, not fixed by
`--literal-pathspecs`.

blamechris:
[...]

PR #7294, opened 2026-08-21
fix(server): stop client strings being parsed as CLI options

### LINKED ISSUE #7290: security(server): getDiff's "flag injection" allowlist admits leading-dash git options
## Summary

`getDiff` in `packages/server/src/ws-file-ops/reader.js:563-568` validates a
client-controlled revision with an allowlist whose comment claims it stops flag
injection:

```js
const rawBase = (typeof base === 'string' && base.trim()) ? base.trim() : 'HEAD'
// Validate ref name to prevent git flag injection
const diffBase = /^[a-zA-Z0-9._\-\/~^@{}:]+$/.test(rawBase) ? rawBase : 'HEAD'
...
await execFileAsync(GIT, ['diff', diffBase], { cwd: cwdReal, ... })
```

`-` is **inside** the character class, so a leading dash is not blocked. Measured
against the regex and then against git 2.54.0:

| payload | regex | git |
|---|---|---|
| `--stat` | passes | parsed as an option, rc=0 |
| `-p` | passes | parsed as an option, rc=0 |
| `--raw` | passes | parsed as an option, rc=0 |
| `--exit-code` | passes | parsed as an option, rc=0 |
| `--ext-diff` | passes | parsed as an option |
| `-O/etc/passwd` | passes | parsed as an option |
| `--output=/tmp/x` | **blocked** | — (`=` is not in the class) |
| `:(top)` | **blocked** | — |

The blast radius **today** is small: only single-token options with no `=` get through,
and there is exactly one client-controlled argv element. But the guard does not do what
its comment says it does, so the safety is nominal — adding any future git option that is
dangerous in single-token form, or a second client-controlled token, makes it live with
no code change here.

This is also the position `--` cannot protect: `diffBase` is a **revision**, not a
pathspec, so the #7281 fix (`--literal-pathspecs`) does not constrain it.

## Suggested fix

- Reject a leading `-` explicitly. The repo already has exactly this check —
  `assertSafeRef` in `packages/server/src/orchestration/git-ops.js:47` — and
  `supervisor.js` validates a ref *in the form git will parse it*. Reuse that rather than
  writing a third spelling.
- Consider `['diff', diffBase, '--']` so the argument cannot drift into a pathspec slot.

## Verification

Prove it red first: assert that a `--stat` base does **not** reach git as an option. A
test that only feeds valid refs passes before and after and proves nothing. Per
`docs/false-safety-guards.md`, the guard needs a negative control that goes red when the
leading-dash check is deleted.

## Provenance

Surfaced by the adversarial argv audit run while fixing #7281 (see #7289), and
independently re-derived by running the regex and git directly. Filed separately because
it is a different injection class in a different argv position, and is not fixed by
`--literal-pathspecs`.

blamechris:
## What

`execFile` with an array argv blocks **shell** injection — no shell ever sees the string.
It does not block **argument** injection: the spawned program still runs its own option
parser over that array, so a client-controlled value beginning with `-` is read as an
option. Three sites got that wrong, in two different ways, and the guards that were
supposed to stop it described stronger checks than they performed.

Closes #7290. Partially addresses #7291 — see *Deliberately left out* below.

## The `getDiff` revision allowlist (#7290)

```js
// Validate ref name to prevent git flag injection
const diffBase = /^[a-zA-Z0-9._\-\/~^@{}:]+$/.test(rawBase) ? rawBase : 'HEAD'
```

`-` is inside the character class, so it is a permitted *character*, and the regex says
nothing about *position*. Measured against git 2.54.0, every single-token option passed
and was then parsed as an option: `--stat`, `-p`, `--raw`, `--exit-code`, `--ext-diff`,
`--word-diff`, `-U99999`, `-O<path>`. Only forms containing `=` were blocked, and only
because `=` is absent from the class — an accident, not a defence.

`base` is unconstrained on the wire: `ClientGetDiffSchema` is
`z.object({ type: z.literal('get_diff') }).passthrough()`, and `handleGetDiff` passes
`msg.base` straight through. One hop, no guard.

**The impact is larger than the issue estimated.** `git diff -O<file>` reads `<file>` as a
diff orderfile, and `getDiff` forwards raw git stderr to the client
(`error: err.message`). That makes `-O` a **file existence/readability oracle over the
entire filesystem, as the daemon user, escaping the session cwd**, with the answer
returned on the wire:

```
-O/etc/passwd            -> (silent: readable)
-O/etc/shadow            -> fatal: failed to read orderfile … No such file or directory
-O~/.ssh                 -> fatal: … Is a directory
-O~/.chroxy/config.json  -> (silent: readable — outside the workspace)
```

### The issue's own suggested fix does not work

#7290 suggests `['diff', diffBase, '--']`. Measured — it does not help, because `--` ends
option parsing at **its own position** and `diffBase` precedes it:

```
git diff --stat --        still applies --stat
git diff --exit-code --   still exits 1
git diff -O/etc/nope --   still reads the orderfile
```

`--literal-pathspecs` (#7281/#7289) does not help either — that constrains the *pathspec*
language, and a revision is not a pathspec. Rejecting the leading dash is the only thing
that works here, so that is what this does. The pre-existing silent fallback to `HEAD` is
kept, so the wire contract for an unusable base is unchanged.

## The `--remote` availability gate (#7291)

`this._remoteAvailable = stdout.includes('--remote')` is a substring match, satisfied by
the longer `--remote-control`. Measured against the **installed** CLI, which advertises
`--remote-control` and `--remote-control-session-name-prefix` and has no `--remote` at all:

```
help.includes('--remote')      -> true    (gate opens)
/--remote(?![\w-])/.test(help) -> false   (correct)
```

So the gate reported the feature available on a CLI that cannot accept it — a check that
reports success without checking, and the thing that was supposed to keep a client prompt
out of the argv in `_spawnRemoteTask`.

The existing test for this gate fed a help text containing **no `--remote` substring at
all**, so the naive check already answered it correctly. It passed before and after this
fix and proved nothing. Replaced by a control that carries `--remote-control` and asserts
`remote === false` while `--teleport` (genuinely present in the same fixture) stays `true`.

## Reject vs `--`: which fix applies

This is the part that is easy to get backwards, so it is stated once in
`utils/argv-safety.js` and referenced from each site:

- A value that **can never legitimately begin with `-`** (a git revision, a branch) →
  **reject**. `getDiff`, `assertSafeRef`.
- A value that **legitimately can** (a user's chat message — `- first point`) → **do not
  reject**; terminate option parsing with `--`, with every flag placed *before* it.
  `_spawnRemoteTask`. Rejecting there would break ordinary prompts.

That `claude` honours `--` was measured directly, with a positive control:

```
claude --nonexistent-flag-xyz     -> error: unknown option   (option-parsed)
claude -- --nonexistent-flag-xyz  -> no such error           (treated as the prompt)
```

## One spelling, not three

#7290 asks for the existing `assertSafeRef` to be reused rather than a third spelling
written. The predicate now lives in `packages/server/src/utils/argv-safety.js` and is
shared by `ws-file-ops/reader.js` and `orchestration/git-ops.js`. `assertSafeRef` survives
only as a thin wrapper that keeps throwing `GitOpsError` and preserves the `empty` vs
`unsafe` message split that `orchestration-git-ops.test.js` pins.

## Verification

Every guard is **mutation-tested** — reverting each one turns its own tests red, and each
has a **positive control** so a guard that refused everything could not pass:

| mutation | result |
|---|---|
| drop `isSafeArgvValue` from `getDiff`, keep the allowlist | **3 red**, positive control green |
| revert the gate to `stdout.includes('--remote')` | **1 red** |
| revert the argv to `['--remote', prompt]` | **2 red** |
| all restored | 29/29 green |

The `getDiff` tests were written **red first** against unfixed code, and each observes a
git behaviour only reachable when the option was actually parsed (the orderfile fatal,
`--exit-code`'s rc=1, `--stat` collapsing the unified diff to zero files) rather than
asserting on the argv array.

- **215/215** across every affected suite: `argv-safety`, `web-task-manager`,
  `web-task-handlers`, `orchestration-git-ops`, `ws-server-file-ops`, `file-handlers`,
  `ws-file-ops-git-paths`, `git-stage-commit`.
- All **8** custom server lints OK. eslint: 0 errors, and no new warnings in any touched
  file.

Catalogued as `docs/false-safety-guards.md` entry 13, which entry 12 already predicted.

## Deliberately left out

- **#7291's `codex-session.js` site** stays open. The fix there is not a one-token
  insertion: the flags come *after* the positional prompt, so adding `--` requires
  **reordering** the argv, and `buildCodexArgs` carries a documented hard-won invariant
  (`--sandbox` must precede the `resume` subcommand). The repo's own spawn-and-assert clap
  harness (`tests/integration/codex-spawn-argv.integration.test.js`) is the only honest
  proof for that change, and it self-skips here because this host's codex install is broken
  (`vendored binary ENOENT`). Shipping an unverified argv reorder against a fragile
  ordering invariant is not worth it; filed as a follow-on instead.
- **`buildRemoteTaskArgs` cannot be exercised against a live binary**, because the shipping
  CLI has no `--remote` at all — and after the gate fix, that path is correctly closed on
  such a CLI. The shape is asserted against the CLI's documented grammar,
  `claude [options] [command] [prompt]`, in which the prompt is positional. Stated
  explicitly in the function's doc comment rather than left implied.

### DIFF
--- a/docs/false-safety-guards.md
+++ b/docs/false-safety-guards.md
@@ -348,7 +348,88 @@ is handed to something that parses it under a different grammar — a pathspec, 
 revision, a glob, a shell word, a URL — the validation proves nothing.** The
 sibling instances that audit turned up are `#7290` (a revision allowlist whose
 comment claims it blocks flags, while `-` is inside its character class) and
-`#7291` (client prompts in positional argv slots with no `--`).
+`#7291` (client prompts in positional argv slots with no `--`) — both
+now catalogued as entry 13.
+
+### 13. The allowlist that permitted what its comment forbade — `#7290`, `#7291`
+
+Entry 12 predicted these two as its siblings. Both are the same shape: a guard
+whose *comment* describes a stronger check than its *code* performs.
+
+`getDiff` validated a client-controlled revision with
+
+```js
+// Validate ref name to prevent git flag injection
+const diffBase = /^[a-zA-Z0-9._\-\/~^@{}:]+$/.test(rawBase) ? rawBase : 'HEAD'
+```
+
+`-` is inside the character class, so it is a permitted *character*, and the
+regex has nothing to say about *position*. Measured against git 2.54.0, every
+single-token option passed and was then parsed as an option: `--stat`, `-p`,
+`--raw`, `--exit-code`, `--ext-diff`, `--word-diff`, `-U99999`, `-O<path>`.
+Only forms containing `=` were blocked, and only because `=` is absent from the
+class — an accident, not a defence.
+
+**The impact was larger than "git errors out".** `git diff -O<file>` reads
+`<file>` as a diff orderfile, and `getDiff` forwards raw git stderr to the
+client (`error: err.message`). So `-O` is a file existence/readability oracle
+over the entire filesystem, as the daemon user, escaping the session cwd
+completely, with the answer returned on the wire:
+
+```
+-O/etc/passwd                 -> (silent: readable)
+-O/etc/shadow                 -> fatal: failed to read orderfile … No such file or directory
+-O~/.ssh                      -> fatal: … Is a directory
+-O~/.chroxy/config.json       -> (silent: readable — outside the workspace)
+```
+
+The field is unconstrained on the wire: `ClientGetDiffSchema` is
+`z.object({ type: z.literal('get_diff') }).passthrough()`, so nothing upstream
+narrows it either.
+
+**The obvious fix does not work, and the issue itself proposed it.** Appending
+a `--` separator — `['diff', diffBase, '--']` — is ineffective, because `--`
+ends option parsing at *its own position* and `diffBase` precedes it:
+
+```
+git diff --stat --        still applies --stat
+git diff --exit-code --   still exits 1
+git diff -O/etc/nope --   still reads the orderfile
+```
+
+`--literal-pathspecs` (entry 12's fix) does not help either: it constrains the
+*pathspec* language, and a revision is not a pathspec. Rejecting the leading
+dash is the only thing that works here.
+
+**Which fix applies is decided by whether a leading `-` is legitimate**, and
+this is the part that is easy to get backwards. A git revision can never begin
+with `-`, so rejection is right. A user's chat message legitimately can
+("- first point"), so rejecting a prompt would break normal use — there, the
+`--` separator *is* the fix, with every flag moved before it. `#7291`'s
+`_spawnRemoteTask` needed the second treatment, not the first.
+
+**And the gate that was supposed to keep the prompt out of that argv reported
+success without checking.** Feature detection was `help.includes('--remote')`,
+a substring match. Measured against the installed CLI, whose help text carries
+`--remote-control` and `--remote-control-session-name-prefix` and no `--remote`
+at all:
+
+```
+help.includes('--remote')      -> true    (gate opens)
+/--remote(?![\w-])/.test(help) -> false   (correct)
+```
+
+The existing test for this gate fed a help text containing no `--remote`
+substring *at all*, so the naive check already answered it correctly. It passed
+before and after the fix — the recurring test shape in this document: a case
+the guard already handles proves nothing about the case it does not.
+
+**The generalisation.** `execFile` with an array argv stops *shell* injection —
+no shell ever sees the string — and it is easy to write a comment claiming that
+covers injection generally. It does not stop *argument* injection: the spawned
+program still runs its own option parser over that array. The two are different
+classes with different fixes, and "we use execFile, not exec" is an answer to
+only one of them.
 
 ## The one that got me while writing this
 
--- a/docs/false-safety-guards.md
+++ b/docs/false-safety-guards.md
@@ -351,7 +351,13 @@ comment claims it blocks flags, while `-` is inside its character class) and
 `#7291` (client prompts in positional argv slots with no `--`) — both
 now catalogued as entry 13.
 
-### 13. The allowlist that permitted what its comment forbade — `#7290`, `#7291`
+### 13. The allowlist that permitted what its comment forbade — `#7290`, `#7291` (partial)
+
+`#7290` is closed. `#7291` is **not** — its `codex-session.js` half is still
+open, because fixing it needs an argv reorder around a documented
+`--sandbox`-before-`resume` invariant that only the repo's spawn-and-assert
+clap harness can verify, and that harness needs a working codex binary. This
+entry describes the half that landed; do not read it as closing `#7291`.
 
 Entry 12 predicted these two as its siblings. Both are the same shape: a guard
 whose *comment* describes a stronger check than its *code* performs.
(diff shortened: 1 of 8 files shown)


blamechris:
## Code Review: PR #7294

Adversarial panel — 6 independent dimension finders, then 2 refuters per finding on a
distinct lens each (does-it-reproduce / scope-and-novelty). 19 agents, 24 raw findings.
Nitpicks skip refutation by policy. Every measurement below was re-derived against a real
binary, not taken from the finder's word.

### Verdict

**Comment** — nothing blocks. One defect the PR *introduced* was found and has been **fixed
in `b79da14d2`**; the two critical-severity survivors are real but **pre-existing** (the code
they indict is byte-identical to `origin/main`), so they go to follow-ups.

### The one regression this PR introduced — fixed

`['--remote', '--', prompt]` is only correct when `--remote` is boolean or optional-arg.
Re-measured against the repo's commander 12.1.0 with a prompt of
`--dangerously-skip-permissions`:

| `--remote` declared as | `['--remote', prompt]` (before) | `['--remote', '--', prompt]` (this PR) |
|---|---|---|
| boolean | **injects** | safe |
| `[name]` optional | **injects** | safe |
| `<name>` **required** | safe — swallowed as the flag's value | **injects** |

Against a required-argument flag the `--` becomes the flag's own value and frees the prompt
to be option-parsed — strictly worse than the code it replaced. Unreachable today (the
shipping CLI has no `--remote`, and this PR's own gate fix is what keeps that path closed),
but it would arm itself on a CLI upgrade with no change on our side.

Fixed by gating on the advertised arity: new `cliHelpFlagArity()`, and `detectFeatures`
refuses the feature when it cannot use a separator safely. Mutation-tested.

### Also fixed in `b79da14d2`

- **`cliHelpAdvertisesFlag` had no leading boundary**, so it returned `true` for `x--remote`,
  `---remote`, and for a `-O` probe against a help line advertising `--O <file>` — the exact
  mirror of the defect the function exists to prevent.
- **The PR over-claimed, which a PR about over-claiming guards should not do.** `getDiff`'s
  comment and catalogue entry 13 said the fix closed a filesystem oracle "over the whole
  disk". It closes the **leading-dash route only** — see the survivors below. Both now say so.
- **`ClientGetDiffSchema` does not exist** — the schema is `GetDiffSchema`. Corrected in three
  places.
- **`git-ops.js` claimed it "used to be the only copy".** False: a third spelling,
  `rejectGitOptionLike`, is live in `k8s.js:322`. Now named.
- **The `-O~/.ssh` row in entry 13 was unreproducible** — a `~` never expands on an argv path,
  so that probe only ever reports "No such file". Replaced with absolute paths.
- **Entry 13 read as closing #7291**; its `codex-session.js` half is still open. Now says so.
- **`assertSafeRef` dropped the original error.** Now preserved as `cause` — set after
  construction, because `GitOpsError`'s constructor takes `{ args, stderr }` and would have
  silently discarded a `cause` passed to it.
- **A test gap the panel was right about:** `buildRemoteTaskArgs` was covered only as a pure
  function while every test stubs `_spawnRemoteTask`, so inlining the argv literal back into
  the call site left the suite **green** — the guard was wired to none of its callers. Added
  a call-site assertion that goes red on exactly that.

### Critical survivors — real, but pre-existing → #7298

The oracle #7290 set out to close **survives via routes needing no dash**, because `:` and `/`
are both in the charset allowlist and git's stderr is forwarded verbatim:

```
base='HEAD:/etc/passwd'  -> fatal: path '/etc/passwd' exists on disk, but not in 'HEAD'
base='HEAD:absent.txt'   -> fatal: path 'absent.txt' does not exist in 'HEAD'
base='/etc/passwd'       -> fatal: '/etc/passwd' is outside repository at '<cwdReal>'
base='/no/such/file'     -> fatal: ambiguous argument ...
```

All pass both halves of the guard. The charset and the `error: err.message` forwarding are
unchanged from `main`, so this PR neither introduces nor widens it — but it did read as
though it sealed it, which is now corrected in the code comment and in entry 13. Closing it
means resolving the base with `rev-parse --verify` and not forwarding raw git stderr: a wider
contract change, tracked as #7298. Note `handleGetDiff` has no `rejectMutationIfBound` gate,
so this is reachable by a bound share-a-session token.

| Item | Issue | Rationale for deferral |
|---|---|---|
| `getDiff` path oracle via `HEAD:<path>` / bare absolute paths; `rev-parse` rework + stop forwarding raw git stderr | [#7298](https://github.com/blamechris/chroxy/issues/7298) | Pre-existing; needs a wire-contract change wider than this fix |
| `'ambiguous argument'` not matched by the `'unknown revision'` recovery | [#7298](https://github.com/blamechris/chroxy/issues/7298) | Same function, same fix — folded in |
| Silent `detectFeatures` catch (its caller's error branch is dead code); dash-permissive `remoteTaskId` regex; post-`destroy()` polling restart | [#7299](https://github.com/blamechris/chroxy/issues/7299) | Pre-existing; diagnosability and lifecycle, not this defect class |
| `Grep` tool pattern → `rg` positional slot; `--pre=` gives program execution through an auto-allowed read-only tool | [#7295](https://github.com/blamechris/chroxy/issues/7295) | Different subsystem; highest-severity find of the sweep |
| Provider prompt argv — `codex` (both branches) and the `gemini` twin | [#7291](https://github.com/blamechris/chroxy/issues/7291) | codex needs a working binary for the clap harness; gemini needs `--prompt=` not `--` |
| `docker run` image, `cloudflared` tunnel name, deploy ref; plus the `k8s.js` `--` misfix | [#7296](https://github.com/blamechris/chroxy/issues/7296) | Lower severity, mostly local-config-only |

### Strengths

- The doctrine header in `argv-safety.js` is the real deliverable: three fix shapes, each
  backed by a measurement against a real binary, and an explicit note that shape 2 *breaks*
  on a `requiresArg` flag — which pre-empts a future "add `--` everywhere" sweep. It also
  records the probe trap (`gemini -p --version` prints a version and looks like injection
  when it is not).
- `git-ops.js`'s delegation is behaviour-preserving across ~35 input classes × 4 `kind`
  values — identical messages, constructor, `err.name` and top stack frame.
- `cliHelpAdvertisesFlag` fixes a genuinely live false-safety gate, confirmed against the
  installed CLI 2.1.235.

### Verification

225/225 across every affected suite, all 8 custom server lints OK, eslint 0 errors with no
new warnings. Four new guards each mutation-tested — dropping the leading boundary, the arity
detection, the arity gate, and the call-site wiring turns 1, 4, 1 and 1 tests red
respectively; all restored green.

Issue #7351, opened 2026-08-24
notifications: Notification.requestPermission() is never called — chroxy's only native-notification path is unreachable, and its test hand-grants the permission


blamechris:
## Summary

A permission prompt appeared, the user was never notified, it sat unanswered for its full timeout, **expired**, and the agent continued **without the tool** — silently degrading the work. The user only discovered it by scrolling back through the transcript afterwards:

> *"Bash: Fix the deadline race properly — Timed out — Permission expired — Claude will continue without this tool"*

Two distinct defects combine here. The second one is the reason no notification arrived, and it is severe: **chroxy's only native-notification path can never fire in production.**

## Defect 1 — `Notification.requestPermission()` is never called, so the feature is dead code

`packages/dashboard/src/hooks/usePermissionNotification.ts:22`:

```ts
if (typeof Notification === 'undefined' || Notification.permission !== 'granted') return
```

It returns early — silently — unless the browser/WKWebView has already granted notification permission. But **nothing in chroxy ever asks for it.** Grepping `packages/dashboard/src` and `packages/desktop/src-tauri/src` for `requestPermission` returns only chroxy's own unrelated `requestPermissionInput` WS action. The Web API call `Notification.requestPermission()` appears **nowhere**.

So `Notification.permission` remains `'default'` forever, the guard on line 22 returns on every invocation, and `new Notification(...)` on line 47 is unreachable. This is the **only** `new Notification` call site in the dashboard — so chroxy has, in practice, **no working native notifications at all**.

### Why no test caught it

`usePermissionNotification.test.ts` sets up the exact precondition that production never establishes:

```ts
globalThis.Notification.permission = 'granted'                                  // line 22
globalThis.Notification.requestPermission = vi.fn().mockResolvedValue('granted') // line 23
```

The suite passes, has passed continuously, and is *incapable* of failing on this: it hand-grants the permission the app never requests, and even stubs the request function the app never calls. This is precisely the pattern catalogued in `docs/false-safety-guards.md` — *"a precondition that is false, so the body never runs and the job is green."* Success and never-running are the same observable outcome.

## Defect 2 — an expired permission degrades the turn silently

When the prompt expired, the tool call was dropped and the agent **continued without it**, producing a result derived from strictly less capability than it asked for. The transcript records this, but nothing surfaces it at the time or afterwards: no notification, no session-level marker, no summary at turn end. The user's words: *"I never got notifications or saw anything to indicate I didn't approve something."*

[...]

Issue #7340, opened 2026-08-23
sessions: background agents that outlive a turn are force-completed at turn end — no signal that work is still running


blamechris:
## Summary

When a session's turn ends while it still has **background agents running** (subagents/tasks the model launched to run past the turn), Chroxy marks them all complete and shows the session as idle. There is no indicator anywhere — tab, sidebar, header — that work is still in flight, so the user cannot tell a finished session from one with four reviewers still running.

Observed during dogfooding: the model's own reply said *'4 in flight: security-reviewer, test-quality-reviewer, conventions-reviewer, ci-monitor'* while Chroxy showed the session idle with no activity signal.

## Root cause — turn end force-completes every tracked agent

Agent tracking exists and is wired for all the relevant providers (`_activeAgents` in `base-session.js`, populated by `cli-session.js`, `sdk-session.js`, `byok-session.js`), and the dashboard already consumes it — `App.tsx:2625` renders `<AgentMonitorPanel />` when `activeAgents.length > 0`, and `agentCount` is passed to the header.

But `base-session.js:1628-1634`, inside `_clearMessageState` (turn end), does this:

```js
// Emit completions for any tracked agents so the app clears badges
if (this._activeAgents.size > 0) {
  for (const agent of this._activeAgents.values()) {
    this.emit('agent_completed', { toolUseId: agent.toolUseId })
  }
  this._activeAgents.clear()
}
```

Every still-running agent is declared complete because the **turn** ended. The comment states the intent plainly — *'so the app clears badges'* — which is right for agents the model abandoned mid-turn (the orphan sweep the surrounding code is doing) and wrong for agents deliberately launched to outlive it. The two cases are indistinguishable at this point in the code, so the safe-looking choice silently discards live state.

The `ActivityIndicator` ('Working… last activity Ns ago') is likewise turn-scoped, so it disappears at the same moment.

## Why it matters

This is the difference between 'the agent finished' and 'the agent is waiting on four reviewers and a CI run'. On mobile especially — where the whole point is glancing at a session from elsewhere — an idle-looking session that is actually mid-review is a wrong answer to the only question the user is asking.

## Suggested direction

1. **Distinguish orphaned from backgrounded.** Track whether an agent was launched as background/detached; sweep only the former at turn end. Absent a reliable signal from the provider, at minimum do not *silently* complete them — emit a distinct event so the UI can show 'still running (turn ended)' rather than nothing.
[...]

Issue #3751, opened 2026-05-11
app: 44pt min touch targets for AskUserQuestion Other free-text Send/Cancel


blamechris:
## Context

Identified during review of PR #3750 (which closes #3746).

## Description

The new free-text input row in `packages/app/src/components/chat/MessageBubble.tsx` (Other → free-text fallback) uses `paddingVertical: 8` plus `fontSize: 14` on its Send and Cancel buttons. That works out to roughly 30pt of vertical hit area — below the React Native iOS HIG / Android Material 44pt minimum touch target. The existing option buttons in the same component use a larger style.

Project guideline (CLAUDE.md → Mobile / RN): `44pt min touch targets`.

## Original Review Comment

> Send and Cancel buttons in the new `promptFreetextRow` are `paddingVertical: 8` + 14px text ≈ ~30pt — below the 44pt min touch target the project asks for elsewhere. Bump vertical padding to ~12 or set a `minHeight: 44` on both `TouchableOpacity`s.

## Acceptance Criteria

- [ ] `promptFreetextSend` and `promptFreetextCancel` styles have effective tap height ≥ 44pt (either `paddingVertical: 12` or `minHeight: 44`)
- [ ] `promptFreetextInput` has the same min height for visual parity
- [ ] No layout regression in MessageBubble option-row prompts on small screens (iPhone SE 3)

## References

- `packages/app/src/components/chat/MessageBubble.tsx` lines 332-373 (new freetext styles)
- CLAUDE.md → Mobile / RN section

PR #7366, opened 2026-08-25
fix(dashboard): notify natively when a session finishes and is waiting on you

### LINKED ISSUE #7347: desktop: no native notification when a session finishes and awaits the user — only permission requests notify
## Summary

When a session completes its work and is waiting on the user (a merge decision, an answer, any hand-back), **no desktop/tray notification fires**. The user only discovers it by switching back to Chroxy and looking. Reported after a long autonomous run finished and sat waiting, unnoticed.

## Root cause — exactly one native-notification trigger exists, and it isn't this one

`packages/dashboard/src/hooks/usePermissionNotification.ts` is the **only** Web-Notification call site in the dashboard (`grep` for `new Notification` / `sendNotification` across `packages/dashboard/src` returns this file alone, line 47). Its own header states the scope: *'fire native notifications for permission requests when the browser/Tauri window is not focused.'*

There is no equivalent for turn-complete / session-idle / awaiting-input. So:

| event | desktop native notification | push (mobile / Discord) |
|---|---|---|
| permission requested | ✅ `usePermissionNotification` | ✅ `ws-permissions.js:382` |
| **session finished, awaiting user** | ❌ **nothing** | ✅ `activity_update` 'Session idle' |

The server-side idle notification **does** exist — `server-cli/push-notification-handler.js:175` sends `activity_update` / 'Session idle' — but it goes to the push sinks (Expo, Discord), never to the desktop's native layer. The Tauri app has `tauri_plugin_notification` initialised (`src-tauri/src/lib.rs:890`) and uses `send_notification` for server lifecycle events, so the capability is present and already wired for other purposes.

`~/.chroxy/desktop-settings.json` has `showNotifications: true`, so this is not a user-configuration problem.

## Why this is the more important of the two

A pending permission at least leaves the session **visibly blocked** — the ⚠ pending badge, the Allow/Deny card. A completed turn awaiting a decision is visually **indistinguishable from an idle session** (see #7340), so it is precisely the state that needs an out-of-app signal. Getting notified for the loud case and not the silent one is backwards.

## Also check: the active-viewer suppression

The server-side idle push is gated (`push-notification-handler.js:152`) on an allow check whose discipline the code describes as *'no-tokens / active-viewers / already-notified'*. Suppressing a phone buzz while the user is demonstrably looking at the session is right for **push**; it is wrong for a **desktop** notification, where 'a client is connected' does not mean the user is looking — in the reported case the Chroxy window was connected but the user was in another application. A desktop notification should key on **window focus** (as `usePermissionNotification` already does) rather than on client connectivity.

## Expected

When a session transitions to awaiting-user while the Chroxy window is unfocused, raise a native notification naming the session and why it stopped ('finished — awaiting your decision', 'question pending'), and click-to-focus that session.

## Verification

Prove it red: assert that a session going idle-awaiting-user with the window unfocused raises a notification. **Confirm this fails on current main.**

Positive controls, both required:
- Window **focused** ⇒ no notification (or it becomes noise on every turn, and users disable notifications wholesale — which loses the permission ones too).
- A session going idle with **nothing pending** ⇒ no notification, so 'awaiting user' stays meaningful.

Found while dogfooding v0.11.0. Related: #7340 (awaiting/idle indistinguishable in-app), #7344 (external-wait signalling).

blamechris:
Closes #7347

## The gap

The dashboard had exactly **one** native-notification trigger — `usePermissionNotification`, for permission requests. A completed turn had none.

That is backwards, and #7347 says why: a pending permission leaves the session *visibly* blocked (the pending badge, the Allow/Deny card), while a finished turn awaiting a decision is indistinguishable in-app from a session nobody ever asked to do anything (#7340). The silent case is precisely the one that needs an out-of-app signal. A long autonomous run finishing and sitting unnoticed is what filed the issue.

(The trigger only became reachable at all with #7351, which added the `Notification.requestPermission()` call chroxy had never made. Added before that, it would have been just as dead as the permission one.)

## The edge, not the state

`useTurnCompleteNotification` fires on the busy → **not**-busy transition. A session that is merely idle is not awaiting anybody; keying on the state would notify for every session on the list the moment the window blurred.

`busy` is read from the per-session server-authoritative `isIdle` in `sessionStates`, which `session_activity` maintains for **every** session (broadcast on `stream_start` and on `result`) — so a background session's completion is announced too, not just the active tab's. Both nearby signals are wrong here and both are easy to reach for:

| signal | why not |
|---|---|
| `sessions[].isBusy` | only a `session_list` snapshot; not rebroadcast on a turn boundary |
| top-level flat `isIdle` | mirrors the **active** session alone |

A source-level guard in `notificationPermissionWiring.test.ts` pins the correct one.

## Gates

Every one of these was verified by **deleting it and confirming a named test goes red** — see below.

- **Window focus**, the same rule `usePermissionNotification` uses. Notifying while the user is looking is how this becomes noise on every turn, and the way users answer noise is by revoking OS notifications wholesale — which takes the permission notifications with it.
- **A session that stopped ON a permission prompt** — already covered by `usePermissionNotification`. One moment, one card.
- **A per-device mute** (Settings → Dashboard), separate from the intervention ping and from the OS permission, for the same reason as above.
- **A dropped socket** — `handleDisconnected` clears transient state and a reconnect re-seeds `isIdle` from the `session_list` snapshot. Neither is a completed turn, and the turn may still be running on the server.

The gates suppress **dispatch only**. Busy bookkeeping runs on every pass, so a mute toggled or a window focused mid-turn cannot leave stale state that fires a phantom notification later — two tests cover exactly that.

## Click-to-focus — and the half that is not here

Wired on the **web** backend via `Notification.onclick` → `window.focus()` + switch to that session.

**Not wired on Tauri.** The plugin's `sendNotification()` returns `void` and exposes no click callback on desktop; `onAction` is built around `registerActionTypes`, an Android/iOS concept. Doing this properly on desktop means a Rust-side notification that can call `window::show_window` — a separate, platform-specific change, filed as a follow-on.

Deliberately not papered over with a fallback. A test pins the gap (`sends but does NOT invoke onClick on the Tauri backend`) so it is discovered by reading the suite rather than by clicking a notification and getting nothing.

## Also here — both in the same feature area

- **`KEY_NOTIFICATION_PERMISSION_ASKED` was never added to `isGlobalKey`** (#7351). An unscoped `clearPersistedState()` — the server-switch path — deleted it, re-arming the automatic OS-permission prompt for a user who had already dismissed one. That is the exact re-prompt loop the flag exists to stop, just hidden behind an action no unit test performed. Regression test included, and proven red without the fix.
- **`.settings-field-checkbox label` was ~20px tall** against the repo's 44pt floor (#3751 family). The label *is* the control here — clicking the text toggles the checkbox — so it is the box that has to clear 44px, not the native glyph. Fixed on the shared class so every toggle in the section clears it and a fourth inherits a compliant target.

## Testing

`297 files / 5169 tests` green, stable across three consecutive full runs (this package has a known leaked-hook flake class, so repeats were the point). `tsc --noEmit` clean; `vite build` clean.

**13 mutants, each killed by its named test.** Restores were byte copies, never `git checkout --`; the harness asserted the replacement landed (`count == 1` + bytes changed) before trusting a red, and asserted the *specific* named test failed rather than just "something failed".

Behavioural guards (`useTurnCompleteNotification.ts`):

| mutant | killed by |
|---|---|
| drop `if (document.hasFocus()) return` | does not notify while the window is focused |
| drop `if (!enabled) return` | does not notify while muted |
| drop `if (session.awaitingPermission) continue` | does not double-notify a session that stopped on a permission prompt |
| drop the `!connected` reset | does not notify across a reconnect that re-seeds the session as idle |
| edge → state (`observedBusy &&` removed) | does not notify for a session that was never busy |
| drop the stale-session prune | forgets a session that disappears mid-turn |
| capture `onNotificationClick` instead of the ref | uses the latest click handler |

Wiring guards (`App.tsx`, CSS, `SettingsPanel.tsx`):

| mutant | killed by |
|---|---|
| delete the hook call from App | imports and mounts useTurnCompleteNotification |
| hardcode `connected: true` | feeds it the live connection state |
| read `sessions[].isBusy` instead of `isIdle` | derives busy from the per-session isIdle flag |
| drop the `awaitingPermission` input | suppresses the alert for a session that stopped on a permission prompt |
| drop the 44px floor | gives .settings-field-checkbox label a 44px minimum |
| drop the checkbox class from the toggle row | renders the turn-complete toggle inside a .settings-field-checkbox row |

Plus the persistence regression: removing the two `isGlobalKey` lines turns both survive-a-server-switch tests red, and each carries a positive control asserting genuinely-scoped state *was* cleared, so neither can pass by `clearPersistedState()` quietly doing nothing.

Every negative assertion in the hook suite is paired with a positive control that fires under the same harness one field apart. Nothing here is async — the effect body runs synchronously inside `render`/`rerender`, so there is no `waitFor` that could pass before the guarded thing had a chance to happen.

## Not verified

The **Tauri plugin path is still unverified against a running desktop build** — inherited from #7351, unchanged by this PR. It is tested against a mocked `window.__TAURI__.notification`. If the global is not injected, the code falls back to the Web API and reports `unsupported` in Settings, so it surfaces rather than hides. Worth a dogfood check.

### DIFF
--- a/packages/dashboard/src/App.tsx
+++ b/packages/dashboard/src/App.tsx
@@ -76,6 +76,7 @@ import { startServer, revealInFinder } from './hooks/useTauriIPC'
 import { usePermissionNotification, type PermissionPromptInfo } from './hooks/usePermissionNotification'
 import { useNotificationPermission } from './hooks/useNotificationPermission'
 import { useInterventionPing } from './hooks/useInterventionPing'
+import { useTurnCompleteNotification, type TurnCompleteSession } from './hooks/useTurnCompleteNotification'
 import { useShortcutDispatch } from './hooks/useShortcutDispatch'
 import { FileOpenPalette } from './components/FileOpenPalette'
 import { SymbolSearchPalette } from './components/SymbolSearchPalette'
@@ -90,7 +91,7 @@ import { useMessageRenderer } from './hooks/useMessageRenderer'
 import { SplitPane } from './components/SplitPane'
 import { ViewSwitcher } from './components/ViewSwitcher'
 import { DEFAULT_PROVIDER, USER_SHELL_PROVIDER } from '@chroxy/protocol'
-import { persistSidebarWidth, loadPersistedSidebarWidth, persistSplitMode, persistShowConsoleTab, loadPersistedShowConsoleTab, persistInterventionPing, loadPersistedInterventionPing, persistCompactChatFilter, loadPersistedCompactChatFilter, loadPersistedSidebarPanelHeight, loadPersistedSidebarPanelView, loadPersistedSidebarPanelCollapsed } from './store/persistence'
+import { persistSidebarWidth, loadPersistedSidebarWidth, persistSplitMode, persistShowConsoleTab, loadPersistedShowConsoleTab, persistInterventionPing, loadPersistedInterventionPing, persistTurnCompleteNotification, loadPersistedTurnCompleteNotification, persistCompactChatFilter, loadPersistedCompactChatFilter, loadPersistedSidebarPanelHeight, loadPersistedSidebarPanelView, loadPersistedSidebarPanelCollapsed } from './store/persistence'
 import { applyOrderById } from './utils/reorderById'
 import { DiffViewerPanel } from './components/DiffViewerPanel'
 import { AgentMonitorPanel } from './components/AgentMonitorPanel'
@@ -480,6 +481,13 @@ export function App() {
   // dedupes by requestId and throttles bursts; `enabled` honors the operator
   // mute toggle (Settings → Dashboard, persisted per-device).
   useInterventionPing(permissionPrompts, { enabled: interventionPingEnabled })
+  // #7347 — turn-complete OS-notification enable/mute. Defaults on; persisted
+  // per-device. Separate from the intervention ping and from the OS permission
+  // itself so muting "the session finished" does not also mute "the session
+  // needs permission" — the hook is wired further down, next to `isConnected`.
+  const [turnCompleteNotificationEnabled, setTurnCompleteNotificationEnabled] = useState(() => {
+    return loadPersistedTurnCompleteNotification()
+  })
 
   // #3698 — derive the user-message history for the InputBar's terminal-
   // style Up/Down navigation. Oldest-first (matches the natural arrival
@@ -2097,6 +2105,42 @@ export function App() {
   // available). Extracted to `useTunnelReady` (#5560).
   const tunnelReady = useTunnelReady(isConnected)
 
+  // #7347 — native notification when a session finishes its turn and is left
+  // waiting on the user. Sits here rather than beside the other notification
+  // hooks (~L470) because it needs `isConnected`, declared just above; the
+  // call is unconditional, so hook order stays stable.
+  //
+  // `busy` comes from the per-session server-authoritative `isIdle` (#4639) in
+  // `sessionStates`, which `session_activity` maintains for EVERY session
+  // (broadcast on stream_start and on result). Two nearby signals are wrong
+  // here and both are easy to reach for: `sessions[].isBusy` is only a
+  // `session_list` snapshot and is not rebroadcast on a turn boundary, and the
+  // top-level flat `isIdle` mirrors the ACTIVE session only. Shallow-equal so
+  // the record keeps its identity — and the hook's effect stays put — unless a
+  // session's busy state actually flips.
+  const sessionBusyById = useConnectionStore(
+    useShallow((s) => {
+      const out: Record<string, boolean> = {}
+      for (const id in s.sessionStates) out[id] = s.sessionStates[id]!.isIdle === false
+      return out
+    }),
+  )
+  const turnCompleteSessions = useMemo<TurnCompleteSession[]>(
+    () =>
+      sessions.map(s => ({
+        sessionId: s.sessionId,
+        name: s.name,
+        busy: sessionBusyById[s.sessionId] ?? false,
+        awaitingPermission: (pendingPermissionCounts[s.sessionId] ?? 0) > 0,
+      })),
+    [sessions, sessionBusyById, pendingPermissionCounts],
+  )
+  useTurnCompleteNotification(turnCompleteSessions, {
+    enabled: turnCompleteNotificationEnabled,
+    connected: isConnected,
+    onNotificationClick: handleSwitchSession,
+  })
+
   // Fetch conversation history when welcome screen is shown
   useEffect(() => {
     if (showWelcome) fetchConversationHistory()
@@ -2403,6 +2447,11 @@ export function App() {
               onToggleInterventionPing={(enabled) => {
                 setInterventionPingEnabled(enabled)
                 persistInterventionPing(enabled)
+              }}
+              turnCompleteNotificationEnabled={turnCompleteNotificationEnabled}
+              onToggleTurnCompleteNotification={(enabled) => {
+                setTurnCompleteNotificationEnabled(enabled)
+                persistTurnCompleteNotification(enabled)
               }}
               notificationPermission={notificationPermission}
             />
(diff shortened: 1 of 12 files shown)


blamechris:
Follow-on filed: #7367 — desktop (Tauri) click-to-focus, which this PR could not deliver because the notification plugin exposes no click callback on desktop. The gap is pinned by `sends but does NOT invoke onClick on the Tauri backend` in `native-notifications.test.ts`; that test is the thing to update when #7367 lands.

blamechris:
## Review record — inline pass

Reviewed at MEDIUM intensity (feature code, test-covered, no security or serialization surface). Everything below was found and fixed on this branch; each fix is its own commit.

### Findings acted on

**1. `useTurnCompleteNotification` was unproven under `StrictMode`.** `src/main.tsx` renders the app inside `<StrictMode>`, which double-invokes effects on mount. This hook tracks the busy → idle edge in a `useRef` and returns no cleanup, so the second invocation observes the bookkeeping the first one did. That is precisely the shape StrictMode catches out, and it fails *silently* in either direction — a duplicate alert, or an edge swallowed at mount. Reasoning said it was safe; reasoning is not the standard here.

Three cases added (`23700434d`): no fire on the double-invoked mount of an already-busy session, none for an already-idle one, and still exactly one fire for a turn that completes afterwards.

**2. A spy could leak across cases.** The new `onClick` tests stub `window.focus`, one of them with a thrower. Each restores at the end of its own body — but an assertion failing *before* that line would have leaked a throwing focus stub into every subsequent test in the file, turning one red into several and burying the real failure. `vi.restoreAllMocks()` now runs in `afterEach` (`23700434d`).

**3. The wiring guards were loose enough to pass with the wiring gone.** `expect(appSource).toMatch(/connected:\s*isConnected/)` scans the *whole* of `App.tsx`. It is unique today, but it is satisfiable by any unrelated component that later takes an identically-named prop — at which point the guard keeps passing with this hook's own wiring deleted. That is exactly the defect class the file exists to catch, reintroduced inside the guard against it.

Now sliced by anchor (`355c1841b`): the `turnCompleteSessions` memo and the `useTurnCompleteNotification(...)` call are extracted and asserted *within*, with a positive control that both slices are non-empty so a stale anchor fails loudly instead of making every assertion vacuous. Two previously-unpinned inputs gained guards while I was there — the mute toggle and the click handler — both proven by hardcoding them.

### Considered and deliberately not changed

- **A prompt that arrives while the window is focused, followed by the user walking away, produces no alert.** `usePermissionNotification` skipped it (correctly — the user saw it), and this hook suppresses on `awaitingPermission`. But the session stays *busy* while a permission is pending (`session_activity` flips to idle on `result`, not on `permission_request`), so there is no idle edge to fire on either way. Not a regression, and out of scope for a turn-completion trigger.
- **Two dashboards open (desktop app + browser tab) both notify.** Pre-existing for permission notifications; not introduced here.
- **`connectionPhase` leaving `connected` for `server_restarting` discards tracking for in-flight turns.** Deliberate and documented — the conservative direction is to lose an alert, not to manufacture one out of a re-seeded snapshot.
- **An empty `sessions[].name` would render `Chroxy: `.** Already a display issue in the tab strip if it can happen; adding an untested fallback here would not fix it where it shows.

### Verification after the review pass

- Mutation harness re-run: **7 behavioural + 8 wiring mutants, all killed by their named tests.** Each asserts the replacement landed (one match, bytes changed) before trusting a red, and that the *named* test was among the failures — not merely that something failed.
- `297 files / 5175 tests` green. `tsc --noEmit` clean.

### Blocked on

Copilot has now errored on **three** attempts (`f82536fe7`, `23700434d`, and re-requested on `355c1841b`) with *"Copilot encountered an error and was unable to review this pull request"* — an explicit error, not a pending or quota-silent review. Merge stays blocked on a real Copilot review landing.

blamechris:
**Gate update at `355c1841b` — re-derived, not carried over from the earlier reading.**

| input | state |
|---|---|
| CI | 21 pass / 1 skipping, all green on this head |
| unresolved review threads | 0 |
| `mergeable` / `mergeStateStatus` | `MERGEABLE` / `CLEAN` |

So the branch-protection gate is satisfied and my earlier "merge stays blocked" line no longer holds — the ruleset accepted it.

**One caveat worth naming rather than letting the green speak for it:** Copilot never produced a substantive review. Three attempts, two of which recorded *"Copilot encountered an error and was unable to review this pull request"* and a third re-request that never landed one. The ruleset counts the errored reviews, so the gate reads clean, but in practice this PR has had **no Copilot input at all** — only the inline pass above. That is a thinner review than the usual pipeline, and it is the reason to look at the diff yourself before merging rather than trusting the checkmarks.

Not merging — interactive session.

Issue #554, opened 2026-02-17
Split chat message on permission prompt boundary


blamechris:
## Context

Observed during manual testing of permission handling.

## Description

When Claude's response includes a tool use that requires permission, the entire turn (text before prompt + text after approval/denial) appears as a single message bubble. After the user approves or denies, Claude's subsequent text "streams" into the existing bubble above the permission card.

**Expected behavior:** The text after a permission prompt response should start a new message bubble, visually separating the pre-prompt explanation from the post-prompt continuation.

**Current flow:**
1. Claude says "Let me try..." (bubble starts)
2. Permission card appears
3. User approves
4. Claude says "Done, the file was..." → appended to the same bubble from step 1

**Desired flow:**
1. Claude says "Let me try..." (bubble 1)
2. Permission card appears
3. User approves
4. Claude says "Done, the file was..." (new bubble 2)

## Acceptance Criteria

- [ ] Text streamed after a permission response starts a new message bubble
- [ ] The permission card visually sits between the two message bubbles
- [ ] Works for both approve and deny flows

Issue #7353, opened 2026-08-24
dashboard: dismissing an expired permission erases all trace of it — the only record of a dropped tool call is one click from gone


blamechris:
## Summary

When a permission request expires, the transcript shows a card — *"Timed out — Permission expired — Claude will continue without this tool"* — with a **Dismiss** button. Clicking Dismiss removes the card entirely, leaving **no trace that a permission was ever requested, or that a tool call was dropped**.

The expired-permission card is currently the *only* place a dropped tool call is visible. Dismissing it destroys the sole record of a capability the agent asked for and did not get.

## Why this is worse than normal dismiss-a-toast behaviour

A dismissable notice is fine when the underlying event is recorded elsewhere. Here it is not:

- The `PermissionAuditLog` is an explicitly **in-memory ring buffer** (`permission-audit.js`), so it does not survive a daemon restart — and it records no delivery information at all (#7352).
- There is no session-level marker that a tool call was dropped (#7340 covers the adjacent case of background agents being force-completed at turn end).
- No notification ever fired to leave a trail elsewhere, because chroxy's only native-notification path is unreachable (#7351).

So after one click the answer to *"did the agent silently lose a tool call during this run?"* becomes unrecoverable from the UI, and unreliable from the server.

## Observed

A single session accumulated **three** expired permissions in about ten minutes:

- `Bash: Fix the deadline race properly`
- `Bash: Fix the symlink test and re-check mutation M`
- `Bash: Commit the restructured fix`

Each means the agent proceeded with less capability than it requested. The third is the clearest illustration of the stakes: a dropped **commit** step leaves work uncommitted while the transcript reads as though the agent moved on. (In this instance the agent recovered and the commit landed on a later attempt — but that was verifiable only by inspecting the worktree from outside chroxy, which is precisely the problem.)

Note the session was in **Auto (skip all prompts)** mode throughout, so none of these prompts should have existed — see #7337.

## Expected

Dismiss should clear the *interactive affordance*, not the *history*. After dismissing, the transcript should still show — in a collapsed or muted form — that a permission was requested for tool X at time T and expired unanswered.

## Suggested fix

1. **Dismiss collapses, never deletes.** Replace the card with a one-line muted marker: *"Permission for Bash expired — tool call dropped."* Keep it in the transcript permanently.
2. **Surface a per-turn summary** when any tool call was dropped, so the count is visible without scrolling the whole transcript.
[...]

PR #7375, opened 2026-08-25
fix(dashboard,server): warn before Auto mode kills a paused turn, and retire the prompt it strands

### LINKED ISSUE #7335: permissions: switching to Auto while a permission prompt is open silently kills the CLI turn — the destructive-warning guard misses the paused-on-prompt case
## Summary

Switching the permission mode to **Auto (skip all prompts)** while a permission prompt is open kills the in-flight turn on `claude-cli` **without showing the destructive-consequence warning that exists for exactly this situation**. The user is left with a stale permission card, an 'already acknowledged'-style message when they click Accept, and a tool call rendered red — with no indication that the turn is dead.

Reported from dogfooding, verbatim: *'it will show up as red... when I click accept after, it will tell me it's already been acknowledged, but it's not clear what was. And you can see this bash command says red. So it's unclear if the whole process has stopped or not.'*

**It has stopped.** Nothing in the UI says so.

## Root cause — the warning is gated on the wrong predicate

`packages/dashboard/src/store/connection.ts:4139`

```ts
const isStreaming = !!get().getActiveSessionState().streamingMessageId;
```

That value feeds `buildAutoModeConfirmMessage({ interruptsTurn, isStreaming })`
(`lib/auto-mode-confirm.ts:51`), which shows the destructive copy **only** when
`interruptsTurn && isStreaming`:

```ts
if (input.interruptsTurn && input.isStreaming) return DESTRUCTIVE_COPY
return BASE_COPY   // "Tools will run without asking for permission."
```

A session **paused on a permission prompt is not streaming** — it is waiting on the user — so `isStreaming` is false, the benign copy is shown, and the user confirms what reads as a harmless preference change. Meanwhile `CliSession` has `interruptsTurnOnAutoSwitch: true` (`cli-session.js:256`): `_onPermissionModeChanged` respawns the `claude -p` subprocess and **drops the in-flight turn**.

**The repo already has the correct predicate**, 670 lines earlier, and even explains why the two-part form is required — `connection.ts:3467`:

```ts
const busy = active.streamingMessageId !== null || active.isIdle === false;
```

with the comment: *'the busy signal must match EXACTLY what the InputBar shows as busy (`isStreaming || isBusy`) … `isIdle` is the server-authoritative working flag (#4639); `streamingMessageId` additionally covers the optimistic pre-status window.'*

The auto-mode confirm uses only the first half of that disjunction. So the guard is correct for every state it was tested in and blind to the one where users most want to flip to Auto — **being prompted is the reason you reach for 'skip all prompts'.** That is the whole point of the #3729 panic-button this dialog is guarding.

## Secondary failures (all downstream of the same event)

1. **Stale permission card.** The pending prompt stays on screen after the mode switch resolves/abandons it.
2. **Uninformative rejection.** Clicking Accept on the stale card reports that it was already handled, without saying *what* was decided or *why* the card is dead. Cf. `input-handlers.js:904`, which logs the analogous 'stale/unknown toolUseId … already resolved' case server-side but surfaces nothing comparable to the user.
3. **Unexplained red tool call.** The dropped turn leaves its `Bash` call in an error state with no message distinguishing 'this command failed' from 'the turn was terminated underneath it'.

## Expected

- The destructive warning fires whenever the provider interrupts on auto-switch **and the session is busy in any sense** — streaming *or* not idle *or* holding a pending permission.
- On confirm, the pending permission card is removed or visibly marked resolved-by-mode-switch.
- The killed turn's tool calls carry an explicit 'turn interrupted by permission-mode switch' state, not a bare red.

## Suggested fix

Replace line 4139 with the same disjunction used at 3467, and extend it to cover a pending permission:

```ts
const st = get().getActiveSessionState();
const isBusy = st.streamingMessageId !== null || st.isIdle === false || hasPendingPermission(st);
```

Better: extract the busy predicate into **one** exported helper and have both call sites use it. Two hand-written copies of 'is this session busy' that disagree is the defect, not the specific missing clause — a third caller will otherwise repeat it.

## Verification — prove it red first

Per `docs/false-safety-guards.md`: a test that flips to Auto while streaming passes today and proves nothing.

- **Negative control:** session paused on a pending permission (`streamingMessageId === null`, `isIdle === false`) + provider with `interruptsTurnOnAutoSwitch: true` ⇒ assert `buildAutoModeConfirmMessage` returns the DESTRUCTIVE copy. **Confirm this fails on current main before fixing.**
- **Positive control:** genuinely idle session (no pending permission, `isIdle === true`) ⇒ still the benign copy; the warning must not become unconditional.
- **Provider control:** same paused state on `claude-sdk`/`claude-tui` (`interruptsTurnOnAutoSwitch: false`) ⇒ benign copy, since those apply the switch in place.

Note a fix that always returns the destructive copy would satisfy the first assertion while destroying the dialog's usefulness — the positive control is what keeps it honest.

Found while dogfooding v0.11.0 on `claude-cli`. Related: #5609 (which introduced this dialog), #3729 (the panic-button).

blamechris:
Closes #7335.

Switching to **Auto** while a permission prompt was open killed the in-flight turn on `claude-cli` behind benign confirm copy, then left the dead prompt on screen with a live Allow button. Reported from dogfooding v0.11.0.

Two independent gaps, same shape — a predicate that is correct for every state it was tested in and blind to the one that matters.

## 1. The confirm copy (dashboard)

`setPermissionMode` computed "is a turn in flight" from `streamingMessageId` alone. The #554 stream-split **clears** that field the moment a `permission_request` arrives, while the server-authoritative `isIdle` stays `false` because the turn really is still running (the CLI child is blocked on its PreToolUse hook). So a session *paused on a prompt* read as idle and got the benign copy — in the one state where users most reach for "skip all prompts", since being prompted is the whole reason you press the #3729 panic-button.

The correct two-part disjunction already existed 670 lines earlier in the same file (#5952, `sendInput`), with a comment explaining why both halves are required. Two hand-written copies of one safety predicate that can drift **is** the defect, so both call sites now use one exported helper in `lib/session-busy`:

| helper | clauses | used by |
|---|---|---|
| `isSessionBusy` | `streamingMessageId \|\| !isIdle` | `sendInput` |
| `hasInterruptibleWork` | the above **+** a live permission prompt | the Auto confirm |

`sendInput` keeps the narrow one deliberately: widening it would optimistically badge a send "Queued" in a state where the server would not queue it. A test pins that the two must continue to differ, so a future "helpful" merge of the two fails.

`AutoModeConfirmInput.isStreaming` is renamed `isBusy`. The name was the bug — the helper was always correct, the call site fed it a lying value.

## 2. The stranded prompt (server)

`_killAndRespawn` kills the child, so every PreToolUse hook it blocked on dies and the daemon tears the pending entry down via `req 'aborted'` **without broadcasting anything**; `_emitInterruptedTurnResult` → `_clearMessageState` then wipes `_pendingPermissionIds` silently. The prompt was dead on both sides and the client was never told — hence the card sitting live until its own 5-minute expiry, and the "already acknowledged" response on Allow.

`_handleHardTimeout` has emitted `permission_expired` for exactly this reason since #2831. It was simply never wired to the **other** ways a turn dies. Extracted as `_expirePendingPermissions` and called *before* the state wipe — the ordering is the whole fix.

The callers were then audited rather than assumed, because a turn dies more ways than the respawn:

| path | wired? | why |
|---|---|---|
| `_handleHardTimeout` | yes | the original (#2831) |
| `_killAndRespawn` | yes | the panic-button, the auto-mode switch, mid-turn `setModel` |
| `_handleChildClose` | yes | a crash, **and user Stop** — SIGINT exits the child without going through `_killAndRespawn`, so this needed wiring of its own |
| `_handleStreamStall` | no | **unreachable** with a permission pending, not merely unlikely: `notifyPermissionPending` clears the stall timer and `_armResultTimeout` bails on `_resultTimeoutPaused` |
| interrupt safety timeout | no | fires when the child *ignored* SIGINT and is still alive, so its hook is still blocked and the prompt still genuinely answerable — expiring there would be the lie this method exists to prevent |
| `destroy()` | no | takes the whole session down, card included; `ws-permissions` already releases the HTTP side |

The respawn path sets `_respawning`, so the close it causes short-circuits above the new call — no double emit, pinned by a test rather than left to the reader.

**No client change was needed**: the dashboard's `permission_expired` handler already appends the expiry note and clears the prompt's `options`, which is what retires the dead Allow button.

## Verification — red first

Per `docs/false-safety-guards.md`, confirmed failing before the fix and passing after:

- The guards are **call-site** tests. A unit test of `buildAutoModeConfirmMessage` passes on the broken build — the helper was never where the bug lived, so that test proves nothing. The existing unit suite now says so in its header.
- **Negative controls** (the bug): paused-on-prompt ⇒ destructive copy; respawn with pending permissions ⇒ `permission_expired` per id.
- **Positive controls**, so a fix that fires unconditionally cannot pass: genuinely idle session, *expired* prompt, *answered* prompt, AskUserQuestion prompt (not a permission), respawn with nothing pending, clean child close with nothing pending, already-resolved permission, and no double-emit either when a hard timeout precedes a respawn or when a respawn is followed by the close it caused.
- **Provider control**: paused-on-prompt on a non-interrupting provider (SDK/TUI) ⇒ benign copy.
- Both original defects were re-applied as mutants after the fix and confirmed to go red.

## Gates

- Dashboard: 299 files / 5193 tests pass · `tsc --noEmit` clean · `vite build` clean
- Server: 107 related `cli-session` tests pass · all 9 `scripts/lint-*.sh` pass
- Two failures in `tests/cli-session.test.js` (`omits CHROXY_PORT/CHROXY_HOOK_SECRET when port is not set`) were verified to fail identically on `origin/main` — a local-environment artifact, not from this change.

## Left for a follow-on

The issue's third expected outcome — tool calls of a killed turn carrying an explicit *"turn interrupted by permission-mode switch"* state rather than a bare red — is not in this PR. It needs an interrupt reason threaded from the respawn through `event-normalizer` into the tool-call render state, which is a separate change from the two predicates above. Filed separately.

### DIFF
--- a/packages/dashboard/src/lib/auto-mode-confirm.test.ts
+++ b/packages/dashboard/src/lib/auto-mode-confirm.test.ts
@@ -1,29 +1,37 @@
+/**
+ * Unit tests for the pure copy-selection helper.
+ *
+ * #7335: these pin the helper only. The `isBusy` input is where the defect
+ * actually lived, so the guard that matters is the CALL-SITE suite in
+ * `store/auto-mode-confirm-busy.test.ts` — every case below passed on the
+ * broken build.
+ */
 import { describe, it, expect } from 'vitest'
 import { buildAutoModeConfirmMessage } from './auto-mode-confirm'
 
 describe('buildAutoModeConfirmMessage (#5609)', () => {
-  it('warns about interrupting the turn when the provider interrupts AND a turn is streaming', () => {
-    const msg = buildAutoModeConfirmMessage({ interruptsTurn: true, isStreaming: true })
+  it('warns about interrupting the turn when the provider interrupts AND the session is busy', () => {
+    const msg = buildAutoModeConfirmMessage({ interruptsTurn: true, isBusy: true })
     expect(msg).toMatch(/INTERRUPT/)
     expect(msg).toMatch(/restart the session/)
     // still explains the bypass consequence
     expect(msg).toMatch(/without asking for permission/)
   })
 
-  it('uses the plain copy when the provider interrupts but no turn is in flight', () => {
-    const msg = buildAutoModeConfirmMessage({ interruptsTurn: true, isStreaming: false })
+  it('uses the plain copy when the provider interrupts but there is no work at risk', () => {
+    const msg = buildAutoModeConfirmMessage({ interruptsTurn: true, isBusy: false })
     expect(msg).not.toMatch(/INTERRUPT/)
     expect(msg).toMatch(/Tools will run without asking for permission/)
   })
 
-  it('uses the plain copy for non-interrupting providers (SDK/TUI) even mid-turn', () => {
-    const msg = buildAutoModeConfirmMessage({ interruptsTurn: false, isStreaming: true })
+  it('uses the plain copy for non-interrupting providers (SDK/TUI) even when busy', () => {
+    const msg = buildAutoModeConfirmMessage({ interruptsTurn: false, isBusy: true })
     expect(msg).not.toMatch(/INTERRUPT/)
     expect(msg).toMatch(/Tools will run without asking for permission/)
   })
 
   it('treats an undefined capability flag as non-interrupting', () => {
-    const msg = buildAutoModeConfirmMessage({ interruptsTurn: undefined, isStreaming: true })
+    const msg = buildAutoModeConfirmMessage({ interruptsTurn: undefined, isBusy: true })
     expect(msg).not.toMatch(/INTERRUPT/)
     expect(msg).toMatch(/Tools will run without asking for permission/)
   })
--- a/packages/dashboard/src/lib/auto-mode-confirm.ts
+++ b/packages/dashboard/src/lib/auto-mode-confirm.ts
@@ -25,11 +25,18 @@ export interface AutoModeConfirmInput {
    */
   interruptsTurn?: boolean
   /**
-   * Whether the active session is currently streaming a response (a turn is in
-   * flight). When false, even a CLI switch has no turn to interrupt, so the
-   * destructive warning is omitted.
+   * Whether the active session has in-flight work the switch would destroy.
+   * When false, even a CLI switch has no turn to interrupt, so the destructive
+   * warning is omitted.
+   *
+   * #7335: this was named `isStreaming`, and the name was the bug — the call
+   * site dutifully fed it `!!streamingMessageId`, which the #554 stream-split
+   * clears the moment a permission prompt arrives, so a session PAUSED on a
+   * prompt looked idle and got benign copy while the CLI respawn dropped its
+   * turn. Callers must source this from `hasInterruptibleWork` (lib/session-busy),
+   * never from the streaming flag alone.
    */
-  isStreaming: boolean
+  isBusy: boolean
 }
 
 const BASE_COPY = 'Switch to Auto mode? Tools will run without asking for permission.'
@@ -48,7 +55,7 @@ const DESTRUCTIVE_COPY =
  * the standard non-destructive copy.
  */
 export function buildAutoModeConfirmMessage(input: AutoModeConfirmInput): string {
-  if (input.interruptsTurn && input.isStreaming) {
+  if (input.interruptsTurn && input.isBusy) {
     return DESTRUCTIVE_COPY
   }
   return BASE_COPY
--- a/packages/dashboard/src/lib/session-busy.test.ts
+++ b/packages/dashboard/src/lib/session-busy.test.ts
@@ -20,6 +20,9 @@ const livePrompt = {
   timestamp: NOW,
   requestId: 'req-1',
   expiresAt: NOW + 60_000,
+  // A live permission prompt always carries its options — that is what makes it
+  // answerable, and `hasInterruptibleWork` keys the "still at risk" test on it.
+  options: ['allow', 'deny'],
 } as unknown as ChatMessage
 
 const idle = { streamingMessageId: null, isIdle: true, messages: [] as ChatMessage[] }
@@ -53,9 +56,25 @@ describe('hasInterruptibleWork (#7335)', () => {
     expect(hasInterruptibleWork(idle, NOW)).toBe(false)
   })
 
-  it('false once the prompt has expired', () => {
+  it('false once the prompt has CLOCK-expired', () => {
     const expired = { ...livePrompt, expiresAt: NOW - 1 } as ChatMessage
     expect(hasInterruptibleWork({ ...idle, messages: [expired] }, NOW)).toBe(false)
+  })
+
+  it('false once the SERVER retired the prompt, even with time left on its clock', () => {
+    // `permission_expired` clears `options` but leaves `answered` unset and
+    // `expiresAt` in the future, so isLivePermissionPrompt still calls this
+    // "live". #7335's server half makes this routine — every auto-switch
+    // expires the open prompt — so without the options filter the dialog would
+    // warn about interrupting a session with nothing in flight.
+    const retired = { ...livePrompt, options: undefined } as ChatMessage
+    expect(hasInterruptibleWork({ ...idle, messages: [retired] }, NOW)).toBe(false)
+  })
+
+  it('a retired prompt does not mask a genuinely busy session', () => {
+    // The options filter must not swallow the busy signal itself.
+    const retired = { ...livePrompt, options: undefined } as ChatMessage
+    expect(hasInterruptibleWork({ streamingMessageId: null, isIdle: false, messages: [retired] }, NOW)).toBe(true)
   })
 
   it('false once the prompt is answered', () => {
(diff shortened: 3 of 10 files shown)


blamechris:
## Code Review: PR #7375

Three-reviewer panel (server lifecycle · dashboard predicate · adversarial false-safety). Every finding below was re-verified against the code before action; one was rejected.

### Verdict

**Request Changes → addressed.** The panel found that the PR's headline claim was false and that it introduced one regression. Both are fixed in `a25f1b58`, along with four other findings.

### Findings

#### Critical

| # | Finding | Disposition |
|---|---|---|
| F1 | **`permission_expired` did NOT retire the dashboard's Allow button.** Claimed in the PR body, in `cli-session.js`, and in a test header. `PermissionPrompt` hardcodes its own buttons and **never reads `options`** — #2853 documents this. The real gate is `expiresAt`, which nothing moved. Expired prompts kept a working Allow for 5 minutes with "(Expired…)" beside them. The claim was true of the *mobile app* and was attributed to the dashboard unchecked. | **Fixed** — handler stamps `expiresAt` to now (not 0: a falsy value drops the card and erases the record, #7353). `answered` deliberately untouched — it is a decision token (#6222/#6223). Pre-existing bug: `permission_expired` already fired from the hard-cap and SDK paths on `main`. |
| C1 | **Regression introduced by this PR.** `_pendingPermissionIds` and `_resultTimeoutPaused` are paired; `notifyPermissionResolved` only unsets the flag for an id still *in* the set. Taking the ids without releasing the flag wedged it true on a reachable Stop-then-crash path, silently killing the inactivity warning, the hard cap **and** the #4467 stream-stall recovery for the rest of the session. Pre-PR the session self-healed. | **Fixed** — one line, plus a test asserting the next turn re-arms its timers. |

#### Suggestions

| # | Finding | Disposition |
|---|---|---|
| F3/F4 | **The hand-written audit table was wrong twice.** The interrupt-safety-timeout skip was excused as "the child ignored SIGINT so the prompt is still answerable" — contradicted by this same file (`claude only aborts the current turn`) — and that path wiped the ids anyway. The plain `result` path, the most common turn end, was absent entirely. | **Fixed** — wired the *funnel* (`_clearMessageState`) rather than hand-enumerated sites. Enumerating by hand is what produced both misses. |
| F2 | **A vacuous test.** "no double emit — pinned by a test" could not distinguish the `_respawning` short-circuit from an already-empty set; deleting the guard left all 8 tests green. | **Fixed** — re-arms an id in the gap; deleting the guard now goes red. |
| S1 | **False rationale.** The comment justifying the two-predicate split claimed the server would not queue a send while a permission is pending. It would — `sendMessage` gates purely on `_isBusy`. The claim also contradicted its own sibling comment, and a test pinned the split citing the wrong reason. | **Fixed** — all three now cite the #5952 InputBar match. |
| S2/F6 | A server-retired prompt kept counting as "work at risk" for up to 5 minutes, and this PR's server half makes that state routine — the dialog would warn about interrupting an idle session. | **Fixed** — root-caused by F1; `hasInterruptibleWork` also ignores option-less prompts as a second signal. |

#### Rejected

- **"The mobile app flips to Auto with zero warning."** Not so: the app omits `confirmed`, so the server's `confirm_permission_mode` round-trip fires and the app renders that warning verbatim (`SettingsBar.tsx`). The server's predicate uses `_isBusy`, which *is* true while paused on a prompt — mobile was already correct for this scenario.

### Deferred Items

| Item | Issue | Rationale |
|---|---|---|
| The daemon never releases its own pending entry; killing the CLI child does not signal the hook's `curl` grandchild, so a reconnect within 5 min re-sends a dead prompt with a live Allow | [#7379](https://github.com/blamechris/chroxy/issues/7379) | Pre-existing (not a regression), cross-layer, and in the security-sensitive permission pipeline — wants its own design and review |
| `InputBar`'s busy props are a third inline copy; `!isIdle` inverts against `isIdle === false` when nullish | [#7378](https://github.com/blamechris/chroxy/issues/7378) | Touches five App.tsx call sites; the header comment was corrected to stop over-claiming meanwhile |
| The app lacks the dashboard's already-resolved suppression, newly exposed by the higher `permission_expired` rate | [#7380](https://github.com/blamechris/chroxy/issues/7380) | Pre-existing asymmetry in a different package |

### Verification

Every fix was mutation-tested against the defect it repairs: reverting the `expiresAt` stamp reds 3 tests; removing the `options` filter reds 2; deleting the `_respawning` guard reds 1 (it reded 0 before); reverting the predicate reds 2 with 5 controls green.

### Gates

Dashboard 300 files / 5202 tests · store-core 60 files / 2166 tests · server 166 tests · `tsc` clean · all 9 `scripts/lint-*.sh` · CI 21 pass / 1 skip on `a25f1b58`.

> **Note on the Copilot gate:** Copilot's only review on this PR reads *"Copilot encountered an error and was unable to review this pull request."* The ruleset counts an errored review, so `mergeStateStatus` reads `CLEAN` while nothing automated has actually read this diff. Re-requested; still not delivered. Worth knowing before treating CLEAN as reviewed.

PR #7395, opened 2026-08-27
fix(server): recognise the Agent tool and wire subagent lifecycle events

blamechris:
## What

A session showed **idle** while four reviewers were still running (#7340).

The issue put the root cause in `_clearMessageState`'s turn-end sweep. That sweep *is* a defect — but it was not what produced the symptom. **Nothing was being swept, because nothing was ever tracked.**

Claude Code names the sub-delegation tool **`Agent`**. Chroxy matched only `Task`.

## Evidence

Captured from live `-p --output-format stream-json` runs of both binaries chroxy supports:

| binary | version | tool name emitted |
|---|---|---|
| standalone CLI (`cli-session`) | 2.1.243 | `Agent` |
| Agent SDK bundled engine (`sdk-session`) | 2.1.114 | `Agent` |

```jsonc
{"name":"Agent","input":{"description":"Probe sleep test","subagent_type":"general-purpose","run_in_background":true}}
{"type":"system","subtype":"task_started","tool_use_id":"toolu_01B3…","task_type":"local_agent","is_backgrounded":true}
{"type":"system","subtype":"task_notification","tool_use_id":"toolu_01B3…","status":"completed"}
```

So `agent_spawned` never fired on either path, and every downstream surface was dark: the dashboard's `AgentMonitorPanel` + header count, the mobile `SettingsBar` "N agents running" badge, activity-registry agent nodes, and Control Room's `cancelActivity`.

`is_backgrounded` is **absent** on a foreground spawn, not `false` — so it is compared with a strict `=== true` in both directions.

## Scope — #7340 stays open

The PR originally also exempted backgrounded subagents from the turn-end sweep, which is what #7340 actually asks for. Review found that exemption unsafe on its own, so it has been **reverted** and #7340 stays open with a spec. Three things must land with it:

1. **Six paths reach `_clearMessageState` because the provider is DEAD** — Stop/SIGINT, child crash, `_killAndRespawn`, SDK hard timeout, stream-stall recovery, `interrupt()`. `task_notification` can never arrive on any of them, so an exempted agent is stranded and the session claims to be working for the daemon's lifetime. The issue names stuck-busy as the *worse* failure; the exemption introduced it on six paths.
2. **`_activity.reset()` exempts only `kind === 'shell'`**, so a backgrounded agent's Control Room node dies at turn end anyway and `cancelActivity` answers `not-found` for as long as it runs.
3. **Both clients wipe `activeAgents` on `history_replay_start`** (dashboard `message-handler.ts:5119`, app `:2453`) and nothing backfills it. An earlier revision of this description claimed no client change was needed — that was wrong. #4466 made this exact carve-out for the sibling field `activeTools`.

The sweep is unconditional again, and that is now a **documented decision**: the block in `_clearMessageState` names all three prerequisites, and tests in `base-session.test.js` and `cli-session-events.test.js` pin the contract so it cannot be reintroduced as a one-liner without a test going red.

## What lands here

1. **`Agent` recognised alongside `Task`.** Both stay live; chroxy runs against whatever `claude` the user has installed.
2. **`task_started` / `task_notification` wired on the CLI path.** They previously fell into the generic system-event branch, which forwarded the literal strings to the client **as chat bubbles**. Beyond ending that noise, this gives the CLI a terminal signal it never had — a subagent that finished early in a long turn used to keep its badge until turn end.
3. **`_completeAgent` / `_trackAgent` on `BaseSession` are the single implementation.** `SdkSession`'s and `ByokSession`'s `_finalizeAgentByToolUseId` delegate rather than keeping a third and fourth copy.
4. **The description cap now has one choke point.** `task_started` was a second producer of that wire field with no clamp, and `ServerAgentSpawnedSchema` declares `description: z.string().optional()` with no max — so nothing downstream would have caught it. `AGENT_DESCRIPTION_MAX` is a named export rather than a third bare `200`.
5. **Two more sites of the same rename.** `NEVER_AUTO_ALLOW` and the orchestration gate's `ALWAYS_DENY` named only `Task`. **Neither omission was exploitable** — a rule must also clear `ELIGIBLE_TOOLS`, and `_decide` ends in `return 'deny'` — but a denylist naming only the retired spelling claims more than its code does. The gate's own test also had `['deny','deny','deny']` hardcoded beside the list it iterates, so *dropping* a tool would have kept it green; now derived from the one list. The parse-failure log branch in `cli-session.js` was the fourth roster line.

## Verification

Per `docs/false-safety-guards.md`, proven red before green:

- New tests were run against unmodified `src/` and fail there. The CLI lifecycle suite replays the **verbatim** event shapes captured from the probe runs. (One — "does not track a backgrounded Bash shell as a subagent" — passes on main and is a guard against removing the `task_type` gate *within* the fix, not a red-before-green test.)
- **Coverage the review found missing**, each verified red under the matching mutation: `removeAllListeners()`'s agent clear (deleting the line left the suite green) and `sdk-session.js`'s `task_started` branch (deleting the `local_agent` gate left it green). The SDK one is source-anchored — the dispatch lives inside the `_callQuery` async-iterator loop — sliced to the branch and asserted *within* the slice, with negative controls proving the slice is narrow, since a file-wide `assert.match` is satisfiable by any unrelated line.
- Full server suite diffed against a baseline run on unmodified main: **no regressions**.
- All 9 custom server lints pass; eslint 0 errors; protocol and store-core suites green.

## Deferred to #7340

When an SDK tool_use block has no `id`, the agent is tracked under a synthesized `${messageId}-tool` while `task_started` carries the real one, so both could register. Not fixed speculatively: the obvious fix (make `task_started` the sole tracker) is wrong, because `cancelActivity` already feature-detects SDK builds that emit no task lifecycle messages at all.

Relates to #7340. Follow-ons filed: #7396 (claude-tui has no agent tracking at all), #7397 (two more unconsumed stream events).


blamechris:
## Review round 2 — scope split

Deep review returned **REQUEST CHANGES** on the turn-end sweep exemption, with three critical findings. I verified all three independently against this branch and they hold:

| # | Finding | Verified |
|---|---|---|
| C1 | Both clients wipe `activeAgents` on `history_replay_start` (`message-handler.ts:5119` / `:2453`), and nothing backfills it | ✅ — the `#4466` carve-out for the sibling field `activeTools` is three lines below |
| C2 | Six paths reach `_clearMessageState` *because the provider is dead*; `task_notification` can never arrive, so an exempted agent is stranded forever | ✅ — nothing on the CLI child-death path completes agents; the only `_completeAgent` call needs a live child |
| C3 | `_activity.reset()` exempts only `kind === 'shell'`, so the Control Room node dies at turn end and `cancelActivity` answers `not-found` | ✅ — `activity-registry.js` `reset()` |

C1 disproves this PR's original claim that no client change was needed. C2 and C3 trade the reported bug for the one #7340 explicitly calls worse.

**Resolution: split.** The sweep exemption is reverted; the tracking fix lands. #7340 stays open and now carries the full spec as a comment. Everything else in the PR is unchanged and independently verified.

Also addressed from the review:

- **N8** — the sweep keys off the map key, not `agent.toolUseId`. The old code ended in an unconditional `clear()`, so a drifted record was still removed; deleting by the field would leak it and complete the wrong id. Test added.
- **N10** — the parse-failure log branch in `cli-session.js` accepted only `Task`. Fourth roster line, fixed with the other three; I re-audited every remaining `'Task'` site in `packages/server/src` (only `byok-session.js:1755` remains, correctly — that is byok's *own* tool, which byok itself names `Task`).
- **S6** — `removeAllListeners()`'s agent clear and `sdk-session.js`'s `task_started` branch had no coverage; deleting either left the suite green. Both now tested, each verified red under the matching mutation.
- **S7** — the "all 15 fail against unmodified `src/`" claim was too broad. Correct: "does not track a backgrounded Bash shell as a subagent" passes on main and is a guard against removing the `task_type` gate *within* the fix. PR body corrected.
- **S4** — the synthesized-`${messageId}-tool` duplicate is real but **deliberately not fixed here**. The obvious fix (make `task_started` the sole tracker) is wrong: `cancelActivity` already feature-detects SDK builds that emit no task lifecycle messages at all. Deferred to #7340 with that reasoning.
- **S5** — `_trackAgent`'s OR-upgrade is the safe direction while the flag is inert, and wrong once it is consumed. Documented on the method and on #7340.

Follow-ons filed: #7396, #7397.

PR #7409, opened 2026-08-27
fix(app,dashboard): key the isPlanPending replay wipe to the replayed session

### LINKED ISSUE #7402: dashboard/app: `isPlanPending` replay wipe targets the active session, not the replayed one
## Context

Found during review of PR #7398, which fixed the sibling bug for `activeAgents` in the same code block.

## Description

On `history_replay_start`, both clients clear transient state via `updateActiveSession`, which resolves `state.activeSessionId` and **ignores the frame's `sessionId`**:

- `packages/dashboard/src/store/message-handler.ts`
- `packages/app/src/store/message-handler.ts`

`replayTargetId` is already computed two lines above from `sharedHistoryReplayStart`.

This matters because `subscribe_sessions` replays every newly-subscribed **background** session (`handlers/session-handlers.js`), and both clients fire it for all non-active sessions on every `session_list` — i.e. on every reconnect. So a background session's replay clears the **active** session's `isPlanPending` / `planAllowedPrompts`, cancelling a pending plan approval the user is looking at.

PR #7398 moved the `activeAgents` wipe onto `replayTargetId` (mirroring the #4909 block just below, which made the same move for `stoppedAt`/`stoppedCode` and documented why). `isPlanPending` was deliberately left alone in that PR to keep its blast radius to the bug it was fixing.

## Acceptance criteria

- [ ] `isPlanPending` / `planAllowedPrompts` are cleared for `replayTargetId`, not `activeSessionId`, on both clients
- [ ] A test that a background session's replay does not cancel the active session's pending plan
- [ ] A test that the single-session case (frame omits `sessionId`) is unchanged — `replayTargetId` falls back to `activeSessionId`

blamechris:
## Problem

On `history_replay_start`, both clients cleared `isPlanPending` / `planAllowedPrompts` via `updateActiveSession`, which resolves `state.activeSessionId` and **ignores the frame's `sessionId`** — even though `replayTargetId` is already computed two lines above.

`subscribe_sessions` replays every newly-subscribed **background** session (`handlers/session-handlers.js`), and both clients fire it for all non-active sessions on every `session_list` — i.e. on every reconnect. So a background session's replay cancelled the **active** session's pending plan approval, on the very reconnect that was supposed to restore it.

Same mis-keying #7340 fixed for `activeAgents` and #4909 fixed for `stoppedAt`/`stoppedCode`, in the same code block. #7340 deliberately left `isPlanPending` alone to keep its blast radius to the bug it was fixing; this is that follow-up.

## Fix

Fold the wipe into the existing `replayTargetId` block in both clients, so one `updateSession` call clears both transient slices for the session the frame is actually replaying. `replayTargetId` falls back to `activeSessionId` when the frame omits `sessionId`, so the single-session case is unchanged. The merged updater guards both slices (`ss.activeAgents?.length ?? 0`, `ss.isPlanPending`) because the replayed session — unlike the active one — can be a partially-built state.

The dashboard's #4466 note (why `activeTools` is deliberately *not* wiped) is preserved and moved with the block.

## Acceptance criteria

- [x] `isPlanPending` / `planAllowedPrompts` cleared for `replayTargetId`, not `activeSessionId`, on **both** clients
- [x] A test that a background session's replay does not cancel the active session's pending plan
- [x] A test that the single-session case (frame omits `sessionId`) is unchanged

## Verification

Five tests per client. Each was **proven red against the pre-fix source with the final fixtures** — sources reverted to `origin/main` with the new tests kept, giving **3 failed / 2 passed on both clients**; the 2 that pass are the "unchanged" controls (active-is-target, no-`sessionId` fallback), which is the split a correct fix should produce.

| Gate | Result |
|---|---|
| dashboard full suite | 301 files / **5252 passed** |
| app full suite | 181 suites / **2359 passed** |
| `tsc --noEmit` (dashboard) | clean |
| `tsc --noEmit` (app, as CI runs it) | clean |

`tsc` caught the first draft of the test fixtures using `string[]` for `planAllowedPrompts`; they now use the real `{ tool, prompt }` shape.

The app suite's "worker process has failed to exit gracefully" warning is **pre-existing** — reproduced on unmodified `origin/main` with these changes reverted.

Closes #7402

### DIFF
--- a/packages/app/src/__tests__/store/message-handler.test.ts
+++ b/packages/app/src/__tests__/store/message-handler.test.ts
@@ -537,6 +537,108 @@ describe('history_replay_start clears stale session_stopped marker (#4909)', () 
   });
 });
 
+// #7402 — the `isPlanPending` / `planAllowedPrompts` wipe on
+// `history_replay_start` must target the session the frame is REPLAYING, not
+// whichever session happens to be active. Same mis-keying #7340 fixed for
+// `activeAgents` in the block immediately above it, and #4909 fixed for
+// `stoppedAt`/`stoppedCode` in the block below.
+//
+// `subscribe_sessions` replays every newly-subscribed BACKGROUND session
+// (server `session-handlers.js`), and both clients fire it for all non-active
+// sessions on every `session_list` — i.e. on every reconnect. Keyed to
+// `activeSessionId`, that burst cancels a pending plan approval the user is
+// looking at, on the very reconnect that is supposed to restore it.
+describe('history_replay_start: isPlanPending wipe targeting (#7402)', () => {
+  afterEach(() => {
+    resetReplayFlags();
+  });
+
+  function seedTwoSessions() {
+    const store = createMockStore({
+      activeSessionId: 's1',
+      sessions: [
+        { sessionId: 's1', name: 'S1' } as any,
+        { sessionId: 's2', name: 'S2' } as any,
+      ],
+      sessionStates: {
+        s1: { ...createEmptySessionState(), isPlanPending: true, planAllowedPrompts: [{ tool: 'ExitPlanMode', prompt: 'go ahead' }] },
+        s2: createEmptySessionState(),
+      },
+    });
+    setStore(store as any);
+    _testMessageHandler.setContext(createMockContext() as any);
+    return store;
+  }
+
+  it("a background session's replay does not cancel the active session's pending plan", () => {
+    const store = seedTwoSessions();
+
+    _testMessageHandler.handle({ type: 'history_replay_start', sessionId: 's2' });
+
+    const state = store.getState();
+    expect(state.sessionStates.s1.isPlanPending).toBe(true);
+    expect(state.sessionStates.s1.planAllowedPrompts).toEqual([{ tool: 'ExitPlanMode', prompt: 'go ahead' }]);
+  });
+
+  it('still clears the active session when it is the one being replayed', () => {
+    const store = seedTwoSessions();
+
+    _testMessageHandler.handle({ type: 'history_replay_start', sessionId: 's1' });
+
+    const state = store.getState();
+    expect(state.sessionStates.s1.isPlanPending).toBe(false);
+    expect(state.sessionStates.s1.planAllowedPrompts).toEqual([]);
+  });
+
+  it("clears a background session's own pending plan when it is the replay target", () => {
+    const store = createMockStore({
+      activeSessionId: 's1',
+      sessions: [
+        { sessionId: 's1', name: 'S1' } as any,
+        { sessionId: 's2', name: 'S2' } as any,
+      ],
+      sessionStates: {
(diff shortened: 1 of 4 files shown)


blamechris:
## Review pass — findings addressed

Both suggestions folded in, both deferred items filed.

### Fixed in `9d6667b12`

**1. The `does not throw when the replayed session state is partially built` test could not fail for its stated reason.** This is the important one — it was a false-safety guard in the PR's own new tests, the exact class `docs/false-safety-guards.md` catalogues. It deleted only `isPlanPending`/`planAllowedPrompts`, neither of which is read unguarded, so the `?? 0` guard the surrounding comment justifies could be deleted outright and all five tests stayed green.

Now deletes `activeAgents` as well — the slice that guard actually defends. **Proven both ways:** with `(ss.activeAgents?.length ?? 0) > 0` mutated to `ss.activeAgents.length > 0`, both clients go **red (1 failed each)**; before the change that same mutant survived all five green.

**2. The transcript-viewer arming comment (`message-handler.ts` ~886) was stale and load-bearing.** It justified "closing the viewer stops RENDERING, never INTERCEPTING" with a claim that a late `history_replay_start` wipes live `activeAgents`/`isPlanPending` — false after #7340, and fully false after this PR. Rewritten to state what is still true: the `message`-frame fallback hazard, plus the two statements above the wipe that still run on the frame's raw id (`_replayingSessions.add` and `reconcileReplayStart`), so a disarmed transcript stream would still pollute replay bookkeeping with a conversationId. The invariant is unchanged; only its justification is now accurate.

**3. Nitpicks.** The dashboard's "mirrors the #4909 block below" cross-reference is dangling — that block lives in the app's copy; the dashboard clears `stoppedAt` in `connection.ts`. Corrected. Also reworded the `#7402` note, which pointed at the *guarding* rationale when the reason it moved is the *keying*.

### Filed, not folded in

| Finding | Issue | Why deferred |
|---|---|---|
| `history_replay_end`'s `answered` sweep is the 4th instance of this mis-keying, and silently resolves **live** permission prompts on reconnect | #7410 | Separate case body, higher severity, deserves its own red-proven tests rather than riding along |
| App's socket-close sweep only clears the active session; dashboard already sweeps all (#5731 T4) | #7411 | Correct fix there is *all sessions*, not *the replay target* — a different fix shape |

#7410 is the one worth reading: `endTargetId` is already computed 20 lines above and unused, and the predicate matches live prompts (`isLivePermissionPrompt` requires `type === 'prompt' && !answered`), so a background session's replay end dismisses an approval the user is looking at.

### Re-verified after the changes

| Gate | Result |
|---|---|
| dashboard full suite | 301 files / **5252 passed** |
| app full suite | 181 suites / **2359 passed** |
| `tsc --noEmit` both | clean |

The review's independent reproduction of the red/green claim (3 fail / 2 pass on both clients against reverted `origin/main` source) matches what I measured.

Issue #7380, opened 2026-08-25
app: permission_expired has no already-resolved suppression, so answering a prompt just before it expires shows a false "Expired" note


blamechris:
Identified during review of PR #7375.

## Context

The dashboard suppresses the #2833 race: if the user already answered a permission and the server's `permission_expired` arrives afterwards, it does **not** append the "(Expired — this permission was already handled or timed out)" note, and instead shows an info toast confirming the response was recorded. See `packages/dashboard/src/store/message-handler.ts` (the `alreadyResolved` branch, gated on `resolvedPermissions[requestId]`).

The mobile app's handler (`packages/app/src/store/message-handler.ts:3154-3183`) has no such branch.

## Why it matters now

Pre-existing asymmetry, but #7375 makes `permission_expired` fire far more often: it now fires whenever a `claude-cli` turn dies — a mode switch, a model switch, Stop, a crash — not just on the five-minute timeout. So the race window is hit much more often than before.

Concrete: a mobile user taps **Allow**; while that response is in flight over a cellular tunnel, the turn is killed (or someone flips to Auto from the dashboard). The app appends "(Expired — this permission was already handled or timed out)" to a prompt the user *did* successfully answer, and gives no indication their tap landed.

## Acceptance criteria

- [ ] The app suppresses the expiry note for a `requestId` the user already resolved locally, matching the dashboard's `alreadyResolved` behaviour.
- [ ] The app surfaces the same reassurance the dashboard does ("already answered — your response was recorded"), rather than silently swallowing the event.
- [ ] Verified red first: a test that resolves a permission then delivers `permission_expired` for it must show the false note before the fix.
- [ ] Positive control: an expiry for a prompt that was **never** answered still appends the note.

Related: #7335, #7375, #2833.

PR #7419, opened 2026-08-27
fix(app,dashboard): key the history_replay_end '(resolved)' sweep to the replayed session

### LINKED ISSUE #7410: dashboard/app: the history_replay_end 'answered' sweep targets the active session, not the replayed one — it silently resolves LIVE permission prompts
## Context

Found during review of PR #7409 (which fixed #7402, the `isPlanPending` instance). This is the **fourth** instance of the same mis-keying in the same pair of files, and the highest-severity one so far.

## Description

In the `history_replay_end` case, both clients sweep messages and stamp unanswered prompts as resolved via `updateActiveSession`, which resolves `state.activeSessionId` and **ignores the frame's `sessionId`**:

- `packages/app/src/store/message-handler.ts` (~2550–2562)
- `packages/dashboard/src/store/message-handler.ts` (~5211–5223)

```ts
updateActiveSession((ss) => { … m.type === 'prompt' && !m.answered ? { ...m, answered: '(resolved)' } … });
```

`endTargetId` is already computed ~20 lines above in the same block (app ~`:2523`, dashboard ~`:5183`) and is simply not used by this sweep — exactly the shape of #4909, #7340 and #7402.

## Why this is worse than #7402

The predicate matches **live** prompts, not just historical ones:

- `isLivePermissionPrompt` (`packages/store-core/src/pending-permissions.ts:109`) is `type === 'prompt' && requestId && expiresAt > now && !answered`
- AskUserQuestion also emits `type: 'prompt'` (`packages/store-core/src/handlers/user-question.ts:174`)

**Trigger:** reconnect → `subscribe_sessions` replays background session B → B's `history_replay_end` arrives while active session A is displaying a live permission prompt → A's prompt is stamped `answered: '(resolved)'` and the user's approval is silently dismissed.

A silently dismissed permission approval is a worse outcome than the cancelled plan approval #7402 fixed, and it happens on every reconnect with more than one session.

## Fix shape

Known and already established three times: key the sweep to `endTargetId` (guarded by a `sessionStates[endTargetId]` existence check), on **both** clients. `endTargetId` already falls back to `activeSessionId` when the frame omits `sessionId`, so the single-session case is unchanged.

## Acceptance criteria

- [ ] The `answered` sweep targets `endTargetId`, not `activeSessionId`, on both clients
- [ ] A test that a background session's `history_replay_end` does not resolve the active session's live permission prompt
- [ ] A test that the single-session case (frame omits `sessionId`) is unchanged
- [ ] Each test proven red against the pre-fix source

## Related

After #7402/#7409 the `history_replay_start` case body is fully target-keyed on both clients. This is the remaining `updateActiveSession` in the replay pair. Worth considering the generalisable rule noted in the #7409 review: **`updateActiveSession` is suspect inside any handler for a message that carries a `sessionId`** — a lint would catch the next one instead of a fifth follow-up issue.

blamechris:
## Summary

On `history_replay_end` both clients stamped `answered: '(resolved)'` on every unanswered prompt through `updateActiveSession`, which resolves `activeSessionId` and **ignores the frame's `sessionId`** — even though `endTargetId` is computed ~20 lines above in the same case body and was simply unused by this sweep.

That silently dismissed a live permission approval in whatever session the user was looking at, on **every multi-session reconnect** (`subscribe_sessions` replays every non-active session on every `session_list`).

Fixing it surfaced a second, independent defect in the same sweep — see "Two defects" below.

## Two defects, three commits

| Commit | Defect |
|---|---|
| `f3a8d75a0` | sweep keyed to `activeSessionId`, not the frame's `endTargetId` |
| `a11dc9bbb` | sweep stamped **live** prompts, not just replayed ones |
| `5fff69a33` | the liveness predicate itself had a hole — replaced with the wire contract |

### Defect 1 — wrong session

Fold the sweep into the existing `endTargetId` block, on both clients. `endTargetId` already falls back to `activeSessionId` when the frame omits `sessionId`, so the single-session case is unchanged; and `updateSession` no-ops on an id with no state — the right outcome for the "session vanished mid-replay" branch, where stamping the active session instead was never correct.

### Defect 2 — live prompts, and why the predicate is `requestId`

The sweep's premise is *"anything in replayed history is already resolved."* Its predicate was `type === 'prompt' && !answered`, which does not express that — it also matches prompts that were never in history at all.

The wire contract settles which prompts those are. Exactly **two** producers emit a `type: 'prompt'` ChatMessage, and history splits them cleanly:

| Producer | In the history ring buffer? | Sets `requestId`? |
|---|---|---|
| `permission_request` | **No** — in the server's `builtinTransient` list (`session-manager.js:3285`) | Yes |
| `user_question` | **Yes** — replayed on every reconnect (`store-core/handlers/user-question.ts`) | No |

So `!m.requestId` is precisely *"could this have come from history?"*, which is the only question this sweep is entitled to ask.

A prompt with a `requestId` present at replay-end always raced the replay rather than being replayed — and that is routine, not exotic. `replayHistory` sends its first 20-entry chunk synchronously and defers the rest through `setImmediate` with back-pressure pauses of up to 30s (`ws-history.js`), while `resendPendingPermissions` runs later in the same post-auth block. On any session with more than one chunk of history the re-sent permission lands *before* `history_replay_end`, and `reconcileReplayEnd`'s baseline slice keeps it deliberately.

**Why not `isLivePermissionPrompt`.** That was the first attempt (`a11dc9bbb`) and review found the hole. It additionally requires `expiresAt > now`, but `expiresAt` is only set when the frame carried `remainingMs` — and `remainingMs` is `.optional()` in `ServerPermissionRequestSchema`, with the shared parser mapping an omitted value to `null`. An emitter that omits it yields a genuinely live, answerable prompt with a `requestId` and no `expiresAt`, which that predicate reports as **not** live — so the sweep stamped it and #7410 recurred on that path. Latent rather than live today (no current emitter omits `remainingMs`), but the predicate was wrong, not lucky.

It also stamped `permission_expired` prompts, whose `answered` is deliberately left empty (*"it is a DECISION token and no decision was made"*) — and post-defect-1 it did so for every replayed background session, not just the active one. `requestId`-keying has neither hole.

**Still open:** a live `user_question` racing a replay is byte-identical to a replayed one, so nothing client-side can separate them. Filed as **#7420** and cross-referenced from the source comment.

## This is the last instance of the mis-keying family

Fourth after #4909 (`stoppedAt`), #7340 (`activeAgents`) and #7402 (`isPlanPending`). Every remaining `updateActiveSession` call in both handlers was audited:

| Site | Why it's correct |
|---|---|
| `server_error`, `stream_end` | the deliberate fallback branch of a target-keyed `if/else` |
| `server_status`, `append_memory_result`, `client_joined`, `client_left`, `web_task_error` | the message type carries no `sessionId` on the wire (verified against the Zod schemas in `packages/protocol`) |

There is no third copy of the sweep — store-core only references it in comments, four of which this PR corrects (they still described it as blanket-stamping the *active* session, and they are the canonical justification for `isPermissionDecision` existing, so a reader taking them at face value could conclude the #7380 guard was unnecessary).

## Proof the tests can fail

Seven tests per client, four mutations, **identical results on both clients**:

| Mutation | Failed | Which tests |
|---|---|---|
| keying reverted to `updateActiveSession` | 2 | keying guard + vanished-session |
| predicate reverted to the liveness check | 1 | the no-`expiresAt` case |
| permission exclusion removed entirely | 2 | both live-prompt cases |
| sweep deleted outright | 3 | the three stamping cases |

Each mutation is caught by exactly the test named after it, and no other.

Two deliberate test-design points:

- **The keying guard uses *historical* prompt fixtures on both sides.** With a live prompt in the active session it would pass with the keying reverted — the permission exclusion would protect it — and would therefore pin nothing. Correct keying is the only thing protecting the active session in that case.
- **Positive controls on every negative assertion.** The live-prompt tests assert `isLivePermissionPrompt` is *true* before dispatching; the no-`expiresAt` test asserts the fixture has a `requestId` and that `isLivePermissionPrompt` is *false* for it — i.e. the test states the exact gap it closes. A malformed fixture fails loudly instead of satisfying the assertion for free (#7409's lesson).

One test — "a background session's replay-end does not resolve the active session's LIVE prompt" — is protected by both mechanisms and cannot go red under any single mutation. It is a **scenario** test pinning the user-visible regression, not a guard, and its comment says so.

## Verification

| Package | Tests | Typecheck |
|---|---|---|
| `packages/app` | 211 passed (was 204) | `tsc --noEmit` clean |
| `packages/dashboard` | 306 passed (was 299) | clean |
| `packages/store-core` | 2174 passed | clean |

Note: the app's `message-handler.test.ts` prints "Jest did not exit one second after the test run has completed". That is **pre-existing** — the pristine `origin/main` file reproduces it at 204/204 — and is not introduced here.

Closes #7410

### DIFF
--- a/packages/app/src/__tests__/store/message-handler.test.ts
+++ b/packages/app/src/__tests__/store/message-handler.test.ts
@@ -662,8 +662,6 @@ describe("history_replay_end: '(resolved)' sweep targeting (#7410)", () => {
     resetReplayFlags();
   });
 
-  const FUTURE = Date.now() + 600_000;
-
   /**
    * An AskUserQuestion-shaped prompt: `type: 'prompt'` with no `requestId` and
    * no `expiresAt`. `user_question` IS in the server's history ring buffer, so
@@ -687,7 +685,27 @@ describe("history_replay_end: '(resolved)' sweep targeting (#7410)", () => {
       type: 'prompt',
       content: 'Allow Write to src/index.ts?',
       requestId,
-      expiresAt: FUTURE,
+      // Computed per call, not once at describe-collection time: a slow suite
+      // would otherwise leave the fixture expired and silently non-live.
+      expiresAt: Date.now() + 600_000,
+      timestamp: 1,
+    } as any;
+  }
+
+  /**
+   * A permission prompt whose frame carried no `remainingMs`. `remainingMs` is
+   * `.optional()` in `ServerPermissionRequestSchema` and the shared parser maps
+   * an omitted value to `null`, so the client leaves `expiresAt` undefined —
+   * giving a genuinely LIVE, answerable prompt for which
+   * `isLivePermissionPrompt` is FALSE. This is the shape that made an earlier
+   * `!isLivePermissionPrompt(...)` predicate stamp a live approval.
+   */
+  function livePromptNoTtl(id: string, requestId: string) {
+    return {
+      id,
+      type: 'prompt',
+      content: 'Allow Bash: rm -rf /tmp/x?',
+      requestId,
       timestamp: 1,
     } as any;
   }
@@ -782,6 +800,38 @@ describe("history_replay_end: '(resolved)' sweep targeting (#7410)", () => {
     const s1Prompt = store.getState().sessionStates.s1.messages[0] as any;
     expect(s1Prompt.answered).toBeUndefined();
     expect(isLivePermissionPrompt(s1Prompt, Date.now())).toBe(true);
+  });
+
+  // The sweep keys off `requestId`, not liveness, and this is why. A permission
+  // frame may omit `remainingMs` (`.optional()` in the schema), which leaves
+  // `expiresAt` undefined and makes `isLivePermissionPrompt` false for a prompt
+  // the user can still answer. A liveness-based predicate stamps it; a
+  // `requestId`-based one cannot.
+  it('does not stamp a live permission prompt that has no expiresAt', () => {
+    const store = seed(historicalPrompt('p1'), livePromptNoTtl('p2', 'req-2'));
+    const before = store.getState().sessionStates.s2.messages[0] as any;
+    // Positive controls: it IS a permission prompt (has a requestId) and it is
+    // NOT recognised as live — i.e. exactly the gap being closed.
+    expect(before.requestId).toBe('req-2');
+    expect(before.expiresAt).toBeUndefined();
+    expect(isLivePermissionPrompt(before, Date.now())).toBe(false);
+
+    _testMessageHandler.handle({ type: 'history_replay_end', sessionId: 's2' });
+
+    expect((store.getState().sessionStates.s2.messages[0] as any).answered).toBeUndefined();
+  });
+
+  // `updateSession` no-ops on an id with no state. Stamping the active session
+  // instead — what the pre-fix code did — was never right for a frame naming a
+  // session this client does not track.
+  it('stamps nothing when the replayed session has no state', () => {
+    const store = seed(historicalPrompt('p1'), historicalPrompt('p2'));
+
+    _testMessageHandler.handle({ type: 'history_replay_end', sessionId: 's-gone' });
+
+    const st = store.getState().sessionStates;
+    expect((st.s1.messages[0] as any).answered).toBeUndefined();
+    expect((st.s2.messages[0] as any).answered).toBeUndefined();
   });
 });
 
--- a/packages/app/src/__tests__/store/message-handler.test.ts
+++ b/packages/app/src/__tests__/store/message-handler.test.ts
@@ -643,6 +643,99 @@ describe('history_replay_start: isPlanPending wipe targeting (#7402)', () => {
     expect(store.getState().sessionStates.s1.isPlanPending).toBe(true);
   });
 });
+
+// #7410 — the `answered: '(resolved)'` sweep on `history_replay_end` must
+// target the session the frame is ENDING, not whichever session happens to be
+// active. Fourth instance of this mis-keying in this file after #4909
+// (`stoppedAt`), #7340 (`activeAgents`) and #7402 (`isPlanPending`), and the
+// worst of the four: the predicate is `type === 'prompt' && !answered`, which
+// matches LIVE prompts as well as replayed ones. `isLivePermissionPrompt`
+// (store-core) is `type === 'prompt' && requestId && expiresAt > now &&
+// !answered`, so a stamp of '(resolved)' takes a live prompt out of the
+// pending-permission surface — and AskUserQuestion emits `type: 'prompt'` too.
+//
+// `subscribe_sessions` replays every background session on every reconnect, so
+// keyed to `activeSessionId` this silently dismissed a permission approval the
+// user was looking at, every time a second session reconnected.
+describe("history_replay_end: '(resolved)' sweep targeting (#7410)", () => {
+  afterEach(() => {
+    resetReplayFlags();
+  });
+
+  const FUTURE = Date.now() + 600_000;
+
+  /** A live, unanswered permission prompt — `isLivePermissionPrompt` is true. */
+  function livePrompt(id: string, requestId: string) {
+    return {
+      id,
+      type: 'prompt',
+      content: 'Allow Write to src/index.ts?',
+      requestId,
+      expiresAt: FUTURE,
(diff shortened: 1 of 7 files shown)


blamechris:
## Follow-up from review — live-prompt exclusion folded in (`a11dc9bbb`)

The inline review found the first commit relocated the harm rather than removing it, and it was right. Verified the premises against the source before acting:

- **`permission_request` is transient.** It is in the server's `builtinTransient` list (`packages/server/src/session-manager.js:3285`), so it is **never** written to the history ring buffer. A live permission prompt present at replay-end has therefore *always* raced the replay — it cannot be a replayed entry.
- **The race window is wide.** `replayHistory` emits 20-entry chunks separated by `setImmediate` yields, with back-pressure pauses of up to 30s on fat payloads (`packages/server/src/ws-history.js:17-47`).
- **`reconcileReplayEnd` keeps racing entries on purpose** — its baseline slice comment says so explicitly (`packages/store-core/src/replay-reconcile.ts:215-218`).

So stamping a live permission prompt inverts the sweep's own premise ("anything in history is already resolved"). Excluded via store-core's `isLivePermissionPrompt` — imported, not re-expressed inline, so there stays exactly one definition of "live permission prompt". `Date.now()` is sampled once per sweep so the `some` and `map` predicates cannot disagree across a millisecond boundary.

### Why one predicate covers both prompt kinds

`user_question` is the **opposite** case — it *is* in the ring buffer and is re-sent on every replay (the `#4613` note in `packages/store-core/src/handlers/user-question.ts:155-171` exists precisely because replayed question events need their original timestamp). So AskUserQuestion prompts must stay stampable, and they carry no `requestId`/`expiresAt`, so `isLivePermissionPrompt` leaves them alone by construction.

The one case that stays undecidable client-side — a **live** AskUserQuestion racing a replay, byte-identical to a replayed one — needs a wire change and is filed as **#7420**, cross-referenced from the source comment.

### Tests restructured so each mechanism is pinned independently

The review's point exposed a second problem: my original "background replay doesn't resolve the active session's live prompt" test used a **live** fixture, so after this hardening it would have passed for two reasons and pinned neither. The keying guard now uses **historical** prompts on both sides — correct keying is the only thing protecting the active session in that case.

Five tests per client, three mutations, identical results on both:

| Mutation | Result |
|---|---|
| keying reverted to `updateActiveSession` | 1 failed / 4 passed — *the keying guard* |
| live-prompt exclusion removed | 1 failed / 4 passed — *the exclusion guard* |
| sweep deleted outright | 3 failed / 2 passed — *the stamping cases* |

Each mutation is caught by exactly the test named after it, and no other.

### Verification

- `packages/app` — 209 passed; `npx tsc --noEmit` clean
- `packages/dashboard` — 304 passed; `npm run typecheck` clean

blamechris:
## Review Comments Addressed

| — | Copilot generated **0** inline comments | n/a | Review body: "Files reviewed: 4/4, Comments generated: 0, Review effort level: Lite" |

**Total:** 0 inline review comments to process.
- Fixed: 0
- False positives: 0
- Follow-up issues created: 0 (from this pass)
- Unresolved conversation threads: **0** (verified via GraphQL `reviewThreads`)

### Caveat worth recording: Copilot reviewed the first commit only

Copilot's review was submitted against **`f3a8d75a0`** at 17:01:55Z. It has not seen `a11dc9bbb` (the live-prompt exclusion) or `5fff69a33` (the `requestId` predicate, the store-core comment corrections, and two of the seven tests per client) — and this repo's ruleset has `review_on_push: false`, so it does not re-review automatically. Re-requesting it via `POST /pulls/7419/requested_reviewers` returned 200 but did not stick.

Its "Approval recommended" therefore covers roughly a third of the final diff, and specifically **not** the subtlest part. The substantive review of the later commits came from the agent review above, which is what found the `expiresAt`/`remainingMs` hole that `5fff69a33` fixes.

Recording this rather than letting a green Copilot badge stand in for coverage it does not have.

### Issue reconciliation

Open `from-review` issues checked against this PR's changes:

- **#7420** — a live `user_question` racing a replay is indistinguishable from a replayed one. **Still open, correctly.** This PR deliberately does not fix it (needs a wire change); the source comment cross-references it.
- **#7421** — `_ctx.isSessionSwitchReplay` is write-only dead state. **Still open, correctly.** Out of scope here; filed from the agent review.

Nothing closed by this pass.

PR #7440, opened 2026-08-28
fix(server): mark transient fork bail-outs indeterminate so the CI watcher keeps its arm (#7435)

### LINKED ISSUE #7435: session-pr-status: a transient `gh` failure on a fork reports the quiet negative, cancelling a CI watch
## Context

Identified during review of PR #7432 (#7427).

## Description

`surveySessionPrStatus`'s fork-widening bail-outs (`packages/server/src/session-pr-status.js`, the paths around the upstream lookup) return `{ pr: null, reason: null }` — the **quiet negative**, meaning *"I looked, and there is no open PR"* — when the underlying `gh` call merely failed transiently.

That posture is right for the **display** slice it was written for (#7422): don't downgrade a usable answer into a spinner. It is wrong for a **state machine**. `SessionCiWatcher._reconcile` reads a null `pr` as an authoritative fact and **drops the arm**:

```js
if (!snapshot?.pr) {
  st.armed = null
  return 'no-pr'
}
```

So a fork-checkout session whose CI is being watched cancels its own watch on the first flaky `gh` call, and the user never gets the completion notification they were waiting on.

## Why now

Pre-existing — #7426's sweep shares the same reading. But #7427 made the survey reachable **on demand, at client-chosen frequency**, so the odds of hitting a transient failure while armed went up substantially, and it is now triggerable rather than merely possible.

## Original review comment

> The fork-widening bail-outs return `{ pr: null, reason: null }` — the **quiet negative** — on a *transient* `gh` failure. `_reconcile` reads that as "no open PR" and **drops the arm**. […] The display posture ("do not downgrade a usable answer") is right for a chip and wrong for a state machine that treats the negative as a fact.

## Acceptance criteria

- [ ] A transient `gh`/git failure on a fork path is distinguishable by the watcher from a genuine "no open PR"
- [ ] `_reconcile` does not drop an arm on the indistinguishable case (a failed reading is not evidence in either direction — the same rule `reason` already gets)
- [ ] The display slice keeps its current behaviour: a usable partial answer is still rendered, not downgraded to "unavailable"
- [ ] Test: an armed session survives a fork bail-out, and a genuine no-PR reading still disarms (positive control)

blamechris:
Fixes #7435: a fork-checkout session whose CI is being watched cancelled its own watch on the first flaky `gh` call, because the fork-widening bail-outs in `surveySessionPrStatus` report a *transient* failure in exactly the shape of the authoritative quiet negative (`pr: null, reason: null`), and `SessionCiWatcher._reconcile` reads that shape as "no open PR" and drops the arm.

## What changed

- **`resolveParentRepo` answers `{ parent }` vs `{ failed: true }`** — "gh answered: this repo has no parent" (authoritative) and "the lookup itself failed" (transient) no longer share a `null`. An answered-but-unusable shape (including an option-parseable parent name, which exists but cannot be queried) counts as failed, not as absence.
- **Four fork bail-outs now set `indeterminate: true`** (server-side-only marker): failed parent lookup, failed upstream query, unparseable upstream output, and a matching-but-unusable upstream row. The two authoritative fork readings — upstream answered with no PR from this fork, and not-a-fork — keep the quiet negative, as does the answered-empty same-repo case.
- **`_reconcile` treats `indeterminate` exactly like `reason`**: `'undeterminable'` — never disarms, never fires. A failed reading is not evidence in either direction.
- **The WS handler strips the marker before the reply goes out.** The wire/display contract is byte-for-byte unchanged: the chip still renders the usable partial answer rather than downgrading to "cannot determine" (the #7422 display posture, pinned by the acceptance criteria).

## Acceptance criteria → tests

- Transient fork failure distinguishable from genuine no-PR → per-site survey tests assert `indeterminate: true` with `reason` still null.
- Arm survives the indistinguishable case → watcher tests for both the sweep path and the `observe()` path prove the arm survives an indeterminate reading and still fires on the later genuine settle.
- Display unchanged → handler test proves the wire reply does not carry the marker while `observe()` receives it verbatim.
- Positive controls → authoritative no-PR readings (same-repo and fork) still read as the quiet negative and still disarm.

## Verification

- Red-first: the rewritten "parent lookup fails" test and every new test failed against the unmodified source.
- Mutation-tested: removing the watcher's `indeterminate` guard fails exactly the two new watcher tests; making a failed parent lookup read as authoritative fails exactly its paired survey test; reverting the handler strip fails exactly the wire-strip test.
- 143/143 across the three touched suites + `ws-server.test.js` (124) green; all 9 `scripts/lint-*.sh` server lints pass.


blamechris:
## Code Review: PR #7440

### Summary

The diagnosis is right and the shape of the fix is right. `{ pr: null, reason: null }` was carrying two incompatible meanings — "gh answered: nothing here" and "I never found out" — and `_reconcile` is a state machine that can only act on the first. Splitting `resolveParentRepo`'s return into `{ parent }` vs `{ failed: true }` puts the distinction where it belongs (at the source of the ambiguity, not at the consumer), and giving `indeterminate` exactly the treatment `reason` already gets means the watcher has one rule for "not evidence" rather than two.

I reproduced the PR body's two headline mutation claims independently, and both landed exactly as stated:

- reverting `session-ci-watcher.js:479` to `if (snapshot?.reason)` → **143 tests, 2 fail**, and they are precisely the two new watcher tests. Nothing else moved.
- reverting the handler strip at `session-pr-status-handlers.js:167-168` to `...snapshot` → **21 tests, 1 fail**, the wire-strip test alone.

Baseline confirmed green here too: 267/267 across the three touched suites plus `ws-server.test.js`. CI is green on `696c6fe37` and the PR is `MERGEABLE`/`CLEAN`.

Two things block, both of them coverage rather than behaviour — but this repo's own doctrine (`docs/false-safety-guards.md`, and CLAUDE.md's *"if you cannot make it fail, it is not a guard"*) treats an unproven guard as a bug, and I have mutation evidence for both. Together they are about five lines of test.

### Strengths

- **The split is at the right layer.** `resolveParentRepo` returning a discriminated result instead of `null` means the ambiguity cannot be reintroduced by a caller that forgets — the type no longer has a shape that can express "failed" and "absent" with the same value.
- **`indeterminate` reuses the existing rule rather than inventing a second one.** `snapshot?.reason || snapshot?.indeterminate === true` keeps one "changes nothing" branch in `_reconcile`. A parallel `if` with its own return value would have been the thing that drifts.
- **The two authoritative fork readings are deliberately left alone** (`session-pr-status.js:534` not-a-fork, `:536` upstream-answered-no-row), and both have a `assert.equal(snap.indeterminate, false)` positive control on them. That is the assertion that stops the fix from degenerating into "mark everything indeterminate", which would have disabled disarming outright while every test stayed green.
- **The wire strip is tested from the outside** — `!('indeterminate' in reply)` plus `observed[0].indeterminate === true` in one test pins both halves, so a "fix" that passed `wireSnapshot` to `observe()` fails immediately.
- **The doc-block edits are load-bearing**, not decoration: the third-state paragraph at `session-pr-status.js:51-58` names the one consumer that must not read the marker as a fact, which is the fact a future reader needs.

### Issues Found

#### Critical (Must Fix)

|---|---|---|---|
| `packages/server/src/session-pr-status.js` | 378, 382, 385 | **Three of the five `{ failed: true }` arms are unpinned — mutation-proven.** I reverted all three to the pre-PR `return { parent: null }` in one pass and ran the three touched suites: **143 tests, 143 pass, 0 fail.** Only the `execFn` throw at :370 is covered (by the rewritten parent-lookup test). The PR body states the claim explicitly — *"An answered-but-unusable shape (including an option-parseable parent name, which exists but cannot be queried) counts as failed, not as absence"* — and nothing proves it. The `isSafeArgvValue` arm at :385 matters most: it is the security guard's *outcome* that changed in this PR (rejected parent → was quiet negative, is now indeterminate), and the existing argv test at `tests/session-pr-status.test.js:441` asserts only on argv, so it stays green either way. Verified shapes → all four currently yield `indeterminate: true`: `{}`, `{"parent":{}}`, `{"parent":"x"}`, `notjson`. | Add three assertions to `tests/session-pr-status.test.js`, in the fork describe block: (1) a `gh repo view` stdout of `'{"parent":{}}'` → `snap.indeterminate === true`; (2) unparseable `gh repo view` stdout → `snap.indeterminate === true`; (3) `assert.equal(snap.indeterminate, true)` appended to the existing `'never puts an option-parseable parent owner/name into an argv slot'` test. Re-run the same three-arm revert afterwards and confirm it now goes red. |
| `packages/server/tests/session-ci-watcher.test.js` | 942-946 | **This PR silently voided the `'no open PR'` entry of the REAL-producer contract test.** That entry answers *every* `gh` invocation with `'[]'` via a blanket `if (bin === '/usr/bin/gh') return { stdout: '[]' }` — including `gh repo view --json parent`. Pre-PR, `JSON.parse('[]')` fell to `typeof owner !== 'string'` → `return null` → the quiet negative. Post-PR it hits the new `!('parent' in parsed)` arm at `session-pr-status.js:378` → `{ failed: true }`. I drove the identical exec table through the real producer and measured: `{"pr":null,"reason":null,"indeterminate":true}`. So the one test in the repo that runs the REAL producer through the watcher's shape contract no longer reaches a quiet negative at all, its label is now false, and neither positive control (`sawPr`/`sawReason` at :958-959) can see it. This is exactly the class the test's own comment says it exists to prevent — *"a guard going quietly inert"* — and #7435's entire subject is keeping that one reading distinguishable. | Give the fork lookup its own stdout in that entry: `if (bin === '/usr/bin/gh' && argv[0] === 'repo') return { stdout: '{"parent":null}' }` before the blanket `gh` arm. Then add a third positive control alongside :958-959 — `let sawQuietNegative = false` / `if (!real.pr && !real.reason && !real.indeterminate) sawQuietNegative = true` / `assert.ok(sawQuietNegative, ...)` — so the next change that swallows this path fails instead of relabelling itself. Consider adding one `indeterminate` entry to the same table while you are in it: the table claims to enumerate *"every shape surveySessionPrStatus actually returns"* and this PR added one. |

|---|---|---|---|
| `packages/protocol/src/schemas/server/session.ts` · `packages/server/src/ws-server.js` | 709-711 · 493 | **The wire's documented meaning is now false, in the one direction a client can act on.** Both say `pr: null` + `reason: null` means *"this branch has no open PR on `origin`, nor (when `origin` is a fork) on its upstream"*. After this PR the server emits that exact shape on four fork paths while knowing it established nothing — and deliberately withholds the marker that says so. `SessionCiChip.tsx:155` renders it as the tooltip **"This branch has no open pull request"**. Keeping the wire byte-identical is the pinned #7422/#7435 posture and I am not arguing with it, but the *documentation* of a contract that just got weaker should record the weakening in the same PR — otherwise the next client feature to treat the quiet negative as a fact reproduces #7435 on the dashboard side, where there is no server-side marker to save it. | Two comment edits, no code: append to the protocol note that on a fork checkout the quiet negative is **best-effort** — a failed upstream widening also reports it — and that the authoritative-vs-transient distinction is held server-side and not transmitted (`#7435`). Mirror one clause into the `ws-server.js:493` roster line. |
| `packages/server/src/session-pr-status.js` | 538-542 | Asymmetry: a matching-but-unusable *fork* row gets `indeterminate`, while the identical same-repo condition at :551 gets a `reason`. Filed — see Deferred Items. | → **#7441** |
| `packages/server/src/handlers/session-pr-status-handlers.js` | 167 | A whole-server invariant ("never on the wire") enforced at one distant call site. Filed — see Deferred Items. | → **#7442** |
| `packages/server/src/session-ci-watcher.js` | 479 | A permanently-undeterminable armed session holds its arm and its every-tick survey cadence forever, with no backoff. Filed — see Deferred Items. | → **#7443** |

#### Nitpicks (Optional)

|---|---|---|---|
| `packages/server/src/session-ci-watcher.js` | 479 | `snapshot?.reason || snapshot?.indeterminate === true` mixes a truthy test and a strict-equality test in one expression. Safe today — the only producer writes a literal boolean, and `observe()`'s input is the server's own survey result, never client data — but the asymmetry reads as accidental and would let a future `indeterminate: 1` slip past silently. | Either make both strict, or drop `=== true` and note in the comment that only `session-pr-status.js` may set it. |
| `packages/server/tests/session-pr-status.test.js` | 530-544 | The `wire contract` test validates the **survey's** output, not the handler's wire output, and `ServerSessionPrStatusSchema` is a plain stripping `z.object` — so it cannot detect an added wire field in either direction, and it now silently strips `indeterminate` from every case. Harmless (the new handler test is the real guard), but the suite name overstates it. | Add one indeterminate-producing table to the cases for completeness, and/or rename to make clear it checks snapshot conformance rather than the emitted frame. |
| `packages/server/src/session-pr-status.js` | 359-363 | The `@returns` union `{parent}|{failed: true}` is accurate, but the call site at :502 reads `parentResult.failed` on a member that does not declare the key. Fine in JS; slightly loose if this file ever gets `checkJs`. | Give both members both keys (`{ parent, failed: false }`), or leave as-is and note the union is discriminated by presence. |

### Deferred Items

| Issue | Title | Why deferred |
|---|---|---|
| [#7441](https://github.com/blamechris/chroxy/issues/7441) | A matching-but-unusable fork row should report a `reason`, like its same-repo twin | Changes the display posture for one case, which #7435's AC3 pinned. Worth doing — at that site the upstream *answered* and returned *our* row, so there is no usable partial answer being protected and the `reason` path preserves `branch`/`repo` identically — but it is a display decision, not part of the watcher fix, and it would also make the marker unnecessary at that one site. |
| [#7442](https://github.com/blamechris/chroxy/issues/7442) | Make the server-only marker structurally unable to reach the wire | The single-call-site strip is complete today (one sender). Making it structural — `toWireSnapshot()`, or a non-enumerable / `Symbol`-keyed property that `JSON.stringify` and object spread both drop — is a design change beyond this fix's scope. Worth noting the dashboard's stripping `safeParse` is a real second line of defence, but it is invisible from the server. |
| [#7443](https://github.com/blamechris/chroxy/issues/7443) | A permanently undeterminable session holds its arm and its every-tick cadence forever | Pre-existing on the `reason` path; this PR widens which failures reach it rather than creating the gap. The fix is backoff-with-arm-retained (disarming after N failures would reintroduce #7435 on a slow flap), which needs its own design. |

### Architecture Notes

**The three-state model is the right call, and the honest cost is worth naming.** `reason` / `indeterminate` / quiet-negative encodes *epistemic* state — how the server knows what it knows — and only `reason` is transmitted. That is a deliberate asymmetry between what the daemon knows and what the client is told, and it is defensible here: the display slice's job is to avoid a spinner, the watcher's job is to avoid acting on a non-fact, and those genuinely want different answers. The cost is that the wire contract is now *weaker than it reads* on fork paths, which is why the doc suggestion above is the one I would fix in this PR rather than defer.

**`resolveParentRepo`'s failure set grew from 2 to 5 without the tests growing.** Pre-PR it had three `return null` sites, all meaning the same thing, so one test covered the lot honestly. The split turned each into a distinct claim, and only one gained a test. That is the mechanism behind Critical 1: refactoring a collapsed return into a discriminated one multiplies the assertions a suite needs, and the suite is the thing that does not automatically follow.

**Over a tunnel, this is the right direction.** The failure mode #7435 describes — one flaky `gh` call silently cancelling a watch the user is waiting on, with no surface anywhere — is exactly the class that is invisible on a desk and infuriating on a phone. Holding the arm through uncertainty is strictly the better default, and #7443 is about paying for that default sensibly rather than reconsidering it.

**Verification note:** all mutations above were applied by scripted replace with an occurrence assertion, run, and restored from a `/tmp` copy. `git status --short` is empty and all three source files match their pre-review checksums.

### Verdict

- [ ] **Approve** — ready to merge
- [x] **Request Changes** — the two Critical items above (~5 lines of test between them). The behaviour is correct and CI is green; what is missing is proof for three guards and a contract test that this diff quietly stopped covering. Both are mutation-demonstrated above, and both are cheap.

blamechris:
Both criticals addressed in 47c41502b, with the prescribed re-verification run:

- **Critical 1 (unpinned arms):** added the shape-pin test (`notjson` / `{}` / `{"parent":{}}` / `{"parent":"x"}` → all `indeterminate: true`) and the outcome assertion on the argv-safety test. Re-ran the exact three-arm revert from the review: **now fails precisely the two pinning tests** (pin test + argv outcome), 38/40 elsewhere green.
- **Critical 2 (voided contract entry):** the `'no open PR'` entry now answers `gh repo view --json parent` with `{"parent":null}`, a `'fork lookup failed'` entry drives the real indeterminate path, and `sawQuietNegative`/`sawIndeterminate` positive controls guard both. Re-applied the blanket-`'[]'` mutation: **the `sawQuietNegative` control goes red** exactly as intended.
- **Suggestion (wire docs):** both spec sites now record the weakening — the protocol note (`session.ts`, dist rebuilt and committed) and the `ws-server.js` roster line say the fork-path quiet negative is best-effort, with the distinction held server-side (#7435).
- **Nitpicks:** strict-`=== true` rationale added to the watcher guard comment; discriminated-by-presence note on `resolveParentRepo`'s @returns; an indeterminate case added to the schema-conformance test. The suggestion-4/5/6 items stay tracked as #7441 / #7442 / #7443.

268/268 across the four touched suites at 47c41502b.

blamechris:
## Re-review at `47c41502b` — Approve

Re-verified all six items. Both of my demonstrations now go red, and each one is caught by the test that was written for it. Verdict updated to **Approve**.

### The two demonstrations, re-run

**1. The `resolveParentRepo` arm revert — now RED.** I mutated each arm *individually* rather than all three at once, because the new loop test asserts fail-fast and a combined revert would let one arm's failure mask the others:

| Mutated arm (alone) | Result | Caught by |
|---|---|---|
| `:378` JSON.parse catch → `{ parent: null }` | **1 fail** | `pins each answered-but-unusable parent shape` (on `notjson`) |
| `:380` `!('parent' in parsed)` → `{ parent: null }` | **2 fail** | that test (on `{}`) **and** the real-producer contract test via `sawIndeterminate` |
| `:384` non-string owner/repo → `{ parent: null }` | **1 fail** | that test, failing specifically `for {"parent":{}}` — so the loop is proven to reach past its first element |
| `:387` `isSafeArgvValue` → `{ parent: null }` | **1 fail** | `never puts an option-parseable parent owner/name into an argv slot` |

All four survived the suite before; all four are now individually pinned. Combined with the pre-existing `execFn`-throw coverage at `:372`, every failure return in `resolveParentRepo` is proven.

**2. Re-voiding the contract entry — now RED.** I deleted the new `if (bin === '/usr/bin/gh' && argv[0] === 'repo') return { stdout: '{"parent":null}' }` line from the `'no open PR'` entry, restoring the blanket `'[]'` that caused the original defect — leaving the `'fork lookup failed'` entry untouched. Result: **1 fail**, `accepts every shape surveySessionPrStatus actually returns — from the REAL producer`, with exactly the intended message:

> `the table must reach the authoritative quiet negative — a new arm that reroutes it must fail here, not relabel it (#7435)`

**The two controls are complementary, and I confirmed neither is redundant.** `sawQuietNegative` is satisfiable by only one entry in the table (every other path sets a `reason`, a `pr`, or `indeterminate`), and my two mutations killed the test through the two controls *separately* — the arm revert via `sawIndeterminate`, the entry re-void via `sawQuietNegative`. Either control alone would have let one of the two regressions through.

### The rest

- **`'fork lookup failed'` entry** — answers `gh repo view` with `{}`, so it lands on the `!('parent' in parsed)` arm and produces the genuine three-state shape from the real producer. Proven live: mutating that arm is what makes `sawIndeterminate` fail.
- **Doc edits** — the protocol note (`session.ts:709-714`) and the roster line (`ws-server.js:493`) both now record that the fork quiet negative is best-effort and that the distinction is held server-side and not transmitted. `packages/protocol/dist/schemas/server/session.js` is rebuilt, tracked, and in sync with the source. `ServerSessionPrStatusSchema` is still a plain stripping `z.object`, so nothing became accidentally strict.
- **Nitpicks** — all three taken: the strict `=== true` now carries the reason it is strict rather than being changed, the `@returns` union documents that it is discriminated by key presence, and the wire-contract test gained an indeterminate case.
- **No new untested behaviour.** I diffed `696c6fe37..47c41502b` restricted to `packages/server/src` + `packages/protocol/src` with comment and blank lines filtered out: the result is empty. Every source change in this commit is a comment, so the entire diff is tests plus documentation.

### Gates

- Targeted suites at `47c41502b`: **268/268 green** (`session-pr-status`, `session-ci-watcher`, `session-pr-status-handler`, `ws-server`) — re-run after restoring every mutation.
- CI: all required checks pass at `47c41502b`; `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`.
- Every mutation was applied by scripted replace with an occurrence assertion and restored from a `/tmp` copy; `git status --short` is empty and the working tree is byte-identical to `47c41502b`.

The three deferred follow-ups (#7441, #7442, #7443) stand as filed and are not blockers.

### Verdict

- [x] **Approve** at `47c41502b`
- [ ] Request Changes
- [ ] Comment

Issue #7270, opened 2026-08-19
Server Windows Tests runs a hardcoded file list, so new cross-platform tests silently never run there


blamechris:
## Summary

`.github/workflows/ci.yml`'s `Server Windows Tests` job passes an explicit list of 8 test files to `node --test`. The list is curated on purpose — most of the server suite needs node-pty / POSIX signals — but it means a newly added, genuinely cross-platform test is **never run on Windows and nobody is told**.

Concretely: #7266 added `tests/setup-sandbox-binding-forms.test.js`, which is entirely about path handling and errno behaviour and is exactly the kind of test Windows coverage is for. It is not in the list, so the Windows job passed in 58s without executing it. The job's green tells you nothing about the file.

This is `docs/false-safety-guards.md` cause #1 — a hardcoded list beside a set that grows — in the CI config rather than in a script.

## Why not just add the file

Adding it is the immediate fix, and the sandbox test is a reasonable candidate (its probes rely on missing-parent `ENOENT`, which needs confirming on a real Windows host). But the *class* is the issue: the next cross-platform test will be omitted the same way.

## Suggested fix

Invert the list, per the doc's own guidance: walk all test files and maintain an EXEMPT list of the POSIX-only ones. A missing entry then fails loudly (an unexpected file runs and breaks) instead of silently (a new file never runs). Same data, opposite failure direction — the `#7222`/`#7235` pattern.

## Acceptance criteria

- [ ] Windows job derives its set by exclusion, not inclusion
- [ ] The exempt list says why each file is exempt
- [ ] `setup-sandbox-binding-forms.test.js` confirmed green on a real Windows host, or explicitly exempted with a reason
- [ ] Prove the gate works: add a throwaway failing test and confirm the Windows job goes red

## Provenance

Adversarial review of #7266.

Issue #7252, opened 2026-08-18
merge-updater-feeds.test.sh is never run by CI


blamechris:
## Context

Surfaced while checking whether #7236's new `scripts/__tests__/lib/` directory could be swept up by a glob over `scripts/__tests__`. It cannot — CI names every scripts test explicitly. But enumerating them turned up one that is named nowhere.

`scripts/__tests__/merge-updater-feeds.test.sh` exists and passes (`exit=0` locally), but the "Scripts Tests" job in `.github/workflows/ci.yml` invokes only nine of the ten test files in that directory. This one is not among them:

```
bump-version · docker-entrypoint · flush-and-exit · compile-skill-targets
is-entry-point · lint-no-raw-color-literals · lint-no-nul-bytes
gen-agents-md · check-release-pr-subject
```

Every reference to `merge-updater-feeds` outside the test itself points at the *script*, not the test:

- `.github/workflows/release.yml:499` — `node scripts/merge-updater-feeds.mjs --output artifacts/updater/latest.json $FEEDS`

So the subject is **release-critical** (it builds the combined Tauri updater `latest.json`) while its test has never gated anything. It appears to have been unwired since it was introduced in `d9a46076f` (#4736).

## Why this matters

This is the repo's own named defect class, from [`docs/false-safety-guards.md`](docs/false-safety-guards.md): a check whose *output* is correct and whose *coverage* is what is wrong. A test suite that is never invoked is indistinguishable from a passing one — the same shape as "a precondition that is false, so the body never runs and the job is green" (#7184, #7198). It survives lint, typecheck and CI precisely because nothing ever asks it a question.

## Suggested fix

Add the step to the "Scripts Tests" job alongside the other nine:

```yaml
      - name: Run merge-updater-feeds.mjs tests
        run: bash scripts/__tests__/merge-updater-feeds.test.sh
```

Then prove the wiring is real rather than assuming it: break `scripts/merge-updater-feeds.mjs` and confirm the job goes **red**. A step that is present but silently no-ops is the same bug one level up.

Worth checking in the same pass whether any *other* directory has test files CI does not name — the failure mode is invisible by construction, so the audit is the fix.

## Acceptance criteria

- [ ] `merge-updater-feeds.test.sh` runs in CI
- [ ] Breaking `merge-updater-feeds.mjs` turns that job red (verified, not assumed)
- [ ] No other test file in `scripts/__tests__/` is unreferenced by CI

Issue #7383, opened 2026-08-25
ci: the 5-minute job tier is too tight for the self-hosted runner — setup-node's cache restore alone can take 4m+, and the cancellation reads as a test failure


blamechris:
## Summary

Jobs in CI's **5-minute tier** are being cancelled at `actions/setup-node` before their real steps run. `gh pr checks` renders a cancelled job as **`fail`**, so every occurrence looks like a red test result until you query the job — which is the expensive part: the natural reading is "my change broke this".

Observed four times across three consecutive runs on two PRs (#7375, #7381), and every affected job is in the 5-minute tier:

| Run / PR | Job | `timeout-minutes` | What actually happened |
|---|---|---|---|
| #7375 | `App Tests` | 5 | **All 2343 tests passed**, 181/181 suites, coverage uploaded — then cancelled |
| #7381 (head 1) | `Claude Hooks Tests` | 5 | cancelled at `setup-node`; no test ever ran |
| #7381 (head 2) | `Claude Hooks Tests` + `Store Core Type Check` | 5 each | both cancelled at `setup-node` |
| #7381 (head 2, re-run) | `Claude Hooks Tests` | 5 | cancelled at `setup-node` again |

## Root cause

`setup-node`'s npm cache restore is slow on the self-hosted runner. Measured on job `98006824887`:

```
setup-node  started=2026-08-25T23:38:50Z  completed=2026-08-25T23:43:01Z   -> 4m11s
```

That leaves under a minute of the 5-minute budget for `Install dependencies` + `Build @chroxy/protocol` + `Run tests`, so the job hits the wall and GitHub cancels it. The `cache-dependency-path: '**/package-lock.json'` glob spans the whole monorepo, which is presumably why the restore is expensive.

This is **not** flakiness in the usual sense — it is a budget that does not cover the observed worst case, so it fails whenever the cache path is cold or the runner is loaded. The affected jobs are exactly the cheap ones, because they are the ones given 5 minutes; `Server Tests` (10), `Desktop Rust Tests` (15) and `Server Windows Tests` (20) absorb the same 4-minute setup without noticing.

The code is fine in every observed case — e.g. `npm test` in `packages/claude-hooks` passes 185/185 locally on the affected commit.

## Why it matters beyond the noise

[...]

PR #7444, opened 2026-08-28
ci: run packages/design-tokens' test suite on every PR (#7408)

### LINKED ISSUE #7408: test(ci): packages/design-tokens' test suite never runs in CI
## Context

Found while wiring the `--test-force-exit` refusal (#7400 / PR #7407) into every package that runs `node --test`.

`packages/design-tokens` has a real suite:

```
$ cd packages/design-tokens && npm test
# tests 6
# pass 6
```

No CI job runs it. `.github/workflows/ci.yml` mentions `design-tokens` exactly once — the staleness check that tells you to edit `src/tokens-data.js` and regenerate — and never invokes its `test` script. Server, Claude Hooks, Protocol, Store Core, Dashboard and App all have jobs; this package does not.

## Why it matters

This is the plain form of the class in `docs/false-safety-guards.md`: a suite that never runs and a green PR are the same observable outcome. The tokens feed the dashboard's generated `theme.css` / `tokens.ts`, so a regression here surfaces as a visual defect rather than a red check.

## Acceptance criteria

- [ ] `packages/design-tokens`' `npm test` runs in CI on every PR that touches it (fold into an existing job or add one — it is 6 tests and takes under a second)
- [ ] Verify it can go RED: break one token assertion on a scratch branch and confirm the check fails
- [ ] While there: check whether any other workspace package's `test` script is likewise unreferenced by the workflow, and say so in the PR either way

blamechris:
Closes #7408.

`packages/design-tokens` has a real six-test suite that no CI job ever invoked — `ci.yml` mentioned the package only in the dashboard staleness check. This is the plain form of `docs/false-safety-guards.md`: a suite that never runs and a green PR are the same observable outcome, and these tokens feed the dashboard's generated `theme.css`/`tokens.ts`, so a regression would surface as a visual defect rather than a red check.

## What changed

One new **Design Tokens Tests** job, modeled line-for-line on the sibling per-package jobs (`runner-target` resolution, Node 22 + npm cache, `npm ci` at the root, `npm test` in the package). It runs unconditionally on every PR, like its siblings — the suite is six sub-second tests, so path-gating would buy nothing and cost a filter to maintain.

## Acceptance criteria

- [x] `npm test` for the package runs in CI on every PR — the new job.
- [x] **Red-proof completed in this PR's own CI history** (the guard lives in CI config, where no local test can prove the wiring — #7270's lesson): RED on `0df81c5fd` — [failing Design Tokens Tests run](https://github.com/blamechris/chroxy/actions/runs/33200183741/job/98947490111) · GREEN on the exact revert `db9e5b138` — [passing run](https://github.com/blamechris/chroxy/actions/runs/33200558894/job/98948732808).
- [x] Survey of other unreferenced test scripts: server / claude-hooks / protocol / store-core / dashboard / app all have jobs; desktop is Rust (Desktop Rust Tests ×2); the repo `scripts/` suite runs under Scripts Tests. **design-tokens was the only `packages/*` gap.** (The `scripts/` suite runs under Scripts Tests, but that job names test files by hand and is missing one — pre-existing, tracked as #7252.)

## Note for the merger

The new check is not in the branch-protection required-contexts list; making it required is a repo-settings change only the owner can do.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -814,6 +814,36 @@ jobs:
       - name: Run tests
         run: npm test
 
+  design-tokens-tests:
+    name: Design Tokens Tests
+    # #7408: the one JS workspace package whose `npm test` no CI job invoked —
+    # the plain form of docs/false-safety-guards.md (a suite that never runs and
+    # a green PR are the same observable outcome; the tokens feed the
+    # dashboard's generated theme.css/tokens.ts, so a regression surfaced as a
+    # visual defect instead of a red check). Six sub-second tests; runs
+    # unconditionally like the sibling per-package jobs.
+    needs: runner-target
+    runs-on: ${{ fromJSON(needs.runner-target.outputs.runner) }}  # SELF_HOSTED_OR_HOSTED (see header)
+    timeout-minutes: 5
+    defaults:
+      run:
+        working-directory: packages/design-tokens
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+
+      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
+        with:
+          node-version: 22
+          cache: ${{ needs.runner-target.outputs.npmcache }}
+          cache-dependency-path: '**/package-lock.json'
+
+      - name: Install dependencies
+        run: npm ci
+        working-directory: .
+
+      - name: Run tests
+        run: npm test
+
   store-core-typecheck:
     name: Store Core Type Check
     needs: runner-target
--- a/packages/design-tokens/test/index.test.js
+++ b/packages/design-tokens/test/index.test.js
@@ -28,7 +28,7 @@ test('navy palette + structural tokens are present with expected values', () => 
   assert.equal(vars['bg-primary'], '#0f0f1a') // navy preserved
   assert.equal(vars['accent-blue'], '#4a9eff')
   assert.equal(vars['text-chat'], '15px') // new chat reading size
-  assert.equal(vars['radius-md'], '999px') // RED-PROOF (#7408): temporary — reverted in the next commit
+  assert.equal(vars['radius-md'], '10px')
   assert.equal(vars['dur-base'], '200ms')
   assert.match(vars['ease-out'], /^cubic-bezier/)
 })
--- a/packages/design-tokens/test/index.test.js
+++ b/packages/design-tokens/test/index.test.js
@@ -28,7 +28,7 @@ test('navy palette + structural tokens are present with expected values', () => 
   assert.equal(vars['bg-primary'], '#0f0f1a') // navy preserved
   assert.equal(vars['accent-blue'], '#4a9eff')
   assert.equal(vars['text-chat'], '15px') // new chat reading size
-  assert.equal(vars['radius-md'], '10px')
+  assert.equal(vars['radius-md'], '999px') // RED-PROOF (#7408): temporary — reverted in the next commit
   assert.equal(vars['dur-base'], '200ms')
   assert.match(vars['ease-out'], /^cubic-bezier/)
 })


blamechris:
Red-proof completed in this PR's own CI history, per the issue's second acceptance criterion:

- **RED** — `0df81c5fd` (deliberately broken assertion): Design Tokens Tests **failure** → https://github.com/blamechris/chroxy/actions/runs/33200183741/job/98947490111
- **GREEN** — `db9e5b138` (revert): Design Tokens Tests **success** → https://github.com/blamechris/chroxy/actions/runs/33200558894/job/98948732808

The new check demonstrably runs the suite and goes red on a real regression. (Body checkbox will be ticked once CI on the head settles — editing the body cancels the in-flight run via the concurrency group.)

blamechris:
## Code Review: PR #7444

### Summary

One new `Design Tokens Tests` job in `.github/workflows/ci.yml` (+30 lines, no other file
touched), closing the #7408 gap where `packages/design-tokens`' six-test suite was never
invoked by CI. I verified the job independently rather than by inspection alone: the
red-proof in the PR's own history, byte-consistency against the sibling jobs, the repo's
three CI-config guard suites, the survey claim, YAML validity, and the skippability of the
needs-chain. It is correct and the red-proof is real.

**Verdict: approve.** No blocking issues. Three suggestions and two nitpicks below, all
optional; two are already filed as follow-ups so they survive this PR closing.

### Strengths

**The red-proof is genuine, and precise about what it proved.** `0df81c5fd` flips
`vars['radius-md']` from `10px` to `999px` in `packages/design-tokens/test/index.test.js`
— one line, in the suite under test. On that commit's run, `Design Tokens Tests` is
`completed/failure` **in the `Run tests` step specifically** (job `98947490111`; `Install
dependencies` succeeded first), while `Store Core Tests` and `Dashboard Tests` are
`success` in the same run. So the red is attributable to the design-tokens assertion, not
to a runner or install-time flake. `db9e5b138` is `completed/success` (job `98948732808`),
and `git diff 5c3ead074..db9e5b138` is **empty** — the revert is exact and nothing of the
mutation survives into the merge candidate. This is the CI-config guard proven the only way
it can be (#7270's lesson), and it was done without being asked twice.

**Byte-consistency with the siblings holds.** Diffed against `claude-hooks-tests` and
`store-core-tests`: identical `actions/checkout@34e1148…` and `actions/setup-node@49933ea…`
pins, `node-version: 22`, `cache: ${{ needs.runner-target.outputs.npmcache }}`,
`cache-dependency-path: '**/package-lock.json'`, `npm ci` with `working-directory: .`,
`npm test` under `defaults.run.working-directory`, and the same trailing
`# SELF_HOSTED_OR_HOSTED (see header)` marker on `runs-on`. The only difference from
`claude-hooks-tests` is the absent `Build @chroxy/protocol` step, which design-tokens
genuinely does not need (its test imports `../src/index.js` and a repo-local hook, nothing
else). `timeout-minutes: 5` matches `claude-hooks-tests` / `app-tests` / `dashboard-typecheck`.

**It clears the repo's own CI-config guards, including the one it could most easily have
tripped.** Ran all three against the branch — 30 tests, 0 failures:
`ci-npm-cache-routing.test.js`, `ci-cache-key.test.js`, `ci-workflow-reader.test.js`. The
#7383 rule ("a new job that hardcodes `cache: npm` silently re-adds a 521 MB per-job
download") is exactly the trap a copy-pasted job falls into, and this one takes the routed
value. All three guards are scan-based with `>=` thresholds rather than hardcoded job
rosters, so adding a job can only help them — nothing needed a count bumped.

**Nothing else in the repo held a roster that needed updating.** There is no aggregate
all-green gate job (the only `needs: [...]` list in the workflow is `renovate-config`'s
`[runner-target, changes]`), so no needs-list edit was missed. `CONTRIBUTING.md`'s check
list at lines 60-66 is explicitly the *required* set and says "Other CI jobs run too", so
it is still accurate with the job not-required — correct to leave alone today.

**The job is not silently skippable.** Parsed all 22 jobs: `design-tokens-tests` has no
`if:`, and its sole dependency `runner-target` is itself unconditional and GitHub-hosted.
It produces its context on every PR (including forks) and on every push to `main` — which
also makes it structurally safe to require later, since a context that is sometimes not
produced is what wedges a PR forever.

**`working-directory` is right, twice over.** The red-proof establishes it end-to-end, and
locally `npm test` in `packages/design-tokens` runs 6 tests in 93 ms. I also checked the
near-miss: because npm resolves scripts against the nearest `package.json` and runs them
from the package root, a `working-directory` pointing *inside* the package (e.g.
`packages/design-tokens/src`) would still run all 6 — and one pointing at the repo root
would fail loudly with `Missing script: test`. Both non-silent.

**YAML is valid** (parses cleanly to 22 jobs) and the comment block follows house style —
issue number, the mechanism, the consequence, placed after `name:` and before `needs:`,
same as `store-core-tests` and `app-tests`. Its factual claims check out: 6 tests, 93 ms.

### Issues Found

#### Critical

| # | File | Issue | Recommendation |
|---|------|-------|----------------|
| — | — | None. | — |

#### Suggestions

| # | File | Issue | Recommendation |
|---|------|-------|----------------|
| S1 | `packages/design-tokens/package.json` (`test` script) | The job now runs the suite — but `npm test` here **exits 0 when it discovers zero test files**, so the check it just created is exposed to the same class it fixes, one level down. Measured on Node v22.22.3 against a scratch copy: unmodified → `# tests 6`, exit 0. `test/index.test.js` renamed to `test/index.spec.js` → still 6, exit 0 (Node's `**/test/**/*.js` pattern catches it — genuinely robust). Moved out of `test/` as `src/tokens.spec.js` → **`# tests 0`, exit 0**. `test/` deleted → **`# tests 0`, exit 0**. The last two are a green CI check with nothing run. `packages/server` already solves this with `scripts/assert-test-count.mjs`. | Out of scope for a CI-wiring PR, so filed as **#7447** (which also covers `packages/claude-hooks`, whose quoted `'./tests/**/*.test.js'` glob likewise exits 0 on zero matches — verified). If you would rather not leave the new check unguarded, a test-count floor on this one script is a small enough fold-in; otherwise #7447 tracks it. |
| S2 | PR body, survey bullet | The `packages/*` half of the survey I verified independently and it holds **exactly**: server / claude-hooks / dashboard / protocol / store-core / app all have jobs, desktop has no `test` script (Rust, covered by `Desktop Rust Tests` ×2), and no `package.json` outside `packages/*` declares a `test` script. But the `scripts/` clause is looser than it reads: `Scripts Tests` names its test files **by hand**, 11 of them, and `scripts/__tests__/` holds 12 — `merge-updater-feeds.test.sh` is not among them, and it guards `scripts/merge-updater-feeds.mjs`, which `release.yml:525` actually runs to build the Tauri updater `latest.json`. It passes today (14/14) — it simply never runs in CI. Already tracked as **#7252**, so not a new gap. | Amend the third bullet to scope "the only gap" to `packages/*` (which is what was surveyed) and reference #7252 for the `scripts/` residue. As written, "design-tokens was the only gap" is the kind of claim a future session re-derives from and stops looking. |
| S3 | `ci.yml:836-838` (`Install dependencies`) | The root `npm ci` is the job's dominant cost and the suite needs nothing from it. Measured on the green run: whole job **18s**, of which `Install dependencies` is **12s** and `Run tests` is sub-second. The suite has zero dependencies — it passes in a checkout with **no `node_modules` at all** (verified). A leaner shape would be `setup-node` with no `cache:` / `cache-dependency-path` and no install, matching `scripts-tests`, which the cache guards already exempt. | Author's call, and uniformity is a real argument the other way — a job that reads like its neighbours is a job the next person copies correctly. Worth noting the failure mode of dropping it is **loud** (`ERR_MODULE_NOT_FOUND` the moment the package gains a dependency), not silent, which is what usually decides this trade-off. Fine to keep as-is. |

#### Nitpicks

| # | File | Issue | Recommendation |
|---|------|-------|----------------|
| N1 | PR body, acceptance criteria | Box 2 ("**Red-proof pending in this PR's own CI history**") is still unticked, and neither run is linked — the proof currently requires `gh api …/check-runs` to find. | Tick it and link the two job URLs (failure: `actions/runs/33200183741/job/98947490111`, success: `actions/runs/33200558894/job/98948732808`). The evidence is the most valuable artifact in this PR; it should be readable from the PR page. |
| N2 | PR body, "Note for the merger" | The not-required note lives only in the PR body, which stops being visible once this merges. | Filed as **#7448** so it outlives the PR. No change needed here. |

### Deferred Items

| Issue | Title | Why deferred |
|-------|-------|--------------|
| [#7447](https://github.com/blamechris/chroxy/issues/7447) | Guard against a zero-test green run in the `node --test` packages (design-tokens, claude-hooks) | S1. Touches two packages' test commands and probably hoists `assert-test-count.mjs` to `scripts/lib/`; that is a different change from wiring a job, and it needs its own red-proof. |
| [#7448](https://github.com/blamechris/chroxy/issues/7448) | Wire "Design Tokens Tests" into required status checks (and update CONTRIBUTING.md's roster) | N2. Branch-protection settings are owner-only and cannot be done from a PR. Carries the `CONTRIBUTING.md` roster update that must land with it, or the required-checks paragraph goes stale. |
| [#7252](https://github.com/blamechris/chroxy/issues/7252) | merge-updater-feeds.test.sh is never run by CI | S2. Pre-existing and already filed; noted here only because this PR's survey is the natural place someone would expect it to have been caught. |

### Architecture Notes

**Concurrency.** The group is `ci-${{ github.head_ref || github.run_id }}` with
`cancel-in-progress: true`, and adding a job changes nothing about it — the group is
per-PR, not per-job. The first commit's run (`5c3ead074`) shows `Design Tokens Tests`,
`Dashboard Tests` and `Store Core Tests` all `cancelled`, which is that supersede working
as designed, not a defect. It does interact with #7448 though: a cancelled run renders as a
failed check, so once this context is *required* it gains the same "cancelled looks red"
behaviour its 11 required siblings already have. That is an argument for making it required
deliberately rather than as a side effect, which #7448 does.

**Discovery, not configuration, is what binds this job to the suite.** The script is a bare
`node --test` with no path argument, so what actually connects the CI job to
`test/index.test.js` is Node's implicit discovery from the cwd. That is robust to a rename
within `test/` and not robust to a move out of it (measured, S1). Worth knowing when
reading the job: the `working-directory` line is doing more load-bearing work than it
looks, because there is no glob beside it to state the intent.

**One more instance of the pattern this repo keeps re-learning.** #7408 was found by asking
"which package's `npm test` does no job invoke". Asking the same question one notch finer —
"which *test file* does no CI step invoke" — is what surfaces `merge-updater-feeds.test.sh`
(#7252), and asking it one notch finer still — "which suite can report zero tests and exit
0" — is #7447. Each level is the same defect wearing a smaller hat, and each was invisible
to the level above. Nothing to change in this PR; it is the reason the survey bullet in S2
is worth scoping honestly rather than closing out.

### Verdict

- [x] **Approve** — ready to merge
- [ ] **Approve with comments** — minor issues, safe to merge
- [ ] **Request changes** — must fix before merge

Checks at head `db9e5b138`: 22 of 23 green, `Server Tests` still running. `Design Tokens
Tests` is `pass`. `mergeStateStatus` is `BLOCKED` pending that check and the usual Copilot
ruleset — re-read it at the current head before escalating.

blamechris:
Review follow-ups:

- **S1** (zero-test green run): agreed and tracked as #7447 — out of scope here, it affects claude-hooks equally.
- **S2** (survey scope): the body's "only gap" claim is now scoped to `packages/*`, with the `scripts/` gap noted as pre-existing #7252.
- **S3** (drop the root `npm ci`): declining — uniformity with the sibling jobs wins, and the failure mode of a missing install is loud (module-not-found), not silent, so there is no false-safety angle. Cost is ~12s on a cached runner.
- **N1**: acceptance box 2 is ticked with both run links (edit made after CI settled — an earlier edit would have cancelled the in-flight run via the concurrency group).
- **N2**: the not-required status is tracked durably in #7448 rather than only in this body.

PR #7453, opened 2026-08-28
fix(store-core): stop replay-end resolving a live AskUserQuestion

### LINKED ISSUE #7420: dashboard/app: a LIVE AskUserQuestion that races a history replay is indistinguishable from a replayed one, and history_replay_end resolves it
## Context

Found while fixing #7410 (PR #7419), which stops `history_replay_end`'s `answered: '(resolved)'` sweep from stamping live **permission** prompts.

That fix is sound for permission prompts because the wire contract decides it: `permission_request` is in the server's `builtinTransient` list (`packages/server/src/session-manager.js:3285`), so it is **never** written to the history ring buffer. A live permission prompt present at replay-end is therefore always one that raced the replay, and `isLivePermissionPrompt` identifies it exactly.

## The remaining case

`user_question` is **not** transient — it *is* in the ring buffer and is re-sent on every replay (see the `#4613` note in `packages/store-core/src/handlers/user-question.ts:155-171`, which exists precisely because replayed question events need their original `timestamp`).

The ChatMessage it builds carries **no `requestId` and no `expiresAt`** (`user-question.ts:172-183`). So:

- a **replayed** AskUserQuestion prompt, and
- a **live** AskUserQuestion that arrived mid-replay

are byte-identical in client state. `isLivePermissionPrompt` returns `false` for both (it requires `requestId` + `expiresAt`), so the sweep stamps both.

## Why the race window is wide

Replay is chunked: `replayHistory` emits 20-entry batches separated by `setImmediate` yields, with back-pressure pauses of up to 30s on fat payloads (`packages/server/src/ws-history.js:17-47`). `reconcileReplayEnd`'s baseline slice deliberately preserves "any racing live" entries (`packages/store-core/src/replay-reconcile.ts:215-218`). So a `user_question` for the replaying session can land mid-replay and be swept at replay-end.

## Impact

Lower than #7410 — an AskUserQuestion has no TTL, and the user can usually re-trigger — but the prompt collapses into a resolved pill and the answer path is lost for that turn.

## Why it isn't fixed in #7419

There is no client-side signal to separate the two. Fixing it needs a wire change, e.g. one of:

1. mark replayed `user_question` frames explicitly (a `replayed: true` / `historySeq` field the live broadcast omits), or
2. give the prompt ChatMessage a liveness token the way permission prompts have `requestId` + `expiresAt`, or
3. have the client record which prompt ids arrived inside the replay window and sweep only those.

(3) is client-only and probably the cheapest, but it needs care: the same `_rebuildBaseline` machinery already tracks the replay window, so the sweep could stamp only entries at or before the baseline rather than filtering by shape.

## Acceptance criteria

- [ ] A live `user_question` arriving during its own session's `history_replay_end` is not stamped `'(resolved)'`
- [ ] A genuinely replayed `user_question` still is
- [ ] Both clients
- [ ] Each test proven red against the pre-fix source

## Related

- #7410 / PR #7419 — the permission-prompt half of this, and the source comment that forward-references this issue
- #7402, #7340, #4909 — the `updateActiveSession` mis-keying family

blamechris:
## The bug

`history_replay_end` sweeps the replayed session's messages and stamps
`answered: '(resolved)'` on every unanswered `type: 'prompt'`, on the premise
that anything in history is already resolved.

#7410 / PR #7419 excluded **permission** prompts by `requestId`, and that
exclusion is exact because the wire contract decides it: `permission_request` is
in the server's `builtinTransient` list (`session-manager.js`), so it is never
written to the history ring buffer — a permission prompt present at replay-end
therefore always *raced* the replay.

`user_question` is **not** transient. It IS in the ring buffer and is re-sent on
every replay (`store-core/handlers/user-question.ts`), *and* it is broadcast
live. The ChatMessage it builds carries **no `requestId` and no `expiresAt`**
(`user-question.ts:172-183`), so:

- a **replayed** AskUserQuestion prompt, and
- a **live** AskUserQuestion that arrived mid-replay

are byte-identical in client state, and the sweep stamped both. The prompt
collapses into a resolved pill and the answer path for that turn is lost.

The race window is wide, not exotic: `replayHistory` emits 20-entry chunks
separated by `setImmediate` yields with back-pressure pauses of up to 30s
(`ws-history.js:17-47`), and `reconcileReplayEnd`'s baseline slice deliberately
preserves any racing live entry (`replay-reconcile.ts`).

## Why option (3)

Options (1) and (2) from the issue both need a wire change (a `replayed: true`
flag, or a server-supplied liveness token on the prompt). (3) is client-only and
needs no protocol version bump, so it is what this PR implements.

The issue's parenthetical suggested sweeping "only entries at or before the
baseline". That does not work, and the code says why: in a full rebuild the
pre-baseline prefix is **discarded** by the atomic swap, and the surviving tail
(`messages.slice(base)`) contains replayed entries *and* racing live ones
interleaved in arrival order. Index position cannot separate them. So the
discriminator has to be recorded at arrival instead — which is still option (3),
using the same module that already tracks the replay window.

## The discriminator, and why it cannot re-break #7419

**`historySeq` presence on the arriving frame.** `replayHistory` surfaces the
server-internal `_seq` as `historySeq` on every entry it replays
(`ws-history.js:1265-1268`), and a live broadcast never carries one. It is not a
new wire field — the same value already drives `recordHistorySeq` — and it is
reliably present: `_pushHistory` stamps `_seq` on append and `setHistory`
re-stamps 1..N on restore, so every ring-buffer entry has one.

`dispatchUserQuestion` (shared by both clients) is the last point at which the
frame is in hand, so it records the ids of questions that arrived **without**
`historySeq` while that session's replay window was open. The sweep skips
exactly those.

Three properties keep #7419 intact:

1. **The `requestId` exclusion is untouched.** It is still the first thing the
   sweep checks, and the new clause only ever *removes* messages from the
   stamped set. A permission prompt cannot become stampable because of this
   change — mutation M2 below confirms the two directions are pinned separately.
2. **The ledger is per-session and per-window.** It is dropped at the next
   `reconcileReplayStart` for that session, so a prompt that pre-dates a replay
   window stays stampable — the long-standing behaviour, unchanged.
3. **The window is tracked for delta replays too.** `_rebuildBaseline` only
   holds full rebuilds; a live question races a delta replay just as easily, so
   `_replayWindowOpen` is its own set.

Also folded in: the sweep predicate itself now has **one implementation**
(`sweepUnansweredPromptsAtReplayEnd` in store-core). Both clients carried
byte-identical copies of it through #7410 and #7419 — the exact drift surface
this repo keeps getting bitten by — and now both just call it.

## Acceptance criteria → tests

| AC | Test |
|----|------|
| A live `user_question` arriving during its own session's `history_replay_end` is not stamped | app + dashboard `does not stamp a live question that arrived mid-replay (delta replay)` and `(full rebuild)`; store-core `skips a prompt that arrived live inside the window…` |
| A genuinely replayed `user_question` still is | app + dashboard `still stamps a question the replay itself delivered`; the full-rebuild test also asserts the replayed prompt IS stamped in the same array; store-core `does NOT record a question the replay delivered (historySeq present)` |
| Both clients | the app and dashboard describes are mirrored case-for-case |
| Each test proven red against the pre-fix source | below |

Extra guards: `still stamps a question that pre-dates the replay window`,
`forgets the previous window: a racer protected once is stamped on the NEXT
replay`, `a racer in session s1 is not protected by session s2's replay window`,
plus store-core unit tests for the ledger (no-op outside a window, closes at
end, per-session, cleared by `resetReplayReconcile`, null-id handling) and for
the sweep's #7410/#7419 exclusions (`never touches a permission prompt`, both
the with-TTL and no-`expiresAt` shapes).

## Red-first evidence

Both client describes were written and run **against the completely unmodified
source** (no store-core change, no client change) before any implementation:

```
app      Tests: 3 failed, 3 passed
dashboard  Tests: 3 failed | 3 passed
```

The three failures per client, identical on both:

- `does not stamp a live question that arrived mid-replay (delta replay)` — `expected '(resolved)' to be undefined`
- `does not stamp a live question that arrived mid-replay (full rebuild)` — same
- `forgets the previous window: a racer protected once is stamped on the NEXT replay` — same (its first half is an AC-(a) assertion)

The three that passed pre-fix are the AC-(b) / unchanged-behaviour pins
(`still stamps a question the replay itself delivered`, `still stamps a question
that pre-dates the replay window`, the cross-session one). They are stated
honestly as such — they are green on both sides of the fix and exist to catch an
*over-broad* fix, and mutation M2 shows they do.

The store-core unit tests cover functions that did not exist pre-fix, so they
carry their red proof by mutation instead.

## Mutation evidence

Each mutation was applied by script with an occurrence-count assert, `grep`-verified
to have landed, and restored from a byte backup with `cp` (absolute paths, sha
verified equal afterwards).

| # | Mutation | Result |
|---|----------|--------|
| M1 | drop the `!live?.has(m.id)` clause from the sweep (revert the fix) | store-core 2 failed / 161 passed · app 3 failed (exactly the pre-fix three) · dashboard 3 failed (same three) |
| M2 | drop the `historySeq` check in the dispatcher — record *every* question as live (an over-broad "never stamp a question" fix) | store-core 1 failed (`does NOT record a question the replay delivered`) · app 2 failed (`still stamps a question the replay itself delivered`, and the full-rebuild test's replayed-prompt assertion) · dashboard the same 2 |
| M3 | stop clearing the ledger at `reconcileReplayStart` | store-core 1 failed (`forgets the previous window at the next start`) · app 1 · dashboard 1 — exactly the paired tests |
| M4 | remove the `_replayWindowOpen` gate in `noteLivePromptDuringReplay` | store-core 4 failed (`no-op outside a replay window`, `closes the window at replay end`, `is per-session`, `resetReplayReconcile drops the window and the ledger`) · app and dashboard **green** |

M4 is worth stating plainly rather than hiding: the client-level "pre-dates the
window" and "s2's window doesn't protect s1" cases survive it, because the
start-time ledger clear independently produces the right answer for both. The
gate is pinned directly at the store-core level, which is where the four
failures land. Two mechanisms, tested separately.

M1 and M2 together are the load-bearing pair: they kill disjoint test sets, so
neither direction of the discriminator is free.

## Test results

| Package | Command | Result |
|---------|---------|--------|
| store-core | `npx vitest run` | **60 files / 2189 tests passed** |
| store-core | `npm run typecheck` | clean |
| app | `npx tsc --noEmit` | clean |
| app | `npx jest src/__tests__/store --coverage=false` | **34 suites / 731 tests passed** |
| app | `npx jest --coverage=false` (full suite, both jest roots) | **181 suites / 2372 tests passed** |
| dashboard | `npx vitest run src/store/` | 79 files / 1158 tests passed, 1 file failed to load |
| dashboard | `npx tsc --noEmit` | 249 errors, **identical count on the pristine tree** (`git stash` comparison) |

The dashboard's load failure and its 249 type errors are one pre-existing
environment gap, not a regression: `@testing-library/react` is declared in
`packages/dashboard/package.json` but absent from the installed
`node_modules`, so 170 test files cannot resolve it and every test callback in
them types as implicit `any`. All 170 failures share that single cause, every
file that *did* load passed (1968 tests, zero test failures), and none of the
type errors are in a file this PR touches. CI installs the dependency.

Related: #7410 / PR #7419 (the permission-prompt half), #7402, #7340, #4909
(the `updateActiveSession` mis-keying family).

Closes #7420

### DIFF
--- a/packages/app/src/__tests__/store/message-handler.test.ts
+++ b/packages/app/src/__tests__/store/message-handler.test.ts
@@ -24,6 +24,7 @@ import {
   PERMISSION_ALREADY_ANSWERED_NOTICE,
   PERMISSION_DECISION_TOKENS,
   isLivePermissionPrompt,
+  resetReplayReconcile,
 } from '@chroxy/store-core';
 import { clearPersistedSession } from '../../store/persistence';
 import { setCallback, clearAllCallbacks } from '../../store/imperative-callbacks';
@@ -832,6 +833,161 @@ describe("history_replay_end: '(resolved)' sweep targeting (#7410)", () => {
     const st = store.getState().sessionStates;
     expect((st.s1.messages[0] as any).answered).toBeUndefined();
     expect((st.s2.messages[0] as any).answered).toBeUndefined();
+  });
+});
+
+// #7420 — the other half of #7410. `permission_request` is server-transient, so
+// a permission prompt present at replay-end always RACED the replay and
+// `!m.requestId` excludes it exactly. `user_question` is NOT transient: it is in
+// the history ring buffer and is re-sent on every replay, and the ChatMessage it
+// builds carries no `requestId` and no `expiresAt`. So a LIVE AskUserQuestion
+// that lands mid-replay is byte-identical in store state to a replayed one, and
+// the pre-fix sweep stamped both — losing the answer path for that turn.
+//
+// The discriminator is the wire frame, recorded at arrival: `replayHistory`
+// stamps every replayed history entry with `historySeq` (ws-history.js) and a
+// live broadcast never carries it. store-core's replay-reconcile module notes
+// the ids of questions that arrived WITHOUT one while the session's replay
+// window was open, and the sweep skips exactly those.
+describe("history_replay_end: '(resolved)' sweep vs a racing live AskUserQuestion (#7420)", () => {
+  beforeEach(() => {
+    // The live-arrival set is module state in store-core; a window left open by
+    // an earlier test would otherwise decide these cases.
+    resetReplayReconcile({ clearCursors: true });
+  });
+
+  afterEach(() => {
+    resetReplayFlags();
+    resetReplayReconcile({ clearCursors: true });
+  });
+
+  function seedOne(messages: any[] = []) {
+    const store = createMockStore({
+      activeSessionId: 's1',
+      sessions: [{ sessionId: 's1', name: 'S1' } as any],
+      sessionStates: { s1: { ...createEmptySessionState(), messages } },
+      sessionNotifications: [],
+    });
+    setStore(store as any);
+    _testMessageHandler.setContext(createMockContext() as any);
+    return store;
+  }
+
+  /**
+   * A `user_question` wire frame. `historySeq` present ⇒ the replay delivered
+   * it; absent ⇒ a live broadcast. That is the ONLY difference between the two
+   * on the wire, which is the whole difficulty of this bug.
+   */
+  const question = (extra: Record<string, unknown> = {}) => ({
+    type: 'user_question',
+    sessionId: 's1',
+    toolUseId: 'q-use-1',
+    questions: [{ question: 'Which approach?', options: [{ label: 'A' }, { label: 'B' }] }],
+    ...extra,
+  });
+
+  const answeredOf = (store: any, i = 0) =>
+    (store.getState().sessionStates.s1.messages[i] as any).answered;
+
+  // --- AC (a): the live racer survives -------------------------------------
+
+  it('does not stamp a live question that arrived mid-replay (delta replay)', () => {
+    const store = seedOne();
+    _testMessageHandler.handle({ type: 'history_replay_start', sessionId: 's1', fullHistory: false });
+    _testMessageHandler.handle(question());
+    // Positive control: the prompt really was appended, so an assertion of
+    // `answered === undefined` below is about the sweep and not about an
+    // absent fixture.
+    expect(store.getState().sessionStates.s1.messages).toHaveLength(1);
+    expect(store.getState().sessionStates.s1.messages[0].type).toBe('prompt');
+
+    _testMessageHandler.handle({ type: 'history_replay_end', sessionId: 's1' });
+
+    expect(answeredOf(store)).toBeUndefined();
+  });
+
+  it('does not stamp a live question that arrived mid-replay (full rebuild)', () => {
+    // A pre-replay message makes the baseline non-zero, so the atomic swap at
+    // replay-end is real (it slices the prefix off) rather than a no-op.
+    const store = seedOne([{ id: 'old-1', type: 'response', content: 'before', timestamp: 1 }]);
+    _testMessageHandler.handle({ type: 'history_replay_start', sessionId: 's1', fullHistory: true });
+    // Replayed first, then the live racer — the interleaving the server's
+    // chunked replay actually produces.
+    _testMessageHandler.handle(question({ historySeq: 3, timestamp: 1000 }));
+    _testMessageHandler.handle(question({ toolUseId: 'q-use-2' }));
+    expect(store.getState().sessionStates.s1.messages).toHaveLength(3);
+
+    _testMessageHandler.handle({ type: 'history_replay_end', sessionId: 's1' });
+
+    const msgs = store.getState().sessionStates.s1.messages;
+    // Swap dropped the pre-replay prefix; both prompts survive it.
+    expect(msgs).toHaveLength(2);
+    expect(answeredOf(store, 0)).toBe('(resolved)');
+    expect(answeredOf(store, 1)).toBeUndefined();
+  });
+
+  // --- AC (b): a genuinely replayed question is still stamped ---------------
+
+  it('still stamps a question the replay itself delivered', () => {
+    const store = seedOne();
+    _testMessageHandler.handle({ type: 'history_replay_start', sessionId: 's1', fullHistory: false });
+    _testMessageHandler.handle(question({ historySeq: 7, timestamp: 1000 }));
+    expect(store.getState().sessionStates.s1.messages).toHaveLength(1);
+
+    _testMessageHandler.handle({ type: 'history_replay_end', sessionId: 's1' });
+
+    expect(answeredOf(store)).toBe('(resolved)');
+  });
+
+  it('still stamps a question that pre-dates the replay window', () => {
+    const store = seedOne();
+    // Live, but BEFORE any replay opened — nothing about this reconnect
+    // vouches for it, so the long-standing sweep behaviour stands.
+    _testMessageHandler.handle(question());
+    _testMessageHandler.handle({ type: 'history_replay_start', sessionId: 's1', fullHistory: false });
+    _testMessageHandler.handle({ type: 'history_replay_end', sessionId: 's1' });
+
+    expect(answeredOf(store)).toBe('(resolved)');
+  });
+
+  it('forgets the previous window: a racer protected once is stamped on the NEXT replay', () => {
+    const store = seedOne();
+    _testMessageHandler.handle({ type: 'history_replay_start', sessionId: 's1', fullHistory: false });
+    _testMessageHandler.handle(question());
+    _testMessageHandler.handle({ type: 'history_replay_end', sessionId: 's1' });
+    expect(answeredOf(store)).toBeUndefined();
+
+    _testMessageHandler.handle({ type: 'history_replay_start', sessionId: 's1', fullHistory: false });
+    _testMessageHandler.handle({ type: 'history_replay_end', sessionId: 's1' });
+
+    expect(answeredOf(store)).toBe('(resolved)');
+  });
+
+  it("a racer in session s1 is not protected by session s2's replay window", () => {
+    const store = createMockStore({
+      activeSessionId: 's1',
+      sessions: [
+        { sessionId: 's1', name: 'S1' } as any,
+        { sessionId: 's2', name: 'S2' } as any,
+      ],
+      sessionStates: {
+        s1: createEmptySessionState(),
+        s2: createEmptySessionState(),
+      },
+      sessionNotifications: [],
+    });
+    setStore(store as any);
+    _testMessageHandler.setContext(createMockContext() as any);
+
+    // Only s2 is replaying; a live question for s1 is not inside any window of
+    // its own, so s1's own replay-end still stamps it.
+    _testMessageHandler.handle({ type: 'history_replay_start', sessionId: 's2', fullHistory: false });
+    _testMessageHandler.handle(question());
+    _testMessageHandler.handle({ type: 'history_replay_end', sessionId: 's2' });
+    _testMessageHandler.handle({ type: 'history_replay_start', sessionId: 's1', fullHistory: false });
+    _testMessageHandler.handle({ type: 'history_replay_end', sessionId: 's1' });
+
+    expect((store.getState().sessionStates.s1.messages[0] as any).answered).toBe('(resolved)');
   });
 });
 
--- a/packages/app/src/store/message-handler.ts
+++ b/packages/app/src/store/message-handler.ts
@@ -65,6 +65,11 @@ import {
   recordHistorySeq,
   reconcileReplayStart,
   reconcileReplayEnd,
+  // #7420 — the shared `history_replay_end` unanswered-prompt sweep. Both
+  // clients carried byte-identical copies of the predicate through #7410/#7419;
+  // it now has one implementation, which is also where the live-arrival ledger
+  // that protects a racing AskUserQuestion is read.
+  sweepUnansweredPromptsAtReplayEnd,
   replayDedupCache,
   resetReplayReconcile,
   handlePermissionRequest as sharedPermissionRequest,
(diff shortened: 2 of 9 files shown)


blamechris:
## Code Review: PR #7453

### Summary

Adversarial review of `fix(store-core): stop replay-end resolving a live AskUserQuestion` at head `33e7c7323`, focused on the one thing the whole fix rests on: **is `historySeq` presence really a sound live-vs-replayed discriminator?**

The fix is well-built. The predicate unification is provably behaviour-preserving, the #7419 exclusion is structurally safe, the red-first evidence reproduces exactly, and the tests carry positive controls rather than bare negative assertions. I re-ran mutation **M1** independently and it lands precisely as the PR body claims.

The discriminator, however, is sound for `replayHistory` — not for *every* replay. `handleRequestFullHistory` is a second replay path that brackets its own `history_replay_start`/`history_replay_end` and forwards ring-buffer entries raw, with no `_seq` → `historySeq` mapping. Every question it replays is now misclassified as a live racer. The failure direction is the benign one, so this is a follow-up (#7454) rather than a blocker, but it is the honest limit of the guard and worth a pointer in the code.

Three further follow-ups filed: a re-entrancy hole (#7455), a ledger-lifetime asymmetry (#7456), and the non-racing half of the same bug (#7457).

### Verification performed

| Check | Result |
|---|---|
| **M1 reproduced** — dropped `&& !(live?.has(m.id) === true)` from `sweepUnansweredPromptsAtReplayEnd` (1 occurrence asserted, grep-verified landed, restored from a byte backup, sha equal) | store-core **2 failed / 161 passed**; app **3 failed / 214 passed**; dashboard **3 failed / 309 passed** — the *same three named tests* per client, and all three AC-(b) pins stayed **green**. Matches the PR body exactly. |
| Post-restore baseline | store-core 163 ✅ · app 217 ✅ · dashboard 312 ✅ · `store-core npx tsc --noEmit` clean |
| Old-copy diff (unification safety) | app `:2605-2606` and dashboard `:5300-5301` at `8cf5dd01e` are **byte-identical**; `REPLAY_RESOLVED_PLACEHOLDER === '(resolved)'` matches what both wrote |
| Server `user_question` emit paths | 3 exist: live broadcast (no `historySeq` ✅), `replayHistory` (`historySeq` ✅), `handleRequestFullHistory` (**no `historySeq`** ⛔ — see C1) |
| `_seq` stamping | no append path skips it (`session-message-history.js:416-419`); `setHistory` re-stamps 1..N (`:174-179`) |
| Client parse path | no Zod parse, no field strip — bare `JSON.parse` → `runDispatch` (app `connection.ts:1348`, dashboard `:2720`) |

### Strengths

- **Red-first evidence is honest and independently confirmed.** M1's kill set is exactly the three named pre-fix failures on both clients, and the three AC-(b) pins stay green under it — so M1 and M2 really do kill disjoint sets and neither direction of the discriminator is free.
- **Negative assertions have positive controls.** Every `toBeUndefined()` is preceded by a length + `type` assertion that the fixture actually landed, and the store-core ledger test asserts `wasPromptLiveDuringReplay(...) === true` before asserting the sweep skipped it. This is the repo's own rule, followed without being asked.
- **#7419 cannot regress structurally, not just empirically.** The new clause is a conjunct appended *after* `!m.requestId` (`replay-reconcile.ts:366-367`), so it can only ever shrink the stamped set. The `#7410`/`#7419` describes still exist and pass in both clients.
- **The discriminator is corroborated by already-shipped code.** `recordHistorySeq(targetId, (msg as {...}).historySeq)` at app `:2610` / dashboard `:5325` already treats `historySeq` as the replay marker inside the same window — the #5555.3 cursor would corrupt if a live frame carried one.
- **Choosing `_replayWindowOpen` over `_rebuildBaseline` is the right call and is pinned.** `_rebuildBaseline` holds full rebuilds only; a live question races a delta replay just as easily, and there is a dedicated test for it.
- **The unification removes a genuine drift surface** — the exact failure class this repo keeps re-encountering — and `reconcileReplayEnd` closing the window *before* the delta early-return (`:262` above `:263`) is a detail that would have been easy to get wrong.
- **StrictMode-safe.** `handleMessage` runs from `socket.onmessage` (dashboard `connection.ts:2757`), not a React effect, and the ledger value is a `Set`, so a repeated record is idempotent. No new double-invoke hazard.

### Critical

| File | Line | Finding |
|---|---|---|
| `packages/server/src/handlers/conversation-handlers.js` | 258-273 | **`request_full_history` is a second replay path that does not stamp `historySeq`, so #7453 misclassifies everything it replays as live.** It sends its own `history_replay_start` (`:258`), then for a `user_question` entry takes the else-branch at `:270` — `ctx.transport.send(ws, { ...entry, sessionId: targetId })` — forwarding the ring-buffer entry **raw**, with no `_seq` → `historySeq` mapping (contrast `ws-history.js:1267-1268`, which does exactly that), then `history_replay_end` (`:273`). The ring buffer really is reached: `getFullHistoryAsync` falls back to it whenever `resumeSessionId` is null (`session-manager.js:2959-2971`) — every non-Claude-CLI provider and any Claude session before its first `init` — and the JSONL reader can only produce `user_input`/`response`/`tool_use`, so the ring-buffer branch is the *sole* source of `user_question` here. The path is user-triggered ("Sync Full History", `packages/app/src/screens/SettingsScreen.tsx:541`). Post-merge, every question it replays lands in `dispatchUserQuestion` without `historySeq`, gets recorded as a live racer, and the sweep skips it — so long-resolved prompts render as answerable pills where they used to read `(resolved)`. **The direction is the safe one** (a stale prompt stays interactive rather than a live one being destroyed), which is why I am not blocking on it — but it does mean "every replayed entry carries `historySeq`" is not a global statement, and the guard is only as good as that statement. Line `:270` also leaks the internal `_seq` onto the wire, which `replayHistory` strips deliberately. → **#7454** (also covers `ServerUserQuestionSchema`, below). Worth folding a one-line pointer into `dispatch-table.ts:1993` so the next reader does not have to re-derive this. |

### Suggestions

| File | Line | Finding |
|---|---|---|
| `packages/store-core/src/replay-reconcile.ts` | 207-208, 262, 312 | **Overlapping replays for one session wipe the ledger and close the window early.** `_replayWindowOpen` is a `Set`, not a refcount, and `reconcileReplayStart` *actively discards* the previous window's evidence (`_liveDuringReplay.delete(sessionId)`). The server can interleave two `forceFull` replays of the same id — `subscribe_sessions` (`session-handlers.js:469`) replays a background session while `switch_session` (`:78`) fires a second replay **unconditionally**, including for an already-subscribed id, and `replayHistory` chunks over `setImmediate` with back-pressure pauses. Wire order `start,start,end,end` then means: `start#2` wipes what `start#1` recorded; `end#1` closes the window while replay #2 is still streaming; a question arriving before `end#2` fails the `:312` gate and `end#2`'s sweep stamps it. Not a regression (pre-fix *all* such questions were stamped) — a hole in the new protection. **This is also the honest answer to M4:** the `_replayWindowOpen` gate is not unobservable client-side, it is observable *exactly here*, and here it is the thing that loses the question — removing the gate would protect it. The fix is neither keeping nor removing the gate but making the window a refcount, clearing the ledger only on the 0→1 transition. → **#7455** |
| `packages/store-core/src/replay-reconcile.ts` | 106-109, 112 | **`_liveDuringReplay` is the only per-session collection in this file with no release on the closing frame, no cap, and no per-session teardown.** `_rebuildBaseline` is deleted at `:267`, `_replayWindowOpen` at `:262`, `_historyCursors` is capped at 64. The ledger is released only by the *next* start for that id or a whole-module `resetReplayReconcile` — which in production is called only from `auth_ok` (app `:1863`, dashboard `:4295`); notably **neither `socket.onclose` nor `socket.onerror` calls it**. The stated bounding argument at `:106-109` ("a subset of the ChatMessages the store is already holding") does not hold once the session is pruned or times out: both clients drop `sessionStates[id]` and touch no store-core state, and there is no per-session teardown export to call. Related: `dispatch-table.ts:1995` records *before* the `hasSession` check on the next line, so a session the store holds no state for can still get an entry. Slow leak, not a runaway — but the comment should say what actually bounds it. → **#7456** |
| `packages/protocol/src/schemas/server/stream.ts` | 652-656 | **`ServerUserQuestionSchema` is strip-by-default and does not declare `historySeq`.** A plain `z.object`, not `.passthrough()`. Nothing parses outbound frames through it today (`validateOutbound` in `ws-outbound-schemas.js:193-208` has zero call sites in `packages/server/src`), so this is latent — but if outbound validation is ever wired up it silently deletes the discriminator and #7420 regresses with every test still green. That is the false-safety shape in `docs/false-safety-guards.md`: success and not-checking become the same observable outcome. `historySeq: z.number().optional()` is cheap insurance. → folded into **#7454** |

### Nitpicks

| File | Line | Finding |
|---|---|---|
| `packages/store-core/src/dispatch-table.ts` | 934 | `historySeq?: number` on `DispatchMessageMap['user_question']` is **inert**. The dispatcher reads it through its own `(msg as { historySeq?: unknown })` cast at `:1993`, and the test helper `dispatch()` takes `Record<string, unknown>` (`dispatch-table.test.ts:248`) — which is why `:445` needs a needless `as never`. Deleting the type field breaks nothing. It is good documentation, but nothing depends on it; dropping the `as never` would at least make the map type load-bearing for that one fixture. |
| `packages/store-core/src/index.ts` | 77 | `wasPromptLiveDuringReplay` is exported from the package barrel but has **no production consumer** — grep finds it only in `replay-reconcile.test.ts` and `dispatch-table.test.ts`. Test-only public surface. |
| `packages/server/src/ws-history.js` | 1268 | The `historySeq` attach is conditional (`typeof _seq === 'number' ? { historySeq: _seq } : {}`) and *omits the key* rather than setting `undefined` — so a `_seq`-less entry would arrive looking exactly like a live frame. No reachable append path produces one today, so latent; noting it because the invariant now rests on `_pushHistory`/`setHistory` discipline rather than on the replay code enforcing it. |
| `packages/server/src/handlers/conversation-handlers.js` | 270 | Same line as C1: `{ ...entry, sessionId }` also puts the server-internal `_seq` on the wire. `getHistory` returns the live array untouched (`session-message-history.js:140-142`). |
| `packages/app/src/__tests__/store/message-handler.test.ts`, `packages/dashboard/src/store/message-handler.test.ts` | 856-861, 3793-3797 | The new `#7420` describes call `resetReplayReconcile` **describe-scoped**, not file-global, in both large test files. Benign today because they clean up in `afterEach`, but module state in these files is otherwise unguarded — a future describe added between them inherits whatever the previous one left. |

### Deferred Items

- **#7457 — a genuinely PENDING AskUserQuestion is stamped `(resolved)` on any ordinary reconnect.** Filed as the non-racing half of the same bug, and arguably the more common user-facing symptom. `user_question` is not in `builtinTransient`, so it is replayed from the ring buffer **with** `historySeq` — correctly classified as "delivered by the replay", and therefore swept — and there is no `resendPendingQuestions` sibling to `resendPendingPermissions` (`ws-permissions.js:769-861`), so the replay is the only way it reaches a reconnecting client. The agent stays blocked while the UI shows a resolved pill; on a full rebuild the atomic swap can remove the prompt entirely. Correctly out of scope for #7453, which cannot fix it client-side for the same reason #7420 needed an arrival-time record.
- **#7455 / #7456** as above — both are lifecycle hardening on state this PR introduces, neither is a regression.
- The E2E harness `handshake-e2e/fake-ws.ts` drives `reconcileReplayStart`/`End` (`:417`, `:443`) but never `noteLivePromptDuringReplay` or `sweepUnansweredPromptsAtReplayEnd`, so the handshake scenarios exercise the replay lifecycle with the #7420 logic absent. Not worth widening this PR for; worth knowing that regressions there are invisible to those tests.

### Architecture Notes

**Recording at arrival is the right call, and the PR's rejection of the issue's own suggestion is correct.** The parenthetical "sweep only entries at or before the baseline" cannot work: in a full rebuild the pre-baseline prefix is discarded by the atomic swap, and the surviving tail (`messages.slice(base)`) interleaves replayed and racing-live entries in arrival order. Index position genuinely cannot separate them, and the PR says so with the code to back it.

**On M4.** The PR reports M4's client-level survival openly, and attributes it to the start-time ledger clear. That attribution is right for the well-formed protocol: the only interval between a ledger clear and a sweep is exactly `reconcileReplayStart` → `reconcileReplayEnd`, and the sweep runs synchronously on the next statement, leaving no gap in which a question can arrive with the window closed. So under `start → end` the gate is genuinely unobservable client-side. It becomes observable under `start,start,end,end` (#7455) — and there the gate is the thing that *loses* the question. That is a more useful reading than "the gate is redundant", and it argues for the refcount rather than for deleting the gate.

**Where the `historySeq` premise is load-bearing elsewhere.** #5555.3's cursor advance (`recordHistorySeq` inside the replay-set gate) would already be corrupted if a live broadcast carried a seq, so this PR is not introducing a new assumption — it is taking a dependency on an existing one. That is the strongest argument for the design, and also the reason C1 matters: an existing assumption that was only *approximately* true is now load-bearing for correctness rather than for a cursor optimisation.

**A note on how C1 got past the tests.** The store-core, app and dashboard suites all model replay through `replayHistory`'s frame shape, which is the shape that *does* carry `historySeq`. The second replay path lives in a different package with a different frame builder, so no client-side test could have caught it — the same "testing the source tree when the artifact is what ships" flavour catalogued in `docs/false-safety-guards.md`. The mitigation is the server-side test in #7454, not more client tests.

### Verdict

- [ ] Approve
- [x] Approve with comments
- [ ] Request changes

The fix is correct for the race it targets, honestly evidenced, and lower-risk than the code it replaces. C1 is a real limit on the discriminator and a real (benign-direction) behaviour change on a shipped path, but the remedy is server-side and outside this PR's stated client-only scope; it is tracked in #7454 along with the schema hardening. The only thing I would fold in here is a one-line comment at `dispatch-table.ts:1993` naming `handleRequestFullHistory` as the known exception, so the invariant's boundary is written down where the next reader will look for it.

Follow-ups filed: #7454 (server `historySeq` gap + protocol schema), #7455 (overlapping-replay re-entrancy), #7456 (ledger lifetime), #7457 (pending question on reconnect).

blamechris:
Review follow-ups: the one open decision is taken — the `dispatch-table.ts` discriminator now carries a KNOWN GAP pointer to #7454 (611a35ee6, comment-only, typecheck clean), so the next reader of the discriminator learns about the `request_full_history` path where it lives rather than from the tracker. Everything else from the review is tracked: #7454 (server `historySeq` on the second replay path + schema hardening), #7455 (refcount the replay window — also the honest home of the M4 client-level survivor), #7456 (ledger lifetime/cap), #7457 (the non-racing reconnect half). The critical's failure direction is benign (unstamped, never falsely resolved), per the review.

Issue #7457, opened 2026-08-28
A genuinely PENDING AskUserQuestion is stamped '(resolved)' on any ordinary reconnect — the non-racing half of #7420


blamechris:
Found while reviewing PR #7453 (#7420). Out of scope for that PR, which fixes the *racing* half.

## The gap

`history_replay_end` stamps `answered: "(resolved)"` on every unanswered `type: "prompt"` it is entitled to touch, on the premise that anything in history is already resolved. #7453 carves out a question that arrived **live during** the replay. It deliberately does not carve out the case where the question is genuinely still pending and the replay itself re-delivers it — and it cannot, because that question is indistinguishable from a resolved one by the discriminator: it comes from the ring buffer, so it carries `historySeq`, so it is correctly classified as "delivered by the replay" and swept.

## Why it is reachable, and common

`user_question` is NOT in the server's `builtinTransient` list, so it IS written to the history ring buffer. There is no `resendPendingQuestions` sibling to `resendPendingPermissions` (`packages/server/src/ws-permissions.js:769-861` handles permissions only, and is what #7410/#7419 keyed the `requestId` exclusion off). So the only way a pending question reaches a reconnecting client is the ordinary history replay — the exact path the sweep then resolves.

The everyday sequence:

1. The agent calls `AskUserQuestion`; the tool call blocks on the resolver in the daemon.
2. The user backgrounds the phone / closes the laptop; the WebSocket drops. The daemon and the pending resolver survive.
3. The user reopens; the client reconnects, `auth_ok` → `replayHistory` re-sends the question with `historySeq`.
4. `history_replay_end` fires and stamps it `(resolved)`.

The question renders as a resolved pill. The user cannot answer it, and the agent stays blocked until the question times out.

On a **full rebuild** it is worse: `reconcileReplayEnd`'s atomic swap (`messages.slice(base)`) also discards any pre-replay copy, so the pending prompt can vanish from the transcript entirely.

## Why this is the harder half

The client cannot fix this alone. A replayed-because-still-pending question and a replayed-because-long-answered question are byte-identical on the wire — the same reason #7420 could not be solved by inspecting store state. It needs the server to say which pending questions are still awaiting an answer.

## Proposed direction

Mirror what already exists for permissions:

1. Track pending `AskUserQuestion` calls per session (keyed by `toolUseId`) alongside `_pendingPermissions`.
[...]

Issue #7456, opened 2026-08-28
store-core: _liveDuringReplay has no release at replay end, no cap, and no per-session teardown


blamechris:
Found while reviewing PR #7453 (#7420). Low severity — a slow leak and an inaccurate comment, not a behavioural bug.

## The asymmetry

`packages/store-core/src/replay-reconcile.ts` holds four per-session collections. Three of them are released; the new ledger is not:

| Collection | Added | Released at `reconcileReplayEnd`? | Cap |
|---|---|---|---|
| `_historyCursors` (`:73`) | per applied entry | n/a | yes — `MAX_CLIENT_HISTORY_CURSORS = 64`, LRU evict |
| `_rebuildBaseline` (`:58`) | full-rebuild start | **yes** (`:267`, and `:214` at a delta start) | no |
| `_replayWindowOpen` (`:84`) | every start | **yes** (`:262`) | no |
| `_liveDuringReplay` (`:112`) | `noteLivePromptDuringReplay` (`:313-318`) | **no** | no |

`_liveDuringReplay` is released only by the *next* `reconcileReplayStart` for that same session id, or by a whole-module `resetReplayReconcile`. That is deliberate — `sweepUnansweredPromptsAtReplayEnd` reads it immediately after `reconcileReplayEnd` returns, so it cannot be cleared at `:262`. But it means a session that has its last replay and is then pruned keeps its `Set` forever.

## The bounding argument in the comment is wrong

`replay-reconcile.ts:106-109`:

> That also bounds it — the entries are a subset of the ChatMessages the store is already holding for the same window — so it needs no cap of its own

The store drops a session's `messages` wholesale on prune (`session_list` reconcile) and on `session_timeout`; both clean `sessionStates` and call `clearPersistedSession`, and neither touches any store-core module state. After that the ledger entries are a subset of nothing.

There is also no per-session teardown API to call: `packages/store-core/src/index.ts` exports only the whole-module `resetReplayReconcile`, and there is no `forgetSession`/`dropSession` anywhere in `packages/store-core/src`.

Additionally, `dispatch-table.ts:1995` records into the ledger *before* the `adapter.hasSession(sessionId)` check on the following line, so a `user_question` for a session the store holds no state for can still write an entry (the window is opened unconditionally for any id — both clients call `reconcileReplayStart` with no existence gate). That widens the key set past the very assumption the comment leans on.

## Rate

[...]

Issue #7455, opened 2026-08-28
store-core: overlapping replays for one session wipe the #7420 live-question ledger and close the window early


blamechris:
Found while reviewing PR #7453 (#7420).

## The shape

#7453 tracks a per-session replay window and a per-session ledger of questions that arrived live inside it (`packages/store-core/src/replay-reconcile.ts:84`, `:112`). Both are keyed by `sessionId` with **membership**, not a refcount:

```ts
// replay-reconcile.ts:207-208 — reconcileReplayStart
_replayWindowOpen.add(sessionId)
_liveDuringReplay.delete(sessionId)          // actively discards the previous window's evidence

// replay-reconcile.ts:262 — reconcileReplayEnd
if (sessionId) _replayWindowOpen.delete(sessionId)
```

That is correct for a strictly serial `start(X) … end(X)` sequence, and every test in the PR uses one.

## The server can interleave two replays of the same session

Both callers force a full rebuild and neither checks whether one is already in flight:

- `packages/server/src/handlers/session-handlers.js:469` — `subscribe_sessions` replays each newly-subscribed background session. Both clients fire `subscribe_sessions` for every non-active session on each `session_list`.
- `packages/server/src/handlers/session-handlers.js:78` — `switch_session` calls `replayHistory(ws, targetId, { forceFull: true })` **unconditionally**, including for a session the client is already subscribed to.

`replayHistory` chunks 20 entries at a time over `setImmediate`, with back-pressure pauses (`ws-history.js` `sendChunk` / `scheduleAfterDrain`), so replay #1 for session X can still be streaming when the user taps into X and starts replay #2. Wire order: `start(X) … start(X) … end(X) … end(X)`.

## What goes wrong

1. `start#2` runs `_liveDuringReplay.delete(X)` — destroying every id recorded during `start#1`.
2. `end#1` runs `_replayWindowOpen.delete(X)` while replay #2 is still streaming. The window is now closed even though a replay is in flight.
3. Any live `user_question` arriving between `end#1` and `end#2` hits the `if (!_replayWindowOpen.has(sessionId)) return` gate at `replay-reconcile.ts:312` and is never recorded.
4. `end#2` runs the sweep with an empty ledger and stamps `answered: "(resolved)"` on it — the exact bug #7420 set out to fix.

[...]

PR #7458, opened 2026-08-28
fix(server): map _seq to historySeq on the request_full_history replay path (#7454)

### LINKED ISSUE #7454: server: request_full_history replays user_question WITHOUT historySeq, disarming the #7420 live/replay discriminator
Found while reviewing PR #7453 (#7420).

## The invariant PR #7453 rests on

#7453 tells a LIVE `AskUserQuestion` apart from a REPLAYED one by the presence of `historySeq` on the arriving frame, recorded at arrival in `packages/store-core/src/dispatch-table.ts:1993-1995`:

```ts
const seq = (msg as { historySeq?: unknown }).historySeq
const deliveredByReplay = typeof seq === "number" && Number.isFinite(seq)
if (!deliveredByReplay) noteLivePromptDuringReplay(sessionId, chatMessage.id)
```

That is exactly right for `replayHistory`, which maps the server-internal `_seq` onto the wire at `packages/server/src/ws-history.js:1267-1268`:

```js
const { _seq, ...wireEntry } = entry
send(ws, { ...wireEntry, sessionId, ...(typeof _seq === "number" ? { historySeq: _seq } : {}) })
```

## The path that does not

`handleRequestFullHistory` is a **second** replay path. It brackets its own `history_replay_start` / `history_replay_end` and forwards ring-buffer entries verbatim — `packages/server/src/handlers/conversation-handlers.js:258-273`:

```js
ctx.transport.send(ws, { type: "history_replay_start", sessionId: targetId, fullHistory: true })
for (const entry of fullHistory) {
  if (entry.type === "user_input" || entry.type === "response" || entry.type === "tool_use") {
    ctx.transport.send(ws, { type: "message", messageType: entry.type, ... })
  } else {
    ctx.transport.send(ws, { ...entry, sessionId: targetId })   // <-- line 270
  }
}
ctx.transport.send(ws, { type: "history_replay_end", sessionId: targetId })
```

A ring-buffer `user_question` entry has `type: "user_question"`, so it takes the else-branch at line 270 and goes out **raw**: no `_seq` → `historySeq` mapping.

The ring buffer really is reached here — `getFullHistoryAsync` falls back to it whenever `resumeSessionId` is null (`packages/server/src/session-manager.js:2959-2971`), which is every non-Claude-CLI provider (BYOK, Codex, Gemini, user-shell) and any Claude session before its first `init`. And the JSONL reader can only ever produce `user_input` / `response` / `tool_use`, so the ring-buffer branch is the sole source of `user_question` on this path.

## Effect

`request_full_history` is user-triggered (the app's "Sync Full History" button, `packages/app/src/screens/SettingsScreen.tsx:541`; also `packages/dashboard/src/store/connection.ts:5102-5109`). Post-#7453, every question it replays is misclassified as a live racer, so `sweepUnansweredPromptsAtReplayEnd` skips all of them and long-resolved AskUserQuestion prompts render as answerable pills instead of `(resolved)`.

The direction is the benign one — a stale prompt stays interactive rather than a live prompt being destroyed — which is why this is a follow-up and not a blocker on #7453. But it does falsify "every replayed entry carries `historySeq`" as a global statement, and the guard is only as good as that statement.

Line 270 also leaks the server-internal `_seq` onto the wire; `replayHistory` strips it deliberately (`ws-history.js:1267`).

## Also: the protocol schema would strip the field

`ServerUserQuestionSchema` (`packages/protocol/src/schemas/server/stream.ts:652-656`) is a plain `z.object`, i.e. strip-by-default, and does not declare `historySeq`:

```ts
export const ServerUserQuestionSchema = z.object({
  type: z.literal("user_question"),
  toolUseId: z.string(),
  questions: z.array(z.any()),
})
```

Nothing parses outbound frames through it today — `validateOutbound` (`packages/server/src/ws-outbound-schemas.js:193-208`) has zero call sites in `packages/server/src` — so this is latent, not live. But if outbound validation is ever wired up, it silently deletes the discriminator and #7420 regresses with every test still green. That is the false-safety shape catalogued in `docs/false-safety-guards.md`.

## Proposed fix

1. `conversation-handlers.js:270` — mirror `ws-history.js:1268`: destructure `_seq` off the entry and re-attach it as `historySeq` when numeric. Fixes the misclassification and the `_seq` wire leak in one line.
2. `packages/protocol/src/schemas/server/stream.ts:652` — add `historySeq: z.number().optional()` (or `.passthrough()`) so a future strict parse cannot disarm the guard.

## Acceptance criteria

- A `user_question` replayed by `request_full_history` arrives with `historySeq` and is stamped `(resolved)` by the replay-end sweep, exactly as one replayed by `replayHistory` is.
- No `_seq` key appears on any frame `handleRequestFullHistory` sends.
- `ServerUserQuestionSchema.parse()` on a replayed frame preserves `historySeq`.
- A server test drives `request_full_history` against a session whose ring buffer holds a `user_question` and asserts `historySeq` on the emitted frame — proven red against current `main` first.

Related: #7420, #7410, PR #7453

blamechris:
Closes #7454 (found in review of PR #7453 / #7420).

`handleRequestFullHistory` brackets its own `history_replay_start`/`history_replay_end` but forwarded ring-buffer entries **raw** — no `_seq` → `historySeq` mapping. On current `main` this means the raw internal counter leaked and the #5555.3 delta-replay cursor (`recordHistorySeq`) never advanced on this path; once PR #7453 lands, it would additionally have made every replayed `user_question` read as a **live racer** to #7420's discriminator (which keys on `historySeq` presence) on every provider whose `getFullHistoryAsync` falls back to the ring buffer — this PR pre-arms that guard so the pair is safe in either merge order (review S3 corrected my original present-tense phrasing). The raw internal `_seq` counter also leaked onto the wire on that branch.

## What changed

- The else-branch now maps `_seq` → `historySeq` **exactly as `replayHistory` does** (destructure off, attach when numeric) and never sends the raw counter.
- `ServerUserQuestionSchema` gains the optional `historySeq` field with a comment naming the guard it protects — a future strip-on-parse consumer can no longer silently disarm #7420 with all tests green (the review's schema-hardening ask). Protocol dist rebuilt and committed.

## Verification

- Red-first: the new test file's main assertion failed against unmodified source (`historySeq` undefined, `_seq` leaking).
- Mutation: reverting the mapping to the raw spread fails exactly the paired test; restored, 3/3.
- Positive control: an entry with **no** `_seq` goes out with **no** `historySeq` key — absence is the live signal and must never be fabricated.
- 340/340 across the new file + `ws-history.test.js` + `ws-schemas.test.js`; all server lints pass.

## Not in scope

#7455 (replay-window refcount), #7456 (ledger lifetime), #7457 (non-racing reconnect half) — tracked separately from the same review.

### DIFF
--- a/packages/protocol/dist/schemas/server/stream.d.ts
+++ b/packages/protocol/dist/schemas/server/stream.d.ts
@@ -405,6 +405,7 @@ export declare const ServerUserQuestionSchema: z.ZodObject<{
     type: z.ZodLiteral<"user_question">;
     toolUseId: z.ZodString;
     questions: z.ZodArray<z.ZodAny>;
+    historySeq: z.ZodOptional<z.ZodNumber>;
 }, z.core.$strip>;
 export declare const ServerAgentBusySchema: z.ZodObject<{
     type: z.ZodLiteral<"agent_busy">;
--- a/packages/protocol/dist/schemas/server/stream.d.ts
+++ b/packages/protocol/dist/schemas/server/stream.d.ts
@@ -105,6 +105,7 @@ export declare const ServerMessageSchema: z.ZodObject<{
     attemptedResumeId: z.ZodOptional<z.ZodString>;
     stdout: z.ZodOptional<z.ZodString>;
     stderr: z.ZodOptional<z.ZodString>;
+    historySeq: z.ZodOptional<z.ZodNumber>;
 }, z.core.$strip>;
 export declare const ServerToolStartSchema: z.ZodObject<{
     type: z.ZodLiteral<"tool_start">;
@@ -113,6 +114,7 @@ export declare const ServerToolStartSchema: z.ZodObject<{
     tool: z.ZodString;
     input: z.ZodAny;
     serverName: z.ZodOptional<z.ZodString>;
+    historySeq: z.ZodOptional<z.ZodNumber>;
 }, z.core.$strip>;
 export declare const ServerToolResultSchema: z.ZodObject<{
     type: z.ZodLiteral<"tool_result">;
@@ -120,6 +122,7 @@ export declare const ServerToolResultSchema: z.ZodObject<{
     result: z.ZodAny;
     truncated: z.ZodOptional<z.ZodBoolean>;
     isError: z.ZodOptional<z.ZodBoolean>;
+    historySeq: z.ZodOptional<z.ZodNumber>;
 }, z.core.$strip>;
 export declare const ServerToolInputDeltaSchema: z.ZodObject<{
     type: z.ZodLiteral<"tool_input_delta">;
@@ -186,6 +189,7 @@ export declare const ServerResultSchema: z.ZodObject<{
             "final-round-prompt": "final-round-prompt";
         }>>;
     }, z.core.$strip>>>;
+    historySeq: z.ZodOptional<z.ZodNumber>;
 }, z.core.$strip>;
 export declare const ServerModelChangedSchema: z.ZodObject<{
     type: z.ZodLiteral<"model_changed">;
--- a/packages/protocol/dist/schemas/server/stream.js
+++ b/packages/protocol/dist/schemas/server/stream.js
@@ -185,6 +185,11 @@ export const ServerMessageSchema = z.object({
     // shape-compatible.
     stdout: z.string().max(8192).optional(),
     stderr: z.string().max(8192).optional(),
+    // #7454/#7458: present on REPLAYED frames only (both replay paths map the
+    // server-internal `_seq` onto the wire; absent on live broadcasts). The
+    // #5555.3 delta-replay cursor — and for user_question the #7420
+    // live-vs-replayed guard — key on this field's presence.
+    historySeq: z.number().optional(),
 });
 export const ServerToolStartSchema = z.object({
     type: z.literal('tool_start'),
@@ -193,6 +198,11 @@ export const ServerToolStartSchema = z.object({
     tool: z.string(),
     input: z.any(),
     serverName: z.string().optional(),
+    // #7454/#7458: present on REPLAYED frames only (both replay paths map the
+    // server-internal `_seq` onto the wire; absent on live broadcasts). The
+    // #5555.3 delta-replay cursor — and for user_question the #7420
+    // live-vs-replayed guard — key on this field's presence.
+    historySeq: z.number().optional(),
 });
 export const ServerToolResultSchema = z.object({
     type: z.literal('tool_result'),
@@ -205,6 +215,11 @@ export const ServerToolResultSchema = z.object({
     // missing value as false. Older servers and non-mcp tools omit it; codex
     // mcpToolCall emits it explicitly (`false` on success).
     isError: z.boolean().optional(),
+    // #7454/#7458: present on REPLAYED frames only (both replay paths map the
+    // server-internal `_seq` onto the wire; absent on live broadcasts). The
+    // #5555.3 delta-replay cursor — and for user_question the #7420
+    // live-vs-replayed guard — key on this field's presence.
+    historySeq: z.number().optional(),
 });
 // #4080 / #4081: incremental partial-JSON chunk for a streaming tool_use
 // `input`. Emitted between `tool_start` and `tool_result` while the
@@ -307,6 +322,11 @@ export const ServerResultSchema = z.object({
     // #6769: occupancy snapshot (see ServerContextOccupancySnapshotSchema).
     // Absent from older servers and from providers with no occupancy signal.
     contextOccupancy: ServerContextOccupancySnapshotSchema.nullable().optional(),
+    // #7454/#7458: present on REPLAYED frames only (both replay paths map the
+    // server-internal `_seq` onto the wire; absent on live broadcasts). The
+    // #5555.3 delta-replay cursor — and for user_question the #7420
+    // live-vs-replayed guard — key on this field's presence.
+    historySeq: z.number().optional(),
 });
 export const ServerModelChangedSchema = z.object({
     type: z.literal('model_changed'),
--- a/packages/protocol/dist/schemas/server/stream.js
+++ b/packages/protocol/dist/schemas/server/stream.js
@@ -608,6 +608,11 @@ export const ServerUserQuestionSchema = z.object({
     type: z.literal('user_question'),
     toolUseId: z.string(),
     questions: z.array(z.any()),
+    // #7454: present on REPLAYED frames only (both replay paths map the server's
+    // internal `_seq` onto the wire); absent on every live broadcast. The #7420
+    // live-vs-replayed discriminator keys on this field's PRESENCE — a
+    // strip-on-parse schema without it would silently disarm that guard.
+    historySeq: z.number().optional(),
 });
 export const ServerAgentBusySchema = z.object({
     type: z.literal('agent_busy'),
--- a/packages/protocol/src/schemas/server/stream.ts
+++ b/packages/protocol/src/schemas/server/stream.ts
@@ -653,6 +653,11 @@ export const ServerUserQuestionSchema = z.object({
   type: z.literal('user_question'),
   toolUseId: z.string(),
   questions: z.array(z.any()),
+  // #7454: present on REPLAYED frames only (both replay paths map the server's
+  // internal `_seq` onto the wire); absent on every live broadcast. The #7420
+  // live-vs-replayed discriminator keys on this field's PRESENCE — a
+  // strip-on-parse schema without it would silently disarm that guard.
+  historySeq: z.number().optional(),
 })
 
 export const ServerAgentBusySchema = z.object({
--- a/packages/protocol/src/schemas/server/stream.ts
+++ b/packages/protocol/src/schemas/server/stream.ts
@@ -198,6 +198,11 @@ export const ServerMessageSchema = z.object({
   // shape-compatible.
   stdout: z.string().max(8192).optional(),
   stderr: z.string().max(8192).optional(),
+  // #7454/#7458: present on REPLAYED frames only (both replay paths map the
+  // server-internal `_seq` onto the wire; absent on live broadcasts). The
+  // #5555.3 delta-replay cursor — and for user_question the #7420
+  // live-vs-replayed guard — key on this field's presence.
+  historySeq: z.number().optional(),
 })
 
 export const ServerToolStartSchema = z.object({
@@ -207,6 +212,11 @@ export const ServerToolStartSchema = z.object({
   tool: z.string(),
   input: z.any(),
   serverName: z.string().optional(),
+  // #7454/#7458: present on REPLAYED frames only (both replay paths map the
+  // server-internal `_seq` onto the wire; absent on live broadcasts). The
+  // #5555.3 delta-replay cursor — and for user_question the #7420
+  // live-vs-replayed guard — key on this field's presence.
+  historySeq: z.number().optional(),
 })
 
 export const ServerToolResultSchema = z.object({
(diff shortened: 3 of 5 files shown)


blamechris:
## Code Review: PR #7458

### Summary

The change is correct and the smallest thing that could work. `handleRequestFullHistory`'s else-branch now performs the same `_seq` → `historySeq` mapping `replayHistory` does, and a normalized diff of the two expressions is identical modulo the callee and the `sessionId` binding:

```
ws-history.js:1267-1268          conversation-handlers.js:274-275
  const { _seq, ...wireEntry } = entry              (same)
  send(ws, { ...wireEntry, sessionId, ...})         ctx.transport.send(ws, { ...wireEntry, sessionId: targetId, ...})
```

I re-derived the reachability argument rather than taking it from the issue, and it holds. `getFullHistoryAsync` (`session-manager.js:2955`) returns JSONL when `resumeSessionId` is set and the file is non-empty, and `parseJsonlContent` emits only `user_input` / `response` / `tool_use` (`jsonl-reader.js:104` / `:129` / `:139`) — which is *exactly* the message-branch condition. So the else-branch is fed **only** by the ring buffer, and every ring entry is `_seq`-stamped (`_pushHistory`, `session-message-history.js:417`; restore re-stamps 1..N in `setHistory`, `:174-179`, called from `session-manager.js:2701`). The two branches partition cleanly on "has a `_seq`", which is what makes the one-line mapping sufficient.

I ran four mutants. Three died, one lived — the surviving one is the only substantive finding, and it is a test-coverage gap, not a defect in the fix.

Verification run in the PR worktree at `75bbf08c`:

| Gate | Result |
|---|---|
| `conversation-full-history-replay` + `ws-history` + `ws-schemas` | 340/340 pass (matches the PR body) |
| `ws-outbound-coverage` + `event-normalizer-schema` | 79/79 pass |
| `handlers/conversation-handlers` + `ws-handler-coverage` | 77/77 pass |
| protocol `schemas` + `handler-coverage` | 287/287 pass |
| all 9 `packages/server/scripts/lint-*.sh` | PASS |
| CI at `75bbf08c` | 20/20 green |
| working tree after mutation testing | clean (`git status --short` empty, sha256 restored) |

### Strengths

- **The positive control is the real thing, and it is load-bearing.** `!('historySeq' in q)` uses `in`, not `=== undefined`, so a fix that emits `historySeq: undefined` is caught too. I ran the exact fabrication mutant the PR body implies (`historySeq: _seq ?? 0`, unconditional) and it fails test 2 and only test 2. Absence *is* the signal here, and the control is the one assertion that keeps it that way.
- **The mutation claimed in the PR body reproduces.** Reverting the else-branch to `{ ...entry, sessionId: targetId }` fails exactly test 1 (2/3 pass). So does the weaker variant — keep the mapping, spread `entry` instead of `wireEntry` — which the `_seq`-leak assertion catches independently of the `historySeq` one. Two distinct assertions, two distinct mutants, no overlap.
- **Stripping `_seq` is provably safe rather than assumed safe.** Zero occurrences of `_seq` in `packages/dashboard/src`, `packages/app/src` and `packages/store-core/src` — nothing on either client ever read the raw counter, so removing it from the wire cannot break a consumer that was (wrongly) relying on it.
- **Protocol dist is CI-gated, not trusted.** `.github/workflows/ci.yml:1026-1030` rebuilds `packages/protocol` and fails on `git diff --exit-code packages/protocol/dist`, and the committed delta here is exactly the schema addition (+5 in `stream.js`, +1 in `stream.d.ts`, nothing else).
- **The schema comment names the guard it protects.** That is the right shape for a field whose whole semantics is "present ⇒ replayed" — a bare `historySeq: z.number().optional()` would read as an ordinary optional and be a natural thing for a future author to prune.

### Critical

None.

### Suggestions

| # | File / line | Finding |
|---|---|---|
| S1 | `packages/server/tests/conversation-full-history-replay.test.js` | **Surviving mutant: the tests pin the fix to one entry type, the fix covers five.** Narrowing the mapping to `...(typeof _seq === 'number' && entry.type === 'user_question' ? { historySeq: _seq } : {})` passes **208/208** across the new file + `ws-history.test.js` + `handlers/conversation-handlers.test.js`. The else-branch actually carries `message`, `tool_start`, `tool_result`, `result` and `user_question` (`session-message-history.js:345/361/372/382/393`) — and the two the clients *actually read `historySeq` off* are `message` (`dashboard:5362`, `app:2647`) and `tool_start` (`dashboard:2565`, `app:2875`), both feeding `recordHistorySeq`. Neither is pinned by any test in this PR. One extra fixture entry (`{ type: 'tool_start', toolUseId: 't9', _seq: 8 }`) plus one assertion kills that mutant and covers the part of the change that is live on `main` today. In scope, ~5 minutes. |
| S2 | `packages/protocol/src/schemas/server/stream.ts:138, 203, 212, 279` | **The schema hardening is one of five.** `historySeq` now rides *every* else-branch frame here, and `replayHistory` has always done the same for every history entry — but only `ServerUserQuestionSchema` (`:652`) declares the field. `ServerMessageSchema`, `ServerToolStartSchema`, `ServerToolResultSchema` and `ServerResultSchema` are plain strip-by-default `z.object`s without it. By the PR's own stated rationale, the two most load-bearing frames are the ones left undeclared: `message` and `tool_start` carry the **#5555.3 delta-replay cursor**, so a future strip-on-parse consumer would silently degrade every reconnect to a full replay — a strictly worse outcome than the `user_question` case this PR guards. Either add the field to all five (a shared `historySeqField` const keeps the comment in one place) or narrow the comment to say it is scoped to the discriminator on purpose. This is the adjacent-field pattern: three roster lines fixed, the fourth walked past. |
| S3 | PR body / commit message | **The narrative asserts a live defect that does not exist on `main`.** The `deliveredByReplay` discriminator lives on #7453's branch (`fix/7420-user-question-replay-sweep`, head `611a35ee6`) and is **not** in `origin/main` — `grep -c deliveredByReplay packages/store-core/src/dispatch-table.ts` returns `0` at `75bbf08c`. So today nothing reads `historySeq` off a `user_question` frame and no prompt is misclassified. What this PR fixes *today* is the raw `_seq` wire leak and the cursor divergence; the discriminator half is **pre-arming** #7453. Both are worth landing — but "read as a live racer" and "the discriminator was disarmed on every provider" are present-tense claims about a state that has not shipped, and a bisecting reader will hunt a symptom that never existed. One sentence ("#7453 is not yet merged; this lands the server half first") settles it. |
| S4 | cross-PR | **Merge-order coupling with #7453.** `611a35ee6` adds a `KNOWN GAP (#7454)` paragraph to `packages/store-core/src/dispatch-table.ts` saying replayed questions "are left unstamped (the benign direction) **until the server closes that gap**". #7458 closes it. Whichever of the two merges second must drop that paragraph, or `main` ships a comment describing a fixed bug as open — at the exact site of the guard, which is the worst place for it. (Its formatting is also mangled on that branch: the inserted sentence runs into the pre-existing text on one long line. Same edit, same PR.) |

### Nitpicks

| # | File / line | Finding |
|---|---|---|
| N1 | `conversation-full-history-replay.test.js:63` | `assert.ok(!('_seq' in rebuilt), …)` is a construction guard, not coverage: the message-branch builds a fixed object literal, so it cannot fail unless the branch is rewritten to spread. Worth keeping as a regression tripwire — worth a half-line comment saying that's what it is, so a later reader doesn't count it as behavioural coverage. The fixture is also a combination production cannot produce (`{ type: 'response', …, _seq: 3 }`): JSONL entries never carry `_seq`, and ring entries that carry `_seq` never have type `response`. Harmless, but it is why the assertion is free. |
| N2 | same test, third case | The symmetric assertion is missing — nothing asserts a message-branch frame carries **no** `historySeq`. Given that the field's entire meaning is presence-based, the "must not appear here" direction is worth one line next to the "must appear there" one. |

### Deferred Items

Both are out of scope for this PR and are filed rather than folded — they are the same "the second replay path doesn't do what the first one does" family as #7454, but different mechanisms:

- **#7459** — `handleRequestFullHistory` doesn't mirror `replayHistory`'s `result` → `agent_idle` synthesis (`ws-history.js:1270-1284`, #4628). The clients' own `case 'result'` deliberately skips the `activeTools` clear while replaying (#4466 gate, `dashboard/message-handler.ts:5444-5459`), so the zombie "Running X" chip is healed by reconnect and session-switch but **not** by "Sync Full History" — the one action a user takes because the view looks wrong.
- **#7460** — the same handler sends the entire history synchronously with no `bufferedAmount` gating, so the #4833 backpressure fix never reached this path; the 1MB `EVICT_THRESHOLD` (`ws-client-sender.js:6`) *closes the client*. Also folds in the two smaller parity gaps: `truncated` omitted from `history_replay_start`, and `latestSeq` omitted from both bracket frames (so `reconcileReplayEnd`'s cursor finalisation at `replay-reconcile.ts:208` never runs on this path).

### Architecture Notes

**The change does more than `user_question`, and I checked that the rest is safe.** On the ring-buffer branch *every* entry now carries `historySeq`, so both clients' `recordHistorySeq` advances the #5555.3 delta-replay cursor on the full-history path too — new behaviour that ships on `main` immediately, unlike the discriminator half. It is safe, and for a specific reason: `recordHistorySeq` (`replay-reconcile.ts:101-116`) ignores non-numeric and non-increasing values, so the cursor is monotonic and cannot regress; and it advances **per entry as each is applied**, so a mid-replay disconnect leaves the cursor at the last entry the client genuinely applied rather than ahead of it. Net effect is convergence with `replayHistory`, not a new divergence — and a small win: a "Sync Full History" now refreshes the cursor instead of leaving it stale.

**`historySeq` has no dispatch effect on either client.** `messageIsReplay` / `toolStartIsReplay` derive from `_replayingSessions` (set by `history_replay_start`), not from the field, so nothing about which handler runs or what it renders changes. The field is read at exactly four sites, all four `recordHistorySeq`.

**Nothing validates outbound frames today, so the schema addition is inert by construction.** `validateOutbound` (`ws-outbound-schemas.js:193`) has zero call sites under `packages/server/src` — and even wired up it discards the parse result (`if (result.success) return { ok: true }`), so it validates without stripping. Zod's default `z.object` already accepted the extra key. Accept/reject behaviour is unchanged in both directions; this is documentation-with-teeth for a future consumer, which is the right reason to add it.

**`_seq` absence iseffectively impossible in production, which is the right place for a positive control.** `_pushHistory` stamps on the way in, `truncateEntry` strips it on serialize, and `setHistory` re-stamps 1..N on restore. So test 2 covers a case the ring buffer essentially cannot produce — and that is exactly what a positive control is for: it proves the fix didn't take the shortcut of always emitting a number, which no production fixture would ever catch.

### Verdict

- [x] **Approve with suggestions** — no blocking issues; S1 and S2 are worth folding in before merge (both in-scope, both small)
- [ ] Request changes
- [ ] Approve as-is

The fix itself is right, minimal, and mutation-proven. S1 is the one I'd actually act on: the shipped behaviour covers five entry types and the tests pin one, and the four unpinned ones include both frames the clients read the field off.

blamechris:
All review items addressed at 714415f0e:

- **S1**: the breadth is now pinned — a `tool_start`/`tool_result` test asserts `historySeq` on the else-branch's other reader types; your exact narrowing mutant (`entry.type === 'user_question' && …`) now fails it. Verified red then restored.
- **S2**: the four sibling schemas carry the same optional `historySeq` with the shared rationale (dist rebuilt and committed) — the adjacent-field pattern closed rather than half-done.
- **S3**: the PR body's present-tense claim is corrected — on current main this PR fixes the `_seq` leak and the #5555.3 cursor, and pre-arms #7453's guard for either merge order.
- **S4**: #7453's branch now states the both-paths invariant instead of the KNOWN GAP (cad292671); recommended merge order puts #7458 first so it is true on arrival.
- **N2**: the symmetric no-`historySeq` assertion on the rebuilt message branch is in.
- N1 noted; #7459/#7460 stay tracked.

341/341 server + 4/4 on the extended file post-restore; protocol dist in sync.

PR #7463, opened 2026-08-28
test(app): extract a typed shared createMockMessageHandlerContext factory

### LINKED ISSUE #7451: app tests: extract a shared createMockMessageHandlerContext factory (4 hand-rolled copies)
## Context

Identified during review of PR #7446 (#7421).

## Description

Four store-test files each hand-build a `MessageHandlerContext`-shaped mock inline: `agent-event-handler.test.ts:~49`, `message-handler-activity.test.ts:~82`, `message-handler-permission-input.test.ts:~66`, `message-handler.test.ts:~69`. PR #7446 had to edit all four in lockstep to remove a single field, and the copies already diverge in field coverage (two carry `replayingSessions`, two do not). No shared mock-context helper exists in either app jest root (`src/test-utils/test-helpers.ts` only covers hook rendering; `__tests__/PushTokenError.test.ts`'s `createMockContext` is an unrelated connection-context shape).

## Original Review Comment

> 4 independent copies of the same object literal must be kept in sync by hand for every MessageHandlerContext field change — this PR is direct proof: a one-field removal touched 4 files instead of 1.

## Acceptance Criteria

- [ ] A single `createMockMessageHandlerContext(overrides?)` in a shared app test-utils module
- [ ] All four suites consume it; a context-field change is a one-file edit
- [ ] The factory derives from (or is typechecked against) the real `MessageHandlerContext` type so a missing field fails at the factory, not far from its cause

blamechris:
## What

Extracts one shared, **typed** `createMockMessageHandlerContext(overrides?)` into
`packages/app/src/test-utils/mock-message-handler-context.ts` and routes every
hand-rolled connection-context mock in `packages/app/src/__tests__/` through it,
plus the one remaining copy in the other jest root.

A `ConnectionContext` field change is now a one-file edit, and — the part the
issue's follow-up comment asks for — it now **fails at the factory** instead of
nowhere.

Closes #7451

## The divergence, and what it turned out to be

The four `createMockContext()` copies named in the issue:

| suite | fields in its local literal |
|---|---|
| `src/__tests__/store/agent-event-handler.test.ts` | `socket`, `serverUrl`, `apiToken`, `connectionId`, `reconnecting`, `connectedAt`, `activeSessionIdAtConnect` |
| `src/__tests__/store/message-handler.test.ts` | same 7 |
| `src/__tests__/store/message-handler-activity.test.ts` | same 7 **+ `replayingSessions`** |
| `src/__tests__/store/message-handler-permission-input.test.ts` | same 7 **+ `replayingSessions`** |
| `__tests__/PushTokenError.test.ts` (5th copy, other root) | `url`, `token`, `isReconnect`, `silent`, `socket` |

The real type is `ConnectionContext` (`@chroxy/store-core`, re-exported from
`src/store/types.ts`):

```ts
interface ConnectionContext { url: string; token: string; isReconnect: boolean; silent: boolean; socket: WebSocket }
```

So the divergence is bigger than "two carry `replayingSessions`": **the four
store copies share exactly one field name with the real type.** All seven of
their fields are absent from `ConnectionContext`, and four of its five real
fields were missing. That is precisely the blind spot the follow-up comment
predicted — the `as any` at each call site meant neither direction diagnosed.

Every stale field is inert, verified before removal:

- `replayingSessions` — the handler only ever reads `_ctx.replayingSessions`,
  its own module-private `MessageHandlerContext`; the injected connection
  context is never consulted for it (`grep -rn 'ctx\.replayingSessions' src/store/`
  returns `_ctx.` sites only). The two suites that carried it drive replay state
  through the real `history_replay_start` path, not through the mock.
- `serverUrl` / `apiToken` / `connectionId` / `activeSessionIdAtConnect` — zero
  occurrences in `src/store/message-handler.ts`.
- `reconnecting` — never read; the real field is `isReconnect`, and the one test
  that needed it was already spreading `{ ...createMockContext(), isReconnect: true }`
  on top. That is now the explicit override `createMockMessageHandlerContext({ isReconnect: true })`.

## Factory contract

```ts
export function createMockMessageHandlerContext(
  overrides: Partial<ConnectionContext> = {},
): ConnectionContext
```

- **Return type is the real `ConnectionContext`**, and the base is a typed object
  literal — so both a missing field (TS2741) and an excess field (TS2353) fail
  inside the factory.
- **One cast, scoped to one field**: `socket` (a real `WebSocket` can't be
  constructed under jest). Defaults to `{ readyState: 1, send: jest.fn(), close: jest.fn() }`.
- **`_testMessageHandler.setContext` already takes `ConnectionContext`**, so the
  parameter type forced no cast — every call site drops its `as any` outright.
- Per-suite differences are explicit `overrides` at the call site: `{ isReconnect: true }`
  for the reconnect test, `{ socket: mockSocket }` where a suite asserts on its own
  socket spy.

## Red-first evidence (both directions, on the final tree)

Mutate the factory, run `npx tsc --noEmit`, restore (`cp` from a backup, not
`git checkout --`). Mutant landing was asserted by grep count each time.

**A — delete a required field (`silent`)** → exit 2, one diagnostic, at the factory:

```
src/test-utils/mock-message-handler-context.ts(29,9): error TS2741:
  Property 'silent' is missing in type '{ url: string; token: string; isReconnect: false; socket: WebSocket; }'
  but required in type 'ConnectionContext'.
```

**B — add an excess field** (shaped from the original bug: re-add the stale
`replayingSessions` the old copies carried) → exit 2, one diagnostic, at the factory:

```
src/test-utils/mock-message-handler-context.ts(32,5): error TS2353:
  Object literal may only specify known properties, and 'replayingSessions' does not exist in type 'ConnectionContext'.
```

A plain bogus field (`bogusField: 'nope'`) produces the same TS2353. Restored → `tsc --noEmit` exit 0.

**Control — the same two mutations against the PRE-CHANGE hand-rolled literal**
(rename `apiToken` → `bogusField` in `message-handler.test.ts`'s old
`createMockContext`, i.e. delete a field *and* add an unknown one in one edit):

```
TSC_EXIT = 0   # zero diagnostics — both mutations invisible
```

The baseline is itself the strongest control: on `main`, four mocks carrying
seven fields that don't exist on `ConnectionContext` and missing four that do
typecheck clean.

## Suites

`packages/app`, node 22, `--coverage=false`:

| run | suites | tests |
|---|---|---|
| baseline `src/__tests__/store` (before) | 34 passed / 34 | 731 passed / 731 |
| after, `src/__tests__/store` + `__tests__/PushTokenError.test.ts` | 35 passed / 35 | 733 passed / 733 |
| after, **full app suite** (`npx jest`, both roots) | 181 passed / 181 | 2372 passed / 2372 |

`+1 suite / +2 tests` is exactly `PushTokenError.test.ts` joining the selection.
No assertion was added, removed, or changed — the diff on
`message-handler.test.ts` is 177 mechanical call-site swaps plus four targeted
edits (import, factory removal, the `isReconnect` override, and `sentTypes`'
parameter type going from `ReturnType<typeof createMockContext>` to
`ConnectionContext`). `npx tsc --noEmit` is clean on the committed tree.

## Both jest roots surveyed

Scope grew slightly past the issue's four, because `src/__tests__/` had two more
files repeating the same literal inline seven times:

| file | sites | action |
|---|---|---|
| `src/__tests__/store/agent-event-handler.test.ts` | 3 | folded in |
| `src/__tests__/store/message-handler.test.ts` | 181 | folded in |
| `src/__tests__/store/message-handler-activity.test.ts` | 9 | folded in |
| `src/__tests__/store/message-handler-permission-input.test.ts` | 7 | folded in |
| `src/__tests__/store/connection.test.ts` | 6 inline literals | folded in (`{ socket: mockSocket }`) |
| `src/__tests__/store/terminal-mirror.test.ts` | 1 inline literal | folded in (`{ socket: mockSocket }`) |
| `__tests__/PushTokenError.test.ts` | 2 | folded in |

`src/__tests__/` now contains **no** hand-rolled connection-context literal.

**Surveyed and deliberately not folded in** — the other root's
`setConnectionContext(...)` family (`__tests__/dispatch-activity.test.ts`,
`__tests__/message-handler.test.ts`, `__tests__/contract-switch.test.ts`,
`__tests__/message-handler.encryption-gate.test.ts`,
`__tests__/AutoResumeOnReconnect.test.ts`). These are a different pattern: they
inject through the production setter, and the only non-trivial field — `socket` —
is a suite-local spy the assertions read back
(`expect(mockSocket.close).toHaveBeenCalled()`). Two of them deliberately pass
`socket: {} as WebSocket`, a stub that would *throw* if the handler touched it;
the factory's default socket would silently convert "must not touch the socket"
into "may touch it", changing what those suites exercise. Routing them through
the factory would mean overriding `socket` (and `url`/`token` where the value is
load-bearing) at every site, leaving `silent`/`isReconnect` as the only shared
content — churn without dedup. `encryption-gate` already has its own local
`ctx(overrides)` factory bound to its `mockSocket`.

### DIFF
--- a/packages/app/__tests__/PushTokenError.test.ts
+++ b/packages/app/__tests__/PushTokenError.test.ts
@@ -1,5 +1,5 @@
 import { _testMessageHandler, setStore } from '../src/store/message-handler';
-import { createMockMessageHandlerContext } from '../src/test-utils/mock-message-handler-context';
+import { createMockConnectionContext } from '../src/test-utils/mock-connection-context';
 import { createEmptySessionState } from '../src/store/utils';
 import type { ConnectionState } from '../src/store/types';
 
@@ -53,7 +53,7 @@ describe('push_token_error handler (#1987)', () => {
       sessionStates: { s1: { ...createEmptySessionState(), messages: [] } },
     });
     setStore(store as any);
-    _testMessageHandler.setContext(createMockMessageHandlerContext());
+    _testMessageHandler.setContext(createMockConnectionContext());
 
     _testMessageHandler.handle({
       type: 'push_token_error',
@@ -73,7 +73,7 @@ describe('push_token_error handler (#1987)', () => {
       sessionStates: { s1: { ...createEmptySessionState(), messages: [] } },
     });
     setStore(store as any);
-    _testMessageHandler.setContext(createMockMessageHandlerContext());
+    _testMessageHandler.setContext(createMockConnectionContext());
 
     _testMessageHandler.handle({
       type: 'push_token_error',
--- a/packages/app/__tests__/PushTokenError.test.ts
+++ b/packages/app/__tests__/PushTokenError.test.ts
@@ -1,4 +1,5 @@
 import { _testMessageHandler, setStore } from '../src/store/message-handler';
+import { createMockMessageHandlerContext } from '../src/test-utils/mock-message-handler-context';
 import { createEmptySessionState } from '../src/store/utils';
 import type { ConnectionState } from '../src/store/types';
 
@@ -30,16 +31,6 @@ function createMockStore(initialState: Partial<ConnectionState>) {
   };
 }
 
-function createMockContext() {
-  return {
-    url: 'wss://test',
-    token: 'test-token',
-    isReconnect: false,
-    silent: false,
-    socket: { send: jest.fn(), close: jest.fn() } as unknown as WebSocket,
-  };
-}
-
 describe('push_token_error handler (#1987)', () => {
   let warnSpy: jest.SpyInstance;
 
@@ -62,7 +53,7 @@ describe('push_token_error handler (#1987)', () => {
       sessionStates: { s1: { ...createEmptySessionState(), messages: [] } },
     });
     setStore(store as any);
-    _testMessageHandler.setContext(createMockContext() as any);
+    _testMessageHandler.setContext(createMockMessageHandlerContext());
 
     _testMessageHandler.handle({
       type: 'push_token_error',
@@ -82,7 +73,7 @@ describe('push_token_error handler (#1987)', () => {
       sessionStates: { s1: { ...createEmptySessionState(), messages: [] } },
     });
     setStore(store as any);
-    _testMessageHandler.setContext(createMockContext() as any);
+    _testMessageHandler.setContext(createMockMessageHandlerContext());
 
     _testMessageHandler.handle({
       type: 'push_token_error',
--- a/packages/app/src/__tests__/store/agent-event-handler.test.ts
+++ b/packages/app/src/__tests__/store/agent-event-handler.test.ts
@@ -12,6 +12,7 @@ import {
   _testMessageHandler,
   setStore,
 } from '../../store/message-handler';
+import { createMockMessageHandlerContext } from '../../test-utils/mock-message-handler-context';
 import { createEmptySessionState } from '../../store/utils';
 import type { ConnectionState } from '../../store/types';
 import type { ChatMessage } from '../../store/connection';
@@ -46,18 +47,6 @@ function createMockStore(initialState: Partial<ConnectionState>) {
   };
 }
 
-function createMockContext() {
-  return {
-    socket: { readyState: 1, send: jest.fn() } as any,
-    serverUrl: 'wss://test.example.com',
-    apiToken: 'test-token',
-    connectionId: 'test-conn-1',
-    reconnecting: false,
-    connectedAt: Date.now(),
-    activeSessionIdAtConnect: null,
-  };
-}
-
 function taskBubble(toolUseId: string): ChatMessage {
   return {
     id: `m-${toolUseId}`,
@@ -84,7 +73,7 @@ describe('agent_event handler (#5060)', () => {
   it('appends a child wire event to the parent Task bubble childAgentEvents[]', () => {
     const store = storeWithTask('parent-1');
     setStore(store as any);
-    _testMessageHandler.setContext(createMockContext() as any);
+    _testMessageHandler.setContext(createMockMessageHandlerContext());
 
     _testMessageHandler.handle({
       type: 'agent_event',
@@ -104,7 +93,7 @@ describe('agent_event handler (#5060)', () => {
   it('accumulates multiple events in arrival order', () => {
     const store = storeWithTask('parent-2');
     setStore(store as any);
-    _testMessageHandler.setContext(createMockContext() as any);
+    _testMessageHandler.setContext(createMockMessageHandlerContext());
 
     _testMessageHandler.handle({
       type: 'agent_event',
@@ -134,7 +123,7 @@ describe('agent_event handler (#5060)', () => {
   it('same-reference no-op when parentToolUseId matches no bubble', () => {
     const store = storeWithTask('parent-3');
     setStore(store as any);
-    _testMessageHandler.setContext(createMockContext() as any);
+    _testMessageHandler.setContext(createMockMessageHandlerContext());
     const before = store.getState().sessionStates.s1.messages;
 
     _testMessageHandler.handle({
--- a/packages/app/src/__tests__/store/agent-event-handler.test.ts
+++ b/packages/app/src/__tests__/store/agent-event-handler.test.ts
@@ -12,7 +12,7 @@ import {
   _testMessageHandler,
   setStore,
 } from '../../store/message-handler';
-import { createMockMessageHandlerContext } from '../../test-utils/mock-message-handler-context';
+import { createMockConnectionContext } from '../../test-utils/mock-connection-context';
 import { createEmptySessionState } from '../../store/utils';
 import type { ConnectionState } from '../../store/types';
 import type { ChatMessage } from '../../store/connection';
@@ -73,7 +73,7 @@ describe('agent_event handler (#5060)', () => {
   it('appends a child wire event to the parent Task bubble childAgentEvents[]', () => {
     const store = storeWithTask('parent-1');
     setStore(store as any);
-    _testMessageHandler.setContext(createMockMessageHandlerContext());
+    _testMessageHandler.setContext(createMockConnectionContext());
 
     _testMessageHandler.handle({
       type: 'agent_event',
@@ -93,7 +93,7 @@ describe('agent_event handler (#5060)', () => {
   it('accumulates multiple events in arrival order', () => {
     const store = storeWithTask('parent-2');
     setStore(store as any);
-    _testMessageHandler.setContext(createMockMessageHandlerContext());
+    _testMessageHandler.setContext(createMockConnectionContext());
 
     _testMessageHandler.handle({
       type: 'agent_event',
@@ -123,7 +123,7 @@ describe('agent_event handler (#5060)', () => {
   it('same-reference no-op when parentToolUseId matches no bubble', () => {
     const store = storeWithTask('parent-3');
     setStore(store as any);
-    _testMessageHandler.setContext(createMockMessageHandlerContext());
+    _testMessageHandler.setContext(createMockConnectionContext());
     const before = store.getState().sessionStates.s1.messages;
 
     _testMessageHandler.handle({
--- a/packages/app/src/__tests__/store/connection.test.ts
+++ b/packages/app/src/__tests__/store/connection.test.ts
@@ -20,6 +20,7 @@ import {
   getEncryptionState,
   setEncryptionState,
 } from '../../store/message-handler';
+import { createMockMessageHandlerContext } from '../../test-utils/mock-message-handler-context';
 import { createKeyPair } from '@chroxy/store-core';
 
 // Reset store between tests
@@ -821,10 +822,7 @@ describe('connectionError and connectionRetryCount state', () => {
 
   it('auth_ok clears both fields', () => {
     const mockSocket = { readyState: 1, send: jest.fn(), close: jest.fn() } as unknown as WebSocket;
-    _testMessageHandler.setContext({
-      url: 'wss://test', token: 'tok', isReconnect: false,
-      silent: false, socket: mockSocket,
-    });
+    _testMessageHandler.setContext(createMockMessageHandlerContext({ socket: mockSocket }));
     useConnectionLifecycleStore.setState({ connectionError: 'Network error', connectionRetryCount: 2 });
     _testMessageHandler.handle({ type: 'auth_ok', serverMode: 'cli' });
     const state = useConnectionLifecycleStore.getState();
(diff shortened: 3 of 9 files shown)


blamechris:
## Adversarial review at `ed16b0671`

Reviewed from the worktree at `ed16b0671`, Node v22.22.3 / darwin-arm64. Method: independent re-derivation of the load-bearing claim (not a re-read of the PR body), then scripted mutations — `/tmp` byte backups, grep-verified as landed, restored by absolute-path `cp`. Worktree is byte-identical to the commit afterwards (`git diff HEAD` → 0 lines, `git status --short` empty).

**The fossil-shape claim is correct, and it is the right claim to have led with.**

### The load-bearing claim — independently verified

I grepped all seven removed fields against `packages/app/src/store/` myself rather than trusting the summary:

| field | occurrences in `src/store/` | verdict |
|---|---|---|
| `serverUrl` | 0 | inert |
| `connectionId` | 0 | inert |
| `activeSessionIdAtConnect` | 0 | inert |
| `apiToken` | 9 — all `connection-lifecycle.ts` store state (`wsUrl`/`apiToken` pair) | not the context |
| `reconnecting` | ~30 — all FSM phase *strings* | not a field |
| `connectedAt` | 0 — every hit is a substring collision with `disconnectedAttemptId` | inert |
| `replayingSessions` | 12 — **all `_ctx.`** | inert |

The `replayingSessions` one is the case that could have made the swap silently change what the suites test, so I chased it to the bottom. There are genuinely two contexts in `message-handler.ts`: the module-private `MessageHandlerContext` built by `createDefaultContext()` (declared L265, `replayingSessions` at L275), and the injected `ConnectionContext` `_connectionContext` (L634). `_testMessageHandler.setContext` (L3852) writes **only** the latter. So `replayingSessions` on the injected mock was never reachable — no suite passed *because* it was there. The two suites that carried it drive replay through the real `history_replay_start` path, exactly as claimed (`message-handler-activity.test.ts:257`).

Conversely, all five real fields *are* read — `ctx.silent` (L1735, L2132, L2142), `ctx.socket` (many), `ctx.isReconnect` (L1858, L1888, L2188), `ctx.url` (L1860, L1906, L1910, L2057, L3653), `ctx.token` (L1870, L2104). So "four of five real fields were missing" was a real hole, not a cosmetic one.

### tsc backstop — both red directions reproduced

Not taken on trust; re-run here, one mutation at a time.

**A — delete `silent`** → exit 2, **1** diagnostic, at the factory:

```
src/test-utils/mock-message-handler-context.ts(29,9): error TS2741: Property 'silent' is missing in type
  '{ url: string; token: string; isReconnect: false; socket: WebSocket; }' but required in type 'ConnectionContext'.
```

Byte-identical to the diagnostic quoted in the PR body.

**B — re-add the stale `replayingSessions`** → exit 2, **1** diagnostic, at the factory:

```
src/test-utils/mock-message-handler-context.ts(34,5): error TS2353: Object literal may only specify known
  properties, and 'replayingSessions' does not exist in type 'ConnectionContext'.
```

Restored → exit 0, 0 diagnostics. (Mutation A doubles as the positive control that `tsc` is actually checking this file at all.)

**And the backstop is wired to the real type, which is the part worth checking.** A typed factory is worthless if `ConnectionContext` resolves to a stale `.d.ts`. It doesn't: `@chroxy/store-core` resolves through the main checkout's workspace symlink with `"types": "src/index.ts"` — the actual source at `packages/store-core/src/types/connection.ts:75`. A real field change trips this factory.

### Suites and arithmetic — verified, not recomputed from the body

```
npx jest src/__tests__/store --coverage=false                                  → 34 passed / 731 passed, exit 0
npx jest src/__tests__/store __tests__/PushTokenError.test.ts --coverage=false → 35 passed / 733 passed, exit 0
```

`+1 suite / +2 tests` is exactly `PushTokenError.test.ts` (1 `describe`, 2 `it`). And the "before" baseline is *necessarily* the same 34/731, because the diff adds and removes **zero** `it(` / `test(` / `describe(` lines across all seven files — so I didn't need a checkout to confirm it.

### Behaviour preservation — all seven diffs are mechanical

`message-handler.test.ts` is 177 identical call-site swaps plus exactly the four structural edits the body claims (import, factory removal, the `isReconnect` override, `sentTypes`' parameter `ReturnType<typeof createMockContext>` → `ConnectionContext`). I filtered the diff down to non-mechanical lines to check this rather than eyeballing 386 changed lines. The other six are pure swaps. The two deliberate `socket: {} as WebSocket` stubs are untouched and there are exactly two, as stated (`dispatch-activity.test.ts:159`, `__tests__/message-handler.test.ts:139`).

**One thing the body is silent on, which I checked and am recording so the next reader doesn't re-derive it.** Fixture *values* did change at 8 sites, and the four store suites went from `url`/`token`/`isReconnect`/`silent` all being `undefined` to real values — on a path where `ctx.url` and `ctx.token` are genuinely read. Verified inert:

- `connection.test.ts` (6) + `terminal-mirror.test.ts` (1): `'wss://test'` → `'wss://test.example.com'`, `'tok'` → `'test-token'`. Zero assertions on `'wss://test'` / `'tok'` / `wsUrl` / `apiToken` / `lastConnectedUrl` / `persistVerifiedConnection` in any changed suite, and both values classify identically as `'tunnel'` under `/^ws:\/\//i`.
- Socket shape also changed — `close` added for four suites, `readyState` added for `PushTokenError`. Inert: no changed suite exercises an identity-refusal / `ctx.socket.close()` path (0 hits for `identity_*` / `auth_failed`), and `push_token_error` is a four-line `console.warn` (L3637-3642).

So "no assertion was added, removed, or changed" is accurate. The fixture delta is the more interesting half and it holds up.

---

## Findings

### S1 — the excess-field backstop stops at the base literal; `overrides` has a hole (suggestion)

The PR body and the doc comment both scope the claim correctly ("fail **inside the factory**", "excess-property check **on a typed literal**"), so this is not an overclaim — it's a hardening opportunity in a file this PR is creating.

I probed five override shapes in one `tsc` pass:

| call | result |
|---|---|
| `createMockMessageHandlerContext({ bogusField: 1 })` | **TS2353** — caught |
| `createMockMessageHandlerContext({ silent: 'yes' })` | **TS2322** — caught |
| `const o = { bogusField: 1 }; …(o)` | **TS2559** (weak-type) — caught |
| `…({ bogusField: 1 } as Partial<ConnectionContext>)` | **no diagnostic** |
| `const o = { silent: true, bogusField: 1 }; …(o)` | **no diagnostic** |

The last row is the one that matters. Excess-property checking only applies to *fresh* object literals, so a **variable-held** override carrying at least one real field passes an unknown field straight through — and because the body is `{ ...base, ...overrides }`, `bogusField` is then genuinely present on the returned object at runtime. That is the fossil condition re-entering through the override door: a `beforeEach` fixture or a table-driven loop can reintroduce a stale field and neither the factory nor the call site says a word.

Every current call site uses an inline literal, so **the PR is safe as it stands**. If you want the door shut, this compiles clean and I verified all seven call sites still typecheck:

```ts
export function createMockMessageHandlerContext<T extends Partial<ConnectionContext>>(
  overrides?: T & Record<Exclude<keyof T, keyof ConnectionContext>, never>,
): ConnectionContext
```

Proven against the same five probes: the variable-held case now fails **TS2345**, and `{ isReconnect: true }` still compiles. Honest trade — it closes that case but *loses* the TS2559 all-bogus-variable case (degenerate; an override with zero real fields). My first attempt at this signature used `= {} as T` and errored **inside the factory** (TS2322), so take the `overrides?:` form above rather than a default-parameter variant. Your call whether the trade is worth it; the status quo is defensible.

### S2 — the survey is presented as exhaustive and misses the largest file (suggestion) → filed as #7465

`__tests__/auth-ok-handler.test.ts` carries **31** inline `const ctx = { url, token, socket, isReconnect, silent }` literals and appears in neither the folded-in table nor the deliberately-declined list. It is the biggest remaining concentration in the repo — more than all five declined files combined (11).

No fossil risk (its literals are the full correct shape), so this is scope, not correctness. The narrow claim — "`src/__tests__/` now contains no hand-rolled connection-context literal" — I verified and it is true; it's the "both jest roots surveyed" framing that reads as complete and isn't.

### N1 — the declined-scope reasoning is sound for four of five files (nitpick, folded into #7465)

The `socket: {} as WebSocket` argument is correct and well made — the factory's default socket really would convert "must not touch the socket" into "may touch it", and that is a genuine change to what those two suites exercise. Right call.

`AutoResumeOnReconnect.test.ts` is the weak member. Its seven literals are `{ url, token, socket, isReconnect } as any` — **missing `silent` entirely**, and cast — i.e. precisely the drifted-shape-with-no-backstop condition #7451 exists to kill. Each collapses to `createMockMessageHandlerContext({ socket, isReconnect: false })`: a net line reduction that *gains* the backstop, not the "churn without dedup" that correctly rules out the other four. (I checked — `silent: undefined` is inert there, `!ctx.silent` is true either way. Latent drift, not a bug.)

### N2 — the factory name collides with a real type in the module it mocks for (nitpick)

`MessageHandlerContext` is not a phantom name: it is a real, distinct interface at `message-handler.ts:265`, and it is what `_ctx` holds. A factory named `createMockMessageHandlerContext` that returns a `ConnectionContext` collides with it in the same module — which is the exact conflation that produced the seven-field fossil in the first place. `createMockConnectionContext` would say what it builds.

Against that: the name tracks #7451's wording, and #7451 inherited the fossil assumption (its AC literally says "typechecked against the real `MessageHandlerContext` type"). The file's doc comment is clear about what it returns. Cosmetic, but the collision is real rather than imagined — worth a moment's thought, not a blocker either way.

### Placement — no issue

`src/test-utils/` already existed with `test-helpers.ts`, so this follows precedent rather than inventing a location. Correct.

---

### Verdict

- [x] Approve
- [ ] Request changes
- [ ] Comment only

**Approve at `ed16b0671`.** The central claim is the load-bearing one and it survives independent verification: the four store mocks really were a seven-field fossil sharing *zero* fields with `ConnectionContext`, every removed field is provably inert, and the compiler backstop is real — I made it go red in both directions and confirmed it resolves against store-core source rather than a stale `.d.ts`. The red-first evidence in the description reproduces exactly, which is not something I get to say often.

S1 is a genuine gap in the guard's reach and the one thing I'd consider folding in before merge (one line, proven above) — but it's optional, because every call site this PR creates is on the safe side of it. S2/N1 are tracked in #7465. N2 is a judgment call I'm happy to lose.

blamechris:
S1 and N2 folded at ad93161f3:

- **S1**: the mapped-`never` generic you supplied is in — proven red on a **variable-held** `{ silent: true, bogusField: 1 }` override (TS2345 naming the key; my first probe red was a wrong-import TS2307 and proved nothing — re-run with the correct path plus a green positive control before claiming it).
- **N2**: renamed to `createMockConnectionContext` (file renamed to match) — you're right that colliding with a real, different type in the very module it mocks for is worth more than continuity with the issue title.
- **S2/N1** stay tracked in #7465 (auth-ok-handler's 31 literals + the AutoResumeOnReconnect `as any` family).

35 suites / 733 tests green, tsc clean at the new head.

Issue #7447, opened 2026-08-28
Guard against a zero-test green run in the `node --test` packages (design-tokens, claude-hooks)


blamechris:
## Problem

`packages/design-tokens` and `packages/claude-hooks` both run their suites through
`node --test` in a shape that **exits 0 when it discovers zero test files**. A suite
that stops being discovered and a passing suite are then the same observable outcome —
the defect class `docs/false-safety-guards.md` exists for, and precisely the one
#7408 / #7444 just fixed one level up (the suite had no CI job at all).

`packages/server` already solves this with `scripts/assert-test-count.mjs` (a `# tests N`
floor). Nothing equivalent guards the two packages below.

## Measured, Node v22.22.3

**`packages/design-tokens`** — `"test": "node --import ../../scripts/lib/no-test-force-exit-hook.mjs --test"`.
Bare `--test` with no path argument: discovery is entirely implicit from the cwd.
Against a scratch copy of the package:

| scenario | `# tests` | exit |
|---|---|---|
| unmodified | 6 | 0 |
| `test/index.test.js` → `test/index.spec.js` | 6 | 0 (still matched by Node's `**/test/**/*.js` pattern) |
| moved out of `test/` as `src/tokens.spec.js` | **0** | **0** |
| `test/` directory deleted | **0** | **0** |

The last two are green CI with nothing run.

**`packages/claude-hooks`** — `node --import ./tests/_setup.mjs --test './tests/**/*.test.js'`.
The glob is quoted, so Node expands it rather than the shell; a glob that matches nothing
is not an error:

```
$ node --test './tests/**/*.test.js'    # empty tests/ dir
# pass 0 / # fail 0
exit 0
```

(Its `--import ./tests/_setup.mjs` does fail loudly if that one file goes missing, so
claude-hooks is partly covered by accident, not by design.)

The other suites are not exposed: `vitest run` (dashboard, store-core) and `jest`
(app) both exit non-zero on "no test files found", and `packages/protocol`'s unquoted
`tests/*.test.js` reaches Node as a literal path when it matches nothing, which errors.

## Suggested fix

Put a test-count floor in front of the two `node --test` commands, the way the server
suite does. `packages/server/scripts/assert-test-count.mjs` already streams output through
unchanged, reads `# tests N` / `# fail M` from the aggregate TAP summary, and fails on a
missing summary, a signal death, `M > 0`, or `N < floor`. It takes the underlying command
[...]

Issue #7421, opened 2026-08-27
app: `_ctx.isSessionSwitchReplay` is write-only dead state, and a comment claims it still gates UX paths


blamechris:
## Context

Identified during review of PR #7419 (#7410).

## Description

`_ctx.isSessionSwitchReplay` in `packages/app/src/store/message-handler.ts` has **eight** references and **zero reads**:

| Line | Kind |
|---|---|
| `:271` | type declaration |
| `:326` | initialiser (`false`) |
| `:772` | write `false` |
| `:1859` | write `false` |
| `:2330` | write `true` |
| `:2450` | write `true` |
| `:2618` | write `false` |
| `:2448` | comment |

Nothing ever reads it. The dashboard has no such flag at all — the only hit there is a stale test comment (`packages/dashboard/src/store/message-handler.test.ts:4052`) referencing a `_isSessionSwitchReplay` that does not exist in that package.

## Why it matters

More than tidiness: the comment at `packages/app/src/store/message-handler.ts:2448` asserts

> `isSessionSwitchReplay` still gates the same UX paths as before.

That is false — it gates nothing. A comment claiming a live invariant for dead state is the shape catalogued in [`docs/false-safety-guards.md`](../blob/main/docs/false-safety-guards.md) (*"a guard whose comment describes a stronger check than its code performs"*). Someone reasoning about replay ordering will believe the flag is load-bearing and preserve write ordering around it that costs nothing to change.

It came up in #7419 precisely this way: the sweep moved to before the `:2618` write, and establishing that the reordering was inert required proving the flag has no readers.

## Suggested fix

Delete the field, its initialiser, and all five writes; fix or delete the `:2448` comment and the stale dashboard test comment at `:4052`.

If it is instead meant to be read and a reader was lost in a refactor, restore the reader — but then it needs a test, since nothing currently observes it.

## Acceptance criteria

- [ ] `grep -rn "isSessionSwitchReplay" packages/app/src packages/dashboard/src` returns nothing, or returns a reader plus a test that fails when the flag is not set
- [ ] The `:2448` comment no longer claims a gate that does not exist
- [ ] App test suite green

## Related

- #7419 / #7410 — where this surfaced

Issue #3121, opened 2026-04-28
App session_switched: hoist pendingSwitchSessionId clear above early return for defensive parity


blamechris:
## Context

Identified during review of PR #3120.

## Description

In `packages/app/src/store/message-handler.ts`, the `session_switched` handler now early-returns when `sharedSessionSwitched(msg)` returns null (missing/non-string/empty sessionId):

\`\`\`ts
case 'session_switched': {
  const switched = sharedSessionSwitched(msg);
  if (!switched) break;                            // <-- early return
  const { newSessionId: sessionId, conversationId: switchConvId } = switched;
  if (_ctx.pendingSwitchSessionId && _ctx.pendingSwitchSessionId === sessionId) {
    _ctx.isSessionSwitchReplay = true;
  }
  _ctx.pendingSwitchSessionId = null;              // <-- skipped on early return
  ...
}
\`\`\`

If a malformed `session_switched` arrives between a user-initiated switch and the real follow-up message, the stale `_ctx.pendingSwitchSessionId` is left in place and could falsely trigger replay-dedup on the next valid switch.

## Original Review Comment

> When `sharedSessionSwitched(msg)` returns null, the `break` skips `_ctx.pendingSwitchSessionId = null` clearing. If a malformed `session_switched` arrives between a user-initiated switch and the real follow-up, the stale `pendingSwitchSessionId` could falsely trigger replay-dedup on the next valid switch. Server-side validation makes this nearly impossible in practice, but consider hoisting the `_ctx.pendingSwitchSessionId = null` clear above the early return for defensive parity with the prior inline behaviour.

## Risk Assessment

Marginal. The server's `session_switched` emit sites all populate `sessionId` from server-validated state (see `packages/server/src/handlers/session-handlers.js` and `ws-history.js`). A malformed payload arriving here would indicate a deeper protocol drift, not a normal case.

## Acceptance Criteria

- [ ] Hoist `_ctx.pendingSwitchSessionId = null;` (and the replay-dedup check) above the `if (!switched) break;` line, OR move the clear into the early-return branch as well.
- [ ] Verify existing app tests still pass (no test currently asserts this edge case; consider adding one).

## Related

- Parent: #2661
- Reviewing PR: #3120

PR #7464, opened 2026-08-28
feat(ci): lint write-only fields on the app's MessageHandlerContext

### LINKED ISSUE #7452: guard: write-only MessageHandlerContext fields have no lint — the #7421 class recurs silently
## Context

Identified during review of PR #7446 (#7421). The PR removed the write-only `isSessionSwitchReplay` flag — and its own diff silently created a second instance of the same class (`pendingSwitchSessionId` lost its only reader), caught only by the review pass and fixed in the same PR (052227e36).

## Description

Nothing automated detects a context/store field that is written but never read. #7421 cost a manual archaeology pass to establish "zero reads"; the recurrence inside the very PR fixing it demonstrates how easily the class re-enters. `docs/false-safety-guards.md` catalogues the adjacent comment-stronger-than-code shape, but there is no lint for the state itself.

## Suggested shape

A script-lint (alongside `packages/server/scripts/lint-*.sh`, but for the app store) that, for each field declared on `MessageHandlerContext` (and the dashboard's module-level `_ctx`-style state), classifies references as read vs write and fails on zero-read fields. A heuristic (assignment-target vs other) is acceptable; an allowlist with justification comments covers intentional cases. Prove it red against a synthetic write-only field before landing (docs/false-safety-guards.md rule).

## Acceptance Criteria

- [ ] Lint fails on a field with writes and no reads in `packages/app/src/store`
- [ ] Proven red: re-introducing `pendingSwitchSessionId` exactly as it stood after PR #7446's first commit makes it fail
- [ ] Wired into CI (and the failure message names the field and its write sites)

blamechris:
## The class

A field on the app store's `_ctx` (`packages/app/src/store/message-handler.ts`) can lose its last
reader — a guard is deleted, a branch simplified, a handler moved to store-core — while every
`_ctx.field = ...` survives. The result compiles, type-checks, lints and passes every test:
**writing to a field nobody reads is indistinguishable from working code.** TypeScript has nothing
to say either — `noUnusedLocals` does not reach interface members, and the field *is* "used", by
its own writes.

It has happened twice in a row in the same file:

| | field | shape |
|---|---|---|
| #7421 | `isSessionSwitchReplay` | 5 write sites, 0 readers — found by a manual archaeology pass |
| PR #7446, first commit (b31d01ca) | `pendingSwitchSessionId` | the **fix** for #7421 deleted its only reader and left the setter + 3 clears; caught only by the review panel, fixed in 052227e3 |

Two instances, the second created while removing the first, with nothing automated in between.

## The heuristic, and what it cannot see

For each field declared on `MessageHandlerContext` (and on base interfaces declared in the same
file — `EncryptionContext`), every `_ctx.<field>` / `ctx.<field>` reference in the app's **non-test**
sources is classified:

- **WRITE** — assignment target: `=` (not `==`/`===`/`=>`), any compound assignment (`+=`, `??=`,
  `>>>=`, …), `++`/`--` in either position, `delete`.
- **READ** — everything else: conditions, right-hand sides, property/method access through the field
  (`_ctx.set.clear()` must read `_ctx.set` first), template interpolations, arguments.

**≥1 write and 0 reads → fail**, naming the field and every write site.

Deliberate design choices, each stated in the script header rather than left implicit:

- **Comments are stripped before classification** — a commented-out reader must not mask a
  write-only field, and commenting out the dead reader is exactly how this arrives. The lexer leaves
  string, template-literal and regex-literal content intact (`${_ctx.field}` is a genuine read;
  `/a\/\/b/` must not be read as a comment).
- **Whole-file matching, never line-anchored.** `_ctx\n  .field` and `_ctx?.field` count. A
  line-anchored grep undercounts exactly this shape.
- **Test files are excluded from BOTH sides.** Not hypothetical: every mock context under
  `src/__tests__/store/` carried `isSessionSwitchReplay: false` for the whole life of that bug. A
  field read only by a test is still write-only in production.
- **Zero references warns, and does not fail.** Zero is the signature of both "dead field" and
  "reached some way this regex cannot see", and the live tree contains the second: `rttSmoother` is
  held on the context so a reset replaces it, while its two consumers close over the *local*
  `const rttSmoother` it was built from. It is never written either, so it is not this class, and
  failing on it would buy an allowlist entry that suppresses a whole category rather than a case.
  It is emitted as a GitHub `::warning::` on every run so it stays visible.
- **Not seen:** aliasing (`const c = _ctx; c.field`), `Object.assign`, spread copies,
  `_ctx['field']`, and destructuring reads. Aliasing lands in the warning bucket (safe direction); a
  field whose *only* reader destructures would be a false positive, resolved by an allowlist entry.
- **No new dependencies.** `scripts/` sits outside every workspace package and installs nothing —
  the Scripts Tests job does not run `npm ci` — so a real TS parse is not available here. The
  heuristic is honest about that rather than pretending to a parse it does not have.

**Allowlist:** entries require a justification string and expire — an entry naming a field that no
longer exists, or one whose field regained a reader, is refused as stale. **There are none today.**

**Fail-loud:** every "cannot check" is **exit 2, never exit 0** — missing declaring file, interface
not found (renamed / moved / converted to a type alias), unterminated body, **zero fields
extracted**, empty scan set, malformed allowlist entry. "Cannot check" silently read as "nothing to
check" is the catalogued false-safety shape (`docs/false-safety-guards.md`) and is the only way this
lint could be green while checking nothing.

## Red first

Proven red by restoring the real regressions in the working tree (byte backup + `cp` restore, every
scripted edit grep-verified, restored file SHA-256 re-checked against the backup).

**1. `git apply -R` of 052227e3 — PR #7446's first commit, exactly:**

```
::error::[write-only-ctx] app/MessageHandlerContext: MessageHandlerContext.pendingSwitchSessionId
  is WRITE-ONLY: 4 write site(s), 0 reads.
      write: packages/app/src/store/message-handler.ts:770
      write: packages/app/src/store/message-handler.ts:775
      write: packages/app/src/store/message-handler.ts:1861
      write: packages/app/src/store/message-handler.ts:2326
EXIT=1
```

The `#3121` comment block at line 2324 *mentions* the field; it is stripped, and correctly does not
count as a read.

**2. `git apply -R` of b31d01ca as well — the pre-#7446 tree, which is the two-sided control:**

```
::error::[write-only-ctx] app/MessageHandlerContext: MessageHandlerContext.isSessionSwitchReplay
  is WRITE-ONLY: 5 write site(s), 0 reads.
      write: .../message-handler.ts:777, 1864, 2335, 2455, 2581
EXIT=1
```

At that commit `pendingSwitchSessionId` still had its reader (the
`_ctx.pendingSwitchSessionId === switched.newSessionId` comparison) — and the lint **passes** it.
The same field is clean at one commit and red at the next, which is the sharpest false-positive
control available.

**3. Restored:** file SHA-256 identical to the pre-mutation backup, `git status` clean, lint exit 0.

## Controls and stats on current `main`

```
[write-only-ctx] app/MessageHandlerContext: 17 field(s), 111 source file(s),
  122 reference(s) classified — 16 read, 0 write-only, 1 unreferenced, 0 allowlisted.
```

- **Positive control:** exit 0 on `main` unmutated.
- **False-positive control (`--verbose`):** `replayingSessions` 10 reads / 0 writes;
  `postPermissionSplits` 5 / 0; `deltaIdRemaps` 8 / 0; `messageQueue` 10 / 0; `deltaFlusher` 8 / 0;
  `heartbeat` 7r/1w; `pendingKeyPair` 9r/9w; `pendingSalt` 7r/8w; `encryptionState` 7r/4w;
  `lastLatencyLogAt` 2r/2w; `terminalWriteTimer` 5r/3w; `pendingTerminalWrites` 2r/3w;
  `serverWillBootstrap` 2r/1w; `deltaServerTs` 5r/0w; `tokenToRender` / `clientRender` 2r/0w each.
  16 of 17 fields carry ≥1 read; the 17th is `rttSmoother` (the closure case above).
- **17 fields** = 14 declared on `MessageHandlerContext` + 3 inherited from `EncryptionContext`.

## The lint's own test

`scripts/__tests__/lint-write-only-ctx-fields.test.mjs` — 50 cases, no framework, exit-code based,
matching the sibling harnesses, with a `MIN_CASES` floor so a case that stops being discovered
breaks the run rather than shrinking it (#7447).

Covers: both real regressions as fixtures; a commented-out reader not rescuing a field; a test-file
reader not rescuing it; 11 read shapes and 7 write shapes individually; comment-lexer cases (a `//`
in a string, an escaped `//` in a regex, division vs regex, offset/line preservation); interface
extraction (nested object-type members, method and index signatures excluded, `extends` followed,
an imported base skipped); **every cannot-check path** (empty interface, missing interface,
unterminated body, empty scan set) at both the function and CLI-exit-code level; all four allowlist
refusals; the unreferenced-warns behaviour; and CLI exit codes 0/1/2 end to end against fixture
trees.

**Section 0 runs before this file imports the module, deliberately.** The module ends in
`process.exit(runCli())` under `isEntryPoint()`. If that guard read TRUE on a plain import, a static
import at the top would run the lint against the real repo and exit — before a single case ran,
printing nothing, exiting 0. A green run and a deleted suite would be identical (the "an entry-point
call site needs its own test file" trap, #7236). So the two call-site cases run out of process
first, behind a hard failure gate, and only then is the module imported. Mutating the guard to a
literal `true` now fails **by name** instead of erasing the run — verified.

**11 mutants, all killed** (module restored byte-identically after each; every mutation asserted to
have applied at exactly one site before the run):

| mutant | killed by |
|---|---|
| `isWrite` hardwired false | every RED case |
| comment stripping disabled | "a commented-out reader does not rescue" |
| zero-field extraction returns `[]` | empty-interface cannot-check + CLI exit 2 |
| allowlist justification unchecked | "an entry WITHOUT a justification is refused" |
| stale-allowlist check disabled | "an entry for a field that regained a reader" |
| empty scan set accepted as clean | "an EMPTY scan set is a cannot-check" |
| unterminated-body throw removed | "an UNTERMINATED interface body" |
| interface-not-found returns a stub | "a MISSING interface is a cannot-check" |
| test-path filter removed | "CLI ignores a reader in a test file" |
| optional-chaining / multi-line separators dropped from the matcher | those two read-shape cases |
| `isEntryPoint` stuck TRUE | section 0's probe |

## Where it runs

Two named steps in the **Scripts Tests** job (`.github/workflows/ci.yml`) — the test, then the lint
itself. That job has `setup-node` and deliberately no `npm ci`, which is exactly what a pure-node
dependency-free source scan needs; it is the same reason `compile-skill-targets --check` and the
AGENTS.md sync gate live there rather than in a package job. `ruby -ryaml` parse-checked, and
`packages/server/tests/ci-{workflow-reader,npm-cache-routing,cache-key}.test.js` re-run green
(30/30) against the edited workflow.

Also verified against the committed tree: `scripts/lint-no-nul-bytes.sh` clean, and
`packages/server/scripts/lint-entry-point-guard.mjs` clean (1989 files walked, 6 sanctioned).

## Deviations from the issue

- The issue's "and the dashboard's module-level `_ctx`-style state" is **not** covered: the
  dashboard keeps its per-connection state in module-level `let` bindings, not on a context object,
  so there is no interface to read fields off and its guard would be a different shape. The
  `TARGETS` table in the script is the place a second entry goes.
- The issue asks for a failure on zero-read fields generally; this fails on **write-only** and
  **warns** on zero-reference, for the `rttSmoother` reason above.

Closes #7452

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1266,6 +1266,25 @@ jobs:
       - name: Run lint-no-nul-bytes golden test
         run: bash scripts/__tests__/lint-no-nul-bytes.test.sh
 
+      - name: Run lint-write-only-ctx-fields tests
+        run: node scripts/__tests__/lint-write-only-ctx-fields.test.mjs
+
+      - name: No write-only fields on the app's MessageHandlerContext
+        # #7452 — a field on the app store's `_ctx` that is written but never
+        # read compiles, type-checks and passes every test: writing to a field
+        # nobody reads is indistinguishable from working code, and
+        # `noUnusedLocals` does not reach interface members. It has happened
+        # twice in a row in the same file — #7421's `isSessionSwitchReplay`,
+        # then `pendingSwitchSessionId`, orphaned by the very PR removing the
+        # first and caught only by a review pass.
+        #
+        # It runs HERE, not in an App job, because it is a pure-node source scan
+        # with no dependencies: this job has setup-node and deliberately no
+        # `npm ci`, which is the same reason `compile-skill-targets --check`
+        # lives here. Exit 2 (cannot check) is a distinct, loud outcome from
+        # exit 0.
+        run: node scripts/lint-write-only-ctx-fields.mjs
+
       - name: Check compiled skill artifacts are in sync with .claude/commands
         # #7253 — .claude/commands/<name>.md is the neutral source, but
         # .claude/skills/<name>/SKILL.md is what Claude actually LOADS (the
--- a/scripts/__tests__/lint-write-only-ctx-fields.test.mjs
+++ b/scripts/__tests__/lint-write-only-ctx-fields.test.mjs
@@ -38,7 +38,7 @@ const SCRIPT = resolve(HERE, '..', 'lint-write-only-ctx-fields.mjs')
 
 // Every case in this file. Bump it when you add one — a case that vanishes
 // should break the run rather than quietly shrink it (#7447).
-const MIN_CASES = 50
+const MIN_CASES = 53
 
 let pass = 0
 let fail = 0
--- a/scripts/__tests__/lint-write-only-ctx-fields.test.mjs
+++ b/scripts/__tests__/lint-write-only-ctx-fields.test.mjs
@@ -505,6 +505,32 @@ test('CLI reports the classification stats it acted on', () => {
   assert(/1 field\(s\), 1 source file\(s\), 2 reference\(s\) classified/.test(r.stdout), r.stdout)
 })
 
+// --- #7464 review C1/S3: the classification windows and the quote lexer ----
+
+test('a comment-padded write still classifies as a WRITE (C1: the 8-byte window)', () => {
+  // Stripping a comment leaves blanks between the reference and `=`; the old
+  // 8-byte lookahead filed this as a READ, and one rescued write silences a
+  // whole write-only field (demonstrated on the real #7421 regression).
+  const stmt = '_ctx.n /* review on #7464, a padded write */ = 1;'
+  const { reads, writes } = classifyReferences(stripComments(stmt), 'n', RECEIVERS)
+  assert(writes.length === 1 && reads.length === 0, `padded write misclassified: ${reads.length}r ${writes.length}w`)
+})
+
+test('a comment-padded delete still classifies as a WRITE (C1: the lookbehind window)', () => {
+  const stmt = 'delete /* the mirror case */ _ctx.n;'
+  const { reads, writes } = classifyReferences(stripComments(stmt), 'n', RECEIVERS)
+  assert(writes.length === 1 && reads.length === 0, `padded delete misclassified: ${reads.length}r ${writes.length}w`)
+})
+
+test("a lone apostrophe in JSX prose cannot open a blind span (S3: quote spans stop at newline)", () => {
+  // Live shape from CreateSessionModal.tsx: `server's` has no closing quote on
+  // its line. The old lexer swallowed everything to the next apostrophe —
+  // including a following `// _ctx.n = 1` line the stripper should have blanked.
+  const src = "const label = <p>the server's daemon</p>;\n// _ctx.n = 1\n_ctx.n = 2;\n"
+  const { reads, writes } = classifyReferences(stripComments(src), 'n', RECEIVERS)
+  assert(writes.length === 1 && reads.length === 0, `blind span altered classification: ${reads.length}r ${writes.length}w`)
+})
+
 // ---------------------------------------------------------------------------
 
 for (const d of tmpDirs) rmSync(d, { recursive: true, force: true })
--- a/scripts/__tests__/lint-write-only-ctx-fields.test.mjs
+++ b/scripts/__tests__/lint-write-only-ctx-fields.test.mjs
@@ -1,0 +1,524 @@
+#!/usr/bin/env node
+/**
+ * lint-write-only-ctx-fields.test.mjs — pins scripts/lint-write-only-ctx-fields.mjs (#7452).
+ *
+ * The lint exists because a write-only context field is INVISIBLE: it compiles,
+ * type-checks and passes every test. Its own failure mode is the same shape —
+ * a classifier that quietly stops matching, or an extractor that quietly finds
+ * zero fields, reports "clean", and is indistinguishable from a healthy run. So
+ * this suite is built around four kinds of case:
+ *
+ *   1. RED — a write-only field must fail, and the failure must NAME the field
+ *      and its write sites. Anchored on the two real regressions (#7421's
+ *      `isSessionSwitchReplay`, PR #7446's `pendingSwitchSessionId`).
+ *   2. GREEN — a field with any reader must pass. Every read SHAPE the app
+ *      actually uses gets its own case, because "no reads found" is how this
+ *      lint produces a false positive, and one missed shape would fail a
+ *      legitimate field.
+ *   3. CANNOT-CHECK — a missing interface, an empty interface, an unterminated
+ *      body, an empty scan set and a malformed allowlist must each exit 2.
+ *      Never 0. "Cannot check" silently read as "nothing to check" is the
+ *      catalogued false-safety shape (docs/false-safety-guards.md) and is the
+ *      only way this lint could be green while checking nothing.
+ *   4. The isEntryPoint CALL SITE, in both directions — see section 0, which
+ *      runs BEFORE this file imports the module and explains why it must.
+ *
+ * No external test framework. Run from repo root:
+ *   node scripts/__tests__/lint-write-only-ctx-fields.test.mjs
+ */
+
+import { spawnSync } from 'node:child_process'
+import { mkdirSync, mkdtempSync, rmSync, writeFileSync } from 'node:fs'
+import { tmpdir } from 'node:os'
+import { dirname, join, resolve } from 'node:path'
+import { fileURLToPath, pathToFileURL } from 'node:url'
+
+const HERE = dirname(fileURLToPath(import.meta.url))
+const SCRIPT = resolve(HERE, '..', 'lint-write-only-ctx-fields.mjs')
+
+// Every case in this file. Bump it when you add one — a case that vanishes
+// should break the run rather than quietly shrink it (#7447).
+const MIN_CASES = 50
+
+let pass = 0
+let fail = 0
+const failures = []
+
+const test = (name, fn) => {
+  try {
+    fn()
+    pass++
+    process.stdout.write(`  ok ${name}\n`)
+  } catch (err) {
+    fail++
+    failures.push({ name, err })
+    process.stdout.write(`  FAIL ${name}: ${err.message}\n`)
+  }
+}
+
+const assert = (cond, msg) => {
+  if (!cond) throw new Error(msg || 'assertion failed')
+}
+
+const throws = (fn, Type, match) => {
+  let caught = null
+  try {
+    fn()
+  } catch (err) {
+    caught = err
+  }
+  assert(caught !== null, 'expected a throw, got none')
+  assert(caught instanceof Type, `expected ${Type.name}, got ${caught.constructor.name}: ${caught.message}`)
+  if (match) assert(match.test(caught.message), `message did not match ${match}: ${caught.message}`)
+}
+
+// ---------------------------------------------------------------------------
+// Fixture trees + CLI driver (needed by section 0, so defined before it)
+// ---------------------------------------------------------------------------
+
+const DECL_REL = 'packages/app/src/store/message-handler.ts'
+const tmpDirs = []
+
+function fixtureRoot(declText, extra = {}) {
+  const dir = mkdtempSync(join(tmpdir(), 'chroxy-woctx-'))
+  tmpDirs.push(dir)
+  mkdirSync(join(dir, dirname(DECL_REL)), { recursive: true })
+  writeFileSync(join(dir, DECL_REL), declText)
+  for (const [rel, text] of Object.entries(extra)) {
+    mkdirSync(join(dir, dirname(rel)), { recursive: true })
+    writeFileSync(join(dir, rel), text)
+  }
+  return dir
+}
+
+const runCliOn = (dir, ...args) =>
+  spawnSync(process.execPath, [SCRIPT, '--root', dir, ...args], { encoding: 'utf8' })
+
+const CLEAN_DECL = `
+interface MessageHandlerContext {
+  flag: boolean;
+}
+let _ctx: MessageHandlerContext = { flag: false };
+export function run(): void { _ctx.flag = true; if (_ctx.flag) console.log('x'); }
+`
+const WRITE_ONLY_DECL = `
+interface MessageHandlerContext {
+  flag: boolean;
+}
+let _ctx: MessageHandlerContext = { flag: false };
+export function run(): void { _ctx.flag = true; }
+`
+
+// ---------------------------------------------------------------------------
+// 0. The isEntryPoint CALL SITE — and why it is FIRST, before the import below.
+//
+// The module ends in `process.exit(runCli())` under `isEntryPoint()`. If that
+// guard ever read TRUE on a plain import, the static import this file would
+// otherwise open with would run the lint against the real repo and exit —
+// before a single case ran, printing nothing, exiting 0 on a clean tree. A
+// green run and a DELETED suite would be the same observable outcome, which is
+// exactly the trap recorded as "an entry-point call site needs its own test
+// file" (#7236): a file that imports the module cannot witness it auto-running.
+//
+// The resolution here is ordering rather than a second file. These two cases
+// run out of process, the failure gate below is hard (exit 1 immediately), and
+// only then does this file import the module. Mutating the guard to a literal
+// `true` therefore fails HERE, by name, instead of erasing the run.
+// ---------------------------------------------------------------------------
+
+test('importing the module does NOT run the lint (a guard stuck TRUE would erase this suite)', () => {
+  const dir = mkdtempSync(join(tmpdir(), 'chroxy-woctx-probe-'))
+  tmpDirs.push(dir)
+  const probe = join(dir, 'probe.mjs')
+  writeFileSync(
+    probe,
+    `import { TARGETS } from ${JSON.stringify(pathToFileURL(SCRIPT).href)}\n` +
+    "process.stdout.write('IMPORT-RETURNED:' + TARGETS.length + '\\n')\n",
+  )
+  const r = spawnSync(process.execPath, [probe], { encoding: 'utf8' })
+  assert(r.status === 0, `probe exited ${r.status}\n${r.stdout}${r.stderr}`)
+  assert(
+    /IMPORT-RETURNED:1/.test(r.stdout),
+    `the import never returned — the module ran and exited on import\n${r.stdout}${r.stderr}`,
+  )
+  assert(
+    !/\[write-only-ctx\]/.test(r.stdout + r.stderr),
+    `importing the module produced lint output\n${r.stdout}${r.stderr}`,
+  )
+})
+
+test('running the module directly DOES run it — positive control for the case above', () => {
+  // Without this, "no lint output on import" would pass just as happily against
+  // a module that produces no output at all, ever.
+  const r = runCliOn(fixtureRoot(CLEAN_DECL))
+  assert(/\[write-only-ctx\]/.test(r.stdout), `direct run produced no lint output\n${r.stdout}${r.stderr}`)
+})
+
+if (fail > 0) {
+  for (const f of failures) process.stdout.write(`\n--- ${f.name}\n${f.err.stack}\n`)
+  process.stdout.write(
+    '\nFAIL: the entry-point call site is broken. Stopping BEFORE importing the module, ' +
+    'because a guard stuck TRUE would exit this process during that import.\n',
+  )
+  process.exit(1)
+}
+
+const {
+  CannotCheckError,
+  analyzeTarget,
+  classifyReferences,
+  extractInterfaceFields,
+  stripComments,
+} = await import(pathToFileURL(SCRIPT).href)
+
+// ---------------------------------------------------------------------------
+// Fixture builder — a minimal stand-in for message-handler.ts
+// ---------------------------------------------------------------------------
+
+const RECEIVERS = ['_ctx', 'ctx']
+
+/** Analyse a single in-memory module as both the declaration and the source. */
+const analyzeOne = (text, opts = {}) =>
+  analyzeTarget({
+    declText: text,
+    interfaceName: 'Ctx',
+    receivers: RECEIVERS,
+    sources: [{ path: 'fixture.ts', text }],
+    ...opts,
+  })
+
+const withField = (decl, body) => `
+interface Ctx {
+  ${decl}
+  keep: number;
+}
+function build(): Ctx {
+  return { ${decl.split(':')[0].trim()}: null as never, keep: 0 };
+}
+let _ctx: Ctx = build();
+export function run(): void {
+${body}
+  void _ctx.keep;
+}
+`
+
+// ---------------------------------------------------------------------------
+// 1. RED — write-only fields fail
+// ---------------------------------------------------------------------------
+
+test('a field with writes and no reads FAILS and names the field', () => {
+  const r = analyzeOne(withField('flag: boolean;', '  _ctx.flag = true;\n  _ctx.flag = false;'))
+  assert(r.failures.length === 1, `expected 1 failure, got ${r.failures.length}`)
+  assert(/Ctx\.flag is WRITE-ONLY/.test(r.failures[0]), r.failures[0])
+})
+
+test('the failure lists EVERY write site with a line number', () => {
+  const r = analyzeOne(withField('flag: boolean;', '  _ctx.flag = true;\n  _ctx.flag = false;'))
+  const sites = r.failures[0].match(/write: fixture\.ts:\d+/g) || []
+  assert(sites.length === 2, `expected 2 write sites, got ${sites.length}: ${r.failures[0]}`)
+})
+
+test('#7421 shape: isSessionSwitchReplay written in four handlers, read nowhere', () => {
+  const src = `
+interface Ctx {
+  isSessionSwitchReplay: boolean;
+  replayingSessions: Set<string>;
+}
+let _ctx: Ctx = { isSessionSwitchReplay: false, replayingSessions: new Set() };
+export function reset(): void { _ctx.isSessionSwitchReplay = false; _ctx.replayingSessions.clear(); }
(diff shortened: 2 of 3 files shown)


blamechris:
## Independent adversarial review — attacking the classifier

Reviewed at head `ffe4972f7`, Node v22.22.3 / darwin-arm64, from the PR worktree. Every mutation ran on `/tmp` tree copies or on a `/tmp` copy of `scripts/`; the worktree and the main checkout were never written to, and **`git status --short` is empty in both** at the end. Real-tree copies were built by `git apply -R` of the actual commits rather than by hand-written fixtures, so the red proofs below are the regressions, not stand-ins for them.

**Every claim in the PR body reproduced.** Both headline red proofs came back byte-exact, line numbers included; the two-sided control is as sharp as advertised; all the fail-loud arms fire against *real-tree* mutations, not just fixtures; section 0's entry-point probe genuinely catches a stuck-TRUE guard; and `TARGETS` is extensible rather than decorative — I pointed a second entry at a real interface across three packages and it ran clean.

**One blocking finding, and it is one number.** `classifyReferences` decides write-vs-read from a fixed 8-byte slice after the match. Any write whose `=` lands past byte 8 is classified as a **read** — the silent false-pass direction, the one the header says must never happen. A *single* such write rescues the whole field. I flipped the actual `pendingSwitchSessionId` regression from red to green by adding an inline comment to one of its four writes.

### Verified independently (re-run, not re-read)

**Red proof 1 — `git apply -R` of `052227e3`** (`/tmp` copy of `packages/app/src`, app hunks only):

```
::error::… MessageHandlerContext.pendingSwitchSessionId is WRITE-ONLY: 4 write site(s), 0 reads.
      write: packages/app/src/store/message-handler.ts:770 / :775 / :1861 / :2326
EXIT=1
```

Identical to the PR body, line numbers included. The `#3121` comment block at 2324 mentions the field and correctly does not count.

**Red proof 2 — the pre-#7446 tree** (`-R 052227e3` then `-R b31d01ca`): `isSessionSwitchReplay` red at 5 write sites (777 / 1864 / 2335 / 2455 / 2581), **and `pendingSwitchSessionId` green at `2r 4w` on the same run**. The two-sided control holds — the same field is clean at one commit and red at the next, which is the strongest false-positive evidence available here and it is real.

**Positive control + stats.** Clean tree, exit 0, `17 field(s), 111 source file(s), 122 reference(s) — 16 read, 0 write-only, 1 unreferenced, 0 allowlisted` — identical to the PR body. Per-field `--verbose` breakdown matches all sixteen listed figures.

**Fail-loud arms, exercised against the REAL tree rather than fixtures** (the suite only covers these on synthetic roots):

| real-tree mutation | exit | message |
|---|---|---|
| `interface MessageHandlerContext` → `HandlerCtx` | **2** | `CANNOT CHECK — interface … not found — renamed, moved, or converted to a type alias?` |
| converted to `type MessageHandlerContext = EncryptionContext & {…}` | **2** | same arm |
| declaring file renamed to `mh.ts` | **2** | `cannot read packages/app/src/store/message-handler.ts (ENOENT)` |
| whole `packages/app/src` removed | **2** | same |
| scan set reduced to the decl file alone | 1 | correct — the field really is write-only in that scan set |

None of these is reachable as an exit 0. The `--root` fallback to `process.cwd()` on a missing value is benign for the same reason: the decl file is not there, so it exits 2 anyway.

**Section 0's entry-point claim — verified the hard way.** Copied `scripts/` *and* `packages/app/src` to `/tmp` so `REPO_ROOT` resolved to a clean tree, meaning a stuck guard would run the lint, print real output and exit **0** — the case that would otherwise be indistinguishable from a passing run. Mutated `if (isEntryPoint(import.meta.url))` → `if (true)`:

```
  FAIL importing the module does NOT run the lint (a guard stuck TRUE would erase this suite):
       the import never returned — the module ran and exited on import
[write-only-ctx] app/MessageHandlerContext: 17 field(s) … OK
harness exit = 1, 1 case ran before the gate stopped it
```

Fails by name, stops before the import, exactly as documented. Restored → 50/50, exit 0.

**CI wiring.** `ruby -ryaml` parses `ci.yml`; both steps are steps 11 and 12 of the `scripts-tests` job (`name: Scripts Tests`), after `actions/checkout` + `setup-node@22`, and the job has **no `npm ci`** — confirmed by walking the parsed step list, not by grep. The lint imports only `node:fs` / `node:path` / `node:url` plus `./lib/is-entry-point.mjs`, which imports the same three. Checkout alone is sufficient.

**`TARGETS` is real, not decorative.** Added a second entry in a `/tmp` copy — `ConnectionContext` from `packages/store-core/src/types/connection.ts:75`, `receivers: ['ctx','_ctx']`, `scanDirs` spanning dashboard + app + store-core — and it ran with no other change: `5 field(s), 416 source file(s), 57 reference(s) — 5 read, 0 write-only`, exit 0 for both targets.

**Scan-set completeness.** 111 of 248 `.ts`/`.tsx` under `packages/app/src`; every one of the 137 exclusions matches `TEST_PATH` on a genuine `__tests__/` path or `.test.`/`.spec.` suffix. The two `.js` files there reference no `ctx` receiver, so `SOURCE_EXT` costs nothing. `packages/app/src/test-utils/` is *not* excluded by `TEST_PATH`, but contains no `ctx` receiver either — no masking today, worth knowing. The header's "no such string exists today" holds: every string-context hit is a template interpolation (a genuine read) or a JSDoc line (stripped).

**`MIN_CASES = 50` is a true floor** — 50 cases discovered, 50 run.

### Issues

#### Critical

| # | file:line | issue |
|---|---|---|
| C1 | `scripts/lint-write-only-ctx-fields.mjs:375` | **The fixed 8-byte lookahead turns a WRITE into a READ, and one such write rescues the whole field.** `const after = strippedText.slice(end, end + 8)` — if the `=` lands at byte 8 or later, `ASSIGN_AHEAD` never sees it and the reference is filed as a **read**. Two reachable spellings, and the first is caused by this PR's own design choice: comments are blanked to spaces *preserving byte offsets*, so `_ctx.f /* why */ = x` becomes `_ctx.f` + 9 spaces + `=`. The second is a continuation-line assignment indented ≥ 7 columns. Demonstrated on the actual regression — see below. Direction is the silent false pass, no annotation, exit 0. **No test pins either window**: widening `end + 8` → `end + 64` *and* `m.index - 12` → `m.index - 64` leaves the suite at **50/50** and the real tree at exit 0, so nothing pushes back on the value at all. |

The demonstration, on the `-R 052227e3` tree (`pendingSwitchSessionId`, 4 writes / 0 reads, correctly **exit 1**):

```diff
  export function setPendingSwitchSessionId(id: string | null): void {
-   _ctx.pendingSwitchSessionId = id;
+   _ctx.pendingSwitchSessionId /* #7421 */ = id;
  }
```

```
[write-only-ctx] … 17 read, 0 write-only, 1 unreferenced
    1r   3w  pendingSwitchSessionId
[write-only-ctx] OK — no context field is written without a reader.
EXIT=0
```

One comment on one of four writes, and the exact field this lint was built to catch passes silently. The continuation-line form does the same; I mapped the boundary:

| `_ctx.pendingSwitchSessionId\n<N spaces>= id;` | 4 | 6 | **7** | 8 | 9 |
|---|---|---|---|---|---|
| lint exit | 1 | 1 | **0** | 0 | 0 |

`DELETE_BEHIND` has the mirror-image bug on the 12-byte `before` window: `delete` + ≥ 7 spaces + `_ctx.f` classifies as a read (`reads=1 writes=0`).

This also makes the header's own claim stronger than the code, which is the `#7290`/`#7291` shape the repo catalogues: *"Multi-line references (`_ctx\n .field`) and optional chaining ARE handled: each file is matched as one string with newline-tolerant separators, never line by line."* True of the reference **matcher**; false of the **classifier**, which is a fixed byte window. A reader who trusts that sentence will assume a multi-line write is seen. It is not, past 6 columns.

**Fix, verified.** `end + 8` → `end + 64` alone is sufficient and safe: 50/50 green, real tree exit 0 with the same 122/16/0/1 stats (no new false positives), and both demonstrations above go red. Scanning forward to the first non-space character instead of slicing a constant would be better still — the window then cannot be a magic number that drifts. Please also add the two cases (`_ctx.f /* c */ = 1` and a deep continuation) to section 3, so the window is pinned by something.

#### Suggestions

**S1 — `_ctx.f = _ctx.f + 1` counts its own right-hand side as a read, negating a rule the header states outright.** The header is explicit: *"`++`/`--` counts as a write and NOT as a read, deliberately. A field whose only consumer is its own increment has no reader in any sense that matters."* The desugared form defeats it — `classifyReferences` returns `reads=1 writes=1`, so the field is self-sustaining forever. Same for the append idiom `_ctx.q = [..._ctx.q, x]` and `_ctx.x = _ctx.x ?? d`. I grepped: **no self-referential `_ctx` write exists on the tree today**, so this is latent, not a live miss — but it belongs in the "WHAT IT CANNOT SEE" list, because a stated rule silently not holding is worse than an admitted gap. A cheap partial: don't count a read that sits on the RHS of an assignment whose target is the same `receiver.field`.

**S2 — destructuring and `for-of` writes are classified as READS, not as "not seen".** Measured:

| shape | result |
|---|---|
| `({ f: _ctx.f } = obj)` | `reads=1 writes=0` |
| `[_ctx.f] = arr` | `reads=1 writes=0` |
| `for (_ctx.f of xs)` | `reads=1 writes=0` |
| `Object.assign(_ctx, { f: 1 })` | `0/0` → warn (documented) |
| `const c = _ctx; c.f = 1` | `0/0` → warn (documented) |
| `_ctx['f'] = 1` | `0/0` → warn (documented) |

The last three are in the gap list and land in the warn bucket, which is the safe direction. The first three are **not** in the gap list and land in the unsafe one — a phantom read. None occurs in the app today. One line in the header, and ideally an "assignment-context I cannot resolve" bucket rather than a silent read.

**S3 — the lexer mis-lexes TSX, and there is a live 7-line blind span on `main` today.** The `'`/`"` branch (`stripComments`, L207-218) skips to the next matching quote with **no newline stop**, so a raw apostrophe in JSX text opens a multi-line span in which `//` comments are never stripped. `packages/app/src/components/CreateSessionModal.tsx:251` does exactly this:

```tsx
<Text style={styles.hint}>Leave empty to use the server's default directory</Text>
```

The lexer does not resume real lexing until **line 258**. I injected `// if (_ctx.demoField) doThing()` into that span and the reference **survives comment-stripping** — i.e. a commented-out reader in there would count as a genuine read, which is precisely what comment-stripping exists to prevent. `</Text>` and `/>` at a line start hit the same class via the regex-literal branch (`\n` is in `REGEX_MAY_FOLLOW`).

Exposure is nil today and I measured it rather than assuming: across all 111 scanned files exactly **23** `//` lines survive stripping, and all 23 are inside `xterm-html.ts`'s embedded-HTML template literal — correct behaviour, not damage. And **zero** counted references on the real tree sit after a `//` on their line. But `.tsx` is in the scan set and the next `TARGETS` entry may well put a receiver there. Cheapest hardening: break the `'`/`"` scan at a newline (a real JS string cannot contain a raw one), which caps a JSX apostrophe at one line instead of seven.

#### Nitpicks

**N1 — the PR body over-claims on aliasing.** *"Aliasing lands in the warning bucket (safe direction)"* holds only when **every** reference is aliased. Writes through `_ctx` with reads through an alias lands in the **failure** bucket — a false positive; the reverse passes silently. The script header is careful about this ("invisible unless `c` is listed in the target's `receivers`"); the PR body is not.

**N2 — the dashboard-exclusion reasoning is inaccurate, though the conclusion is right.** Both the header (L128-132) and the PR body say the dashboard *"keeps its per-connection state in module-level `let` bindings rather than on a context object, so there is no interface for this lint to read fields off."* There is one: `ConnectionContext`, declared at `packages/store-core/src/types/connection.ts:75` and threaded as `ctx`/`_ctx` through every dashboard handler — that is how I built the second-target probe above. What actually makes it uninteresting is that it holds connection *identity* (`url` / `token` / `isReconnect` / `silent` / `socket`), not mutable state; the mutable state really is in module-level `let`/`const` bindings (`_store`, `_evaluatorPending`, `_pendingTrustGrants`, …). Suggest rewording to say that, since "there is no interface" is checkable and false.

**N3 — `MIN_CASES` sits exactly on the current count (50/50).** Correct as a floor, but it means adding a case without bumping the constant is silently allowed while removing one is caught. That is the intended asymmetry; noting it only so the next editor does not read 50 as "the number of cases".

### Follow-ups filed

- **#7467** — the dashboard half of #7452 (module-level write-only connection state) is explicitly deferred by this PR while `Closes #7452` closes the issue that asked for it. Filed so it is not dropped unrecorded.
- **#7468** — `rttSmoother` is held on the interface but consumed through a closure over the local, so it is a permanent `::warning::` on every run. A guard whose output always contains a warning trains people to skip the warnings.

### Verdict

- [ ] Approve
- [ ] Approve with comments
- [x] **Request changes** — C1 only

Everything else here is genuinely good work: the fail-loud design is real and I could not find an exit-0 path through any "cannot check", the two-sided control is the sharpest evidence this kind of heuristic can offer, section 0 is correct and correctly justified, and the gap list is unusually honest. C1 is blocking for the reason this PR itself argues: a guard whose green can be produced by an inline comment on one of a field's writes is a guard whose green means nothing, and the exact regression it was built to catch passes under it. The fix is one constant, I verified it end to end (50/50, real tree clean, both demonstrations red), and it wants two test cases so the window stops being unpinned. S1-S3 are in-scope to the file this PR adds and are small enough to fold in; N1/N2 are wording.

blamechris:
C1 and S3 folded at d5ff30a95:

- **C1**: both windows widened to 64 bytes with your rationale in a comment at the site, and **pinned** — a comment-padded write and a comment-padded `delete` case; narrowing the windows back fails exactly those two (51/53).
- **S3**: quote spans now stop at newline (template literals excepted); your CreateSessionModal shape is the pinned case — removing the stop fails exactly it (52/53).
- **S1/S2/N1**: documented as known gaps rather than fixed — the self-referential-RHS and destructuring-write shapes have zero live instances and their honest fix is a real parser; the gap list is the contract. **N2**: fair hit — the dashboard-exclusion comment's reasoning was wrong even though its conclusion holds; #7467 (your filing) is the right home for that half. **N3**: now 53 cases over the 50 floor.

53/53, lint exit 0 on the real tree, identical classification stats. Please re-run your tree-level demo (apply -R 052227e3 + the inline-comment rescue) at this head — it should now stay red.

blamechris:
## Re-review at `d5ff30a95` — C1 and S3 closed, verdict updated

Re-verified from the PR worktree at `d5ff30a95`, Node v22.22.3 / darwin-arm64. Same method as before: `/tmp` tree copies built by `git apply -R` of the real commit, every mutation occurrence-asserted and grep-verified as landed, the module restored from a byte backup (sha256 `bfac50be7d4a18d3…` identical before and after every mutant). **`git status --short` is empty in the PR worktree, the main checkout and my own worktree.**

Both fixes are real, and both are pinned by tests that fail under the original mutation. **Approving.**

### C1 — the demo that flipped green at `ffe4972f7` now stays red

Rebuilt the tree (`git apply -R 052227e3`, app hunks only) — baseline red, 4 write sites, exit 1. Then the exact rescue that defeated the previous head, landed at exactly one of the four sites:

```diff
  export function setPendingSwitchSessionId(id: string | null): void {
-   _ctx.pendingSwitchSessionId = id;
+   _ctx.pendingSwitchSessionId /* #7421 */ = id;
  }
```

| head | classification | exit |
|---|---|---|
| `ffe4972f7` | `1r 3w` | **0** — silently green |
| `d5ff30a95` | `0r 4w` | **1** — WRITE-ONLY, all four sites named |

Continuation-line boundary swept again, one mutation site asserted per run:

| indent | 4 | 6 | 7 | 8 | 12 | 20 | 40 | 55 | 60 | 63 | 64 | 70 |
|---|---|---|---|---|---|---|---|---|---|---|---|---|
| `ffe4972f7` | 1 | 1 | **0** | 0 | 0 | 0 | 0 | 0 | 0 | 0 | 0 | 0 |
| `d5ff30a95` | 1 | 1 | 1 | 1 | 1 | 1 | 1 | 1 | 1 | 0 | 0 | 0 |

The escape moves from 7 columns to 63, i.e. it now needs ≥64 bytes of stripped span between the reference and its operator. I measured the operative headroom rather than trusting the comment: across all 111 scanned files, the **largest real gap between any `_ctx.<field>` reference and its assignment operator is 1 byte**. 64x margin. (The comment's "any stripped span this repo produces" is loose read literally — there is a 76-byte single-line inline block comment in a `.tsx` render tree — but nothing of the sort sits anywhere near a `_ctx` assignment, so the operative claim holds.) The mirror case is fixed too: `delete` + 11 spaces + `_ctx.f` now classifies `0r 1w`, was `1r 0w`.

### S3 — the live blind span is closed

Re-ran the probe against the real `CreateSessionModal.tsx:251` (`the server's default directory`), injecting the same line into the span:

```
injected `// if (_ctx.demoField) doThing()` at line 253
  reference survives stripping : false      (was TRUE at ffe4972f7)
  the line, stripped           : "                                            "
  //-comment lines left intact in this file: 0   (was 7 lines of blind span)
```

No over-stripping regression. Tree-wide, exactly **23** `//` lines still survive `stripComments`, unchanged, and all 23 are inside `xterm-html.ts`'s embedded-HTML template literal — correctly exempted by the `c !== '\`'` guard. A `${_ctx.f}` inside a multi-line template still counts as a read, and an escaped-newline string continuation still works because the `\\` check precedes the new newline stop.

### The pins fail under the original mutations

Module restored byte-identically after each; each mutation asserted at exactly one site.

| mutant | result |
|---|---|
| lookahead `64` → `8` | **52/53** — kills exactly *a comment-padded write still classifies as a WRITE* (`1r 0w`) |
| lookbehind `64` → `12` | **52/53** — kills exactly *a comment-padded delete still classifies as a WRITE* (`1r 0w`) |
| both narrowed (the `ffe4972f7` state) | **51/53** — kills exactly those two |
| newline stop deleted from the quote branch | **52/53** — kills exactly *a lone apostrophe in JSX prose cannot open a blind span* |

Each mutant kills its own case and nothing else, which is what makes them pins rather than incidental coverage.

### Real tree unchanged

`53/53`, lint exit 0, and the classification is byte-identical to the previous head — `17 field(s), 111 source file(s), 122 reference(s) — 16 read, 0 write-only, 1 unreferenced, 0 allowlisted`, with every per-field `Nr Nw` figure the same. No new false positives. I re-ran the full 27-shape classification matrix; every previously-correct shape still classifies the same way.

### One correction, not a blocker

The reply says S1/S2/N1 are *"documented as known gaps rather than fixed."* **The header is byte-identical to `ffe4972f7`** — `git diff ffe4972f7..d5ff30a95 -- scripts/lint-write-only-ctx-fields.mjs` changes zero header lines, and the "WHAT IT CANNOT SEE" list still carries exactly its original five bullets. So the decision was taken but not yet written down:

- **S1** appears nowhere, and the header still makes the *stronger* positive claim at L44-45 — *"`++`/`--` counts as a write and NOT as a read, deliberately. A field whose only consumer is its own increment has no reader in any sense that matters."* `_ctx.f = _ctx.f + 1` still returns `1r 1w`, so the desugared form silently does not obey it. That is the same comment-stronger-than-code shape the PR exists to guard against, in the paragraph asserting the opposite.
- **S2**'s unsafe direction is not covered. The DESTRUCTURING bullet describes only `const { field } = _ctx` — the read side, the *safe* direction, framed as a false positive. `({ f: _ctx.f } = o)`, `[_ctx.f] = a` and `for (_ctx.f of xs)` all return `1r 0w`, a phantom read, and are unlisted.
- **N1** — the PR body still reads *"Aliasing lands in the warning bucket (safe direction)"*, which holds only when every reference is aliased.

I agree with the call not to fix these: zero live instances (re-confirmed at this head), and an honest fix is a parser. But the gap list is the contract the header itself points at — *"read this before trusting a green run"* — so a gap that is not in it is not documented. Three comment lines and one PR-body clause; in-scope and small enough to fold here rather than carry an issue.

**N3** — `MIN_CASES` is still `50` against 53 discovered cases. Defensible as a floor, and the constant's own comment says *"Bump it when you add one"*; the three cases that could now vanish unnoticed are precisely the three added to pin this fix. One character, entirely your call.

### Verdict

- [x] **Approve** at `d5ff30a95` — C1 and S3 are genuinely fixed, both pinned, and the demo that defeated the previous head now stays red
- [ ] Approve with comments
- [ ] Request changes

The blocking finding is closed the way it should be: the fix is at the site with the reason next to it, the window is pinned in both directions, and the escape moved from a shape any inline comment produces to one 64x beyond anything in the tree. S3 closed a hole I could demonstrate on `main`. The S1/S2/N1 header lines are the only thing I would still fold in before merge, and they are comments — nothing about them affects a green run today. Follow-ups #7467 and #7468 stand as filed.

blamechris:
Correction folded at fd2d8b5eb: the re-verification was right that my reply claimed documentation that did not exist — the header was byte-identical. The gap list now actually carries S1 (self-referential RHS), S2's write direction, and N1's mixed-alias limit, and `MIN_CASES` is 53 — the three cases that could vanish unnoticed at 50 are exactly the three pinning this round's fixes. 53/53, lint exit 0, stats unchanged.

Issue #7436, opened 2026-08-28
ws: rate-limit `session_pr_status_request` now that it mutates daemon watch state


blamechris:
## Context

Identified during review of PR #7432 (#7427).

## Description

`session_pr_status_request` has no server-side rate limit. Its `inFlight` guard is keyed on `(client, session)` and only bars **concurrent** surveys — back-to-back requests are unrestricted, and the dashboard's Refresh button passes no freshness window, so it re-surveys on every click.

That was acceptable while the handler was **read-only**: the cost was a `git` + `gh` spawn per click, bounded by one in flight per client per session.

#7427 changed the calculus — the handler now **mutates daemon-side watch state** via `SessionCiWatcher.observe()`. `observe()` is written to be safe under an un-throttled caller (it arms only, refuses an already-fired pair, and writes a deferral stamp that is deliberately *not* the sweep's sort key — both of those were defects caught in the same review), but "the callee is careful" is a weaker guarantee than a bound on the caller, and the next external ingest path will have to re-derive it.

Note `docs/security/bearer-token-authority.md` treats a pairing-bound token as authoritative for its own session, so a **per-session** throttle is the natural shape — it does not need to distinguish token classes.

## Original review comment

> There is no server-side rate limit on `session_pr_status_request` — `isInFlight` is per-`(client, session)` and only blocks *concurrent* surveys, not back-to-back ones. […] Pre-existing from #7422, but the endpoint now **mutates daemon watch state**, which changes the calculus.

## Acceptance criteria

- [ ] A per-session minimum interval between surveys, replying with a schema-valid snapshot carrying a `reason` when refused (never silence — the handler's one-reply-per-request property must hold)
- [ ] The interval does not defeat the Refresh button's purpose: a user waiting on CI can still get a fresh reading at a sane cadence
- [ ] Consider surfacing `maxSurveysPerTick` in the `sessionCi` config block alongside `intervalMs` / `discoveryIntervalMs`, so an operator hitting sweep contention has an escape hatch (currently hardcoded at 4)
- [ ] Test: back-to-back requests are throttled; the refusal is a valid `session_pr_status` with a reason; the watcher is not armed on a refused request

Issue #7425, opened 2026-08-27
app: mobile parity for the session PR/CI status chip (session_pr_status)


blamechris:
## Context

#7422 shipped the session PR/CI status chip for the dashboard. The mobile app has no equivalent, so `session_pr_status` is registered **dashboard-only** in both coverage guards:

- `packages/protocol/tests/handler-coverage.test.js` — `PLATFORM_SPECIFIC: 'dashboard'`
- `packages/store-core/src/contract-fixtures/protocol-type-coverage-lint.test.ts` — `DASHBOARD_ONLY`

Both entries name this issue as the tracked follow-up. This is the standard shape for a dashboard-first surface (same as `git_create_pr_result`, `shell_pending_approval`, and the IDE surfaces).

## What's needed

The server side is done and platform-agnostic — `session_pr_status_request` → `session_pr_status` is already session-scoped, so a pairing-bound mobile client can ask about the session it is bound to without any new gate. The work is client-side:

1. Store handling for `session_pr_status`, keyed by session id (the dashboard's `handleSessionPrStatus` is the reference — a reply can land after a session switch and must stay filed under the session it describes).
2. A chip in the session header, or wherever the mobile header has room.
3. Remove both allowlist entries; the coverage guards then require it.

## Carry these three contracts across — they are the point of the feature

1. **`checks.state: 'none'` is not a green.** A push can create no run at all; absence of failure is not success.
2. **Check state and merge state are separate pills, never one "ready to merge?" badge.** The case that motivated #7344 had 21/21 checks green while `mergeStateStatus` was `BLOCKED` on an unresolved review thread.
3. **`pr: null` + `reason: null` (no open PR) must render differently from `pr: null` + a `reason` (could not determine).** Cannot-determine must never read as an implied green.

`packages/dashboard/src/components/SessionCiChip.tsx` and its tests are the reference for all three, including the ones that go red when each is broken.

## Mobile-specific

- **48dp tap targets**, not 44 — Material's floor is stricter, and the chip has a Refresh control plus a PR link (see CLAUDE.md "Tap targets").
- The dashboard has no `hitSlop`; the mobile app does, and an information-dense header row is exactly the case it exists for (#4876).

Related: #7422 (the dashboard slice), #7344 (parent), #7423, #7424.

PR #7469, opened 2026-08-28
feat: count a session PR's unresolved review threads on demand

### LINKED ISSUE #7430: sessions: fetch the unresolved-thread count on demand for the CI prefill line (#7423 follow-on)
## Context

#7423's example prefill line carries an unresolved-thread count:

```
CI green on 5fff69a — 21/21, mergeStateStatus CLEAN, 0 unresolved threads
```

The PR that closed #7423 ships the line without it. `session_pr_status` does not
carry the count, and the obvious place to add it — `surveySessionPrStatus` — is
the wrong one now.

## Why it was not added to the survey

`gh pr list --json` has no thread count; it needs a GraphQL
`reviewThreads(first:N){ nodes { isResolved } }` read. #7423 correctly noted this
is not head-SHA-dependent, so it does not violate #7422's one-call-per-head
consistency rule.

What changed is the survey's caller list. Since #7426 the same
`surveySessionPrStatus` runs on a **daemon-side sweep across every session**
(`SessionCiWatcher`, `tickIntervalMs` + `discoveryIntervalMs`), not only when a
dashboard asks. Adding a second `gh` invocation there doubles the subprocess and
network cost of a background poll — to enrich a string that is only ever built
when a user clicks. The cost lands on every session, every tick, for a field one
click wants.

`merge.reviewDecision` (already on the snapshot, already in the prefill line)
carries the review state in the meantime — `CHANGES_REQUESTED` /
`REVIEW_REQUIRED` / `APPROVED`. What it does not give is the *count*, which is
the thing that explains a `BLOCKED` merge state.

## The shape this wants instead

An **on-demand** read, paid for at click time rather than on every sweep. That is
a new request/response pair rather than a field on the existing snapshot, which
is why it is its own issue and not a one-line addition:

- a `session_pr_threads_request` / `session_pr_threads` pair, or an opt-in flag
  on `session_pr_status_request` that the sweep never sets;
- the server-side GraphQL read, degrading exactly as the rest of
  `session-pr-status.js` does (no `gh`, non-GitHub remote, failed call → a
  `reason`, never a fabricated `0`);
- `formatCiPrefill` gains an optional clause. **`0 unresolved threads` and
  `unresolved-thread count unavailable` must render differently** — a missing
  count that prints as zero is the same false-green this whole feature exists to
  avoid.

## Care needed

- Adding a field to `session_pr_status` trips three coverage guards
  (`packages/store-core/src/contract-fixtures/protocol-type-coverage-lint.test.ts`
  and friends) — see the both-clients note in CLAUDE.md. A new message type trips
  the `Server -> Client:` doc-block roster too.
- The count is a function of the PR, not of the head SHA, so it can be read
  separately without straddling a push — but it goes stale on its own schedule
  and should carry its own timestamp rather than borrowing `generatedAt`.

Related: #7423 (the prefill line), #7422 (the survey), #7426/#7424 (the sweep
that made the survey's cost per-session), #7344 (parent).

blamechris:
## The shape, and why this one

The issue offered two shapes. This takes the **new request/reply pair**
(`session_pr_threads_request` → `session_pr_threads`) rather than an opt-in flag
on `session_pr_status_request`, for two reasons:

- **The sweep's survey is untouched.** Since #7426 `surveySessionPrStatus` runs
  on a daemon-side sweep across every session. A flag the sweep never sets would
  work, but it leaves the count *reachable* from the survey's code path — and
  #7422's one-`gh`-call-per-head consistency rule is easier to hold when the
  extra call cannot be reached from there at all. The sweep now makes exactly
  the calls it made before, by construction.
- **`session_pr_status` does not widen.** Adding four fields to that snapshot
  trips the three coverage guards the issue names, and — more substantively —
  would put the count under `generatedAt`, which the issue is explicit it must
  not borrow. The count is a function of the PR, not of the head SHA, so it goes
  stale on its own schedule and carries its own `countedAt`.

## Degradation table

Every path replies exactly once, with a schema-valid message. No path can emit a
count it did not take.

| condition | `prNumber` | `unresolvedCount` / `totalCount` | `truncated` | `reason` |
|---|---|---|---|---|
| bound client asks about another session | `null` | `null` | `false` | not authorized … |
| unknown session (unbound client) | `null` | `null` | `false` | no such session |
| this client already counting | `null` | `null` | `false` | a review-thread count is already running … |
| throttled, nothing cached yet | `null` | `null` | `false` | counted moments ago … |
| throttled, cache present | — | — | — | **the cached reading, replayed verbatim** |
| no cwd / not a repo / reaped worktree / detached HEAD / no GitHub remote / no `gh` / `gh pr list` failed | `null` | `null` | `false` | **passed through from `surveySessionPrStatus`, verbatim** |
| branch has no open PR (the quiet negative) | `null` | `null` | `false` | no open pull request for this branch … |
| repo owner/name would be option-parsed | `N` | `null` | `false` | repository owner or name cannot be passed safely to gh |
| `gh api graphql` failed | `N` | `null` | `false` | `gh api graphql failed: <stderr line>`, absolute paths redacted |
| output not JSON | `N` | `null` | `false` | gh api graphql produced unparseable output |
| response carries `errors`, or a node has no boolean `isResolved`, or page completeness cannot be established | `N` | `null` | `false` | gh api graphql returned no usable review-thread data |
| counted | `N` | `k` / `m` | `false` | `null` |
| counted, more threads than one page | `N` | `k` / `m` | `true` | `null` |

Resolution is **delegated**, not re-derived: the survey calls
`surveySessionPrStatus` and inherits its whole degradation table, including the
fork path (a cross-repository PR is listed on the base repo and disambiguated by
head owner). The cost is one extra `gh pr list` per counted click, bounded by the
throttle; the alternative was a second copy of resolution logic that would drift.

The three "refuse rather than guess" rows are deliberate. Every guess available
there is a guess in the dangerous direction — an unreadable node absorbed as
resolved, a partial response counted as whole, an unknown page state read as
complete — and each **under-reports** unresolved threads, i.e. manufactures the
green.

## The false-green pin, and the evidence it bites

`docs/false-safety-guards.md` asks that a new guard be proven to fail. The wrong
rendering was written first, on purpose, and the pin caught it:

```
FAIL  src/lib/ci-prefill.test.ts > #7430 > THE FALSE-GREEN PIN:
      a count that was not taken NEVER renders as a zero
AssertionError: expected 'CI status for PR #7423 (head 5fff69a)…'
                not to match /(?<!at least )\b0 unresolved threads\b/
+ Received:
"… checks green (21 passed, 0 failed, 0 pending, 0 skipped of 21) —
 merge state CLEAN — review APPROVED — 0 unresolved threads"
```

That is a snapshot whose `reason` is `gh CLI not found on PATH`, rendering as a
definite zero beside a green check clause. The shipped formatter renders it as
`unresolved-thread count unavailable: gh CLI not found on PATH`, and the two
strings are asserted to differ.

The pin's regex excludes `at least 0 …` (the truncated lower bound), and that
exclusion has its own two-sided control, so it cannot quietly widen later.

## Mutation results — 12 applied, 12 killed

Byte backups taken first, restored by absolute-path `cp`, each restore verified
`diff`-identical; every scripted edit asserted its match count before the run.

| # | mutation | killed by |
|---|---|---|
| M1 | degradation arm removed + `unresolvedCount ?? 0` | 3 prefill tests |
| M2 | truncation arm dropped | truncated-zero test |
| M3 | PR-number match guard removed | different-PR test |
| M4 | socket-drop reset removed from `onclose` | anchored source guard |
| M5 | `COUNT_MIN_INTERVAL_MS` → 0 | 2 throttle tests |
| M6 | throttled replay → degrade | 3 tests |
| M7 | `gate.rollback()` removed | 2 tests |
| M8 | authority check moved after the session lookup | 2 tests |
| M9 | `degraded()` emits `0` instead of `null` | 7 tests |
| M10 | GraphQL `errors` ignored | 2 tests |
| M11 | unreadable node treated as resolved | 1 test |
| M12 | unknown page state assumed complete | 1 test |

Two further controls, run the same way: removing `'session_pr_threads'` from
`DASHBOARD_ONLY` fails the store-core coverage lint with the type named, and the
`Server -> Client:` roster line was observed to turn `handler-coverage.test.js`
red (3 subtests) before the allowlists were updated — so both guards are known to
see this type rather than being assumed to.

## Roster / guard updates

- `ws-server.js` `Server -> Client:` doc block — required, the roster is grepped.
- `packages/protocol/tests/handler-coverage.test.js` → `PLATFORM_SPECIFIC`.
- `packages/store-core/src/contract-fixtures/protocol-type-coverage-lint.test.ts`
  → `DASHBOARD_ONLY`.
- `coverage-lint.test.ts`'s `PENDING_CONTRACT_TYPES` is **not** touched: that set
  is for both-clients types, and this one is dashboard-only.

**Mobile app:** no app code. The type is declared dashboard-only in both
allowlists with the documented `'dashboard'` mapping — its only consumer is the
CI chip's prefill action (#7431), which is dashboard-only, and mobile has no CI
chip at all, so parity rides with #7425 rather than being a new gap. Both
allowlist entries say so.

## Review round 2 — what changed after #7469's adversarial review

Four commits on top of the original three. Full detail in the reply comment.

- **C1 — `Dashboard Type Check` was red on `7217bdc43`, and this body claimed it clean.** The new dispatch tests called `handleMessage(msg, get, set, ctx)` against a 1-2 argument signature: 6× TS2554 + 1× TS6133. They ran anyway because argument 2 was merely truthy, so the tests read as though they injected store accessors and did not. Split into the sibling pair's two files and driven through the real two-argument API. `tsc --noEmit` exits 0.
- **C2 — the "unavailable" clause was unreachable.** `threadsClause` tested `prNumber` before the count, and every degraded reply the server emits carries `prNumber: null`, so the clause silently vanished for all four failures a user can provoke. Guards reordered: a *reason* has no PR to be mis-attributed to, a *number* does. The pin is re-aimed at shapes the server can actually emit, with a join test over all four real reasons and source-level assertions against the producer.
- **S1 — a degraded reply no longer erases a good count**, on both sides: the store keeps the last counted reading (with its own `countedAt`) and attaches the new reason, and the handler no longer caches degraded readings into the replay window.
- **S2 — `docs/security/bearer-token-authority.md` §4** now names both session-scoped surveys, records that this one is read-only while its sibling mutates watch state, and moves the throttle's cross-client visibility there as the authority-model fact it is.
- Nitpicks: the anti-schedule guard sweeps all dashboard sources, its scan is factored out so the controls run the guard's own code, and it gained the timer-half control. `requestSessionPrThreads`'s JSDoc states correlation is by `sessionId` only.

## Also in here: the throttle is now shared, not copied

`survey-throttle.js` extracts the window / cached-replay / compare-and-restore
rollback machinery #7436 and #7445 built for `session_pr_status_request`, and
both handlers open their own instance. `session-pr-status-handlers.js` moves onto
it with no behaviour change — its 31 tests, including the "slow survey's late
failure cannot destroy a newer client's stamp" regression, stay green untouched.

## Suites

Re-run at `e0191248a` after the review. **Exit codes captured directly, not through a pipe** — the original table's `clean` row was produced by a `tsc | head` pipeline whose `$?` was `head`'s, and by a reading taken 3m before the last test file landed. Both are fixed; see the review reply for the post-mortem.

| suite | exit | result |
|---|---|---|
| dashboard `npx tsc --noEmit` | **0** | clean |
| `packages/protocol` full `npm test` | 0 | 392 passed |
| `packages/store-core` full `npx vitest run` | 0 | 60 files, 2189 passed |
| server: threads + threads-handler + status + status-handler + ws-handler-coverage + ws-message-handlers + ws-server | 0 | 319 passed |
| dashboard full `npx vitest run` | 0 | 308 files, 5382 passed |
| all nine `packages/server/scripts/lint-*.sh` | 0 | pass |
| server `npx eslint src/` | 0 | 0 errors (10 pre-existing warnings) |

Closes #7430


blamechris:
## Code Review: PR #7469

### Summary

The design is right and most of it is very carefully built. The count is a separate on-demand pair that the #7426 sweep never sends; the survey **delegates** resolution to `surveySessionPrStatus` instead of re-deriving it; `countedAt` is its own clock; `prNumber` rides along so a consumer can check the join; and the "refuse rather than guess" arms in `summariseReviewThreads` each refuse in the safe direction. I independently re-ran the roster controls and all three bite.

**The throttle extraction is clean, and I checked it the way the brief asked.** `packages/server/tests/session-pr-status-handler.test.js` has a **zero-byte diff** — the 31 existing tests are not merely still green, they were not touched. Line-by-line the moved code is semantically identical: `gate.cached` is `prior.snapshot ?? null` and `if (gate.cached)` has the same truthiness as `if (prior.snapshot)`; `commit()` is `record.snapshot = snapshot`; `rollback()` is the same compare-and-restore, including the "A's late failure must not destroy B's newer stamp" identity check. The two handlers each open their **own** `createSurveyThrottle()`, so no window is shared. There is no `#7440` `indeterminate` interaction to get wrong, because the threads path never spreads a status snapshot — it builds its own `baseSnapshot` and copies only `status.reason`, so the server-only marker is structurally out of reach rather than stripped.

**Two things block.** One is that **CI is red on this diff, not on a flake**: `Dashboard Type Check` fails on `7217bdc43`, in this PR's own new test file, and the PR body claims that suite clean. The other is that the documented **"unavailable"** rendering is unreachable for essentially every degraded reading the server actually emits, because `threadsClause` checks `prNumber` before it checks `reason` — so the clause the issue was filed to produce silently disappears in the cases that need it most, and the pin that is supposed to guard this is asserted against a shape the server can barely produce.

**Deviation 8 (the transient server-test failure): I could not reproduce it.** 11 clean runs — 3× the 7-file batch (`session-pr-threads`, `-handler`, `session-pr-status`, `-handler`, `ws-handler-coverage`, `ws-message-handlers`, `ws-server`) at **317/317** each, and 8× a 5-file batch. I went looking specifically for the shape a shared-module extraction would produce — a throttle record leaking across tests via the module-level `WeakMap` — and it is structurally impossible here: every `makeCtx()` in both handler test files mints a fresh `createMockSessionManager()`, so the `WeakMap` key is new per test and no `_nowMs`-free test can inherit a live window. I'd treat it as environmental (a live daemon is listening on `:8765` on this host, which is a known source of one-off `ws-server` failures) rather than as this diff, but it is unreproduced, not explained.

### Strengths

- **The delegation is the right call and the PR-number race is handled honestly.** Re-running `surveySessionPrStatus` can legitimately resolve a *different* PR than the snapshot the user is looking at. `snapshot.prNumber = number` records what was actually counted (`session-pr-threads.js:218`) and `threadsClause` refuses a mismatch, so a count can never be narrated against another PR — and the store is session-keyed on top of that, so a reply landing after a tab switch cannot cross sessions either. Both halves are tested.
- **`summariseReviewThreads`'s truncation logic is genuinely two-signal.** `hasNextPage === true` → truncated; `false` → cross-checked against `totalCount > nodes.length` (so a contradictory response is treated as truncated, the safe direction); `hasNextPage` unreadable → fall back to `totalCount`; **neither readable → `null`**, not "complete". That last arm is the `docs/false-safety-guards.md` "cannot check ≠ nothing to check" class, refused correctly, and it has its own test.
- **The truncated rendering is a real lower bound.** `at least 0 unresolved threads — only part of 150 review threads were read` is the right answer to the 100-resolved-on-page-one case, and `COUNTED_ZERO`'s lookbehind is controlled from **both** sides (`'roughly 0 unresolved threads'` must match; `'at least 0 …'` must not), so the exclusion cannot silently widen. I re-ran that pair; both hold.
- **The roster controls are armed — I re-ran all three rather than taking the body's word.** Removing `'session_pr_threads'` from `DASHBOARD_ONLY` → store-core coverage lint **1 failed / 4 passed**, naming the type. Removing it from `PLATFORM_SPECIFIC` → protocol `handler-coverage.test.js` **1 failed**. Deleting the `Server -> Client:` doc-block line from `ws-server.js` → protocol `handler-coverage.test.js` **2 subtests + parent = 3 `not ok` lines** (matching the body's "3 subtests"). Note that neither `ws-outbound-coverage.test.js` nor `ws-schemas.test.js` sees the doc-block line — the protocol test is the only guard on it, which is worth knowing.
- **`probeGh` is exported deliberately un-memoised**, with the reason written down (a daemon outlives a `gh` install). That is the correct call and the kind of thing that usually gets "optimised" later without it.
- **`EXEC_OPTS` shared rather than re-minted**, so the GraphQL call inherits the same timeout and `maxBuffer` and cannot drift from its sibling.
- Every datum reaching `gh` rides in a `key=value` token and `owner`/`name` still go through `isSafeArgvValue` anyway — belt and braces on the #7290/#7295 class, and the exec-failure reason is `redactAbsolutePaths`'d before it reaches a pairing-bound client.

### Issues Found

#### Critical (Must Fix)

|---|---|---|---|
| `packages/dashboard/src/store/session-pr-threads.test.ts` | 130, 144, 162, 174, 177, 179, 189 | **`Dashboard Type Check` is RED on `7217bdc43`, and the PR body reports that gate clean.** `npx tsc --noEmit` in `packages/dashboard` exits **2** with **7 errors**, all in this PR's new file: `error TS6133: 'msg' is declared but its value is never read` (`:130`) and 6× `error TS2554: Expected 1-2 arguments, but got 4`. The real signature is `handleMessage(raw: unknown, ctxOverride?: ConnectionContext)` (`message-handler.ts:4225`); the calls pass **four**. Reproduced locally, and the CI job agrees (`Dashboard Type Check  fail  2m2s`). The suites table's `dashboard npx tsc --noEmit \| clean` row is not what the head commit does. Worth naming *why* the runtime tests still pass: the second argument (`useConnectionStore.getState`, a function) is not a `ConnectionContext` at all — it merely happens to be truthy, which satisfies `const ctx = ctxOverride ?? _connectionContext; if (!ctx) return;`. Arguments 3 and 4 are discarded entirely, so the test **reads as though it injects its own `get`/`set`** when it does not; the handler is in fact driven by the dispatcher's real store accessors. | Follow the sibling idiom already in this directory — `dispatch-session-pr-status.test.ts:72` uses `handleMessage(snapshot(), ctx() as never)` with a real two-arg call and a proper `ctx()` factory. Drop the unused `msg` parameter from `handle()` (or use it). Then re-run `npm run typecheck` in `packages/dashboard` and confirm exit 0 before re-reporting the suites table. |
| `packages/dashboard/src/lib/ci-prefill.ts` | 171-179 | **The documented "unavailable" rendering is unreachable for the degraded readings the server actually emits — the clause silently vanishes instead.** `threadsClause` checks `if (threads.prNumber !== prNumber) return null` (`:172`) **before** the `reason` guard (`:174`). Every degraded reply the handler builds via `degraded()` carries `prNumber: null`, as does every reading that returns before `session-pr-threads.js:218`. So `null !== 7423` → the whole clause is dropped and the line makes **no thread claim at all**. I probed the formatter directly against a status snapshot with a live PR (transcript below), and all of these render the pre-#7430 line verbatim: `IN_PROGRESS_REASON` (a second click while the first count is in flight — and `handlePrefillSessionPrStatus` fires a request on **every** click, including ones `runCiPrefill` then refuses for a busy draft), `RATE_LIMITED_REASON` (two clicks inside 5 s before any count has completed — the *first-click* path, since the auto-pull deliberately does not fetch a count), `review-thread count failed: …`, and — the one that matters most — **the status-survey reason passed through at `session-pr-threads.js:205`**, which is exactly the double-`gh` coherence race: the dashboard holds a good status snapshot with a PR, the threads survey's *own* `surveySessionPrStatus` re-run fails transiently, and the user gets `merge state BLOCKED` with the explanation silently missing. This contradicts the file's own contract twice over: the header promises three renderings of which "unavailable … says so, in words", and `threadsClause`'s docstring says `null` is returned "**for exactly two inputs**" — no count supplied, or a *different* PR. A `prNumber: null` degraded reading is neither, which is the `false-safety-guards.md` "comment describes a stronger check than its code performs" shape. **And the pin cannot see it:** the FALSE-GREEN PIN (`ci-prefill.test.ts:317`) uses `prNumber: 7423` with `reason: 'gh CLI not found on PATH'` — a pairing `surveySessionPrThreads` can only produce if `gh` is removed from `PATH` *between* the resolution call and `probeGh` (`:228` runs after `:218`), since an actually-missing `gh` degrades the status survey first, leaves `status.pr` null, and makes `formatCiPrefill` return `null` outright. Server-side, `session-pr-threads.test.js:193` correctly pins `prNumber === null` on the passthrough path. Both halves are green; nothing joins them. | Order the guards so the caveat is reachable: check `reason`/`unresolvedCount === null` **first** and return the "unavailable" clause whenever `prNumber` is `null` (there is no PR to mis-attribute a *reason* to — the mis-attribution risk `:172` guards is specific to a **number** landing against the wrong PR). Keep `:172` as-is for the counted branch. Then add the join test that is currently missing on both sides: a degraded reading with `prNumber: null` + a real handler reason must render `unresolved-thread count unavailable: …`, and re-run the four reasons above. Worth a companion server-side assertion that the handler's `degraded()` shapes are the ones the formatter is tested against. |

|---|---|---|---|
| `packages/dashboard/src/store/message-handler.ts` · `packages/server/src/handlers/session-pr-threads-handlers.js` | 3251-3271 · 202 | **A degraded reply overwrites a good count, and (per Critical 2) then renders as nothing.** `handleSessionPrThreads` stores the message whole — correct in isolation, and the reason is real — but combined with the `prNumber` guard the net effect is that an `IN_PROGRESS` refusal *erases* a count the user was looking at, rather than degrading it. Server-side the same shape exists one level up: `gate.commit(snapshot)` caches **degraded** readings too, so one transient `gh api graphql` failure is replayed to every client for the full 5 s window. Neither is a false green and both mirror the sibling handler, so I am not calling it blocking — but the sibling's equivalent was #7445's Critical 1, and the mitigation there (replay, don't degrade) does not help when the replayed thing is itself a degradation. | Cheapest fix rides on Critical 2: once a `prNumber: null` reading renders as "unavailable" rather than vanishing, the overwrite is at least *visible*. Beyond that, consider not committing a `reason`-bearing snapshot to the replay cache (keep the last **good** count as the replay value and let the fresh failure go out under this request's id), and/or having the store keep the prior good count when the incoming reply is degraded. Either needs a test that a degraded reply cannot blank a previously-good count. |
| `docs/security/bearer-token-authority.md` | §4 | **§4 asks to be updated by exactly this PR, and was not.** It says verbatim: *"If you add another session-scoped survey, follow that shape — **and note it here**, because the §9 checklist's step 2 … would be the wrong answer for a survey."* This is the second session-scoped survey; the doc is untouched and still calls `session_pr_status_request` "**the first** session-scoped survey" with no mention of `session_pr_threads_request`. The new handler's own doc block cites §4 as its authority for the check-authority-before-lookup ordering, so the doc is precisely where a future reader lands. Two facts specifically want writing down: this endpoint is **read-only** (unlike its sibling it arms no watcher), and it inherits the same cross-client-visible per-session throttle. | Add `session_pr_threads_request` to the "Session-scoped SURVEYS reject with a degraded snapshot" subsection — one or two sentences naming the read-only property and the shared `survey-throttle.js`. Same PR. |

#### Nitpicks

|---|---|---|---|
| `packages/dashboard/src/lib/ci-prefill-wiring.test.ts` | 152-171 | The anti-schedule guard is real and **fail-closed** — with neither token in the window both `lastIndexOf` return `-1` and `-1 < -1` is false, so it fails rather than passing vacuously; I checked that. Two scope limits are worth stating in the comment rather than discovering later: it scans **`App.tsx` only**, so the same call from any other dashboard component is invisible to it (true today — `requestSessionPrThreads` appears nowhere else — but that is the property doing the work, not the scan); and the 600-char lookbehind means a call deep inside a long `useEffect` body escapes if a nearer `useCallback(` token happens to sit in the window. The positive control also only exercises the `useEffect`/`useCallback` comparison, not the `setInterval`/`setTimeout` half. | Widen to a glob over `src/**/*.tsx` (the store action is the resource being protected, not the file), and add a one-line control for the timer half. |
| `packages/dashboard/src/store/connection.ts` | 1355 | `requestSessionPrThreads` never sends a `requestId`, so the reply's `requestId` is always `null` and the store cannot correlate. Matches `requestSessionPrStatus`, so it is consistent rather than wrong — but the schema, the handler and the doc block all carry the field, which reads as though correlation exists. | Either send one or note in the JSDoc that correlation is by `sessionId` only, as with the sibling. |
| `packages/dashboard/src/lib/ci-prefill.ts` | 190-192 | `(counted 2026-08-27T00:05:00.000Z)` puts a raw ISO timestamp inline in a line a human also reads. Consistent with `generatedAt` in the prefix, so it is a house-style question, not a defect. | None needed; flagging only because the line now carries two ISO timestamps. |

### Verification note

Every mutation was applied by scripted replace with an occurrence assertion (`MUTATION DID NOT APPLY` → exit 9 if the match count was zero), run, then restored by absolute-path `cp` from a `/tmp` byte backup with `git status --short` confirmed empty after each. Four files were mutated and restored: `protocol-type-coverage-lint.test.ts`, `handler-coverage.test.js`, `ws-server.js`, and a temporary probe test in `packages/dashboard/src/lib/` which was deleted. Final `git status --short` in the worktree is **empty** — no tracked or untracked residue. `packages/protocol/dist` was rebuilt with `npm run build` and is byte-identical to what is committed, so the checked-in dist is in sync.

Suites I ran (Node 22, worktree `manual-7430`, workspace symlinks confirmed pointing at this worktree's packages rather than the main checkout):

| suite | result |
|---|---|
| `packages/protocol` full `npm test` | **392 passed** (matches the body) |
| server: threads + threads-handler + status + status-handler + ws-handler-coverage + ws-message-handlers + ws-server | **317 passed, ×3 consecutive runs** |
| server: 5-file subset, ×8 consecutive runs | **0 failures / 8** |
| `packages/store-core` `src/contract-fixtures/` | 124 passed |
| dashboard `src/store/` | 1182 passed (81 files) |
| dashboard `ci-prefill` + `ci-prefill-wiring` + `session-pr-threads` | 58 passed |
| dashboard `npx tsc --noEmit` | **FAIL — exit 2, 7 errors** (Critical 1) |
| all 9 `packages/server/scripts/lint-*.sh` | pass |

Probe transcript for Critical 2 — `formatCiPrefill(status_with_PR_7423, degraded_reading_with_prNumber_null)`, four reasons, identical output every time:

```
"CI status for PR #7423 (head abcdef1) as of 2026-08-27T00:00:00.000Z: checks green
 (21 passed, 0 failed, 0 pending, 0 skipped of 21) — merge state BLOCKED — review APPROVED"
```

No thread clause, no caveat, for `IN_PROGRESS_REASON`, `RATE_LIMITED_REASON`, a passed-through `gh` reason, and `review-thread count failed: boom`.

### Architecture Notes

**The `prNumber` join is the right mechanism and Critical 2 is a scoping error inside it, not an argument against it.** Joining on the PR number is exactly how you keep a second, independently-clocked reading honest against a first — and it is why the double-`gh` cost buys something real rather than just costing. The bug is that a **reason** and a **number** need different join rules: a number attributed to the wrong PR is a fabrication, but a *reason* has no PR to be wrong about. Collapsing both under one `prNumber` equality test is what makes the safe-direction guard eat the honest-degradation path.

**The extraction moved a property that now wants a home.** `survey-throttle.js` is a good module — the record is compare-and-restore, the owner-keyed `WeakMap` gives test isolation for free, and the "replay, don't degrade" refusal shape is argued from #7445 rather than reinvented. But the throttle's cross-client visibility is an **authority-model** fact, not an implementation detail, and it now applies to two endpoints from one shared file. §4 is where that belongs (see Suggestion 2), and it is a better place for it than a comment in a module neither handler owns.

**On test honesty:** the roster controls, the `COUNTED_ZERO` two-sided control, and the anti-schedule positive control are all genuine and all reproduce. The gap is one level up — the FALSE-GREEN PIN validates the *experiment* (does the formatter print a zero?) without validating the *fixture* (can the server produce this shape?). The server test and the dashboard test each pin their own half of `prNumber` and disagree, and nothing in the suite is positioned to notice. That is the same class as the PR's own M12 ("unknown page state assumed complete") — it just landed on the wire seam instead of inside a parser.

### Verdict

- [ ] **Approve** — ready to merge
- [x] **Request Changes** — Critical 1 is non-negotiable: `Dashboard Type Check` is red on the head commit, in this PR's own file, and the PR body reports that gate clean. Critical 2 is the substantive one: the clause the issue exists to produce silently vanishes for the reachable failure modes, the guard's comment describes a narrower rule than its code applies, and the pin that should catch it is asserted against a shape the server essentially cannot emit. Both are small fixes with a test each. The throttle extraction is faithful (zero-diff test file, verified line-by-line), the roster guards are armed, the GraphQL refusal arms are correct, and deviation 8 did not reproduce in 11 runs.

blamechris:
Both criticals fixed, both suggestions taken, plus the nitpicks. Head is now `e0191248a`.

## First: what produced the false "clean", because that is the more serious failure

The review is right that the suites table was wrong, and it was wrong for two compounding reasons — neither of which was a judgement call. Both are catalogued defect classes I walked straight into.

**1. The reading was stale.** I ran `npx tsc --noEmit` in `packages/dashboard` at **15:54:51**, immediately after fixing a workspace-resolution problem, and it was genuinely clean at that moment. `packages/dashboard/src/store/session-pr-threads.test.ts` — the file with all 7 errors — was written at **15:58:14**, three and a half minutes later. I never re-ran the typecheck after it landed, and reported the earlier reading as though it described the head commit. Verifiable from the file mtimes in the worktree; the symlink whose creation immediately preceded that run is stamped 15:54:51.

**2. My exit-code plumbing never reported `tsc` at all.** Every one of those runs was `npx tsc --noEmit 2>&1 | head -N; echo "EXIT=$?"` — so `$?` is `head`'s status, not `tsc`'s. One of those runs literally printed `EXIT=0` while `tsc` was emitting errors. That is `CLAUDE.md`'s "check the exit code, not the output — note `cmd | grep -c FAIL` reports grep's status" verbatim, and it meant nothing in my harness could have caught the stale reading either. The re-run table below captures exit codes with no pipe in the way.

The honest summary: I reported a measurement taken before the thing being measured existed, through a pipeline that could not have failed.

## C1 — `Dashboard Type Check`

`npx tsc --noEmit` in `packages/dashboard` now **exits 0**:

```
$ ( cd packages/dashboard && npx tsc --noEmit ); echo "TSC_EXIT=$?"
TSC_EXIT=0
```

Your diagnosis of *why* the runtime tests passed was exactly right and is now written into the file's header so it cannot be re-introduced: argument 2 was a function, hence truthy, hence satisfied `ctxOverride ?? _connectionContext`; arguments 3 and 4 were discarded; the tests **read as though they injected their own `get`/`set` and did not**.

Split into the sibling pair's shape rather than patched in place, since that is what makes the idiom obvious:
- `session-pr-threads-request.test.ts` — the action, `vi.resetModules()` + dynamic store import (mirrors `session-pr-status-request.test.ts`)
- `dispatch-session-pr-threads.test.ts` — the reply, `setStore` + two-argument `handleMessage(msg, ctx() as never)` (mirrors `dispatch-session-pr-status.test.ts`)

## C2 — the unreachable caveat

Reproduced your probe before touching anything. All four reachable reasons, identical output, no clause:

```
REASON: a review-thread count is already running …
  -> CI status for PR #7423 (head 5fff69a) as of 2026-08-27T00:00:00.000Z: checks green
     (21 passed, 0 failed, 0 pending, 0 skipped of 21) — merge state BLOCKED — review APPROVED
```

**Your architecture note is the fix.** A *reason* and a *number* need different join rules, so the guards are now ordered by what is being rendered rather than by convenience:

1. an explicitly different `prNumber` → no claim, in **either** direction (a reason belonging to another PR's count says nothing about this one either);
2. `unresolvedCount === null` → the **unavailable** clause, reached with `prNumber: null`, which is what every degraded reply has;
3. from there a **number** is being rendered, so the join must be positive — `prNumber !== prNumber` still refuses. Your "keep `:172` as-is for the counted branch" is kept, and it has its own test: a count that cannot name its PR is refused.

**The pin is re-aimed at shapes the server can emit.** You were right that `{ prNumber: 7423, reason: 'gh CLI not found' }` needs `gh` to vanish between the resolution call and `probeGh`. Red-first, on the reachable shape:

```
FAIL  THE FALSE-GREEN PIN: a count that was not taken NEVER renders as a zero
AssertionError: expected 'CI status for PR #7423 …' to contain 'unresolved-thread count unavailable'
Received: "… — merge state CLEAN — review APPROVED"
```

**The join you asked for exists on both sides.** `serverDegraded()` is shape-coupled to the handler's `degraded()`, `SERVER_REASONS` carries the four real strings, and three source-level tests fail if the producer drifts: that `degraded()`'s body really does set `prNumber: null` / `unresolvedCount: null` / `totalCount: null` / `truncated: false` (anchored to the function, with a positive control that the slice is not the whole file), and that every reason string this file pins is still emitted somewhere in `session-pr-threads-handlers.js` + `session-pr-threads.js` + `session-pr-status.js`.

**Guard order mutation-checked, and the false-green direction shown on a reachable input.** Reverting the order (`prNumber` tested before the count) → **3 failed**, including the pin and the join. Separately, with the order correct but `?? 0` restored, the reachable degraded fixture prints:

```
… — review APPROVED — 0 unresolved threads (counted 2026-08-27T00:05:00.000Z)
  — a newer unresolved-thread count was unavailable: a review-thread count is already running for this client
```

— the false green, on the shape the server actually sends, caught by the pin.

## S1 — a degraded reply no longer erases a good count

Taken on **both** sides, since you noted the store fix alone only makes the overwrite visible.

- **Store**: a degraded reply for a session that already has a count keeps the count, its `prNumber` and its **original `countedAt`**, and takes only the new `reason`. Per session id — `sess-A`'s count is never borrowed for `sess-B` — and a fresh good reading still replaces wholesale (positive control, so the retention cannot pin the chip forever).
- **Handler**: `gate.commit()` is now conditional on the reading carrying a count. The window is still **stamped** for a failed read — it reached `gh` and spent the budget the throttle protects — and only a *thrown* survey rolls back. That distinction has its own positive control, so "never stamp on failure" cannot satisfy the test.
- **Formatter**: the retained pairing renders as the number **plus** the failed-refresh caveat (`2 unresolved threads (counted …) — a newer unresolved-thread count was unavailable: …`). Dropping the number would throw away the only count the user has; dropping the caveat would present a stale reading as current. This replaces the previous "reason wins over number" rule, which was written for a shape the server did not emit and is now a shape the *store* emits deliberately.

## S2 — §4

Done, including the phrasing. §4 now names **both** session-scoped surveys, and writes down the two facts that differ and change what a reviewer must check: `session_pr_status_request` **mutates daemon state** (it arms the CI watcher — which is why #7436 throttled it, and it is not read-only despite reading like one), while `session_pr_threads_request` is **read-only** and is throttled purely for subprocess cost.

Your architecture note about the throttle is taken literally: the cross-client visibility moved out of the module comment and into §4 as the authority-model fact it is, with the reason it is *inside* the sharing model (all clients of one session are authoritative for it) and why no host-level survey may ever be keyed that way.

## Nitpicks

- The anti-schedule guard now sweeps `src/**/*.{ts,tsx}` rather than `App.tsx` — you are right that the store action is the resource, not the file — with a control that the glob actually matched call sites, so an empty sweep cannot pass silently.
- Its scan is factored into one function the controls call, so they exercise the guard's own code rather than a paraphrase, and it has the **timer-half control** it was missing (`setInterval` and `setTimeout`) plus a negative control that a plain click handler passes.
- The 600-character lookbehind limit and the file-scope limit are now stated in the comment rather than left to be discovered.
- `requestSessionPrThreads`'s JSDoc states that correlation is by `sessionId` only and the reply's `requestId` is always null.
- The inline ISO timestamp is left as-is, per your note that it is house-style consistency with `generatedAt`.

## Deviation 8

Accepted as unexplained rather than defended. Your reasoning that a leaked throttle record is structurally impossible matches mine, and 11 clean runs against my 3 is better evidence than I had. A live daemon on `:8765` on this host is the likeliest cause. I have removed the claim rather than restating it.

## Re-run gate, exit codes captured directly

| suite | exit | result |
|---|---|---|
| dashboard `npx tsc --noEmit` | **0** | clean |
| `packages/protocol` full `npm test` | 0 | 392 passed |
| `packages/store-core` full `npx vitest run` | 0 | 60 files, 2189 passed |
| server: threads + threads-handler + status + status-handler + ws-handler-coverage + ws-message-handlers + ws-server | 0 | 319 passed |
| dashboard full `npx vitest run` | 0 | 308 files, 5382 passed |
| all nine `lint-*.sh` | 0 | pass |
| server `npx eslint src/` | 0 | 0 errors (10 pre-existing warnings) |

Seven new mutants this round, all killed, each applied by scripted replace with an occurrence assertion and restored by absolute-path `cp` verified `diff`-identical: guard order reverted (3 failed) · `?? 0` on a reachable fixture (3 failed) · store keep-last-good removed (3 failed) · handler caches degraded again (2 failed) · truncation arm dropped, re-run against the new formatter (1 failed) · `degraded()` emits `0`, re-run (7 failed) · stale caveat dropped from the retained rendering (1 failed). `git status --short` empty after each.

blamechris:
Held finding fixed structurally, at `bee634cd8`. You were right on both halves: the guard was wired to one of two call sites, and §4 was claiming the property for both.

## The finding, confirmed

`session-pr-status-handlers.js:238` was `gate.commit(snapshot)` — unconditional. So the status survey cached reason-bearing readings and replayed them to every client of the session for the rest of the window, while §4 said *"only a reading that actually carries data is cached"* about both surveys. Classic guard-wired-to-only-some-callers, made worse by a security doc that had already described the intended behaviour as the actual one.

## The fix is in the gate, not a second call-site check

A second call-site guard shares the failure mode of the first, so the decision is now a **required argument**:

```js
commit(snapshot, { replayable })   // TypeError unless replayable is a boolean
```

No default in either direction — a default is exactly how a caller that never considered the question inherits a policy silently. The gate stays domain-blind; each caller states its own rule:

- **status**: `replayable: !snapshot.reason`. Deliberately *not* a check on `pr`: an `indeterminate` fork bail-out (#7435) carries `reason: null` and is byte-identical on the wire to an authoritative "no open PR", so replaying it shows exactly what a fresh survey would have. Suppressing it would degrade a usable display for nothing. That case has its own regression test, and it passed before and after — it is the one thing the `!reason` rule must *not* catch.
- **threads**: additionally requires an actual count, since a count-free reading has nothing to say however it arose.

`replayable: false` **retains** the previously cached good reading rather than overwriting or clearing it — you asked me to decide, and I mirrored the threads-side S1: for status the prior good reading is what the cache should keep. With no prior good reading it falls back to the pre-cache behaviour (`RATE_LIMITED`), which is honest, and the next request past the 5s window re-surveys and gets the real reason.

## Red-first, your probe as a test

```
not ok 8 - #7469 — a DEGRADED survey is not cached, and does not overwrite a cached good reading
  error: the replay must be the last GOOD reading, not the cached failure
not ok 9 - #7469 — a degraded survey with NO prior good reading degrades rather than replaying
# tests 34  # pass 32  # fail 2
```

Plus a new `packages/server/tests/survey-throttle.test.js` for the gate itself — 6 of its 13 red before the change.

## Mutations — 4 applied, 4 killed

| mutation | result |
|---|---|
| `replayable` defaults to `true`, throw removed | 3 failed |
| `replayable: false` **clears** the cache instead of keeping the prior good reading | 4 failed |
| status caller reverts to unconditional `replayable: true` | 2 failed |
| **status caller drops the option entirely** (the forgotten-caller case) | **25 failed** |

The last one is the structural guarantee: forgetting the decision is no longer a silent cache, it is a crash on the first survey. A third session-scoped survey cannot inherit this defect by omission.

## Which existing status tests changed

**None.** `session-pr-status-handler.test.js` is **67 insertions, 0 deletions** — all 31 pre-existing tests are byte-identical and still green. 31 + 3 new = 34. The zero-byte-diff property you verified last round is gone only in the sense that the file grew; nothing in it was rewritten to accommodate the change.

## §4

Rewritten so the claim is true, and so the security content is the fail-closed property rather than the convention. It now says both surveys cache only replay-worthy readings, that the decision is a required argument of the shared gate with no default, that `replayable: false` retains the prior good reading, and — since this is where a future reader lands — *why* it is shaped that way, citing this review and the `false-safety-guards.md` class by name.

## Gate, bare exit codes

| suite | exit | result |
|---|---|---|
| server `session-pr-status-handler` (full) | 0 | 34 passed |
| server: survey-throttle + threads + threads-handler + status + status-handler + ws-handler-coverage + ws-message-handlers + ws-server | 0 | 335 passed |
| server `session-ci-watcher` (consumes the status snapshot) | 0 | 87 passed |
| `packages/protocol` full `npm test` | 0 | 392 passed |
| all nine `lint-*.sh` | 0 | pass |
| server `npx eslint src/` | 0 | 0 errors (10 pre-existing warnings) |

Dashboard untouched this round — the change is server + doc only.

## One CI note, not acted on

`Dashboard Tests` on `e0191248a` came back **`cancelled`**, which `gh pr checks` renders as `fail`. It is not a test failure: the job's conclusion is `cancelled` after 11m4s while the rest of run 33221362138 kept going, so it was not a whole-run concurrency cancel either. Flagged rather than re-run.

blamechris:
## Re-review: PR #7469 @ `bee634cd8`

Everything I raised is fixed. I re-ran my own demonstrations rather than reading the diff, and added two mutants of my own; all of it holds.

### The §4 finding is fixed STRUCTURALLY, which is the right fix

My last round's blocker was one clause: `commit()`'s "only cache a reading worth replaying" rule lived at the threads call site, so it covered **one of the gate's two callers** while §4 claimed the property for both. The fix does not add a second call-site guard — it moves the decision into the gate as a **required** argument, which is the only version of this that a third caller cannot inherit by omission.

**The forgotten-caller mutant is loud.** Dropping the option at the status call site (`gate.commit(snapshot)`) → **25 failed / 22 passed**, and the failure is the intended `TypeError: survey-throttle: commit() requires an explicit boolean 'replayable'`, taking down even the basic happy paths. Forgetting crashes on the first survey rather than silently caching. That is fail-closed in the direction that matters.

### The live probe, re-run against the real status handler

| case | setup | second client inside the window |
|---|---|---|
| A | good reading, then a degraded one | the **last GOOD** (`pr 7423`, `generatedAt: T-good`) — never the cached failure; the requester that paid still got the truth |
| B | first reading degrades, nothing cached | **`RATE_LIMITED`**, not the cached failure |
| C | `indeterminate` fork bail-out (`reason: null`, `pr: null`) | **replayed**, and `indeterminate` still stripped from the wire in both the original and the replay |

Case C is the one worth calling out: it confirms the caller's `!snapshot.reason` test is right where a `pr`-based one would have been wrong, and that routing an `indeterminate` reading through the replay path does not leak the server-only marker (#7435 / #7440 intact).

### Two mutants of my own on the retention semantics

- cache regardless of `replayable` → **7 failed**
- `replayable: false` **clears** instead of **retains** (`record.snapshot = null`) → **4 failed**

The second was my real concern. "Keeps the previously cached good reading" is a claim that is easy to write in prose and easy to leave untested, and it is genuinely distinguished from "clears" by behaviour here, not just asserted.

### §4 now reads true for both callers

Checked clause by clause against the code: required boolean, no default in either direction; `replayable: false` returns early leaving the carried-forward prior in place; both per-caller rules stated correctly including the `indeterminate` rationale. The `RATE_LIMITED` sentence was also correctly re-scoped from "before the session's first reading completes" to "arriving with nothing cached to replay" — which is now the precise condition, since a first reading that was not replayable *also* leaves nothing cached. And the paragraph names the original defect and its `false-safety-guards.md` class rather than quietly correcting it, which is what makes it useful to the next person.

### Process notes

`session-pr-status-handler.test.js` is **67 insertions / 0 deletions / 0 modified lines** — the 31 pre-existing tests are byte-identical, 3 added. The new `survey-throttle.test.js` covers the missing bag, `undefined`, and non-boolean values, has a positive control that an explicit boolean does *not* throw (so the throw-tests cannot pass against an always-throwing `commit()`), and the self-disclosed test bug — asserting `cached` off the admitted branch of the return union, where it reads `undefined` — is fixed with the reason written into the test. Disclosing that was the right call.

### Verification note

Three mutations this round, each applied by scripted replace with an occurrence assertion, restored by absolute-path `cp` from `/tmp` byte backups, with `survey-throttle.js` and `session-pr-status-handlers.js` confirmed byte-identical to `bee634cd8` afterwards. `git status --short` empty. Both probes ran from `/tmp` and wrote nothing into the repo.

Suites at this head: server targeted batch (`survey-throttle` + threads + threads-handler + status + status-handler + ws-handler-coverage + ws-message-handlers + ws-server) **335 passed ×2**; all 9 `packages/server/scripts/lint-*.sh` pass; dashboard `tsc --noEmit` exit 0. This commit touches no files under `packages/{dashboard,protocol,store-core}`, so the earlier greens stand unchanged — dashboard full **308 files / 5382**, protocol **392**, store-core contract-fixtures **124**.

### Verdict

- [x] **Approve** — ready to merge. All three rounds of findings are resolved and each fix is pinned by a control I re-ran myself. The last one is the one I'd highlight: the reachable-failure gap and the one-of-two-callers gap were both fixed by changing the *shape* of the interface rather than by adding another check, so neither can recur by omission.
- [ ] **Request Changes**

Issue #7466, opened 2026-08-28
dashboard: clicking a stale permission prompt jumps to the Devices tab and wedges the tab bar in a flashing no-tab-selected refresh loop


blamechris:
## Observed (dogfood, 2026-08-28)

Dashboard v0.11.0 against the local daemon (running yesterday's main, ~`8cf5dd01`), two live claude-cli sessions ("Chroxy", "DockKeeper"), permission mode **Approve**.

A **stale accept/deny permission prompt** was still rendered on a session. Clicking it (deliberately — the staleness was known, to see what would happen):

1. **Navigated to a different tab — Devices** ("Paired devices / No paired devices"), which has no connection to the permission flow.
2. The header text then **flashes in what looks like a refresh/loading loop**.
3. The session tab bar is left in a **no-tab-selected** state while the loop runs (the Devices panel renders, but the tab-strip selection state is broken).

Screenshot captured at the wedged state: Devices panel open, both session tabs unselected-looking, "CI unavailable" chip in the header (unrelated — the running daemon predates today's CI-status merges).

## Three distinct defects likely tangled here

- **A stale prompt was clickable at all.** #2833 (closed) fixed already-answered prompts re-rendering as interactive after tab switches; this looks like the sibling case for *expired* prompts (no requestId left server-side). #7353 (open) covers the trace-erasure UX for expired permissions — this is not that; here the click ACTS.
- **The click navigated to Devices.** Suspicion: the prompt's element was unmounted/replaced mid-click and the pointer event landed on whatever took its place (tab strip overflow area — Devices is the last visible tab in the strip, exactly where a dying overlay's coordinates could land), OR an error path from the dead `permission_response` (`permission_expired`?) triggers navigation. Needs tracing from the prompt's click handler.
- **The refresh/flash loop with no tab selected.** Whatever landed the navigation left the active-tab state pointing at nothing while the Devices panel mounts/unmounts or refetches repeatedly. Repo memory notes the dashboard renders under StrictMode and that edge-tracking hooks need StrictMode-safe state — a mount loop here fits that family.

## Repro (as performed)

1. Let a permission prompt go stale in a session (mode Approve; let the request expire server-side or let the turn move on).
2. Click Accept or Deny on the still-rendered prompt.
3. Observe the jump to Devices, the flashing header, and the unselected tab strip.

## Acceptance criteria

- [ ] A prompt whose requestId is no longer pending renders inert (or as the #7353-style historical record) — never as live buttons
[...]

Issue #1710, opened 2026-03-08
Auto-switch to session tab when interacting with cross-session permission


blamechris:
## Context

When a permission request arrives for session B but the user is viewing session A, the permission prompt correctly appears in session B's message history. However, if the user taps Allow/Deny (e.g. via notification or if they happen to see it), the app stays on session A's tab.

## Expected Behavior

When a user interacts with a permission prompt (Allow/Deny/Always Allow) that belongs to a different session than the currently active tab, the app should automatically switch to that session's tab so the user sees the result of their action in context.

## Applies To

- Mobile app (packages/app)
- Desktop dashboard (packages/server/src/dashboard-next)

## Acceptance Criteria

- [ ] Tapping Allow/Deny on a permission from a non-active session switches to that session's tab
- [ ] Tab switch happens after the response is sent (not before)
- [ ] No switch if the permission already belongs to the active session

Issue #7471, opened 2026-08-29
ci: long self-hosted jobs are being cancelled mid-run under load — four carcasses in one evening


blamechris:
## Pattern (2026-08-28 evening, all on the self-hosted mac runner)

Four long CI jobs concluded `cancelled` mid-run with no human cancel, no push to their branch, and the rest of their run continuing:

| run | job | branch/head | note |
|---|---|---|---|
| 33214113786 | Server Tests | fix/7436-survey-rate-limit @ db1b02cdb | only unfinished job; rerun passed |
| 33216427073 | Server Tests | test/7451-shared-mock-context @ ad93161f3 | app-only diff; rerun passed |
| 33221362138 | Dashboard Tests | feat/7430-thread-count @ e0191248a | cancelled at 11m4s while siblings continued |
| (current) | Server Tests | feat/7430-thread-count @ bee634cd8 | cancelled again; rerun in flight |

`gh pr checks` renders these as `fail`, which twice sent triage down the wrong path before the run-level conclusion was read (the stale-red-check / absent-run family from the triage memory).

## Why it matters

A cancelled long job + green siblings reads as a code failure at the PR gate, costs a manual rerun each time, and — worse — trains people to rerun-until-green. The box was under heavy load at the time (14 repo runners + an agent fleet); the working hypothesis is the runner service losing/recycling long jobs under resource pressure (job-level `timeout-minutes` was NOT exceeded on the 11m4s Dashboard Tests cancel; concurrency groups are per-branch and no same-branch push occurred).

## Acceptance criteria

- [ ] Root cause identified from the runner's own logs (`~/actions-runner/_diag` and the github-runners variants) for at least one of the four runs above
- [ ] Either the cause is fixed (runner service config / resource limits / job sharding) or the long jobs are made resilient (e.g. split Server Tests, or a retry-on-cancelled policy that is LOUD in the log rather than silent)
- [ ] The triage runbook gains a line: a `fail` rendering in `gh pr checks` must be disambiguated against the run-level conclusion before any code-level triage

blamechris:
Correction to the working hypothesis, from the second consecutive `Server Tests` cancellation at `bee634cd8` (run 33222642208), characterized precisely during #7469's review:

- The cancelled jobs ran on the **self-hosted Linux runner** `chroxy-linux-9ca9e00142fe` — NOT the Mac this issue originally blamed (the Mac's agent-fleet load is exonerated for these).
[...]

PR #7472, opened 2026-08-29
fix(dashboard): stale permission banners, a dangling active session, and the Devices-panel refresh loop

blamechris:
All file:line references are against the branch point, `origin/main` @ `e3fd9731e`.

Investigation-first pass on the #7466 dogfood report. Three symptoms were reported as
possibly-tangled; **two are proven and fixed here, a fourth defect was found while
tracing the third, and one symptom is still unproven — this PR says so plainly** rather
than patching by guesswork. `Related to #7466` (not `Closes`) — the
issue stays open on the unproven third.

## Mechanism table

| # | Reported symptom | Mechanism | Evidence | Status |
|---|---|---|---|---|
| 1 | A **stale accept/deny prompt was still rendered** and still acted | The cross-session banner had **no staleness gate at all**. `NotificationBanners.tsx:71` rendered Allow/Deny for any unread notification with `eventType: 'permission'` + a `requestId`; `SessionNotification` (`store/types.ts:529`) carries no expiry, so the banner depended entirely on an inbound `permission_expired`/`permission_resolved` to stamp `readAt`. The inline `PermissionPrompt` has three gates on the same decision (`answered` / its own `remaining <= 0` countdown / `connected` — `PermissionPrompt.tsx:169`, `:204`). When no expiry event arrived, the banner kept live buttons while the inline prompt for the SAME request had already gone inert. | `components/NotificationBanners.tsx:71-89`, `store/types.ts:529-553`, `components/PermissionPrompt.tsx:169,204` | **PROVEN — fixed** |
| 2 | The click **navigated to the Devices tab** | See "What is NOT proven" below. Nothing in the dashboard sets `viewMode` to `'devices'` except the Devices button itself. A *session*-navigation-on-click does exist and is fixed (row 3a), and the precondition for a mis-click is fixed (row 4), but the view-tab jump itself has no named mechanism. | — | **UNPROVEN — precondition removed, mechanism not named** |
| 4 | (not reported directly — found while tracing 2) A failed answer still **deleted** the banner | `App.tsx:2059-2067` called `dismissSessionNotification` regardless of `respondToPermission`'s return. `sendPermissionResponse` returns `false` without throwing on a down or OPEN→CLOSING socket (`connection.ts:3722`, `:3752` — #5699/#6308), and `dismissSessionNotification` **removes** the row (`connection.ts:5239-5243`), it does not mark it read. So a click that achieved nothing destroyed the only record of the request — and collapsed the banner block under the cursor. | `App.tsx:2059-2067`, `store/connection.ts:3722,3752,5239` | **PROVEN — fixed** |
| 3a | The **session tab bar left with no tab selected** | `sendPermissionResponse` auto-switches to the prompt's owning session. The owner id comes from `sessionNotifications`, falling back to a scan of `sessionStates` — **both of which outlive the session**: notifications are never pruned on close, and `sessionStates` retains a closed session's transcript. `switchSession` then writes `activeSessionId` **unconditionally**, with no membership check against `sessions`. `SessionBar` marks a tab active purely by `s.sessionId === activeSessionId`, so a dangling id renders the whole strip unselected. | `store/connection.ts:3780` (pre-fix), `store/connection.ts:4828`/`:4864`, `App.tsx:1379` | **PROVEN — fixed** |
| 3b | The **header flashing in a refresh/loading loop** | `PairedDevicesPanel` (the Devices panel) computes `resolvedFetch = fetchImpl ?? ((i,n) => window.fetch(i,n))` inline, feeds it to `refresh`'s `useCallback` deps, and mounts with `useEffect(() => { void refresh() }, [refresh])`. With `fetchImpl` omitted — **exactly what production does** — the `??` allocates a fresh arrow every render, so `refresh` changes identity every render, so the effect re-fires every render, and its own `setState` renders again. An **unbounded fetch/render loop** that flips the panel header's `Loading…`/`Refresh` label continuously. `SnapshotsPanel` has the identical defect. | `components/PairedDevicesPanel.tsx:156-179,230-232` (pre-fix), `components/SnapshotsPanel.tsx:173-196,243-245` (pre-fix), rendered propless at `App.tsx:2758` / `:2749` | **PROVEN — fixed** |

### The false-safety that hid 3b

All **26** existing tests across `PairedDevicesPanel.test.tsx` / `SnapshotsPanel.test.tsx`
pass `fetchImpl` as a prop, which pins the identity and makes the loop impossible.
Production passes nothing. The suite was green because it never rendered the shape that
ships — the "testing the source tree when the artifact is what ships" family in
`docs/false-safety-guards.md`. The new tests render the **production shape (no props)**
and are StrictMode-honest, since `main.tsx` mounts under `<StrictMode>`.

## What is NOT proven — the jump to Devices

Two independent sweeps of every `viewMode` writer found **no path that can select the
Devices tab except the Devices button itself** (`components/ViewSwitcher.tsx:152`).
Ruled out: keyboard shortcuts (`Cmd+1..9` switches *session* tabs; the only view-tab
shortcut toggles chat↔terminal, `useShortcutDispatch.ts:279`), the command palette
(`store/commands.ts` sets only chat/terminal/files), `openFileInBrowser` (files), the
StreamStallChip "View logs" action (system), the Tauri tray menu (console), the
ErrorBoundary (`window.location.reload()` only), and every handler in
`store/message-handler.ts` — which contains **zero** occurrences of `viewMode`.

Note one thing that surprised me and kills the most attractive alternative: the
localStorage rehydration of `viewMode` at `store/connection.ts:2347` is **dead code in
this package**. It lives inside `loadSavedConnection` (`connection.ts:2311`), whose only
caller in the monorepo is the *mobile app's* separate store
(`packages/app/src/screens/ConnectScreen.tsx:162`). The dashboard never restores
`viewMode` — not on reconnect, not even on boot. The two post-boot
`loadPersistedState()` calls (`switchServer` `:5327`, `connectLocal` `:5369`) apply only
`activeSessionId`. There is also no zustand `persist` middleware and no `storage` event
listener, so no cross-tab sync. So "the click triggered a reconnect that restored a
previously-persisted 'devices'" is not merely unproven — it is impossible.

So the click did not *navigate* to Devices through any code path I can name. The
remaining hypothesis is the reporter's own: a **physical mis-click caused by layout
shift**. `NotificationBanners` returns `null` when the last unread banner goes
(`NotificationBanners.tsx:45`), and `.notification-banners` is a normal-flow,
`flex-shrink: 0` block sitting **immediately above** `.view-switch-wrapper`
(`App.tsx:2567` → `:2581`, `theme/components.css:4999`). Answering the banner therefore
pulls the tab strip **up into the pointer's position**.

Quantified, because "roughly one banner height" is not a claim anyone can check:
`.notification-banner` is `padding: 6px 12px` at `--text-sm: 12px` with a body
`line-height: 1.5` → **~30px** tall. `.view-tab` is `padding: 6px 16px` at
`--text-base: 13px` plus a 2px bottom border → the strip is **~36px** tall. So the
collapse moves the tab strip up by roughly *five sixths of a tab row* — a pointer
resting on the banner's action row ends up squarely inside the strip, not merely near
it. Horizontally it is weaker, and the honest version is worth stating: the banner's
Allow/Deny sit at the far RIGHT (`.notification-banner-actions` under
`justify-content: space-between`), but in the DEFAULT layout the tabs at the right edge
are Checkpoints and Diff, not Devices — `.view-switch-spacer { flex: 1 }`
(`ViewSwitcher.tsx:153`) sits *after* Devices and pushes that tail group over. Devices
lands under the Allow button only when the strip overflows AND has been partially
drag-scrolled (`.view-switch` is `overflow-x: auto`; the spacer's `flex: 1` and the tail
group's `margin-left: auto` both collapse to 0 on overflow, so all 13-15 tabs pack in DOM
order and which tab occupies a given x is a pure function of `scrollLeft`).

And the same click cannot be retargeted at all: `mousedown` / `mouseup` / `click` are all
dispatched before React commits the dismissal. A **second** click is required — which is
precisely why fix 4 matters. Before it, clicking Allow on a stale banner over a dropped
socket did nothing except make the banner vanish, and an operator whose control just
"didn't work" clicks it again, now over the tab strip.

I can prove the adjacency and the collapse; I cannot prove that a single click
retargets onto a specific tab, and jsdom has no layout with which to test it. So the
fix here removes the *precondition* — a stale banner no longer offers a button to click
at all — rather than inventing a mechanism. The issue stays open with a comment naming
exactly what remains and how to instrument it.

## Fixes

1. **`fix(dashboard): stop the Devices and Snapshots panels refetching forever`** — memoize
   `resolvedFetch` on the prop in both panels. `resolvedGetToken` deliberately gets **no**
   memo: `getAuthToken` is a module import and the ternary changes exactly when the prop
   does, so `useMemo(..., [getToken])` is a provable no-op. It was written,
   mutation-tested, **survived its own mutant**, and was removed rather than shipped as a
   guard with no behaviour.
2. **`fix(dashboard): never make a session that no longer exists the active one`** —
   membership-check the auto-switch target against `sessions`. The legitimate
   cross-session jump is preserved and pinned as a positive control; the
   `permission_response` itself still goes to the server, which stays the authority on
   whether the request is pending.
3. **`fix(dashboard): a permission banner that is no longer pending loses its buttons`** —
   gate Allow/Deny on `isPermissionNotificationActionable`, which reads the prompt out of
   `sessionStates` and applies `isLivePermissionPrompt` from `@chroxy/store-core` — the
   **same** signal that already drives the tab badge and jump-to-pending, so no second
   notion of "still pending" is introduced (#7390's one-signal rule). **Conservative on
   absence**: a notification whose prompt the store has never seen keeps its buttons. The
   row survives with a "No longer pending" marker and a Dismiss, so the record of what was
   asked is not erased (#7353). The prop is **required with no default** — defaulting to
   "actionable" is the hole being closed. Re-checked inside `onClick` as well as at
   render, because the render gate holds the `Date.now()` of the *last* render and nothing
   re-renders on the mere passage of a deadline.
4. **`fix(dashboard): a permission banner whose answer never sent must not vanish`** —
   dismiss only when `respondToPermission` reports the answer went on the wire. `'queued'`
   is not a failure, so only an explicit `false` holds the row; the over-strict
   `!== 'sent'` variant has its own mutant.

## Red-first + mutation evidence

Every guard was proven red before the fix and then attacked.

| Guard | Red before fix | Mutant | Result |
|---|---|---|---|
| `resolvedFetch` memo (both panels) | `expected 21 to be 1` (and 14/16/21 across the four cases) — the loop never settles; the first shape of the test hit a 30s `act()` timeout because React never quiesced | **A** revert to the inline `??` in `PairedDevicesPanel` | **killed** — 4 tests, `expected 13 to be 1` |
| `resolvedGetToken` memo | — | **B** revert to the bare expression | **SURVIVED** → the refinement has no behaviour; **deleted** rather than shipped |
| `resolvedFetch` memo (final files, re-run after the B deletion) | — | **A2** revert both panels | **killed** — all 6 tests, `expected 28 to be 1` |
| auto-switch membership check | `expected 'sess-dead' to be 'sess-live'` ×2 | **C** drop the check (`targetIsLive = !!targetSid`) | **killed** — 2 tests |
| ↳ positive control not vacuous | — | **D** invert the check (`!sessions.some(...)`) | **killed** — 3 tests, incl. `expected 'sess-live' to be 'sess-other'` |
| banner render gate | Allow/Deny still rendered; `expected [ …2 buttons ] to have a length of 1` | **E** drop `&& isPermissionActionable(n)` | **killed** — 3 tests |
| `isPermissionNotificationActionable` negatives | `TypeError: isPermissionNotificationActionable is not a function` (×7) | **F** `return true` unconditionally | **killed** — 4 tests |
| ↳ conservative-on-absence branch | — | **G** flip the absence fallback to `return false` | **killed** — 1 test (the CONSERVATIVE case) |
| click-time re-check | — | **H** drop the `onClick` guard | **killed** — 1 test (predicate flipped without a re-render) |
| dismiss-only-on-send | `expected "vi.fn()" to not be called at all, but actually been called 1 times` ×2 | **I** revert the Approve guard | **killed** — 1 test |
| ↳ `'queued'` is not a failure | — | **J** tighten to `!== 'sent'` | **killed** — 1 test (the queued positive control) |

The A2 re-run matters: the fix changed after mutant A (the B deletion), so the original
mutation was re-applied to the *final* file rather than trusting the earlier kill.

## Suites (bare exit codes, captured without a pipe)

| Command | Exit |
|---|---|
| `npx tsc --noEmit` | `0` |
| `npx vitest run src/store/` | `0` — 81 files, **1175** tests |
| `npx vitest run` (whole dashboard) | `0` — 309 files, **5367** tests |
| `npx vitest run` (the 5 touched component files) | `0` — 56 tests |
| `npx vitest run src/App.test.tsx` | `0` — 123 tests |
| `npm run build` (vite) | `0` |

`git status --short` is clean; the committed tree is what was verified.

## Notes / follow-ons

- Two pre-existing `#1710` fixtures seeded `sessionStates` **without** a `sessions`
  roster — not a state the server can produce for a live session — and are corrected
  here. The wire-order test's `switchIdx === -1 ? Infinity` also made it pass just as
  happily when **no** switch was sent at all; it now asserts the switch actually occurred
  before checking the ordering.
- **Tap targets, out of scope but recorded:** `.notification-banner-btn` is
  `padding: 2px 10px` at `--text-xs` — well under the 44pt floor. Growing it would change
  every banner's height (and the banner height is itself the layout-shift surface in the
  unproven symptom above), so it is deliberately untouched here rather than folded into a
  bug fix.

### DIFF
--- a/packages/dashboard/src/App.test.tsx
+++ b/packages/dashboard/src/App.test.tsx
@@ -3159,4 +3159,66 @@ describe('#7466 — cross-session permission banner, failed send', () => {
     fireEvent.click(screen.getByLabelText('Deny'))
     expect(dismissSessionNotification).toHaveBeenCalledWith('n-1')
   })
+
+  // #7466 review finding 1 — the App->banner WIRING of the third gate. The
+  // component's own behaviour is pinned in NotificationBannersStale.test.tsx;
+  // this pins that App actually feeds it `connectionPhase`, which is the
+  // "guard wired to only some of its callers" failure the component tests
+  // cannot see.
+  it('disables the banner buttons while the socket is down', () => {
+    stateOverrides = connectedWithBanner({ connectionPhase: 'reconnecting' })
+    render(<App />)
+    expect(screen.getByLabelText('Allow')).toBeDisabled()
+    expect(screen.getByLabelText('Deny')).toBeDisabled()
+    expect(screen.getByTestId('notification-banner-disconnected-hint')).toBeInTheDocument()
+  })
+
+  it('POSITIVE CONTROL: the same banner is enabled while connected', () => {
+    stateOverrides = connectedWithBanner()
+    render(<App />)
+    expect(screen.getByLabelText('Allow')).toBeEnabled()
+    expect(screen.queryByTestId('notification-banner-disconnected-hint')).not.toBeInTheDocument()
+  })
+
+  // #7466 review nit 5 — an INERT permission row retires by mark-read, not by
+  // deletion. Wiring-level: the component gets `onMarkRead`, and App must hand
+  // it `markSessionNotificationRead` rather than the removing `onDismiss`.
+  it('marks an inert permission row read instead of deleting it', () => {
+    const markSessionNotificationRead = vi.fn()
+    const dismissSessionNotification = vi.fn()
+    stateOverrides = connectedWithBanner({
+      markSessionNotificationRead,
+      dismissSessionNotification,
+      // A prompt the store has SEEN and knows is expired -> 'not-pending'.
+      // The rest of the SessionState shape is present because App reads several
+      // of these fields off the active session while rendering.
+      sessionStates: {
+        s1: {
+          messages: [{
+            id: 'm-1', type: 'prompt', content: 'Bash: rm -rf /tmp/x',
+            timestamp: 1, requestId: 'req-abc', expiresAt: Date.now() - 1,
+          }],
+          streamingMessageId: null,
+          activeModel: null,
+          permissionMode: null,
+          contextUsage: null,
+          sessionCost: null,
+          isIdle: true,
+          activeAgents: [],
+          isPlanPending: false,
+          systemMessages: [],
+          childAgentEvents: [],
+          sessionRules: [],
+          claudeReady: true,
+        },
+      },
+    })
+    render(<App />)
+
+    expect(screen.getByTestId('notification-banner-stale')).toBeInTheDocument()
+    expect(screen.queryByLabelText('Allow')).not.toBeInTheDocument()
+    fireEvent.click(screen.getByLabelText('Dismiss'))
+    expect(markSessionNotificationRead).toHaveBeenCalledWith('n-1')
+    expect(dismissSessionNotification).not.toHaveBeenCalled()
+  })
 })
--- a/packages/dashboard/src/App.test.tsx
+++ b/packages/dashboard/src/App.test.tsx
@@ -3077,3 +3077,86 @@ describe('App-level view-mode effect (#5998)', () => {
     expect(subscribeTerminalMirror).toHaveBeenLastCalledWith('s1')
   })
 })
+
+// ---------------------------------------------------------------------------
+// #7466 — a banner whose answer never went on the wire must not vanish
+// ---------------------------------------------------------------------------
+describe('#7466 — cross-session permission banner, failed send', () => {
+  const bannerSession = {
+    sessionId: 's1', name: 'Chroxy', cwd: '/tmp', type: 'cli' as const,
+    hasTerminal: true, model: null, permissionMode: null, isBusy: false,
+    createdAt: Date.now(), conversationId: null,
+  }
+
+  const livePermissionBanner = {
+    id: 'n-1',
+    sessionId: 's1',
+    sessionName: 'Chroxy',
+    eventType: 'permission' as const,
+    message: 'Bash: rm -rf /tmp/x',
+    timestamp: Date.now(),
+    requestId: 'req-abc',
+  }
+
+  function connectedWithBanner(overrides: Record<string, unknown> = {}) {
+    return {
+      connectionPhase: 'connected',
+      sessions: [bannerSession],
+      activeSessionId: 's1',
+      // No prompt message for req-abc anywhere: the banner's staleness gate is
+      // conservative on absence, so the buttons render. That keeps this case
+      // about the SEND result and nothing else.
+      sessionStates: {},
+      sessionNotifications: [livePermissionBanner],
+      ...overrides,
+    }
+  }
+
+  it('keeps the banner when sendPermissionResponse reports the send failed', () => {
+    // #6308/#5699: the socket can be OPEN at render and CLOSING by the click, and
+    // sendPermissionResponse returns false rather than throwing. Dismissing anyway
+    // deleted the row outright (dismissSessionNotification filters it out of the
+    // store) — so the operator's only record of the request was destroyed by a
+    // click that did nothing, and the banner stack collapsing under the cursor is
+    // what invites the second, mis-aimed click.
+    const sendPermissionResponse = vi.fn(() => false as const)
+    const dismissSessionNotification = vi.fn()
+    stateOverrides = connectedWithBanner({ sendPermissionResponse, dismissSessionNotification })
+    render(<App />)
+
+    fireEvent.click(screen.getByLabelText('Allow'))
+    expect(sendPermissionResponse).toHaveBeenCalledWith('req-abc', 'allow')
+    expect(dismissSessionNotification).not.toHaveBeenCalled()
+  })
+
+  it('keeps the banner when a Deny fails to send', () => {
+    const sendPermissionResponse = vi.fn(() => false as const)
+    const dismissSessionNotification = vi.fn()
+    stateOverrides = connectedWithBanner({ sendPermissionResponse, dismissSessionNotification })
+    render(<App />)
+
+    fireEvent.click(screen.getByLabelText('Deny'))
+    expect(sendPermissionResponse).toHaveBeenCalledWith('req-abc', 'deny')
+    expect(dismissSessionNotification).not.toHaveBeenCalled()
+  })
+
+  it('POSITIVE CONTROL: dismisses the banner when the answer WAS sent', () => {
+    const sendPermissionResponse = vi.fn(() => 'sent' as const)
+    const dismissSessionNotification = vi.fn()
+    stateOverrides = connectedWithBanner({ sendPermissionResponse, dismissSessionNotification })
+    render(<App />)
+
+    fireEvent.click(screen.getByLabelText('Allow'))
+    expect(dismissSessionNotification).toHaveBeenCalledWith('n-1')
+  })
+
+  it("POSITIVE CONTROL: dismisses on 'queued' too — a queued answer is not a failure", () => {
+    const sendPermissionResponse = vi.fn(() => 'queued' as const)
+    const dismissSessionNotification = vi.fn()
+    stateOverrides = connectedWithBanner({ sendPermissionResponse, dismissSessionNotification })
+    render(<App />)
+
+    fireEvent.click(screen.getByLabelText('Deny'))
+    expect(dismissSessionNotification).toHaveBeenCalledWith('n-1')
+  })
+})
--- a/packages/dashboard/src/App.tsx
+++ b/packages/dashboard/src/App.tsx
@@ -57,7 +57,7 @@ import { BillingWarningBanner } from './components/BillingWarningBanner'
 import { ConnectionAnnouncer } from './components/ConnectionAnnouncer'
 import { StdinDisabledBanner } from './components/StdinDisabledBanner'
 import { WelcomeScreen } from './components/WelcomeScreen'
-import { NotificationBanners, isPermissionNotificationActionable } from './components/NotificationBanners'
+import { NotificationBanners, permissionNotificationStatus } from './components/NotificationBanners'
 import { PendingPairRequests } from './components/PendingPairRequests'
 import { type ToastItem } from './components/Toast'
 import { FileBrowserPanel } from './components/FileBrowserPanel'
@@ -2049,19 +2049,26 @@ export function App() {
     return result
   }, [sendUserQuestionResponse])
 
-  // #7466 — the banner's staleness gate. `sessionStates` is the store's own
-  // record of every prompt, and `isLivePermissionPrompt` (inside the helper) is
-  // the SAME signal the tab badge and jump-to-pending already use, so this adds
-  // no second notion of "still pending". Read at render time: `sessionStates` is
-  // replaced on every WS event, so a prompt the server retires (#7335 stamps
-  // `expiresAt`) disarms its banner on the next event, and a prompt that simply
-  // times out disarms on the next render past its deadline — the same way the
-  // inline PermissionPrompt's own countdown retires its buttons. The CLICK-time
-  // re-check lives in NotificationBanners itself (it calls the predicate again
+  // #7466 — the banner's staleness + connectivity gate. `sessionStates` is the
+  // store's own record of every prompt, and `isLivePermissionPrompt` (inside the
+  // helper) is the SAME signal the tab badge and jump-to-pending already use, so
+  // this adds no second notion of "still pending". Read at render time:
+  // `sessionStates` is replaced on every WS event, so a prompt the server retires
+  // (#7335 stamps `expiresAt`) disarms its banner on the next event, and a prompt
+  // that simply times out disarms on the next render past its deadline — the same
+  // way the inline PermissionPrompt's own countdown retires its buttons. The
+  // CLICK-time re-check lives in NotificationBanners itself (it calls this again
   // inside onClick): one place to gate, and reachable from a component test.
-  const isPermissionActionable = useCallback(
-    (n: SessionNotification) => isPermissionNotificationActionable(n, sessionStates, Date.now()),
-    [sessionStates],
+  //
+  // `connectionPhase === 'connected'` is the third gate, matching
+  // PermissionPrompt's `connected` — without it the banner showed ENABLED
+  // Allow/Deny over a dead socket, and since the handlers below correctly refuse
+  // to dismiss when the send fails, that click produced no observable change at
+  // all.
+  const permissionStatus = useCallback(
+    (n: SessionNotification) =>
+      permissionNotificationStatus(n, sessionStates, Date.now(), connectionPhase === 'connected'),
+    [sessionStates, connectionPhase],
   )
 
   // #6222/#6224: respondToPermission (sendPermissionResponse) now marks the
@@ -2593,8 +2600,9 @@ export function App() {
             onApprove={handleBannerApprove}
             onDeny={handleBannerDeny}
             onDismiss={dismissSessionNotification}
+            onMarkRead={markSessionNotificationRead}
             onSwitchSession={handleSwitchSession}
-            isPermissionActionable={isPermissionActionable}
+            permissionStatus={permissionStatus}
           />
         )}
 
--- a/packages/dashboard/src/App.tsx
+++ b/packages/dashboard/src/App.tsx
@@ -2072,13 +2072,21 @@ export function App() {
   // consumers expect) and unconditional (it ran even when the send was refused
   // while disconnected, falsely clearing the prompt). Dropped in favour of the
   // single choke point.
+  // #7466 — dismiss ONLY when the answer actually went on the wire.
+  // `sendPermissionResponse` returns false without throwing when the socket is
+  // down or flipped OPEN -> CLOSING mid-send (#5699 / #6308), and
+  // `dismissSessionNotification` REMOVES the row from the store outright. So a
+  // click that achieved nothing used to delete the operator's only record of
+  // the request, and the banner stack collapsing under the cursor is what
+  // invites the second, mis-aimed click. 'queued' is not a failure — the
+  // message will send — so only an explicit `false` holds the row.
   const handleBannerApprove = useCallback((requestId: string, notificationId: string) => {
-    respondToPermission(requestId, 'allow')
+    if (respondToPermission(requestId, 'allow') === false) return
     dismissSessionNotification(notificationId)
   }, [respondToPermission, dismissSessionNotification])
 
   const handleBannerDeny = useCallback((requestId: string, notificationId: string) => {
-    respondToPermission(requestId, 'deny')
+    if (respondToPermission(requestId, 'deny') === false) return
     dismissSessionNotification(notificationId)
   }, [respondToPermission, dismissSessionNotification])
 
--- a/packages/dashboard/src/App.tsx
+++ b/packages/dashboard/src/App.tsx
@@ -24,6 +24,7 @@ import {
 } from '@chroxy/store-core'
 import { useConnectionStore } from './store/connection'
 import type { BaseSessionState, ContextOccupancy } from '@chroxy/store-core'
+import type { SessionNotification } from './store/types'
 
 import { Sidebar, type RepoNode, type ContextMenuTarget } from './components/Sidebar'
 import { resolveActivePrimaryClientId } from './components/ViewersIndicator'
@@ -56,7 +57,7 @@ import { BillingWarningBanner } from './components/BillingWarningBanner'
 import { ConnectionAnnouncer } from './components/ConnectionAnnouncer'
 import { StdinDisabledBanner } from './components/StdinDisabledBanner'
 import { WelcomeScreen } from './components/WelcomeScreen'
-import { NotificationBanners } from './components/NotificationBanners'
+import { NotificationBanners, isPermissionNotificationActionable } from './components/NotificationBanners'
 import { PendingPairRequests } from './components/PendingPairRequests'
 import { type ToastItem } from './components/Toast'
 import { FileBrowserPanel } from './components/FileBrowserPanel'
@@ -2048,6 +2049,21 @@ export function App() {
     return result
   }, [sendUserQuestionResponse])
 
+  // #7466 — the banner's staleness gate. `sessionStates` is the store's own
+  // record of every prompt, and `isLivePermissionPrompt` (inside the helper) is
+  // the SAME signal the tab badge and jump-to-pending already use, so this adds
+  // no second notion of "still pending". Read at render time: `sessionStates` is
+  // replaced on every WS event, so a prompt the server retires (#7335 stamps
+  // `expiresAt`) disarms its banner on the next event, and a prompt that simply
+  // times out disarms on the next render past its deadline — the same way the
+  // inline PermissionPrompt's own countdown retires its buttons. The CLICK-time
+  // re-check lives in NotificationBanners itself (it calls the predicate again
+  // inside onClick): one place to gate, and reachable from a component test.
+  const isPermissionActionable = useCallback(
+    (n: SessionNotification) => isPermissionNotificationActionable(n, sessionStates, Date.now()),
+    [sessionStates],
+  )
+
   // #6222/#6224: respondToPermission (sendPermissionResponse) now marks the
   // prompt answered with the canonical decision token itself — only when the
   // answer actually went over the wire (after the disconnected-socket guard).
@@ -2570,6 +2586,7 @@ export function App() {
             onDeny={handleBannerDeny}
             onDismiss={dismissSessionNotification}
             onSwitchSession={handleSwitchSession}
+            isPermissionActionable={isPermissionActionable}
           />
         )}
 
--- a/packages/dashboard/src/components/NotificationBanners.test.tsx
+++ b/packages/dashboard/src/components/NotificationBanners.test.tsx
@@ -1,11 +1,11 @@
 /**
  * NotificationBanners — cross-session notification banners with quick-approve (#1369)
  *
- * Every render here passes `isPermissionActionable={() => true}` (#7466): these
- * cases predate the staleness gate and are about the LIVE-prompt rendering they
- * were written for. The gate's own behaviour — including that the prop is
- * required precisely so no call site can silently reopen the hole — is pinned in
- * NotificationBannersStale.test.tsx.
+ * Every render here passes `permissionStatus={() => 'actionable'}` (#7466): these
+ * cases predate the staleness/connectivity gate and are about the LIVE-prompt
+ * rendering they were written for. The gate's own behaviour — including that the
+ * prop is required precisely so no call site can silently reopen the hole — is
+ * pinned in NotificationBannersStale.test.tsx.
  */
 import { describe, it, expect, vi, afterEach } from 'vitest'
 import { render, screen, fireEvent, cleanup } from '@testing-library/react'
@@ -38,7 +38,8 @@ describe('NotificationBanners', () => {
         onDeny={vi.fn()}
         onDismiss={vi.fn()}
         onSwitchSession={vi.fn()}
-        isPermissionActionable={() => true}
+        onMarkRead={vi.fn()}
+        permissionStatus={() => 'actionable' as const}
       />
     )
     expect(container.querySelector('.notification-banners')).not.toBeInTheDocument()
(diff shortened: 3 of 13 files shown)


blamechris:
## Code Review: PR #7472

Reviewed at head `7640fe5ac`, worktree `/.claude/worktrees/manual-7466`, read-only except for
mutation runs (all restored; `git status --short` clean, byte-identical to backups by md5).

**Summary:** the four fixes are correct, and I could not break any of them. I independently
re-ran three of the claimed mutants and reproduced the kills with the same shapes, and I
independently confirmed the false-safety diagnosis behind 3b. Everything below is additive —
two gaps filed as issues, two nits, and one addition to the unproven symptom.

---

### Verified independently

**The false-safety claim behind 3b is real, and stronger than "26 tests pinned it away".**
With the inline `??` restored in `PairedDevicesPanel` only, the *pre-existing*
`PairedDevicesPanel.test.tsx` runs **11 passed / exit 0** with the unbounded loop fully present,
while `PanelRefreshLoop.test.tsx` goes red at `expected 34 to be 1`. The new file is the only
thing standing between this defect and a green suite.

**Mutation replication** (mine, not read off the table):

| Mutant | Result |
|---|---|
| **A** inline `??` restored in `PairedDevicesPanel` | **killed** — 4 tests, `expected 34 to be 1`, `45 to be 2`, `57 to be 1`, `65 to be 1`; SnapshotsPanel's 2 stayed green (clean isolation) |
| **E** drop `&& isPermissionActionable(n)` | **killed** — 3 tests, `expected [ <button …(3)></button>, …(1) ] to have a length of 1 but got 2` |
| **C** `targetIsLive = !!targetSid` | **killed** — 2 tests, `expected 'sess-dead' to be 'sess-live'` ×2 |

**Suites, bare exit codes, no pipe:** `npx tsc --noEmit` -> `0`; `npx vitest run` (whole
dashboard) -> `0`, **309 files / 5367 tests**. Matches the PR body exactly.

**The membership target is the right one.** `SessionBar`'s tabs derive from `sessions` and
`isActive: s.sessionId === activeSessionId` (`App.tsx:1379`), so checking against `sessions` is
precisely the predicate that decides whether a tab can be selected. Not an approximation of it.

**The `expiresAt`-absence hole I went looking for does not exist.** `isLivePermissionPrompt`
requires `!!m.expiresAt`, and `remainingMs` is `.optional()` on the wire
(`protocol/src/schemas/server/stream.ts:458`) -> a schema-legal `permission_request` that omits
it produces a prompt with **no** `expiresAt`, which the new gate judges "not live". That looked
like a live prompt being silently disarmed. It isn't: the inline renderer gates on the same
thing — `storeMsg.requestId && storeMsg.expiresAt && !storeMsg.answered`
(`hooks/useMessageRenderer.tsx:131`) — so it renders no interactive prompt for that shape
either. Banner and inline agree. Worth the sentence because the equivalence is not obvious.
(Also checked the `null` normalisation in `handlePermissionRequest` — no `Date.now() + undefined`
`NaN` path.)

**Scope on 3b is right, and I checked the whole list.** Four components carry the
`fetchImpl ?? inline-arrow` shape, not two: `PoolStatsPanel.tsx:98` and `PagesPanel.tsx:74` also
have it. Both are **immune**, because they route the mount effect through a `refreshRef`
(`refreshRef.current = refresh`, deps `[interval]` / `[]`) instead of `[refresh]`. Fixing only
the two that lack that indirection is correct. Noting it because a future edit that "tidies"
either of those effects to `[refresh]` reintroduces the loop with no test to catch it.

**The #1710 fixtures genuinely tightened.** `expect(switchIdx).toBeGreaterThan(-1)` is a real
assertion where `switchIdx === -1 ? Infinity` was a hole that passed when no switch was sent at
all. The added `sessions` roster is accommodation, but justified accommodation — `sessionStates`
without a matching roster row is not a state the server produces for a live session, and the new
assertion pays for it.

---

### Findings

**1. Major — the banner still lacks the inline prompt's third gate (`connected`), and fix 4 makes the resulting dead click fully silent.** -> filed as #7474

The mechanism table cites the inline prompt's three gates; the new gate reuses two of them
(`answered`, expiry, both via `isLivePermissionPrompt`). `NotificationBanners` renders whenever
`sessionNotifications.length > 0` (`App.tsx:2589`), with no reference to connection state — so
while the socket is down the banner shows **enabled** Allow/Deny. `sendPermissionResponse`'s own
comment states the contract this leaves unmet (`connection.ts:3718-3721`):

> the answer buttons + keyboard shortcuts gate on `connected` in PermissionPrompt.tsx (respond())
> **so the operator gets clear feedback rather than a dead click**

The inline prompt honours it — `disabled={submitting || !connected}` plus
`perm-disconnected-hint`, "Disconnected — reconnect to answer." (`PermissionPrompt.tsx:302-356`).

Fix 4 is right to stop the record being deleted, but the consequence is that a disconnected click
now produces **no observable change whatsoever** — previously the row at least vanished (wrongly,
but visibly). That is the exact state this PR's own analysis names as the trigger:

> the banner stack collapsing under the cursor is what invites the second, mis-aimed click

The same applies to the click-time re-check: when it refuses, nothing in the store changes, so
React does not re-render, so the buttons stay visibly enabled and the next click is equally inert.
`NotificationBannersStale.test.tsx`'s "refuses the CLICK" case asserts `onApprove` was not
called — correct — and there is no UI change it *could* assert, because there is none.

`connectionPhase === 'connected'` is already derived two lines away (`isSocketConnected`,
`App.tsx:1547`), so this is ~3 lines if you'd rather fold it in than defer.

**2. Major — the membership check guards 1 of 4 `switchSession` callers.** -> filed as #7475

`switchSession` still writes `activeSessionId` unconditionally (`connection.ts:4837`, `:4874`);
only the `sendPermissionResponse` call site checks. The other three:

- `OrchestrationRunsSection.tsx:110` "Open session" — the button is gated on `node.sessionId`
  being present, **not** on node status, so a completed run's node still offers it after its
  sub-session is destroyed. Same dead id, same unselected tab strip.
- `ControlRoomView.tsx:411` -> `CrossSessionMissionControl` / `ActivityTree` jump-to-intervene.
- `message-handler.ts:1470` `switchToRestoredSession` — this one legitimately **must not** get
  the check (a restored session can precede its `session_list` entry), which is exactly why the
  predicate wants to live at the choke point with an explicit opt-out rather than be copied
  per-call-site and missed at the next one.

This is the "guard wired to only some of its callers" family already in
`docs/false-safety-guards.md` (#7262). The PR fixes the reported reproduction; the defect stays
reachable from at least one other button.

**3. Minor — a surviving mutant on the shipped memo.** Mutating `[fetchImpl]` -> `[]` in
`PairedDevicesPanel`'s `useMemo` leaves `PanelRefreshLoop.test.tsx` + `PairedDevicesPanel.test.tsx`
at **17 passed / exit 0**. The dep list is unpinned: nothing proves `resolvedFetch` tracks a
*changing* `fetchImpl`. The file already has the right shape for this — the
"fresh inline `getToken`" positive control — it just wasn't mirrored for `fetchImpl`. Bumping
`fetchImpl` identity from `Bumper` and asserting exactly one extra fetch kills it. (Behaviourally
equivalent for today's callers, since production passes no prop — but it's a shipped guard whose
dep list no test defends, and this repo's rule is to name the differing input or cut it. The input
is nameable.)

**4. Nit — `useMemo` is a performance hint, not a semantic guarantee.** React's docs are explicit
that a memoized value may be discarded and recomputed; here identity stability *is* the
correctness property, not an optimisation. A module-level constant removes the class outright,
needs no dep list, is shorter than the memo, and makes finding 3 unfalsifiable by construction:

```ts
const WINDOW_FETCH: typeof fetch = (input, init) => window.fetch(input, init)
// ...
const resolvedFetch = fetchImpl ?? WINDOW_FETCH
```

**5. Nit — the inert row's only affordance deletes the record it was kept to preserve.**
The stale row offers `Dismiss` -> `dismissSessionNotification`, which **removes** the entry
(`connection.ts:5239-5241`), where `markSessionNotificationRead` (`:5255`) stamps `readAt` and
keeps it in the #4890 widget history. Given the rationale is "the record of what was asked is not
erased (#7353)", mark-read looks like the better verb for a row that is now purely a record.
Pre-existing `Dismiss` semantics, so entirely your call.

---

### On the unproven symptom 2

The write-up is honest and the devtools instrumentation suggestion is actionable — I'd merge it
as-is. An independent sweep (every `viewMode` writer, `setViewMode` call site, global listener,
Tauri bridge + Rust menu map, palette/shortcut dispatch, index-based selection, and
pairing/permission routing) **confirmed** the claim: nothing writes `viewMode = 'devices'` except
`ViewSwitcher.tsx:152`. I could not falsify it. Two additions:

- **The vertical collapse is understated.** The geometry section quantifies one banner (~30px),
  but `MAX_VISIBLE = 3` and `App.tsx:2589` unmounts the whole component when the last unread
  goes — so the worst case is ~3 rows plus the container, not ~30px.
- **There is a horizontal vector that needs no drag-scroll.** The PR calls the horizontal case
  weaker, requiring overflow *and* a partial drag-scroll. But answering a permission can raise
  `unreadSystemCount`, which changes the `.system-badge` width (`ViewSwitcher.tsx:144`); the
  measuring effect at `ViewSwitcher.tsx:62` lists `unreadSystemCount` in its deps and toggles the
  24px scroll arrows — so tabs shift **horizontally** as a direct consequence of the same click.
  That is a named mechanism for the half the write-up could only call weak, and it is testable
  the same way you proposed testing the vertical one.
- Agreed the `connection.ts:2347` rehydration is dead in this package. Worth noting the *writer*
  is not dead: `persistViewMode` writes the global unscoped `chroxy_view_mode` on every
  `setViewMode` and `VALID_VIEW_MODES` whitelists `'devices'`, so the moment anyone wires
  `loadSavedConnection` into the dashboard, a clickless path to Devices exists. A comment at the
  call site would be cheap insurance.

---

### CI

`Server Tests` is **red** on this head. This diff touches 12 files, all under
`packages/dashboard/`, so it cannot be the cause; `Server Windows Tests` passed on the same
commit, and main at the branch point (`e3fd9731e`) shows the inverse pair (Windows red, Server
Tests green) — the flaky-self-hosted-runner signature tracked in #7471. Not attributable, but it
is a required check and the gate is not satisfied: re-run it before merging rather than reasoning
past it. `App Tests` and `Release PR Subject` were still pending at review time.

---

### Verdict
- [x] Comment - Feedback only, author decides

Nothing shipped here is wrong, and the evidence discipline is the strongest part of it —
the A2 re-run after the B deletion, and deleting a guard that survived its own mutant rather
than shipping it, are both the right calls. Findings 1 and 2 are gaps rather than defects and
are filed (#7474, #7475); finding 1 is the one I'd most want folded in, since it's three lines
and it closes the dead click this PR's own reasoning identifies as the trigger. Merge is gated
on the `Server Tests` re-run regardless.

blamechris:
Thanks — folded in findings 1, 3, 4 and 5. Head is now `17d758c67` (4 commits added).
The independent mutation replication was useful: your `[fetchImpl]` -> `[]` mutant
reproduced here at **17 passed / exit 0** before I touched it.

## Fold 1 — the third gate (`connected`)

`682d40a71` .. `3621a3247` — you were right that fix 4 made the dead click worse, not
better, and that the PR's own reasoning named that as the trigger.

Implemented as a **three-state status rather than a boolean**, because "not answerable"
has two meanings that must not render alike:

| status | render |
|---|---|
| `not-pending` | inert record, "No longer pending", retire it |
| `disconnected` | buttons **stay**, `disabled`, "Disconnected — reconnect to answer." |
| `actionable` | enabled |

Collapsing `disconnected` into `not-pending` would tell the operator a live request was
gone and offer to retire a record still awaiting an answer — so pending-ness is checked
first (an expired request is dead whether or not the socket is up), and the disconnected
row keeps its buttons in exactly the disabled + hint shape `PermissionPrompt` uses, with
the same wording.

**On "one shared implementation":** I tried, and I don't think it's available. The two
gates answer the same question from genuinely different vantage points —
`PermissionPrompt` reads a monotonic `performance.now()` countdown anchored at *its own
mount* (#3619), and a banner has no mount lifetime to anchor one to, so it answers from
store state. Extracting a common helper would either force the banner to fake a mount or
force the prompt to abandon the monotonic clock #3619 exists for. So: two
implementations, cross-referenced **both ways** (`17d758c67` adds the return leg from
`PermissionPrompt.tsx`, naming the two functions that must move together).

**Red-first:** mutant **M** restores the pre-fold behaviour exactly — `permissionNotificationStatus`
ignores `connected` — and the two new predicate cases go red (`is 'disconnected' for a
LIVE prompt while the socket is down`, and the conservative-absence variant). The
absence path mattered more than I expected: it is the *most common* banner shape, so
without it the hole reopens for almost every real banner.

Component + wiring mutants, all killed:

| Mutant | Result |
|---|---|
| **L** drop `disabled` from both buttons | killed — 1 |
| **M** ignore `connected` in the predicate | killed — 2 |
| **N** collapse `disconnected` into the `not-pending` branch | killed — 4 |
| **O** drop the click-time re-check | killed — 2 |
| **Q** App hardcodes `connected: true` (gate right, wiring absent) | killed — 1 |

Q is there because of your finding 2's framing: a gate correct in the component and
unwired in App is the same defect class, invisible to component tests. There is now an
App-level test for the wiring.

One honest note: mutant **L** killed only *one* test, not two. `cannot fire
onApprove/onDeny even if a click is dispatched at it` survives it, because the click-time
re-check also blocks that path — a second guard making the first guard's test vacuous.
Each guard still has a test that kills its own mutant (`disables Allow and Deny` for L,
the TOCTOU case for O), so I left both rather than thin the defence.

## Fold 2 — findings 3 + 4 together

`682d40a71`. Took nit 4's suggestion, and it subsumes finding 3: a module-level
`WINDOW_FETCH` has **no dep list to mutate**, so your `[fetchImpl]` -> `[]` mutant is
unwritable rather than merely unkilled. Your framing of `useMemo` as a hint and not a
guarantee is the better argument of the two — identity stability is the correctness
property here, not an optimisation.

Unwritable isn't unnecessary, so the input the dep list was meant to defend is pinned
anyway: `a CHANGING fetchImpl prop re-arms exactly once`, mirroring the `getToken`
control as you noted. It's killed by mutant **K** (`resolvedFetch = WINDOW_FETCH`,
ignoring the prop). And I re-ran the *original* loop mutation against the constant-based
files — **mutant A3, still red across all six cases** — rather than assume the earlier
kill carried over a changed fix.

## Fold 3 — nit 5

Taken; it isn't really optional given the rationale it contradicts. The inert row's
Dismiss now calls `markSessionNotificationRead` (stamps `readAt`, banner stack drops it,
#4890 widget keeps it) instead of `dismissSessionNotification` (removes outright).
**Scoped to inert permission rows** — `completed` / `question` / `error` banners keep the
pre-existing delete, since widening it would be an unrequested change to three other
banner types. Both directions pinned; mutants **P** (component reverts to delete) and
**R** (App wires the removing dismiss as `onMarkRead`) both killed.

## Deferred, as agreed

Finding 2's other three `switchSession` callers -> #7475 (agreed the predicate wants to
live at the choke point with an explicit opt-out for `switchToRestoredSession`, not
copied per call site). Broader disconnect-feedback design -> #7474.

Not folded: your suggestion of a comment at the dead `loadSavedConnection` call site.
Real point, but it's outside these three and belongs with whoever wires it.

## Verification

| Command | Exit |
|---|---|
| `npx tsc --noEmit` | `0` |
| `npx vitest run src/store/` | `0` — 81 files, 1175 tests |
| `npx vitest run` (5 touched component files) | `0` — 66 tests |
| `npx vitest run src/App.test.tsx` | `0` — 126 tests |
| `npx vitest run` (whole dashboard) | `0` — 309 files, **5380** tests |
| `npm run build` | `0` |

`git status --short` clean; committed tree is what was verified. Every mutation restored
by absolute-path `cp` from byte backups (`App.tsx` re-verified md5-identical after the
Q/R pair).

Re-running `Server Tests` per your note — dashboard-only diff, Windows green on the same
head, #7471 signature. Not reasoning past it, just not chasing it either.

blamechris:
## Re-verify at `17d758c67`

All three folds hold. I re-ran the mutants rather than reading the table, and added two of my
own — one of which was aimed squarely at the fold-1 state machine. Verdict updated to **Approve**
on the code; the merge gate is a separate matter, see CI at the bottom.

### 1. Fold 1 — the three-state status

**The ordering is correct, and — better than I expected — it is pinned rather than merely
reasoned.** I wrote my own mutant **ORD**, swapping the two checks so `connected` is tested
before pending-ness:

```
if (!connected) return 'disconnected'
if (!isStillPending(...)) return 'not-pending'
```

**Killed** — 1 test, `prefers 'not-pending' over 'disconnected' for a DEAD prompt on a dead
socket`. So the comment's argument ("an expired request is dead whether or not the socket is up")
is defended by an assertion, not just prose. That was the thing I most wanted to check and it is
the strongest part of this fold.

Walking the orderings: **disconnected + expired** -> `not-pending` (the test above);
**answered-elsewhere while disconnected** -> `not-pending`, by composition of
`is 'not-pending' once the prompt carries a decision` with the precedence the ORD mutant pins;
**pending + disconnected** -> `disconnected`; **absent prompt + disconnected** -> `disconnected`,
which has its own case (`is 'disconnected' on the conservative-absence path too`). That last one
is the only corner that can overpromise — an absent prompt that is in fact long dead reads
"reconnect to answer" — and it self-corrects the moment the store learns the prompt, which is the
right way for the conservative-on-absence tradeoff to fail.

**Tap targets: no regression, and nothing new to file.** `.notification-banner-btn` is untouched
at `padding: 2px 10px` / `--text-xs`; the disabled state inherits the same box. That is the
pre-existing sub-44pt problem already tracked as #7473, not something this fold introduces.

**Disabled semantics: correct primitive, and consistent by design.** Native `disabled` on a
`<button>` is the right choice. Worth naming the consequence for the record: a disabled button
leaves the tab order, and the reason sits in a *sibling* `<span role="status">` rather than being
linked by `aria-describedby` — so a keyboard/SR user tabbing the banner stack encounters neither
the button nor, unless the live region fires while they are there, the explanation. That is
*exactly* what the inline prompt does with `perm-disconnected-hint`, which is the "one vocabulary
for one condition" choice paying its own cost. Fixing it is a two-surface change and belongs in
its own issue, not here.

**On the one-implementation refusal — the argument checks out.** `expiresAtRef = useRef(performance.now() + remainingMs)`
is anchored at *that* prompt's mount, and the banner answers from wall-clock `Date.now()` against
`m.expiresAt`. Two different clocks, and the banner has no mount to anchor the monotonic one to.
A shared helper really would force one surface to abandon the thing #3619 exists for. The
cross-references landed **both ways**, each naming the specific functions
(`NotificationBanners` -> `respond()` + `showButtons`; `PermissionPrompt` -> `permissionNotificationStatus`).

**No shared fixture or contract test couples them. I judge that acceptable, with a named
residual — not a finding.** The reasoning: the countdown genuinely cannot be shared, the three
conditions are individually pinned on *each* surface, and the exposure is drift over time rather
than a live defect. But it is worth being precise about what is uncovered, since "three gates"
enumerated in prose in two files is the shape `docs/false-safety-guards.md` warns about: nothing
fails if a fourth gate is added to one surface and not the other. If you ever want to close that,
the cheap version is a table-driven test over the two *shareable* conditions — answered, and
`connected` — asserting both surfaces decline for each, driven by one list. The countdown row
would stay per-surface. Entirely optional; I would not hold the PR for it.

### 2. Fold 2 — `WINDOW_FETCH`

- `useMemo` is gone from both panels (it survives only in the comments explaining why), and the
  module constant is present in both. **My `[fetchImpl]` -> `[]` mutant is genuinely unwritable**
  — there is no dep list to mutate. Confirmed, not assumed.
- **K** (`resolvedFetch = WINDOW_FETCH`, prop ignored) -> **killed**, and it is precisely aimed:
  it kills **exactly one** test, the new `a CHANGING fetchImpl prop re-arms exactly once`
  (`expected [] to deeply equal [ +0 ]`), leaving the other six green. That is the input the old
  dep list was supposed to defend, now pinned by the one test that can see it.
- **A3** (original loop mutation against the constant-based files) -> **red across all six
  cases**, as claimed: `expected 21 to be 1`, `24 to be 2`, `28 to be 1`, `33 to be 2`,
  `19 to be 1`, `18 to be 1`. Re-deriving this against the changed fix rather than trusting the
  earlier kill was the right call and it holds.

### 3. Fold 3 — mark-read

Scope is permission-only, structurally: exactly one `onMarkRead(n.id)` call site, inside the
`not-pending` branch, itself nested inside `isPermission`; `onDismiss(n.id)` survives untouched
in the `!isPermission` branch, pinned by `a NON-permission banner still deletes on Dismiss`.

Retention verified end to end rather than by the commit message: `markSessionNotificationRead`
stamps `readAt` and leaves the row in `sessionNotifications`; `NotificationsWidget` maps over
**all** notifications and only *styles* by `readAt === undefined`; the banner stack filters on
`readAt === undefined` and so drops it. The record survives in the widget. Correct.

### 4. The mutant L disclosure

Both halves check out, and the count is understated in the safe direction.

- **L** (drop `disabled` from both buttons) -> I measure **2** kills, not 1:
  `disables Allow and Deny rather than leaving them enabled` (component) **and**
  `disables the banner buttons while the socket is down` (App wiring).
- The disclosed vacuity is real: `cannot fire onApprove/onDeny even if a click is dispatched at
  it` is among the 146 that **survive** L, exactly as described.
- **O** (drop the click-time re-check) -> **killed**, 2 tests
  (`refuses the CLICK when the request died since the last render`,
  `refuses the CLICK when the socket dropped since the last render`).

Keeping both guards is the right call: they cover different windows — `disabled` covers the
rendered state, the re-check covers the render-to-click gap — and each has a mutant only it can
kill, so neither is the untestable-refinement case.

**One thing I would add.** The vacuity is disclosed in this PR thread, which is ephemeral; the
test at `NotificationBannersStale.test.tsx:373` carries no note that it passes for the *other*
guard's reason. Someone mutating `disabled` in six months will see it green and conclude the
disabled attribute is unnecessary — which is precisely the inference `docs/false-safety-guards.md`
exists to prevent. A one-line comment at that test ("survives a `disabled` mutant; the click-time
re-check covers this path — the disabled attribute is pinned by `disables Allow and Deny` above")
puts the finding where the next reader will be standing. Not a blocker.

### 5. Suites

| Command | Exit |
|---|---|
| `npx tsc --noEmit` | `0` |
| `npx vitest run` — the 8 touched files | `0` — 374 tests |
| `npx vitest run` (whole dashboard) | `0` — 309 files, **5380** tests |

One thing worth recording rather than quietly re-running: my **first** whole-dashboard run came
back **1 failed / 5379 passed**, and the failure was
`InputBar.test.tsx > with >200 files, arrow-down lands on the first MCP resource right after the
capped rows (#6844 review)` — **`Test timed out in 30000ms`** at 30808ms. Unrelated file, and a
timeout on a `>200 files` scale case. I checked rather than assumed: the file passes alone at
**176/176, exit 0**, and the whole suite re-runs clean at 309/5380, exit 0. Load artifact under a
loaded host, not a defect — but it is a 30s budget on a scale test running on a box that is also
a CI runner, so it will do this again.

### CI — not settled

Run `33226067604` is still open at the time of writing: `Dashboard Tests` queued;
`Server Lint`, `Server Tests`, `Store Core Tests` in progress. `Scripts Tests` surfaces as `fail`
in `gh pr checks`, but the job's own conclusion is **`cancelled`** with **zero failed steps** —
the #7471 signature again, not a code failure. Not re-run, per the note that it is being handled
separately. **Required contexts are not settled, so the merge gate is not satisfied yet** — that
is a gate, not a code objection.

### Verdict

Approving the code. Every claim in the fold-in reply that I could test, I tested, and all of them
held; the two additions beyond my prescriptions — the three-state split and the App-level wiring
mutant Q — are both better than what I asked for, and the ordering they had to invent to make the
split work is the one part I tried hardest to break and could not. The only outstanding items are
the one-line comment at the vacuous test (optional) and CI settling green (outside the diff).

Issue #7478, opened 2026-08-29
dashboard: pendingServerSeed is session-keyed and survives both session destroy and a server switch


blamechris:
## Problem

Found while auditing the full session-keyed map roster for #7470.

`pendingServerSeed: Record<sessionId, string>` (`packages/dashboard/src/store/types.ts:1059`)
holds a server-provided composer SEED, stashed by the `session_switched`
create-confirm handler (`message-handler.ts:2221`) and drained by
`takePendingServerSeed` (`connection.ts:1637`) from App's create-confirm effect
(`App.tsx:1214`).

It is the only session-keyed map in the connection store that is not cleaned up
on any of the three session-went-away paths. #7470 covered the five PR/CI maps
at all three; this one was deliberately left out of that PR's scope because it
belongs to a different feature (#5553 session presets), not because it is
correct.

Two ways an entry survives its session:

1. **Never drained.** The drain happens only if App's create-confirm effect
   runs for that session. A session destroyed before the composer is touched
   leaves its seed behind, and the `session_list` `removedIds` block does not
   remove it.
2. **Across a server switch.** `forgetSession` and `_resetSessionMemory` both
   clear `sessions`, `sessionStates`, `activity` and (since #7470) the PR/CI
   maps — but not this. Session ids are minted per daemon, so a stale entry can
   be drained against a *different* server's session of the same id, pre-filling
   one machine's composer with another machine's seed text.

(2) is the one worth fixing promptly: it is a wrong value presented to the user,
not just retained memory.

## Suggested fix

Same three sites #7470 used, and the markers are already there to co-locate with:

- `message-handler.ts` `#7470 prune-block-start/end` — `pruneSessionKeyedMap(get().pendingServerSeed, removedIds)`
- `connection.ts` `#7470 forget-reset-start/end` — `pendingServerSeed: {}`
- `connection.ts` `#7470 switch-reset-start/end` — `pendingServerSeed: {}`

## Acceptance

- Destroying a session removes its `pendingServerSeed` entry; another session's
  entry is untouched (positive control).
- `forgetSession` and `_resetSessionMemory` both clear the map.
- A per-site assertion each, so clearing two of three sites fails naming the
  third — the shape `session-destroy-prunes-pr-maps.test.ts` already uses.

Related to #7470.

blamechris:
[...]

Issue #7450, opened 2026-08-28
session_pr_status throttle: unpruned stamp map + invisible to /diagnostics — reuse RateLimiter or prune on session_destroyed


blamechris:
## Problem

#7445 (#7436) adds a per-session survey throttle to
`packages/server/src/handlers/session-pr-status-handlers.js` as a module-level
`WeakMap<sessionManager, Map<sessionId, lastSurveyStartedAtMs>>`. It works, but
it sits beside an existing, better-instrumented abstraction and gives up two
properties that one already has.

**1. It is not bounded the way its comment claims.** The doc says:

> A destroyed session's stamp lingers until the manager itself is collected —
> one number per ever-surveyed session id, bounded and tiny.

The session manager is the daemon-lifetime singleton (`ws-server.js:1127`,
assigned once), so "until the manager is collected" means *never* in a running
daemon. The map is therefore bounded by **every session id ever surveyed over the
daemon's lifetime**, not by the live session count. Small per entry, but
monotonic — and `SessionCiWatcher._state`, two files over, prunes exactly this
case with a comment stating the requirement:

> Bounded by the live session count — entries are pruned in `tick()` when a
> session disappears, so a long-running daemon does not accumulate the ids of
> every session it has ever seen.

Two modules in the same feature, opposite choices, one of them documented as a
requirement.

**2. It is invisible to operators.** `packages/server/src/rate-limiter.js`'s
`RateLimiter` already provides keyed sliding-window limiting with a bounded entry
map (`maxEntries` + FIFO eviction, #3979), a `name` for attribution (#3996), and
eviction stats surfaced through `/diagnostics` via `collectRateLimiters(server)`
(`diagnostics.js:74`). The new throttle appears in none of that: an operator
seeing refused refreshes has no readout telling them a limiter is shedding.

## Proposal

Either:

- **(a) Reuse `RateLimiter`** — construct one with `name: 'session-pr-survey'`,
  `windowMs: SURVEY_MIN_INTERVAL_MS`, `maxMessages: 1`, `burst: 0`, keyed on the
  session id, owned by `WsServer` and registered with `collectRateLimiters` so it
  shows up in `/diagnostics`. Check `RateLimiter.check()`'s `{ allowed,
  retryAfterMs }` fits the handler's "reply with a degraded snapshot" shape —
  `retryAfterMs` would let the refusal reason carry a real number rather than
  "a few seconds"; or

- **(b) Keep the local map but prune it** — drop a session's stamp on
[...]

Issue #7344, opened 2026-08-23
sessions: surface CI completion for the session's branch/PR — notify the user and wake the agent instead of leaving it to poll


blamechris:
## Summary

When an agent opens a PR and says *'I'll report back when CI lands'*, Chroxy has no idea CI exists. The agent is left to poll `gh pr checks` on its own schedule, the user gets no signal, and the session looks idle (#7340). Observed: CI went fully green at 22:14 UTC and 15+ minutes later the agent had not noticed and the user had to ask whether anything was still happening.

The user should be told when a run finishes, and the **agent** should be told too, rather than burning turns polling.

## The auth question is already answered — no new flow needed

Chroxy **already** reads GitHub state via the `gh` CLI, using the user's existing local `gh auth`, in `control-room/integrations.js` (#5501/#5502):

- `gh run list -R <owner>/<repo>` — workflow runs, owner/repo derived from the git remote
- `gh api .../releases/latest` — one call per refresh
- `gh run rerun <id> -R <owner>/<repo>` — a mutation, already wired
- `GH_MISSING_NOTE` / `NO_GITHUB_REMOTE_REASON` — the not-configured paths are already modelled, with timeouts so *'anything through this is treated as a failure with a reason, never a hang'*

So this needs **no OAuth app, no PAT prompt, no token storage**. Reuse the same mechanism, and inherit its existing degradation behaviour when `gh` is absent or the remote isn't GitHub.

## What's missing

Today's integration is Control-Room-scoped and repo-relay-workflow-specific. The gap is three joins:

1. **Session → PR.** A session knows its branch (worktree sessions especially). Resolve the open PR for that branch (`gh pr view --json number,headRefOid`) and remember it.
2. **Poll → completion event.** Watch check state for the PR's **current head SHA** and emit an event on settle. Report the *terminal* state — green, failed, or blocked-with-green-CI — not merely 'finished'. (`mergeStateStatus` is a function of the head SHA; a reading taken before a push is stale.)
3. **Completion → both audiences.**
   - **User:** route through the existing notification pipeline (`push.js`, the Discord sink, the notifications module) so it reaches the phone. 'CI green on #7339' is exactly the notification class that already exists.
   - **Agent:** inject it into the session as an event so the model learns CI settled without spending a turn asking. This is the higher-value half — it converts an open polling loop into a wake-up.

## Why the agent half matters for cost

[...]

PR #7481, opened 2026-08-29
fix(dashboard): prune the per-session PR/CI maps when a session goes away

### LINKED ISSUE #7470: dashboard: per-session PR/CI maps (sessionPrStatus, sessionPrThreads) are never pruned on session destroy
## Problem

The dashboard store keeps four per-session maps for the PR/CI surface, keyed by
session id, and **none of them is ever pruned when a session is destroyed**:

- `sessionPrStatus` / `sessionPrStatusLoading` / `sessionPrStatusRequestedAt` (#7344)
- `sessionPrThreads` / `sessionPrThreadsLoading` (#7430)

`packages/dashboard/src/store/connection.ts` clears the three *loading*-shaped
maps on a socket drop (`:2911-2929`), which is the right fix for the
stranded-control problem it was written for — but that is a reconnect reset, not
a lifecycle prune. Nothing removes an entry when the session it describes goes
away. A long-lived dashboard tab that creates and destroys sessions therefore
accumulates one snapshot per ever-surveyed session id for the life of the tab,
and each `sessionPrStatus` entry is a full PR/CI snapshot rather than a
timestamp.

This is not new in #7430 — that PR mirrored the existing #7344 shape
deliberately and consistently, which is correct. Filing it separately because
the fix belongs to the surface as a whole, not to either PR.

## Relationship to #7450

#7450 is the **server**-side statement of the same defect: the survey throttle's
stamp map is unpruned and invisible to `/diagnostics`. #7469 has since extracted
that map into `packages/server/src/handlers/survey-throttle.js` and opened a
**second** instance of it for the threads handler, so #7450's fix now has one
place to land and two owners to cover. This issue is the client-side half:
different process, different lifetime, different fix.

## Suggested fix

Prune all five maps wherever the store already handles a session going away
(the `session_destroyed` / session-removal path), the same way other
session-keyed store slices are cleaned up. A bounded LRU would also work, but
lifecycle-driven deletion is simpler and matches what the rest of the store
does.

## Acceptance

- Destroying a session removes its entry from all five maps.
- A test that creates two sessions, populates both maps, destroys one, and
  asserts only the survivor's key remains — with a positive control that the
  surviving entry is untouched, so the assertion cannot pass by clearing
  everything.

Found during review of #7469.

blamechris:
Closes #7470

## The map roster (enumerated first)

Every `Record<sessionId, …>`-shaped map in the dashboard connection store,
audited in full rather than fixing only the two the issue named:

| Map | Issue | Keyed by | Before | Now |
|---|---|---|---|---|
| `sessionPrStatus` | #7344 | session id | never removed | pruned at all 3 sites |
| `sessionPrStatusLoading` | #7344 | session id | reset on socket drop only | pruned at all 3 sites |
| `sessionPrStatusRequestedAt` | #7344 | session id | reset on socket drop only | pruned at all 3 sites |
| `sessionPrThreads` | #7430 | session id | never removed | pruned at all 3 sites |
| `sessionPrThreadsLoading` | #7430 | session id | reset on socket drop only | pruned at all 3 sites |

Audited and deliberately **out**:

| Map | Why |
|---|---|
| `sessionStates` | already pruned on `session_list` removal + both resets |
| `activity` (Control Room tree) | already pruned, via `clearSessionActivity` (#5163) |
| `sessionNotifications` | an array, and deliberately retained on close — the #7466 comment in `connection.ts` depends on that retention |
| `sessionPresetSnapshots` | matches the `sessionPr` prefix by accident; keyed by **cwd**, not session id (#5553) |
| `pendingServerSeed` | genuinely session-keyed and genuinely unpruned, but a different feature family (#5553) — filed as **#7478** rather than folded in silently |

## Three sites, not one

The issue asked for "the `session_destroyed` / session-removal path". There is
no `session_destroyed` in the dashboard — `destroySession` only puts a frame on
the wire, and the server's next `session_list` is what tells this tab a session
is gone. Auditing for that turned up three spellings of "the session went away",
and covering only the first would have been the same adjacent-field mistake one
level up:

1. **`session_list` removedIds** (`message-handler.ts`) — one session closed.
   Pruned beside `sessionStates` and `activity`, which is where the store
   already does this.
2. **`forgetSession`** (`connection.ts`) — disconnect + forget.
3. **`_resetSessionMemory`** (`connection.ts`) — `switchServer`.

(2) and (3) are worse than unbounded growth. These maps are keyed by session id,
session ids are minted per daemon, and `switchServer` keeps the old server's
persisted data on purpose — so a surviving entry can be read against a
**different** server's session of the same id, rendering one machine's CI state
on another machine's chip. That is a wrong answer, not just retained memory.

`sessionStates` is pruned rather than retained on all three paths, so no
retain/prune split was needed — the PR/CI maps go exactly where it goes.

### Red first

`handleMessage` takes an **object**, not a JSON string, and my first fixture
passed `JSON.stringify(...)` — the handler returned at its `typeof raw !== 'object'`
guard and every test "failed" without reaching the code under test. The second
red run crashed inside `clearSessionActivity` because the fixture omitted
`activity`. Both were discarded: **a crash and an early return are not a red
assertion.** The recorded red is the third run.

`session_list` prune site — fix reverted, fixture correct:

```
Tests  11 failed | 2 passed (13)
AssertionError: expected { type: 'session_pr_status', …(10) } to be undefined
AssertionError: expected true to be undefined
AssertionError: expected 1000 to be undefined
AssertionError: expected { type: 'session_pr_threads', …(8) } to be undefined
AssertionError: expected true to be undefined
```

The 2 passing pre-fix are both controls by design: the roster-extraction control
and the no-removal no-op.

Reset sites, before the `connection.ts` change:

```
Tests  10 failed | 4 passed (14)
AssertionError: expected { 'sess-A': 1000, 'sess-B': 2000 } to deeply equal {}
```

After: **38 passed (38)** across both new files.

### Mutation matrix

Twelve single-map mutants — drop exactly one map's cleanup at one site — each
killed by an assertion that **names that map**, so no blanket assertion is
satisfiable by partial pruning:

| Mutant | Killed by |
|---|---|
| drop `sessionPrStatus` @ session_list | `prunes sessionPrStatus …` + `[session_list removedIds] cleans up sessionPrStatus` |
| drop `sessionPrThreads` @ session_list | `prunes sessionPrThreads …` + roster |
| drop each of 5 @ `forgetSession` | `clears <map>` + `[forgetSession] cleans up <map>` |
| drop each of 5 @ `_resetSessionMemory` | `clears <map>` + `[_resetSessionMemory] cleans up <map>` |

Three further mutants:

- **Blanket clear** — replace all five prunes with `= {}`: **5 failed**, one per
  map, caught by the positive controls (the survivor's snapshot must come through
  by reference, its `prNumber` must still be 2, its throttle stamp still 2000). A
  fix that wipes everything cannot pass.
- **Sixth family map** — add `sessionPrMutantSixth: Record<string, number>` to
  `types.ts` with no cleanup: **3 failed**, one per site, each naming the field.
- **Reference churn** — make `pruneSessionKeyedMap` always return a new object:
  killed by `keeps a map's reference when the removed session was never in THAT map`.

Two guards I had to fix because a mutant **survived** them, both recorded here
because a mutation matrix that only reports kills is not evidence:

1. The first reference-stability test drove a `session_list` with **no**
   removals, so the handler's `removedIds.length > 0` guard skipped the block
   entirely and the helper never ran — a false precondition, green for the wrong
   reason. Replaced with a removal the map does not track.
2. Generalising the roster guard to three sites weakened its assertion from
   `toContain('patch.<field> =')` to `toContain(field)` — and
   `sessionPrStatusLoading: {}` **contains the substring** `sessionPrStatus`, so
   2 of 10 mutants survived it. Now a token match
   (`(^|[^A-Za-z0-9_$])<field>\s*[:=][^=]`); the matrix above is the re-run after
   tightening, with all 12 killed.

### Adjacent-field guard

The roster test derives the family from the **producer** — the `ConnectionState`
declaration in `types.ts` — not from a list written in the test, and asserts each
member is assigned inside each site's marked block. `sessionPresetSnapshots` is
excluded **by name with its reason** rather than by narrowing the regex until
only today's five match. The extraction has its own control (a containment check,
so a legitimate sixth map fails the prune assertion rather than the control).

Blocks are sliced by marker before matching — a file-wide grep for
`sessionPrStatus` in `message-handler.ts` is satisfied by dozens of unrelated
lines.

## Suites

| Suite | Command | Result |
|---|---|---|
| Dashboard typecheck | `npx tsc --noEmit` (bare exit code) | **exit 0** |
| Dashboard store | `npx vitest run src/store/` | **85 files, 1230 tests passed** |
| Full dashboard | `npx vitest run` | **313 files, 5460 tests passed** |
| `session-destroy-prunes-pr-maps.test.ts` | individually | **24 passed** |
| `session-reset-clears-pr-maps.test.ts` | individually | **14 passed** |
| `dispatch-control-room-activity.test.ts` | individually | **14 passed** |

The worktree has no `packages/dashboard/node_modules`, so `tsc` initially
reported 251 errors — all `Cannot find module '@testing-library/react'` /
`react-dom` / `react-resizable-panels` and their implicit-any cascade. Confirmed
environmental by diffing against the **pristine** tree: byte-identical error
lists, so the change adds zero. Re-run with those four packages resolvable, the
typecheck exits **0**.

## Note on the touched existing test

`dispatch-control-room-activity.test.ts`'s partial mock state omitted the five
maps, so the new prune hit `undefined`. The maps are non-optional on
`ConnectionState` and always `{}` in the real store, so this is a fixture gap,
not a production shape — filled in rather than making `pruneSessionKeyedMap`
undefined-tolerant, which would have written `undefined` into the store to hide
a test's omission.

## Not in this PR

- #7478 — `pendingServerSeed`, the one session-keyed map left uncleaned. Same
  defect, different feature; the three `#7470` markers this PR adds are where its
  fix lands.
- #7450 remains the server-side half of the same shape and is untouched here.

### DIFF
--- a/packages/dashboard/src/store/connection.ts
+++ b/packages/dashboard/src/store/connection.ts
@@ -3247,6 +3247,19 @@ export const useConnectionStore = create<ConnectionState>((set, get) => ({
       // #5163: drop the Control Room tree on disconnect/forget — a fresh
       // connection re-seeds it from activity_snapshot on subscribe.
       activity: createEmptyActivityState(),
+      // #7470 forget-reset-start — the per-session PR/CI maps go with the
+      // session state they describe. They are keyed by SESSION ID, and session
+      // ids are minted per daemon, so an entry that outlived its server can be
+      // read against a DIFFERENT server's session of the same id — one
+      // machine's CI state rendered on another machine's chip. That is a wrong
+      // answer, not just retained memory, which is why these are cleared here
+      // and not only pruned on the session_list removal path (message-handler.ts).
+      sessionPrStatus: {},
+      sessionPrStatusLoading: {},
+      sessionPrStatusRequestedAt: {},
+      sessionPrThreads: {},
+      sessionPrThreadsLoading: {},
+      // #7470 forget-reset-end
       cancellingActivityIds: new Set<string>(),
       // #5500: drop reindex pending/result state with the rest of the
       // connection-scoped Control Room state.
@@ -3302,6 +3315,17 @@ export const useConnectionStore = create<ConnectionState>((set, get) => ({
       // #5163: drop the Control Room tree on disconnect/forget — a fresh
       // connection re-seeds it from activity_snapshot on subscribe.
       activity: createEmptyActivityState(),
+      // #7470 switch-reset-start — same reasoning as forgetSession above, and
+      // this is the path where the cross-server read is REACHABLE rather than
+      // theoretical: switchServer keeps the old server's persisted data on
+      // purpose, so without this the new server's session ids can collide with
+      // the old server's PR/CI entries.
+      sessionPrStatus: {},
+      sessionPrStatusLoading: {},
+      sessionPrStatusRequestedAt: {},
+      sessionPrThreads: {},
+      sessionPrThreadsLoading: {},
+      // #7470 switch-reset-end
       cancellingActivityIds: new Set<string>(),
       // #5500: drop reindex pending/result state with the rest of the
       // connection-scoped Control Room state.
--- a/packages/dashboard/src/store/connection.ts
+++ b/packages/dashboard/src/store/connection.ts
@@ -3248,12 +3248,22 @@ export const useConnectionStore = create<ConnectionState>((set, get) => ({
       // connection re-seeds it from activity_snapshot on subscribe.
       activity: createEmptyActivityState(),
       // #7470 forget-reset-start — the per-session PR/CI maps go with the
-      // session state they describe. They are keyed by SESSION ID, and session
-      // ids are minted per daemon, so an entry that outlived its server can be
-      // read against a DIFFERENT server's session of the same id — one
-      // machine's CI state rendered on another machine's chip. That is a wrong
-      // answer, not just retained memory, which is why these are cleared here
-      // and not only pruned on the session_list removal path (message-handler.ts).
+      // session roster this action empties.
+      //
+      // Clearing them HERE is not belt-and-braces over the `session_list`
+      // removedIds prune (message-handler.ts): it is the only thing that can
+      // reach them. `removedIds` is a diff against `Object.keys(sessionStates)`,
+      // and this block empties `sessionStates` in the same patch — so from the
+      // next snapshot onwards there is nothing to diff, `removedIds` is `[]`,
+      // and any entry left behind is permanently unprunable for the life of the
+      // tab. Same mechanism at `_resetSessionMemory` below and at `auth_ok`'s
+      // non-reconnect branch (message-handler.ts).
+      //
+      // Session ids carry no daemon namespace — they are 16 random bytes
+      // (server/session-manager.js), so entropy is the ONLY thing separating
+      // two servers' id spaces. That is ample (128 bits) and a collision is not
+      // the hazard here; it is simply not a property worth leaning on when the
+      // correct lifetime for these entries is "this connection".
       sessionPrStatus: {},
       sessionPrStatusLoading: {},
       sessionPrStatusRequestedAt: {},
@@ -3315,11 +3325,12 @@ export const useConnectionStore = create<ConnectionState>((set, get) => ({
       // #5163: drop the Control Room tree on disconnect/forget — a fresh
       // connection re-seeds it from activity_snapshot on subscribe.
       activity: createEmptyActivityState(),
-      // #7470 switch-reset-start — same reasoning as forgetSession above, and
-      // this is the path where the cross-server read is REACHABLE rather than
-      // theoretical: switchServer keeps the old server's persisted data on
-      // purpose, so without this the new server's session ids can collide with
-      // the old server's PR/CI entries.
+      // #7470 switch-reset-start — same mechanism as forgetSession above: this
+      // block empties `sessionStates`, which is the set `removedIds` diffs
+      // against, so anything left behind here can never be pruned by a later
+      // `session_list`. switchServer additionally KEEPS the old server's
+      // persisted data on purpose, so leaving these would carry one server's
+      // PR/CI readings into a session list drawn from another.
       sessionPrStatus: {},
       sessionPrStatusLoading: {},
       sessionPrStatusRequestedAt: {},
--- a/packages/dashboard/src/store/dispatch-control-room-activity.test.ts
+++ b/packages/dashboard/src/store/dispatch-control-room-activity.test.ts
@@ -79,6 +79,15 @@ function baseState(): Partial<ConnectionState> {
     sessionStates: { [SESSION_ID]: createEmptySessionState() },
     activity: createEmptyActivityState(),
     messages: [],
+    // #7470 — the `session_list` removedIds block now prunes the per-session
+    // PR/CI maps alongside the activity tree. They are non-optional on
+    // `ConnectionState` and always `{}` in the real store, so a partial mock
+    // that omits them is a fixture gap, not a production shape.
+    sessionPrStatus: {},
+    sessionPrStatusLoading: {},
+    sessionPrStatusRequestedAt: {},
+    sessionPrThreads: {},
+    sessionPrThreadsLoading: {},
   }
 }
 
--- a/packages/dashboard/src/store/message-handler.ts
+++ b/packages/dashboard/src/store/message-handler.ts
@@ -4447,6 +4447,29 @@ export function handleMessage(raw: unknown, ctxOverride?: ConnectionContext): vo
           sessions: [],
           activeSessionId: null,
           sessionStates: {},
+          // #7470 authok-reset-start — the per-session PR/CI maps go with the
+          // roster this branch empties, and this is a SITE rather than a case
+          // the removedIds prune below covers. `removedIds` is a diff against
+          // `Object.keys(sessionStates)` (store-core `buildSessionListPatches`),
+          // so once the roster is empty there is nothing to diff: the next
+          // `session_list` yields `removedIds = []`, the `removedIds.length > 0`
+          // guard skips the prune block, and anything left here is permanently
+          // unprunable for the life of the tab.
+          //
+          // Reached by Disconnect → Connect to the same server: `disconnect()`
+          // clears `lastConnectedUrl` while preserving the session roster, so
+          // `isReconnect = (lastConnectedUrl === url)` is false and this branch
+          // runs. `scheduleRetry` re-entering `connect()` with `_retryCount > 0`
+          // is a second route to it. Only the NON-reconnect branch clears — a
+          // silent reconnect keeps `sessionStates`, so the snapshots it
+          // describes are still true and blanking them would wipe a chip the
+          // user is reading on every transient drop.
+          sessionPrStatus: {},
+          sessionPrStatusLoading: {},
+          sessionPrStatusRequestedAt: {},
+          sessionPrThreads: {},
+          sessionPrThreadsLoading: {},
+          // #7470 authok-reset-end
           customAgents: [],
         });
       }
@@ -4726,9 +4749,19 @@ export function handleMessage(raw: unknown, ctxOverride?: ConnectionContext): vo
         // session it had ever surveyed, for the life of the connection.
         //
         // Pruned HERE, beside `sessionStates` and `activity` above, because
-        // this removedIds block is the store's one session-went-away path —
-        // `destroySession` only puts a frame on the wire; the server's next
-        // `session_list` is what tells this tab the session is gone.
+        // this is where ONE session going away is observed — `destroySession`
+        // only puts a frame on the wire; the server's next `session_list` is
+        // what tells this tab the session is gone.
+        //
+        // It is NOT the only such path, and reading it as one is how PR #7481's
+        // review found a fourth. `removedIds` is a diff against
+        // `Object.keys(sessionStates)`, so this block only ever sees sessions
+        // that are still in the roster when the snapshot lands. The three sites
+        // that empty the roster WHOLESALE — `auth_ok`'s non-reconnect branch
+        // above, and `forgetSession` / `_resetSessionMemory` in connection.ts —
+        // leave nothing to diff, so each must clear these maps itself. All four
+        // are held by the SITES table in
+        // `session-destroy-prunes-pr-maps.test.ts`.
         //
         // Each map is assigned EXPLICITLY rather than looped over a list of
         // field names: a name list is the same hardcoded roster with the
--- a/packages/dashboard/src/store/message-handler.ts
+++ b/packages/dashboard/src/store/message-handler.ts
@@ -196,7 +196,7 @@ import type {
   MCPResourceItem,
   McpServerOpResult,
 } from './types';
-import { createEmptySessionState } from './utils';
+import { createEmptySessionState, pruneSessionKeyedMap } from './utils';
 import { clearPersistedSession } from './persistence';
 
 // ---------------------------------------------------------------------------
@@ -4716,6 +4716,36 @@ export function handleMessage(raw: unknown, ctxOverride?: ConnectionContext): vo
           nextActivity = clearSessionActivity(nextActivity, id);
         }
         if (nextActivity !== get().activity) patch.activity = nextActivity;
+        // #7470 prune-block-start — the per-session PR/CI maps.
+        //
+        // Five maps, all keyed by session id, and until now nothing removed an
+        // entry when the session it describes went away: connection.ts's #6153
+        // sweep resets the LOADING-shaped ones on a socket drop, but that is a
+        // reconnect reset for a stranded control, not a lifecycle prune. A
+        // long-lived tab therefore accumulated one full PR/CI snapshot per
+        // session it had ever surveyed, for the life of the connection.
+        //
+        // Pruned HERE, beside `sessionStates` and `activity` above, because
+        // this removedIds block is the store's one session-went-away path —
+        // `destroySession` only puts a frame on the wire; the server's next
+        // `session_list` is what tells this tab the session is gone.
+        //
+        // Each map is assigned EXPLICITLY rather than looped over a list of
+        // field names: a name list is the same hardcoded roster with the
+        // compiler switched off, and a dropped entry in one reads as a missing
+        // string instead of a missing statement. The roster is held honest by
+        // `session-destroy-prunes-pr-maps.test.ts`, which extracts the family
+        // from types.ts and asserts each member appears between these markers.
+        //
+        // `pruneSessionKeyedMap` returns the SAME reference when the map held
+        // none of the removed ids, so this cannot churn `useShallow` consumers
+        // on a snapshot that only removed a session they don't track.
+        patch.sessionPrStatus = pruneSessionKeyedMap(get().sessionPrStatus, removedIds);
+        patch.sessionPrStatusLoading = pruneSessionKeyedMap(get().sessionPrStatusLoading, removedIds);
+        patch.sessionPrStatusRequestedAt = pruneSessionKeyedMap(get().sessionPrStatusRequestedAt, removedIds);
+        patch.sessionPrThreads = pruneSessionKeyedMap(get().sessionPrThreads, removedIds);
+        patch.sessionPrThreadsLoading = pruneSessionKeyedMap(get().sessionPrThreadsLoading, removedIds);
+        // #7470 prune-block-end
         // If the active session was removed, switch to next available
         if (initialActiveId && removedIds.includes(initialActiveId)) {
           const remaining = Object.keys(newStates);
--- a/packages/dashboard/src/store/session-destroy-prunes-pr-maps.test.ts
+++ b/packages/dashboard/src/store/session-destroy-prunes-pr-maps.test.ts
@@ -16,24 +16,39 @@
  * went away, so a long-lived tab accumulated one snapshot per ever-surveyed
  * session id for the life of the connection.
  *
- * The store's one session-went-away path is the `session_list` handler's
- * `removedIds` block, where `sessionStates` and the Control Room `activity`
- * tree are already pruned. The fix lives there.
+ * ## Four sites, one mechanism
+ *
+ * "The session went away" has FOUR spellings here. The obvious one is the
+ * `session_list` handler's `removedIds` block, where `sessionStates` and the
+ * Control Room `activity` tree are already pruned. The other three empty the
+ * roster WHOLESALE — `auth_ok`'s non-reconnect branch, `forgetSession`,
+ * `_resetSessionMemory` — and because `removedIds` is a diff against
+ * `Object.keys(sessionStates)`, emptying the roster leaves nothing to diff:
+ * anything not cleared in the same patch is PERMANENTLY unprunable for the life
+ * of the tab. So each of the four clears these maps itself.
+ *
+ * The fourth (`auth_ok`) was missed by the first version of this file and found
+ * in review of PR #7481. It is the most ordinary gesture of the four —
+ * Disconnect, then Connect to the same server — which is worth remembering
+ * before the next "surely the removedIds prune covers this" simplification.
  *
  * ## What these tests are shaped to catch
  *
- * A per-map assertion for EACH of the five, never one blanket "the maps no
- * longer mention sess-dead" check: a blanket assertion is satisfied by pruning
- * four of five, which is exactly the adjacent-field failure this issue is an
- * instance of. Mutating away any single prune call must turn exactly one
- * assertion red, and the message must name the map.
+ * A per-map assertion for EACH of the five at EACH site, never one blanket "the
+ * maps no longer mention sess-dead" check: a blanket assertion is satisfied by
+ * pruning four of five, which is exactly the adjacent-field failure this issue
+ * is an instance of. Mutating away any single clear must turn exactly one
+ * assertion red, and the message must name the map AND the site.
  *
  * Plus a POSITIVE CONTROL in every case: the surviving session's entries must
- * come through byte-identical. Without it, `set({ sessionPrStatus: {} })` —
- * clearing everything — would pass every "dead key is gone" assertion.
+ * come through byte-identical, and a silent RECONNECT must keep all five. Both
+ * exist so that `set({ sessionPrStatus: {} })` — clearing everything, always —
+ * cannot pass.
  *
- * Plus a roster guard derived from `types.ts` rather than from a list written
- * here, so a SIXTH map added to the family without a prune goes red.
+ * Plus a roster guard that classifies every session-keyed-shaped collection on
+ * `ConnectionState`, so a new one is red until someone says how it is keyed.
+ * See its own docstring below for why the first version of that guard did not
+ * work.
  */
 import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest'
 import { readFileSync } from 'fs'
@@ -49,7 +64,7 @@ vi.mock('./crypto', () => ({
 vi.mock('./persistence', () => ({ clearPersistedSession: vi.fn() }))
 
 import { handleMessage, setStore, clearDeltaBuffers, clearPermissionSplits, stopHeartbeat, resetReplayFlags } from './message-handler'
-import { createEmptySessionState } from './utils'
+import { createEmptySessionState, pruneSessionKeyedMap } from './utils'
 import type { ConnectionState } from './types'
 import { createEmptyActivityState } from '@chroxy/store-core'
 import type { ServerSessionPrStatusMessage, ServerSessionPrThreadsMessage } from '@chroxy/protocol'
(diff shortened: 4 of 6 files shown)


blamechris:
## Code Review: PR #7481

### Summary

The three sites this PR fixes are the right three, the fix at each of them is correct, and I
re-ran the evidence rather than taking the body's word for it: every claimed suite reproduces
exactly (`tsc --noEmit` exit **0**, `vitest run src/store/` **85 files / 1230 tests** exit 0, the
two new files **38/38** exit 0), and 6 of my 7 mutants died naming both the map and the site. The
two self-caught guard weaknesses are genuinely fixed — I mutated `pruneSessionKeyedMap` to always
return a new object and the replacement reference test killed it, which is the property the first
weakness lost. The exclusion table holds up: `sessionPresetSnapshots` really is keyed by cwd, and
the `sessionNotifications` retention really does not create a joint read that pruning can break.

**Two things block, and they are the same shape as each other: a claim of completeness that the
code does not have.**

The first is that **there is a fourth "the session went away" path**, and it is reachable by the
most ordinary gesture in the dashboard — **Disconnect, then Connect to the same server.** On that
path `auth_ok` empties `sessionStates` *before* the first `session_list` arrives, so `removedIds`
computes to `[]`, the `removedIds.length > 0` guard skips the new prune block entirely, and the
five maps keep their entries **permanently** — structurally beyond the reach of the only prune
path that exists. I proved it with a throwaway test (transcript below). That makes the commit
message's *"this removedIds block is the store's one session-went-away path"* — which is now a
permanent comment at `message-handler.ts:4729-4731` — false, and leaves #7470's own acceptance
partly unmet on the path a user hits every day.

The second is that the **roster guard covers a name prefix, not the session-keyed family**, while
its docstring claims the family. The PR's own "sixth family map" mutant is named
`sessionPrMutantSixth` — shaped to the guard. I reproduced it (3 failed, one per site) and then ran
the *same* mutant renamed `sessionCiChecks`: **38/38 green, survives.** The live proof that this
matters is `pendingServerSeed` (#7478) — genuinely session-keyed, declared in the same file, and
invisible to this guard.

Neither is a regression; both are claims that will be trusted by the next reader. The fixes are
small.

### Strengths

- **The three sites are correct and each one genuinely bites.** I dropped exactly one map at each
  site and every mutant died with an assertion naming *both* the map and the site: `sessionPrThreads`
  @ `session_list` → 3 failed; `sessionPrStatusRequestedAt` @ `forgetSession` → 2 failed;
  `sessionPrThreadsLoading` @ `_resetSessionMemory` → 2 failed. No blanket assertion is satisfiable
  by partial pruning, which is the property the file's header promises.
- **The positive controls are real, and stronger than the body claims.** The blanket-clear mutant
  (all five prunes → `= {}`) produced **6** failures, not 5 — the five per-map survivor controls
  *plus* `keeps a map's reference when the removed session was never in THAT map`. A fix that wipes
  everything cannot pass.
- **Self-caught weakness #1 is fixed, and I verified the replacement is load-bearing.** Mutating
  `pruneSessionKeyedMap`'s `return next ?? map` to `return next ?? { ...map }` is killed by exactly
  one test — the reference-stability test that drives a *real* removal. The retired version (a
  `session_list` with no removals) could not have caught it, because `removedIds.length > 0` was
  false and the helper never ran. Disclosing that in the body rather than quietly replacing it was
  the right call.
- **The stale-allowlist class is covered, in its strong form.** I renamed `sessionPrThreads` →
  `prThreadsBySession` in `types.ts` only. The type-derived roster silently lost the member and the
  parametrised loop quietly shrank from 38 to 35 tests — and the **hand-listed `arrayContaining`
  control caught it**, failing by name. Hand-listed control + type-derived loop is exactly the
  belt-and-braces shape that makes a rename impossible to sneak past, and it works. Worth saying
  out loud because the control looks redundant and is not.
- **The `sessionNotifications` exclusion is sound — I checked for the joint read.**
  `markSessionNotificationRead` / `markAllSessionNotificationsRead` / `dismissSessionNotification`
  (`connection.ts:5291-5323`) key purely on the notification `id`; the #7466 session-switch path
  gates on `sessions` membership (`connection.ts:3842`), not on any of the five maps. Nothing
  pruning these maps can break the #7472 flow.
- **`sessionPresetSnapshots` keyed by cwd — claim verified at the write site**, not inferred from
  the name: `message-handler.ts:3566-3568`, `const { cwd, preset } = parsed.data; … { …, [cwd]: preset }`.
  Excluding it *by name with its reason* rather than by narrowing the regex is the right call and
  is what makes M6a fail loudly instead of vanishing.
- **The filled fixture is the right call, and the argument is sound.** All five fields are
  non-optional on `ConnectionState` (no `?:` on any of them). An undefined-tolerant
  `pruneSessionKeyedMap` would have to return `undefined`, and `patch.sessionPrStatus = undefined`
  shallow-merges `undefined` into a required field — converting a test fixture's omission into a
  production crash surface for every consumer that does `Object.keys(...)`. Filling
  `dispatch-control-room-activity.test.ts`'s partial mock is correct.
- **#7478 filed rather than folded in silently**, correctly labelled, with the three `#7470` markers
  positioned where its fix lands. That is the follow-on protocol working as intended.
- **UI consumers are safe.** `App.tsx:682-688` read only `?.[activeSessionId] ?? null`; when the
  active session is the removed one, `activeSessionId` is reassigned in the same patch. No render
  regression from the new prune.

### Issues Found

#### Critical (Must Fix)

|---|---|---|---|
| `packages/dashboard/src/store/message-handler.ts` | 4434-4451 · 4729-4731 | **A fourth session-went-away path, reachable by Disconnect → Connect to the same server, on which the maps are stranded permanently.** `auth_ok`'s non-reconnect branch sets `sessions: []`, `activeSessionId: null`, `sessionStates: {}` — and leaves all five PR/CI maps. `removedIds` is a diff against `Object.keys(get().sessionStates)` (`message-handler.ts:4685-4688` → `packages/store-core/src/handlers/session-list.ts:215-218`), so with the roster already emptied the next `session_list` yields `removedIds = []`, the `removedIds.length > 0` guard skips the whole new block, and the entries are beyond the reach of the only prune path for the life of the tab. Every link in the reachability chain is verified: `disconnect()` calls `setLastConnectedUrl(null)` (`connection.ts:3063`) while **preserving** `wsUrl`, `sessions`, `sessionStates` and all five maps (`connection.ts:3122-3128` — the *"Preserve messages, terminalBuffer, sessions, activeSessionId, sessionStates"* comment); `retryConnection` → `connectToServer` → `connect(server.wsUrl, …)` runs with `_retryCount === 0` and `currentUrl === url`, so the `forgetSession` backstop at `connection.ts:2423-2427` does **not** fire; `isReconnect = lastConnectedUrl === url` is then `null === url` → **false**, so the wipe branch runs. The auto-reconnect ladder reaches the same branch by a second route: `scheduleRetry` re-resolves the endpoint and re-enters `connect()` with `_retryCount > 0` (`connection.ts:2556-2565`), which skips that same `_retryCount === 0` backstop — so a repointed `wsUrl` (a fresh Quick Tunnel URL after a daemon restart) lands there too. This makes *"this removedIds block is the store's one session-went-away path"* (`message-handler.ts:4729-4731`, and the commit message's "three spellings") **false as written**, and it is the durable artifact the next reader will trust. | Clear the five in the `auth_ok` **else** branch, beside `sessionStates: {}` — one site, covers both routes, and it is where the PR's own principle ("these maps go with the session state they describe") lands exactly. `disconnect()` is the alternative and already clears `sessionNotifications: []` in the same spirit, but it does not cover the repointed-URL ladder. Then correct the "one path" / "three spellings" wording in the block comment and the commit message, and extend the roster test's `SITES` table to four so the new site is held by the same guard. A regression test is cheap: seed the maps, drive a `session_list` with an **empty** `sessionStates`, and assert the dead ids are gone. |
| `packages/dashboard/src/store/session-destroy-prunes-pr-maps.test.ts` | 317 · 35-36 · 283-286 | **The roster guard's claim is broader than its check — it is a `sessionPr*` name-prefix roster, not a session-keyed-family roster.** The extraction is `/^\s{2}(sessionPr\w*):\s*Record<string,/gm`. The docstrings say *"every session-keyed PR/CI map"* and *"a SIXTH map added to the family without a prune goes red"*, and the PR body reports the sixth-map mutant as evidence of that property — but that mutant is named `sessionPrMutantSixth`, i.e. derived from the guard rather than from the defect. **Measured, not theorised:** I ran `sessionPrMutantSixth: Record<string, number>` added to `ConnectionState` with no cleanup → **3 failed**, one per site, exactly as claimed. I then ran the identical mutant renamed **`sessionCiChecks: Record<string, boolean>`** — same declaration point, same shape, genuinely session-keyed — → **38 passed (38), survives silently.** This is `docs/false-safety-guards.md`'s *"guard whose comment describes a stronger check than its code performs"*, and it is not hypothetical: `pendingServerSeed` (#7478) is session-keyed, declared in the same `types.ts`, and this guard cannot see it. | Cheapest honest fix: narrow both docstrings (and the PR body's mutation matrix row) to say the guard covers the **`sessionPr*`-prefixed** maps only, and that "session-keyed" is not what it tests. The stronger fix is to make the roster derive from something the family actually shares rather than a name prefix — e.g. an explicit `/** @sessionKeyed */` marker on the declarations with a matching opt-out set, so a new session-keyed map is red by default. Either way the mutant that certifies it must be named *outside* the pattern, or it is testing the guard against itself. |

|---|---|---|---|
| `packages/dashboard/src/store/connection.ts` | 3250-3255 · 3318-3322 | **The cross-server-collision rationale is overstated, and it is now permanent in two code comments plus the commit message.** Both blocks assert that "session ids are minted per daemon, so an entry that outlived its server can be read against a DIFFERENT server's session of the same id — one machine's CI state rendered on another machine's chip". Session ids are `randomBytes(16).toString('hex')` (`packages/server/src/session-manager.js:1059`) — 128 bits of CSPRNG, *more* entropy than a UUIDv4; the only other mint (`:2801`) and the `preserve` restore path (`:1053-1058`) both replay or re-draw that same shape. Two daemons producing the same id is not a real hazard, so "a wrong answer, not just retained memory" does not follow. **The clear at both sites is still right** — and there is a stronger, provable reason already sitting in this diff: `forgetSession` and `_resetSessionMemory` both empty `sessionStates`, and `removedIds` diffs against `sessionStates`, so anything left behind becomes **permanently unprunable** — the identical mechanism as Critical 1. | Replace the collision argument with the unprunable-orphan argument at both sites and in the commit message. It is the same conclusion reached from something that is true, and it generalises to Critical 1's site for free. (Worth keeping one sentence noting that ids carry no daemon namespace, so entropy is the *only* separation — that part is accurate and is a fair reason not to rely on it.) |
| `packages/dashboard/src/store/message-handler.ts` | 4705-4748 | **The audit table omits `cancellingActivityIds`, a session-scoped collection missing from exactly the site #7470 named.** It is a `Set<string>` keyed `${sessionId}:${activityId}` (`connection.ts:3689-3694`, keyed that way deliberately per #5277) — cleared in `forgetSession` (`:3263`), `_resetSessionMemory` (`:3329`) and on socket drop (`:2838-2839`), but **not** pruned in the `removedIds` block. It appears in the PR body's table neither as included nor as excluded-with-a-reason, so a reader cannot tell whether it was considered. Low impact — it strands a "Cancelling…" flag on a row that no longer renders, and the next socket drop clears it — but the PR's value is largely its claim to have audited the roster in full. | Out of scope to fix here; filed as **#7483** with the reasoning. Worth one row in the exclusion table saying it was seen and deferred, so the audit reads as complete rather than as having stopped at `Record<>`-shaped fields. |

#### Nitpicks

|---|---|---|---|
| `packages/dashboard/src/store/utils.ts` | 53 | `pruneSessionKeyedMap` tests membership with `if (!(id in map)) continue`, and `in` walks the prototype chain. `prune({a:1}, ['toString'])` returns a **new** object with identical contents — defeating the same-reference guarantee the docstring calls *"load-bearing, not an optimisation detail"*. Same for `__proto__`, `constructor`, `hasOwnProperty`. Unreachable today (ids are `[a-f0-9]{32}`), but the helper is generic, exported, and its contract is the reference identity. | `Object.prototype.hasOwnProperty.call(map, id)`. One word, and it makes the docstring's guarantee unconditional rather than true-by-luck about the id alphabet. |
| PR body | "Note on the touched existing test" | The body says the #7466 comment *"depends on that retention"* of `sessionNotifications`. Reading `connection.ts:3831-3841`, the comment *documents the retention as the hazard the membership check defends against* — it does not require it. The exclusion is still correct (an array, not a session-keyed map, and nothing joins it to the five); only the justification is inverted. | Reword to "the #7466 membership check exists precisely because both lookup sources outlive the session". |

### Verification note

Every mutation was applied by a Python scripted replace with an **occurrence assertion** (wrong match
count → non-zero exit, nothing written), then proved to have landed by `diff` against a `/tmp` byte
backup before the suite ran, then restored by absolute-path `cp` from that backup with a
byte-identical `diff` confirmed after each. Four files were mutated and restored across seven
mutants — `message-handler.ts`, `connection.ts`, `types.ts`, `utils.ts` — all four confirmed
byte-identical to `9af2ed81e` at the end. One probe test was written into
`packages/dashboard/src/store/` and deleted. **Final `git status --short` in the worktree is empty.**
No `git checkout`/`switch`/`stash` was run at any point.

**Mutation matrix (7 mutants, mine):**

| Mutant | Result |
|---|---|
| drop `sessionPrThreads` prune @ `session_list` | **killed** — 3 failed, incl. `[session_list removedIds] cleans up sessionPrThreads` |
| drop `sessionPrStatusRequestedAt` @ `forgetSession` | **killed** — 2 failed, incl. `[forgetSession] cleans up sessionPrStatusRequestedAt` |
| drop `sessionPrThreadsLoading` @ `_resetSessionMemory` | **killed** — 2 failed, incl. `[_resetSessionMemory] cleans up sessionPrThreadsLoading` |
| blanket clear — all five prunes → `= {}` | **killed** — **6** failed (5 survivor controls + the reference test) |
| `pruneSessionKeyedMap` always returns a new object | **killed** — 1 failed, `keeps a map's reference when the removed session was never in THAT map` |
| sixth map named `sessionPrMutantSixth`, no cleanup | **killed** — 3 failed, one per site (reproduces the body's claim) |
| sixth map named **`sessionCiChecks`**, no cleanup | **SURVIVED — 38 passed (38)** → Critical 2 |
| rename `sessionPrThreads` → `prThreadsBySession` in `types.ts` | **killed by the control** — loop silently shrank 38 → 35; `finds the known PR/CI maps in types.ts` caught it |

**Critical 1, reproduced.** A throwaway test driving `session_list` against an already-emptied
`sessionStates` (the state `auth_ok`'s else-branch leaves behind) — asserting the *stale* ids are
still present — **passes**, confirming the prune block is unreachable on that path:

```
✓ a session_list arriving after a roster wipe produces NO removedIds,
  so the maps keep dead ids forever            (1 passed)
```

**Suites at `9af2ed81e`** (worktree `manual-7470`; it has no `node_modules` of its own and resolves
the monorepo root's by ancestor walk, which is exactly why the four packages below are the ones
missing):

| Suite | Command | Result |
|---|---|---|
| Dashboard typecheck | `npx tsc --noEmit` (bare exit) | **exit 0** |
| Dashboard store | `npx vitest run src/store/` | **85 files / 1230 tests**, exit 0 |
| Both new files | `npx vitest run …pr-maps.test.ts` | **38 passed (38)**, exit 0 |

**The 251-error typecheck claim independently confirmed environmental.** Before resolving the
missing packages, `tsc` exits 2 with 251 errors; the distinct missing modules are exactly
`@testing-library/react`, `react-dom`, `react-dom/client`, `react-resizable-panels`, and the rest
is their implicit-any cascade. **Zero** of the 251 errors are in any file this PR touches. With
those four made resolvable in the worktree, `tsc --noEmit` exits **0** — matching the body. (One
transient red in my first store run, `commands.test.ts` failing on `react` 19.1.0 vs `react-dom`
19.2.5, was an artifact of my own package copy, not this diff; it is green once the versions match,
giving the 85/1230 above.)

### Verdict

- [ ] **Approve**
- [x] **Request Changes** — the fix at all three sites is correct and the evidence reproduces, but the PR earns its length by claiming an exhaustive audit and misses by one: a fourth went-away path, reachable by Disconnect → Connect, on which the entries become permanently unprunable — and the guard written to catch a missed sixth map is certified by a mutant named after the guard's own pattern, and survives one that is not. Both fixes are small; the wording corrections matter as much as the code, because both claims are going into permanent comments.

blamechris:
All four findings addressed in `100596b5e`. Both criticals reproduced red before fixing, and both of your surviving-mutant proofs now die.

| Finding | Status | Where |
|---|---|---|
| **C1** — fourth went-away path (`auth_ok`) | Fixed + 8 tests | `message-handler.ts` `#7470 authok-reset-*`; comment at the prune block corrected |
| **C2** — roster guard is a name-prefix roster | Guard replaced (structural, classify-or-fail) | `session-destroy-prunes-pr-maps.test.ts` |
| **S3** — cross-server collision rationale false | Rewritten at both sites + commit message | `connection.ts` `forget-reset` / `switch-reset` |
| **S4** — `cancellingActivityIds` missing from the audit | Now a first-class bucket entry in the guard | `SESSION_KEYED_DEFERRED` |
| **N1** — `in` walks the prototype chain | `Object.prototype.hasOwnProperty.call` + 3 tests | `utils.ts` |
| **N2** — "depends on" inverted | Corrected below | this comment |

---

## C1 — the fourth site

You were right, and the diagnosis was exact. Fixed in `auth_ok`'s **else** branch, beside the `sessionStates: {}` wipe — one site, both routes, and it is where this PR's own principle lands.

**Red first**, five assertions, one per map:

```
Tests  5 failed | 28 passed (33)
> #7470 a fresh auth_ok clears the per-session PR/CI maps > clears sessionPrStatus
AssertionError: expected { 'sess-dead': { …(10) }, …(1) } to deeply equal {}
… clears sessionPrStatusLoading      expected { Object (sess-dead, sess-live) } to deeply equal {}
… clears sessionPrStatusRequestedAt  expected { Object (sess-dead, sess-live) } to deeply equal {}
… clears sessionPrThreads            expected { 'sess-dead': { …(9) }, …(1) } to deeply equal {}
… clears sessionPrThreadsLoading     expected { Object (sess-dead, sess-live) } to deeply equal {}
```

Three of the eight are not regression tests for the fix, and are deliberate:

- **`POSITIVE CONTROL: a silent RECONNECT keeps them`** — the reconnect branch preserves the roster, so the snapshots are still true and clearing there would blank a chip on every transient drop. This is what stops the fix degenerating into "clear on every `auth_ok`", which would satisfy all five assertions above.
- **`MECHANISM: after a roster wipe, no session_list can ever prune the stale ids`** — your repro, kept as a permanent property. It passes **before and after**, which is the point: it states in one place why the fourth site cannot be folded into the `removedIds` block, for the next reader who tries to simplify it away.
- A fixture control, so "empty afterwards" can't pass because the seed never landed.

**The `SITES` table is now four**, and the mutation matrix is 5 maps × 4 sites = **20 cells, all killed**, each naming both the map and the site:

```
[authok-reset/sessionPrStatus]            2 failed — [auth_ok fresh connect] cleans up sessionPrStatus · clears sessionPrStatus
[authok-reset/sessionPrStatusLoading]     2 failed — …
[authok-reset/sessionPrStatusRequestedAt] 2 failed — …
[authok-reset/sessionPrThreads]           2 failed — …
[authok-reset/sessionPrThreadsLoading]    2 failed — …
```
(the other 15 cells reproduce your three, unchanged.)

**The false comment is corrected.** `message-handler.ts:4729-4731` no longer says "the store's one session-went-away path"; it says this is where *one* session going away is observed, names the three roster-wiping sites, and states the mechanism that unifies them. The commit message's "three spellings" is likewise gone.

## C2 — the guard was checking its own pattern

Conceded entirely; this was the more embarrassing of the two, since it is the failure mode the guard exists to catch, committed inside the guard. Docstring narrowing would have been the cheap fix — I took the stronger one.

**The guard is now structural**: it extracts every `Record<string, …>` **and `Set<string>`** declared on the `ConnectionState` interface (sliced to the interface, since types.ts has other two-space-member interfaces), and requires each of the 37 to land in exactly one bucket:

1. `CLEANED` — the five, asserted at all four sites
2. `SESSION_KEYED_ELSEWHERE` — `sessionStates`, the roster itself
3. `SESSION_KEYED_DEFERRED` — `pendingServerSeed` (#7478), `cancellingActivityIds` (#7483)
4. `NOT_SESSION_KEYED` — 29 entries, each with the key it is **actually** keyed by, quoted from its own declaration in types.ts rather than inferred from its name (name inference is what produced this defect)

**Your mutant, run verbatim:**

```
Tests  1 failed | 38 passed (39)
> classification: every session-keyed-shaped collection is in exactly one bucket
AssertionError: new session-keyed-shaped field(s) on ConnectionState. Add each to
CLEANED (and to all four site blocks), to SESSION_KEYED_DEFERRED with a tracking
issue, or to NOT_SESSION_KEYED with the key it is actually keyed by — quoted from
its declaration, not inferred from its name.: expected [ 'sessionCiChecks' ] to deeply equal []
```

Three more against the new guard:

| Mutant | Result |
|---|---|
| `sessionPrMutantSixth` (the old, pattern-shaped one) | killed — same classification failure, no longer special |
| `busySessionIds: Set<string>` | **killed** — proves the `Set` shape is covered, which the old guard could not express at all, and which is exactly #7483's shape |
| delete `credentialTestResults` from types.ts | killed — the reverse check catches a **stale allowlist** |
| your rename `sessionPrThreads` → `prThreadsBySession` | still killed, now by 2 tests (control + classification) |

A new collection is red until someone states how it is keyed — whatever it is named.

## S4 — `cancellingActivityIds`

Fair, and it is the sharper version of C2: a `Set<string>` was a shape my `Record<>`-only extraction could not see, so "was it considered?" was genuinely unanswerable. It is now bucket 3 with its key (`${sessionId}:${activityId}`) and #7483, alongside `pendingServerSeed`/#7478 — visible as *session-keyed and deliberately not yet cleaned*, not filed under "not session-keyed".

## S3 — the collision argument

Withdrawn; it was wrong and I should have checked the mint before writing it into two permanent comments. Both now give the unprunable-orphan reason:

> Clearing them HERE is not belt-and-braces over the `session_list` removedIds prune: it is the only thing that can reach them. `removedIds` is a diff against `Object.keys(sessionStates)`, and this block empties `sessionStates` in the same patch — so from the next snapshot onwards there is nothing to diff, `removedIds` is `[]`, and any entry left behind is permanently unprunable for the life of the tab.

Your suggestion to keep the accurate half is taken: the comment notes that ids carry no daemon namespace, so entropy (128 bits, ample) is the *only* separation — a fair reason not to lean on it, without claiming a collision hazard.

## N1 — `in` vs own-property

Fixed, with three tests: same-reference on a non-member, **not fooled by `toString` / `constructor` / `hasOwnProperty` / `valueOf`**, and a positive control that it still prunes a real own key (the first two are satisfiable by a helper that never prunes anything). Reverting to `in` kills it:

```
> is not fooled by inherited keys (`in` would clone here)
AssertionError: toString must not clone: expected { 'sess-a': 1 } to be { 'sess-a': 1 } // Object.is equality
```

One correction to your suggestion, caught by running the typecheck for its bare exit code: `Object.hasOwn` is **ES2022** and this package is `"lib": ["ES2020", …]`, so it fails `tsc --noEmit` with TS2550. Your exact spelling — `Object.prototype.hasOwnProperty.call` — is what shipped.

## N2 — the inverted justification

Correct, and the PR body has it backwards. The #7466 comment **documents the retention as the hazard the membership check defends against**; it does not require it. The exclusion stands on its own terms (an array, not a session-keyed map, and nothing joins it to the five).

Two other body corrections, since the body was reviewed as it stood: it says **three** sites — there are four; and its mutation-matrix row for the sixth-map mutant claimed a family property the guard did not have at the time. Both are superseded by this comment and by the commit message on `100596b5e`.

## Suites at `100596b5e`

| Suite | Command | Result |
|---|---|---|
| Dashboard typecheck | `npx tsc --noEmit` (bare exit) | **exit 0** |
| Dashboard store | `npx vitest run src/store/` | **85 files / 1248 tests**, exit 0 |
| Full dashboard | `npx vitest run` | **313 files / 5478 tests**, exit 0 |
| `session-destroy-prunes-pr-maps.test.ts` | individually | **42 passed**, exit 0 |
| `session-reset-clears-pr-maps.test.ts` | individually | **14 passed**, exit 0 |

Every mutation applied by a scripted replace with an occurrence assertion, proved landed before the suite ran, and restored by `cp` from a `/tmp` byte backup with `diff -q` confirming byte-identity after each — all four mutated files (`message-handler.ts`, `connection.ts`, `types.ts`, `utils.ts`) verified identical to the green tree at the end. No `git checkout` / `switch` / `stash`. Worktree `git status --short` is clean.

blamechris:
## Re-review at `100596b5e` — all four addressed, verdict updated to Approve

I re-ran every proof rather than reading the reply: the 20-cell matrix cell by cell, both of my
surviving mutants, and one new probe per fix aimed at the way each could have been satisfied
without being true. Everything holds.

### C1 — the fourth site

**Fixed, and the fix is pinned against the degeneration it invites.** The clear is in `auth_ok`'s
**else** branch beside `sessionStates: {}` — the site I'd have chosen, because it covers the
`scheduleRetry` route as well as Disconnect → Connect.

The obvious way to "pass" C1 is to clear on *every* `auth_ok`, which satisfies all five new
assertions and silently blanks a chip the user is reading on each transient drop. So I built that
mutant — moved the five clears into the shared `connectedState` so the reconnect branch clears too:

```
PROBE[clear-on-every-authok] KILLED | Tests 1 failed | 55 passed (56)
  × POSITIVE CONTROL: a silent RECONNECT keeps them, because it keeps sessionStates
```

That is the control doing exactly the job it was written for, and it is why the reconnect asymmetry
is a design decision here rather than an accident.

**The 20-cell matrix, run per cell on my side — 20 killed, 0 survived.** Each cell drops exactly one
map at exactly one site (token-anchored replace with an occurrence assertion, so a cell that fails
to apply aborts rather than reporting a false kill):

| Site | 5 maps | Result |
|---|---|---|
| `session_list` removedIds | all five | **5 killed** (3-4 failing assertions each) |
| `auth_ok` fresh connect | all five | **5 killed** (2 each) |
| `forgetSession` | all five | **5 killed** (2 each) |
| `_resetSessionMemory` | all five | **5 killed** (2 each) |

Every cell failed with **2+ assertions naming the field**, one behavioural and one from the `SITES`
table — so no cell is carried by a blanket assertion. `==== MATRIX RESULT: killed=20 survived=0 ====`.

**The `MECHANISM` test is the right call.** It is my repro kept as a permanent property, and it
passes both before and after — which is precisely what makes it useful. It is the only artifact in
the diff that states *why* the fourth site cannot be folded into the `removedIds` block, and it is
aimed at the future reader who tries to simplify it away rather than at this change. Keeping a test
that cannot fail on the current diff is usually wrong; here it is documentation with a compiler
attached, and the reason is written above it.

The corrected block comment now says this is where *one* session going away is observed, names the
three roster-wiping sites, and gives the mechanism that unifies them. The false claim is gone.

### C2 — the roster guard

**Conceded and over-fixed, in the good direction.** My mutant, run verbatim on my side:

```
PROBE[sessionCiChecks: Record<string, boolean>] KILLED | Tests 1 failed | 55 passed (56)
  × classification: every session-keyed-shaped collection is in exactly one bucket
```

The failure shape quoted in the reply is accurate — I reproduced it independently. Two more:

- **stale allowlist** — deleting `credentialTestResults` from `types.ts` while its bucket entry
  remains → **killed** by the reverse check.
- **the rename** — `sessionPrThreads` → `prThreadsBySession` → still killed, now by two tests.

The guard went from 5 name-prefixed fields to **37 structurally extracted** ones, each of which must
be classified with the key it is *actually* keyed by, quoted from its declaration. Red by default is
the right default, and the four buckets are the honest shape: "session-keyed and deliberately not
yet cleaned" is now a state a field can be *in*, rather than something that has to be inferred from
its absence.

#### One residue, and I'd call it a note rather than a finding

The extraction is `/^ {2}(\w+)\??: (?:Record<string,|Set<string>)/`. A field declared
`Map<string, …>` is not matched. I ran it:

```
PROBE[sessionCiChecks: Map<string, boolean>] *** SURVIVED *** | Tests 56 passed (56)
```

Why I am not raising it as a finding, having measured rather than assumed:

- **The shape is unpopulated.** I swept the `ConnectionState` body: **0** `Map<string,`, **0**
  index-signature (`{ [k: string]: … }`), **0** `Partial<Record<`, **0** `Readonly<Record<`, **0**
  declarations wrapping to a second line. All 37 collections are inline `Record<string,` or
  `Set<string>`. Nothing is hidden today.
- **It cannot swallow an existing field.** I converted a *classified* field
  (`resolvedPermissions: Record<…>` → `Map<…>`) expecting it to vanish from the roster. It goes
  **red on two tests** — the extraction control names it, and the stale-allowlist reverse check
  catches the bucket entry left dangling. So the gap admits only a brand-new `Map` field, never a
  conversion. That is a genuine strength of the reverse check and worth knowing it has.
- A `Map` on Zustand state breaks `useShallow` structural equality and JSON persistence, which is
  presumably why there are none.

If it's ever worth closing, it is one alternation plus a line in the control — cheaper as a
follow-up than as a rev on this PR.

### S3 — the collision argument

**Withdrawn correctly, and the replacement is provable from this diff.** I grepped both files for a
residual claim: the only remaining mention is the corrected sentence, which says a collision *is
not* the hazard. The accurate half is kept and honestly bounded — ids carry no daemon namespace, so
entropy (128 bits) is the only separation, which is a fair reason not to lean on it without
asserting a hazard that isn't there. The unprunable-orphan reason is the better one on the merits:
it is true, it is demonstrable from the code in this diff, and it generalises to the `auth_ok` site
for free — which is presumably why the fourth site was easier to see once it was written down.

### S4 — `cancellingActivityIds`

Now bucket 3 with its composite key and #7483, beside `pendingServerSeed`/#7478. The point stands
that the `Set` shape was what made "was it considered?" unanswerable before; both are visible as
deliberate deferrals now rather than absent.

### N1 — `in` vs own-property

Reverting `Object.prototype.hasOwnProperty.call` to `in`:

```
PROBE[revert-to-in] KILLED | Tests 1 failed | 55 passed (56)
  × is not fooled by inherited keys (`in` would clone here)
```

Three tests, and the third is the one that matters — a positive control that the helper still prunes
a real own key, without which the first two are satisfiable by a helper that never prunes anything.
The `Object.hasOwn` / ES2020 `lib` catch is a good one, and it was found the right way: by running
the typecheck for its bare exit code rather than assuming the shorter spelling compiles.

### N2

Agreed; superseded by the reply and the commit message.

### Verification note

Twenty-six mutants this round (20 matrix cells + 6 probes), each applied by a scripted replace with
an occurrence assertion, proved landed by `diff` against a `/tmp` byte backup before the suite ran,
and restored by absolute-path `cp` with byte-identity confirmed after each. All four mutated
files — `message-handler.ts`, `connection.ts`, `types.ts`, `utils.ts` — verified identical to
`100596b5e` at the end. No `git checkout` / `switch` / `stash`. Final `git status --short` in the
worktree is **empty**.

Suites at `100596b5e`, bare exit codes:

| Suite | Result |
|---|---|
| `npx tsc --noEmit` | **exit 0** (0 lines of output) |
| `npx vitest run src/store/` | **85 files / 1248 tests**, exit 0 |
| `session-destroy-prunes-pr-maps.test.ts` + `session-reset-clears-pr-maps.test.ts` | **56 passed**, exit 0 |
| `npx vitest run` (full dashboard) | **313 files / 5478 tests**, exit 0 |

Reproducing the typecheck needs the same note as last round: this worktree carries no
`packages/dashboard/node_modules` and resolves the monorepo root's by ancestor walk, so
`@testing-library/react`, `react-dom`, `react-dom/client` and `react-resizable-panels` are
unresolvable until they are provisioned — 251 errors, **none in any file this PR touches**, exactly
as the body says. With them resolvable, `tsc` exits 0.

### Verdict

- [x] **Approve** — ready to merge. Both criticals are fixed at the site I'd have chosen, and both
  fixes are pinned by controls I mutated myself rather than by the assertions they were written to
  satisfy: the reconnect control kills the clear-on-every-`auth_ok` degeneration, and the
  classification test kills the mutant that survived the previous guard. The 20-cell matrix is
  20/20 with every cell naming its map and its site. The one thing I'd have accepted as a
  docstring narrowing was fixed structurally instead, and the remaining `Map<string, …>` gap is
  unpopulated, cannot swallow an existing field, and is a follow-up at most.
- [ ] **Request Changes**

Issue #7477, opened 2026-08-29
store-core: overlapping full rebuilds mis-slice the deferred swap and DROP a live message appended between the two starts


blamechris:
Found while implementing #7455 — my client-level regression test for the overlapping-replay interleave tripped over this, which is a *different* defect in the same file.

## The shape

`_rebuildBaseline` (`packages/store-core/src/replay-reconcile.ts`) has the same non-refcount shape #7455 just removed from `_replayWindowOpen`: `reconcileReplayStart` **overwrites** it on every full rebuild, and `reconcileReplayEnd` **deletes** it on the first end.

#7455 scoped this out on the grounds that its failure mode is "a cosmetic mis-swap". It is not cosmetic — it drops a message.

## Reproduction (store-core, no client, no #7420 ledger involved)

```ts
reconcileReplayStart('s1', true, 0)                  // replay #1 — baseline 0
const messages = [{ id: 'live-q' }]                  // a LIVE arrival appended during replay #1
reconcileReplayStart('s1', true, messages.length)    // replay #2 — baseline OVERWRITTEN to 1
reconcileReplayEnd('s1', messages).swappedMessages   // => []     ← 'live-q' is GONE
reconcileReplayEnd('s1', messages).swappedMessages   // => null   (baseline already deleted)
```

`end#1` slices at replay **#2's** baseline, so everything appended between the two starts — which is exactly the live-racer window `#7420` exists to protect — is sliced off and applied to the store as a single atomic swap. The message is not merely stamped `'(resolved)'`; it is removed from `sessionStates[id].messages` entirely.

The interleave is the one #7455 documented and did not invent: `subscribe_sessions` (`session-handlers.js:469`) and `switch_session` (`:78`) both call `replayHistory(..., { forceFull: true })` without checking whether one is already in flight, and `replayHistory` chunks over `setImmediate` with back-pressure pauses.

## Why #7455 did not cover it

#7455 refcounted the *window* (`_replayWindowOpen`) so the live-arrival ledger composes across overlapping replays. It deliberately left `_rebuildBaseline` alone. With the window fixed, a racer that arrives in the **overlap** or in the **tail** is now correctly protected; a racer that arrives **before the second start** is protected by the ledger and then thrown away by the swap.

## Proposed fix

[...]

PR #7482, opened 2026-08-29
fix(store-core): refcount the replay window, and give its live-question ledger a lifetime

### LINKED ISSUE #7455: store-core: overlapping replays for one session wipe the #7420 live-question ledger and close the window early

### LINKED ISSUE #7456: store-core: _liveDuringReplay has no release at replay end, no cap, and no per-session teardown

blamechris:
Two lifecycle defects in `packages/store-core/src/replay-reconcile.ts`, both filed from the same review of PR #7453 (the #7420 live-vs-replayed AskUserQuestion discriminator). They live in the same collections and are fixed together.

Closes #7455

Closes #7456

---

## Defect 1 (#7455) — overlapping replays of one session re-open #7420 in the tail

`_replayWindowOpen` was a `Set`: membership, not a count. The server can have **two replays of one session in flight at once** and neither caller checks — `subscribe_sessions` replays every newly-subscribed background session (`session-handlers.js:469`) and `switch_session` calls `replayHistory(ws, id, { forceFull: true })` unconditionally (`:78`) — while `replayHistory` chunks 20 entries at a time over `setImmediate` with back-pressure pauses. Wire order: `start(X) start(X) end(X) end(X)`.

With a Set:

1. `start#2` ran `_liveDuringReplay.delete(X)`, destroying the still-open outer window's evidence.
2. `end#1` closed the window while replay #2 was still streaming.
3. Every live `user_question` between `end#1` and `end#2` fell through the `if (!_replayWindowOpen.has(id)) return` gate.
4. `end#2` swept with an empty ledger and stamped `answered: '(resolved)'` on them — **the exact bug #7420 fixed**, in the one case the gate exists for.

**Fix shape.** The window is a `Map<sessionId, number>` refcount: `+1` at `reconcileReplayStart`, `-1` at `reconcileReplayEnd`, open while `> 0`. The ledger is dropped on the **0→1 transition only**, so a second *sequential* replay still forgets the previous window (unchanged, still tested) while a *nested* start leaves the outer window's ids alone.

## Defect 2 (#7456) — the ledger had no release, no cap and no teardown

Three of the module's four per-session collections were released; `_liveDuringReplay` was not. It was dropped only by the *next* `reconcileReplayStart` for the same id, or by a whole-module `resetReplayReconcile` — and `resetReplayReconcile` had exactly one production caller on each client, `auth_ok`. Neither `socket.onclose` nor `socket.onerror` called it.

The comment's bounding argument was also wrong: it claimed the store's own message retention bounds the map, but the store drops a session's `messages` wholesale on the `session_list` prune and on `session_timeout` without touching module state, and `dispatchUserQuestion` records into the ledger *before* its `adapter.hasSession` check.

**Fix shape**, all three parts:

- **Release at replay end.** The map now holds a session only while its window is OPEN. When the last window closes, the ids move to a single-slot handoff (`_sweepableLedger`) that `sweepUnansweredPromptsAtReplayEnd` reads on the very next statement — the sweep stays a pure, repeatable read (no consuming read, no ordering hazard) and nothing is retained per-session afterwards. One slot, overwritten by the next replay end anywhere: bounded by construction.
- **Teardown.** New `dropReplaySessionState(sessionId)` export, called from the `session_list` prune and `session_timeout` handlers in **both** clients beside the `clearPersistedSession(id)` they already ran; it also drops the history cursor, since `removedIds` means the session is gone server-side. And `resetReplayReconcile()` now runs from `socket.onclose` / `socket.onerror` on both clients, after the stale-socket guard. Cursors are deliberately kept — they are what makes the reconnect a delta replay (#5555.3).
- **A cap that is never silent.** `MAX_LIVE_REPLAY_LEDGERS = 64`, LRU-evicting like the cursor map, with a `console.warn` naming the module, the cap, the evicted session and the consequence (an evicted ledger un-protects its questions, so the next sweep stamps them). The log text is pinned by a test.

The refcount makes the transport teardown load-bearing rather than tidy: a `history_replay_start` with no matching `_end` strands a `+1`, so every later end decrements from a too-high base and the window never closes again for that session.

## Red-first evidence

Every behaviour was proven failing before its fix.

| Behaviour | Red evidence |
|---|---|
| #7455 interleave (`start start note end note end`) | `AssertionError: expected false to be true` in `replay-reconcile.test.ts` — `q2` (the tail racer) was never recorded |
| #7455 nested start wipes the outer ledger | `AssertionError: expected false to be true` — `q0` gone after `start#2` |
| #7456 `session_timeout` leaks the ledger (app) | `expected true to be false` at `wasPromptLiveDuringReplay` |
| #7456 `session_list` prune leaks the ledger (app) | `expected true to be false` |
| #7456 `session_timeout` / prune leak the ledger (dashboard) | `expected true to be false` (×2) |
| #7456 `socket.onclose` / `onerror` leak the window (app) | `expected 1 to be +0` at `getReplayWindowDepth` (×2) |
| #7456 `socket.onclose` / `onerror` leak the window (dashboard) | `expected 1 to be +0` (×2) |

Each connection-teardown suite also carries a **negative control** — a stale socket's close must NOT tear down the current attempt's state — which passed before and after, so the fix is the guarded reset and not a blunt one.

## Mutation evidence

Every mutant was applied to the fixed tree, grep-verified as landed, and restored from a byte backup.

| # | Mutation | Killed by |
|---|---|---|
| M1 | window back to membership (`depth = 1`, unconditional delete at end) | store-core ×4, app ×1, dashboard ×1 |
| M2 | ledger destroyed on every start (`depth >= 1`) | store-core ×1 (the nested-start test), dashboard ×1 |
| M3a | `dropReplaySessionState` calls removed from the app | app ×2 |
| M3b | `dropReplaySessionState` calls removed from the dashboard | dashboard ×2 |
| M4 | `resetReplayReconcile()` removed from `onclose`/`onerror`, both clients | app ×2, dashboard ×2 |
| M5 | cap truncates silently (`console.warn` deleted) | store-core ×1 |
| M6 | no release at replay end (ledger left in the per-session map) | store-core ×2 |
| M7 | release at end but no handoff to the sweep | store-core ×6 **including two pre-existing #7420 tests**, app ×4 — this is the proof the release does not break the sweep's read |

M2's client row was added on review. The client suites as first written could not catch it, but the reason is narrower than "by design": the obvious variant (racer *before* `start#2`) cannot be written with two **full** starts, because end#1 slices that racer out of `messages` entirely — that is #7477, not the ledger. With two **delta** starts `_rebuildBaseline` is never set, nothing is swapped, and the test works. It is green clean and red under M2 with `AssertionError: expected '(resolved)' to be undefined`. Overlapping pure-delta replays are not the common wire shape (the connect handshake replays the active session with a cursor; `subscribe_sessions`, `switch_session` and `request_full_history` all force full), so it guards the logic rather than reproducing a production sequence.

## Suite table (bare exit codes, log-redirected)

| Suite | Command | Exit | Result |
|---|---|---|---|
| store-core | `npx vitest run` | `0` | 60 files, 2203 tests |
| store-core typecheck | `npm run typecheck` | `0` | clean |
| app store | `npx jest src/__tests__/store --coverage=false` | `0` | 35 suites, 737 tests |
| app typecheck | `npx tsc --noEmit` | `0` | clean |
| dashboard store | `npx vitest run src/store/` | `0` | 84 files, 1199 tests |
| dashboard typecheck | `npx tsc --noEmit` | `0` | clean |

Existing #7420 / #5555.3 / #5555.4 / #5588 assertions are untouched and green — none was wrong, and the "forgets the previous window at the next start" test in particular still pins that a second *sequential* replay drops the previous ledger.

## Follow-on filed

**#7477** — the new interleave test surfaced a third defect in the same file. `_rebuildBaseline` has the identical non-refcount shape, and #7455 assumed its overlap is "a cosmetic mis-swap". It is not: `start#2` overwrites the baseline and `end#1` then slices a message appended **between the two starts** straight out of `sessionStates[id].messages`. Reproduced standalone (no client, no ledger) and filed rather than folded in, so this PR stays on the window; the module comment and both client tests point at it.

### DIFF
--- a/packages/app/src/__tests__/store/connection-replay-teardown.test.ts
+++ b/packages/app/src/__tests__/store/connection-replay-teardown.test.ts
@@ -1,0 +1,178 @@
+/**
+ * #7456 — the replay window and its live-arrival ledger are per-CONNECTION
+ * state, and until now nothing released them on a transport drop.
+ *
+ * `resetReplayReconcile()` was called from exactly one production path on each
+ * client — `auth_ok` — so a socket that dropped mid-replay left the session's
+ * window open and its ledger populated until the NEXT successful auth, which on
+ * a backgrounded app may never come.
+ *
+ * Since #7455 the window is a REFCOUNT, which makes this load-bearing rather
+ * than merely untidy: a `history_replay_start` with no matching `_end` strands
+ * a +1, so every later end decrements from a too-high base and the window never
+ * closes again for that session — the ledger is never released and every later
+ * prompt is protected forever.
+ *
+ * Cursors are deliberately NOT cleared here: they are what makes the reconnect
+ * a delta replay instead of a full rebuild (#5555.3).
+ *
+ * Harness mirrors connection-reconnect-backoff.test.ts.
+ */
+jest.mock('expo-secure-store', () => ({
+  getItemAsync: jest.fn(() => Promise.resolve('dev-id-123')),
+  setItemAsync: jest.fn(() => Promise.resolve()),
+}));
+
+import * as SecureStore from 'expo-secure-store';
+import {
+  reconcileReplayStart,
+  recordHistorySeq,
+  resetReplayReconcile,
+  noteLivePromptDuringReplay,
+  wasPromptLiveDuringReplay,
+  getReplayWindowDepth,
+  getLiveReplayLedgerSessionIds,
+  getHistoryCursor,
+  isRebuildInProgress,
+} from '@chroxy/store-core';
+import { useConnectionStore, __resetDeviceIdCacheForTests } from '../../store/connection';
+import { useConnectionLifecycleStore } from '../../store/connection-lifecycle';
+import { resetReconnectAttempt } from '../../store/message-handler';
+import { clearAllCallbacks } from '../../store/imperative-callbacks';
+
+function flushPromises(): Promise<void> {
+  return new Promise((resolve) =>
+    jest.requireActual<typeof globalThis>('timers').setImmediate(resolve),
+  );
+}
+
+function mockResponse(status: number, body?: unknown): Response {
+  return {
+    ok: status >= 200 && status < 300,
+    status,
+    json: () => Promise.resolve(body ?? {}),
+    text: () => Promise.resolve(JSON.stringify(body ?? {})),
+  } as unknown as Response;
+}
+
+interface FakeSocket {
+  url: string;
+  readyState: number;
+  onopen: (() => void) | null;
+  onclose: ((event?: unknown) => void) | null;
+  onerror: ((event?: unknown) => void) | null;
+  onmessage: ((event: unknown) => void) | null;
+  send: jest.Mock;
+  close: jest.Mock;
+}
+
+function installMockWebSocket(): FakeSocket[] {
+  const instances: FakeSocket[] = [];
+  // @ts-expect-error — mock WebSocket constructor
+  global.WebSocket = class MockWebSocket {
+    static OPEN = 1;
+    url: string;
+    readyState = 0;
+    onopen: (() => void) | null = null;
+    onclose: ((event?: unknown) => void) | null = null;
+    onerror: ((event?: unknown) => void) | null = null;
+    onmessage: ((event: unknown) => void) | null = null;
+    send = jest.fn();
+    close = jest.fn();
+    constructor(url: string) {
+      this.url = url;
+      instances.push(this as unknown as FakeSocket);
+    }
+  };
+  return instances;
+}
+
+const originalFetch = global.fetch;
+const originalWebSocket = global.WebSocket;
+
+beforeEach(() => {
+  jest.useFakeTimers();
+  clearAllCallbacks();
+  __resetDeviceIdCacheForTests();
+  resetReconnectAttempt();
+  resetReplayReconcile({ clearCursors: true });
+  (SecureStore.getItemAsync as jest.Mock).mockReset();
+  (SecureStore.getItemAsync as jest.Mock).mockResolvedValue('dev-id-123');
+  (SecureStore.setItemAsync as jest.Mock).mockResolvedValue(undefined);
+  useConnectionStore.setState({ sessionStates: {}, activeSessionId: null, socket: null });
+  useConnectionLifecycleStore.setState({
+    connectionPhase: 'disconnected',
+    connectionError: null,
+    connectionRetryCount: 0,
+    wsUrl: null,
+  });
+});
+
+afterEach(() => {
+  useConnectionStore.getState().disconnect();
+  resetReplayReconcile({ clearCursors: true });
+  global.fetch = originalFetch;
+  global.WebSocket = originalWebSocket;
+  jest.useRealTimers();
+});
+
+/** Open a connection and return the freshly constructed (current-attempt) socket. */
+async function openConnectedSocket(): Promise<FakeSocket> {
+  global.fetch = jest.fn().mockResolvedValue(mockResponse(200, { status: 'ok' }));
+  const instances = installMockWebSocket();
+  useConnectionStore.getState().connect('wss://tunnel.example.com', 'tok', { silent: true });
+  await flushPromises();
+  useConnectionLifecycleStore.setState({ connectionPhase: 'connected' });
+  return instances[instances.length - 1]!;
+}
+
+/** Open a replay window for `s1` with one live question inside it. */
+function openReplayWindowWithRacer(): void {
+  reconcileReplayStart('s1', true, 0);
+  noteLivePromptDuringReplay('s1', 'live-q');
+  recordHistorySeq('s1', 42);
+  // Positive controls — the fixture took effect, so the assertions below are
+  // about the teardown and not about state that was never there.
+  expect(getReplayWindowDepth('s1')).toBe(1);
+  expect(wasPromptLiveDuringReplay('s1', 'live-q')).toBe(true);
+  expect(isRebuildInProgress('s1')).toBe(true);
+}
+
+describe('transport teardown releases the replay window + live-arrival ledger (#7456)', () => {
+  it('socket.onclose releases the window and ledger but keeps the history cursor', async () => {
+    const socket = await openConnectedSocket();
+    openReplayWindowWithRacer();
+
+    socket.onclose?.({ code: 1006 });
+
+    expect(getReplayWindowDepth('s1')).toBe(0);
+    expect(getLiveReplayLedgerSessionIds()).toEqual([]);
+    expect(wasPromptLiveDuringReplay('s1', 'live-q')).toBe(false);
+    expect(isRebuildInProgress('s1')).toBe(false);
+    expect(getHistoryCursor('s1')).toBe(42);
+  });
+
+  it('socket.onerror releases the window and ledger but keeps the history cursor', async () => {
+    const socket = await openConnectedSocket();
+    openReplayWindowWithRacer();
+
+    socket.onerror?.({ message: 'boom' });
+
+    expect(getReplayWindowDepth('s1')).toBe(0);
+    expect(getLiveReplayLedgerSessionIds()).toEqual([]);
+    expect(wasPromptLiveDuringReplay('s1', 'live-q')).toBe(false);
+    expect(getHistoryCursor('s1')).toBe(42);
+  });
+
+  it('a stale socket close does not tear down the CURRENT attempt state', async () => {
+    const stale = await openConnectedSocket();
+    // A new attempt supersedes it (bumps the module-level attempt id).
+    await openConnectedSocket();
+    openReplayWindowWithRacer();
+
+    stale.onclose?.({ code: 1006 });
+
+    expect(getReplayWindowDepth('s1')).toBe(1);
+    expect(wasPromptLiveDuringReplay('s1', 'live-q')).toBe(true);
+  });
+});
--- a/packages/app/src/__tests__/store/message-handler.test.ts
+++ b/packages/app/src/__tests__/store/message-handler.test.ts
@@ -26,6 +26,9 @@ import {
   PERMISSION_DECISION_TOKENS,
   isLivePermissionPrompt,
   resetReplayReconcile,
+  wasPromptLiveDuringReplay,
+  getReplayWindowDepth,
+  getLiveReplayLedgerSessionIds,
 } from '@chroxy/store-core';
 import { clearPersistedSession } from '../../store/persistence';
 import { setCallback, clearAllCallbacks } from '../../store/imperative-callbacks';
@@ -975,6 +978,88 @@ describe("history_replay_end: '(resolved)' sweep vs a racing live AskUserQuestio
     _testMessageHandler.handle({ type: 'history_replay_end', sessionId: 's1' });
 
     expect((store.getState().sessionStates.s1.messages[0] as any).answered).toBe('(resolved)');
+  });
+
+  // #7455 — the server can have two replays of ONE session in flight at once:
+  // `subscribe_sessions` replays every newly-subscribed background session and
+  // `switch_session` calls `replayHistory(..., { forceFull: true })`
+  // unconditionally, so the wire order is start start end end. Before the
+  // window became a refcount the FIRST end closed it while replay #2 was still
+  // streaming, and every live question in that tail was stamped '(resolved)'.
+  it('protects a racer that arrives between the two ends of overlapping replays (#7455)', () => {
+    const store = seedOne();
+    _testMessageHandler.handle({ type: 'history_replay_start', sessionId: 's1', fullHistory: true });
+    _testMessageHandler.handle({ type: 'history_replay_start', sessionId: 's1', fullHistory: true });
+    // A racer inside the OVERLAP...
+    _testMessageHandler.handle(question({ toolUseId: 'q-use-1' }));
+    _testMessageHandler.handle({ type: 'history_replay_end', sessionId: 's1' });
+    // ...and one in the TAIL, after end#1, while replay #2 is still streaming.
+    // The tail racer is the one the pre-#7455 Set lost.
+    _testMessageHandler.handle(question({ toolUseId: 'q-use-2' }));
+    _testMessageHandler.handle({ type: 'history_replay_end', sessionId: 's1' });
+
+    // A racer arriving BEFORE the second start is deliberately not covered
+    // here: `_rebuildBaseline` has the same non-refcount shape and the second
+    // start overwrites it, so end#1's atomic swap slices that racer out of
+    // `messages` entirely. That is a separate defect in the same file, filed
+    // as #7477 — the ledger protects the message, the swap then drops it.
+    //
+    // Positive control: both questions really are in the store, so the
+    // `undefined` assertions below are about the sweep and not a missing fixture.
+    expect(store.getState().sessionStates.s1.messages).toHaveLength(2);
+    expect(answeredOf(store, 0)).toBeUndefined();
+    expect(answeredOf(store, 1)).toBeUndefined();
+    // Balanced pairs ⇒ the window is closed again.
+    expect(getReplayWindowDepth('s1')).toBe(0);
+  });
+
+  // #7456 — the store drops a session's messages wholesale on prune and on
+  // timeout; module state here used to survive both with nothing left to
+  // correspond to.
+  it('session_timeout drops the session live-arrival ledger (#7456)', () => {
+    const store = seedOne();
+    _testMessageHandler.handle({ type: 'history_replay_start', sessionId: 's1', fullHistory: false });
+    _testMessageHandler.handle(question());
+    const liveId = (store.getState().sessionStates.s1.messages[0] as any).id;
+    expect(wasPromptLiveDuringReplay('s1', liveId)).toBe(true);
+
+    _testMessageHandler.handle({
+      type: 'session_timeout',
+      sessionId: 's1',
+      name: 'S1',
+      idleMs: 600000,
+    });
+
+    expect(wasPromptLiveDuringReplay('s1', liveId)).toBe(false);
+    expect(getReplayWindowDepth('s1')).toBe(0);
+    expect(getLiveReplayLedgerSessionIds()).not.toContain('s1');
+  });
+
+  it('a session_list prune drops the pruned session live-arrival ledger (#7456)', () => {
+    const store = createMockStore({
+      activeSessionId: 's1',
+      sessions: [
+        { sessionId: 's1', name: 'S1' } as any,
+        { sessionId: 's2', name: 'S2' } as any,
+      ],
+      sessionStates: { s1: createEmptySessionState(), s2: createEmptySessionState() },
+      sessionNotifications: [],
+    });
+    setStore(store as any);
+    _testMessageHandler.setContext(createMockConnectionContext());
+
+    _testMessageHandler.handle({ type: 'history_replay_start', sessionId: 's1', fullHistory: false });
+    _testMessageHandler.handle(question());
+    const liveId = (store.getState().sessionStates.s1.messages[0] as any).id;
+    expect(wasPromptLiveDuringReplay('s1', liveId)).toBe(true);
+
+    // s1 dropped out of the authoritative list.
+    _testMessageHandler.handle({ type: 'session_list', sessions: [{ sessionId: 's2', name: 'S2' }] });
+
+    expect(store.getState().sessionStates).not.toHaveProperty('s1');
+    expect(wasPromptLiveDuringReplay('s1', liveId)).toBe(false);
+    expect(getReplayWindowDepth('s1')).toBe(0);
+    expect(getLiveReplayLedgerSessionIds()).not.toContain('s1');
   });
 });
 
--- a/packages/app/src/store/connection.ts
+++ b/packages/app/src/store/connection.ts
@@ -1398,6 +1398,16 @@ export const useConnectionStore = create<ConnectionState>((set, get) => ({
       const wasConnected = useConnectionLifecycleStore.getState().connectionPhase === 'connected';
       set({ socket: null });
 
+      // #7456 — the replay window and its live-arrival ledger are
+      // per-CONNECTION state, like the rebuild baseline: release them on the
+      // transport drop rather than waiting for the next `auth_ok`, which on a
+      // backgrounded app may never arrive. Load-bearing since #7455 made the
+      // window a refcount — a `history_replay_start` with no matching `_end`
+      // would otherwise strand a +1 and the window would never close again for
+      // that session. Cursors are deliberately KEPT (they are what makes the
+      // reconnect a delta replay, #5555.3).
+      resetReplayReconcile();
+
       // Clear transient streaming/plan state so stale UI doesn't persist
       clearPermissionSplits();
       updateActiveSession((ss) => {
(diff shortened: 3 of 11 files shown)


blamechris:
## Inspector review — PR #7482 @ `0b2ddb750`

Reviewed from the PR worktree, Node v22 / darwin-arm64. I re-ran three of the seven mutants
rather than reading the table, wrote a fourth of my own aimed at the one the PR reports as
surviving client-side, and spent most of the effort trying to construct a `reconcileReplayStart`
with no matching end. **I could not find one that survives teardown**, and the direction the
design fails in is the safe one. Nothing blocking.

Both mutated files were restored from byte backups — `replay-reconcile.ts` sha256
`7a455fd67567ebb205da9e4ea962f5429ddcb9f00c2278a08d3f97f06ff4a81a` identical before and after
every mutant — and **`git status --short` is empty in the PR worktree**.

### Verdict

- [x] **Approve with suggestions** — no blocking issues. S1 is the only one I'd consider folding in (7 lines, harness already present); S2/N1/N2 are optional.
- [ ] Request Changes
- [ ] Comment — feedback only

---

## 1. Stuck-open windows — I went looking and every path terminates in a socket close

This was the highest-risk question, because the refcount converts a stranded `+1` from "untidy"
into "this session's window never closes again". Enumerating every way a `history_replay_start`
can lose its `_end`:

| Path | Outcome |
|---|---|
| Socket leaves OPEN mid-replay | `sendChunk`'s `if (ws.readyState !== 1) return` and `scheduleAfterDrain`'s same guard abandon the replay with no `_end` — but the client's `onclose` fires → `resetReplayReconcile()`. **Covered.** |
| Drain stalls on a half-open socket | `BACKPRESSURE_MAX_WAIT_MS` (30s) → `ws.close(1013)` (`ws-history.js:150`) → same path. **Covered.** |
| **Server crashes, client socket stays OPEN** | The case the brief asked about. `connection-runtime.ts:256` — pong timeout calls `socket.close()`, which fires the client's own `onclose`. **Covered.** |
| User-initiated `disconnect()` (nulls `socket.onclose`) | Calls `resetReplayReconcile({ clearCursors: true })` itself (app `:1557`, dashboard `:3126`). **Covered.** |
| Two starts, one end for the same session | The server cannot emit this. `replayHistory` has exactly one `finishReplay` and #7340 routed *both* exits (empty slice, drained final chunk) through it. |
| Start and end keyed to different sessions | Both frames always carry an explicit `sessionId` (`ws-history.js:1232`, `:1240`), and `replayHistory` returns before the start frame when `getHistory(sessionId)` is empty — so the `|| activeSessionId` fallbacks in both handlers cannot mis-key a *pair*. |
| `session_timeout` / prune **during** an open window | `dropReplaySessionState` deletes the refcount key outright rather than decrementing, so a still-streaming replay #2's `end` decrements from zero. `closeReplayWindow` handles depth 0 without going negative — pinned by *"an unmatched end cannot drive the depth negative"*. **Correct.** |

**Direction, judged honestly:** a leaked open window fails **OPEN** — `noteLivePromptDuringReplay`
records everything, `ledgerFor` returns a non-empty set, the sweep skips. Prompts stay answerable.
The dangerous direction is a spuriously *closed* window, and that is exactly what the four new
`resetReplayReconcile()` calls could have introduced — which is why the stale-socket guard placement
matters, and why the negative control in each teardown suite (*"a stale socket close does not tear
down the CURRENT attempt state"*) is the most load-bearing test in the PR. I verified the placement
at all four sites: the `myAttemptId !== connectionAttemptId` early-return precedes
`resetReplayReconcile()` in every one.

**Teardown roster is complete.** This is the failure mode I most expected (fix three roster lines,
walk past the fourth). I grepped both clients for every path that removes an entry from
`sessionStates`: there are exactly **two** — the `session_list` prune and `session_timeout` — and
both are wired in both clients. The other `createEmptySessionState()` sites (app `:3058`,
dashboard `:2235`, `:4785`) are all guarded *creations*, not wipes. No fourth line.

One deliberate widening worth recording: `dropReplaySessionState` also drops the history **cursor**,
and for a *bound* client `session_list` is filtered to the bound session (`ws-history.js:603`), so
every other session lands in `removedIds`. That is consistent — the prune already ran
`clearPersistedSession` for exactly those ids — and the cost of a dropped cursor is one full replay.

## 2. The single-slot `_sweepableLedger` — the "next statement" claim holds, and here is the proof

The write→read window is `reconcileReplayEnd(...)` → *(optional)* `updateSession(swappedMessages)`
→ `updateSession(sweep)`, entirely inside one `handleMessage` invocation in both clients
(app `:2521`/`:2556`, dashboard `:5295`/`:5331`). The only thing that runs in between is a zustand
`set`, which notifies subscribers **synchronously** — so I enumerated them. There are three
(app `connection.ts:2641`, dashboard `:5494` and `:5516`), all persistence/registry writers; none
re-enters `handleMessage` or touches `replay-reconcile`. There is no batched or queued path that
splits the pair either: the delta flusher (`applyDeltaBatch`) writes store state only, and the
terminal-write batcher is `setTimeout`-scheduled. **A second session's replay-end cannot land
between the slot write and the sweep read.**

Two things to record about the shape, neither a finding:

- The slot is keyed by `sessionId` *inside* the value, and `ledgerFor` requires an exact match. If
  the invariant were ever broken, a mismatch yields `undefined` → the sweep stamps. That is
  **fail-closed** — the harmful direction. The invariant is enforced by a comment and by tests, not
  by the interface. M7 does pin the coupling from the client side (app ×4), so a reordering of those
  two statements goes red; I'd call that adequate. The cheap hardening, if it ever comes up again,
  is to *return* the released ledger from `reconcileReplayEnd` (returning is not consuming, so the
  `updateSession`-early-return objection that ruled out a consuming read does not apply).
- `closeReplayWindow` writes `null` when the closing window held nothing, so an end for session B
  clobbers a pending slot for session A. Unreachable for the reason above, and the alternative
  (leaving A's slot in place) would be the fail-*open* variant of the same assumption.

## 3. The sequential gap between `end#1` and `start#2` — real, reachable, and **already filed as #7457**

The brief asked whether #7420 re-enters in the gap. It does, and it is not new here. A
`user_question` that arrives with no window open is not recorded (correct per the gate's contract),
and the next `history_replay_end` for that session stamps it `'(resolved)'`. That is reachable on
the most natural path there is: `switch_session` calls `replayHistory(..., { forceFull: true })`
unconditionally (`session-handlers.js:78`) — i.e. the user tapping into the background session *to
answer the question* triggers the replay that kills the answer path. And there is no rescue:
`resendPendingPermissions` walks only `_pendingPermissions` (`ws-permissions.js:769`), while
`user_question` **is** pushed to the ring buffer (`session-message-history.js:393`).

I checked before filing anything: **#7457** ("A genuinely PENDING AskUserQuestion is stamped
'(resolved)' on any ordinary reconnect — the non-racing half of #7420") is exactly this and is
open. This PR neither widens nor narrows it — the pre-#7455 `Set` had the identical gap. No new
issue.

## 4. LRU is genuine, not FIFO mislabeled

`noteLivePromptDuringReplay` does delete-then-set on the touch path (`if (existing) delete`), so the
just-noted session moves to the Map tail, and the `while (size > MAX)` loop evicts from the head.
That is real recency, and *"re-noting an existing session refreshes its recency"* pins it against
the FIFO variant. Note that recency is by **note**, not by window open — correct, since a session
with no live arrival has no key to evict.

## 5. Mutation re-runs — the table is accurate

Re-run from the fixed tree, each grep-verified as landed, each restored from the byte backup:

| Mutant | Applied | Result |
|---|---|---|
| **M1** — back to membership (`depth = 1`; `closeReplayWindow` always closes) | ✅ | store-core **×4**, app **×1**, dashboard **×1** — matches the table exactly |
| **M6** — no release at end (ledger left in the per-session map) | ✅ | store-core **×2** |
| **M7** — release at end, no handoff (`_sweepableLedger` stays null) | ✅ | store-core **×6**, and I confirmed **two of the six are pre-existing #7420 tests** (`skips a prompt that arrived live inside the window…`, `forgets the previous window at the next start`) — this is the strongest evidence in the PR that the release did not break the sweep's read |

## S1 (suggestion) — a client-level M2 test **is** writable; "survive by design" is true only of the full-rebuild shape

The PR says clients survive M2 (`depth >= 1`, ledger destroyed on every start) because "their racers
arrive after `start#2`". That is true of the tests as written, but the reason the obvious variant
(racer *before* `start#2`) can't be written with two **full** starts is #7477 — end#1 slices the
racer out of `messages` entirely. With two **delta** starts `_rebuildBaseline` is never set, nothing
is swapped, and the test works. I wrote it into the dashboard harness and ran it both ways:

```ts
it('a racer before the second DELTA start stays protected', () => {
  seedOne()
  handleMessage({ type: 'history_replay_start', sessionId: 's1', fullHistory: false }, ctx() as any)
  handleMessage(question({ toolUseId: 'q-use-0' }) as any, ctx() as any)
  handleMessage({ type: 'history_replay_start', sessionId: 's1', fullHistory: false }, ctx() as any)
  handleMessage({ type: 'history_replay_end', sessionId: 's1' }, ctx() as any)
  handleMessage({ type: 'history_replay_end', sessionId: 's1' }, ctx() as any)
  expect((store.getState() as any).sessionStates.s1.messages).toHaveLength(1)
  expect(answeredOf(0)).toBeUndefined()
})
```

- unmutated → `Tests 1 passed`, exit `0`
- under M2 (`if (depth === 1)` → `if (depth >= 1)`) → `AssertionError: expected '(resolved)' to be undefined`, exit `1`

So the 0→1-only clear is currently pinned at the unit level only, and it *can* be pinned at the
integration level for seven lines. I'll note that overlapping pure-delta replays are not the common
wire shape (the connect handshake replays the active session with a cursor; `subscribe_sessions`,
`switch_session` and `request_full_history` all force full), so this is a guard on the logic rather
than a reproduction of a production sequence. Fold in or don't — but it's worth correcting the PR
body's framing, since "clients can't catch M2" isn't quite the reason.

## S2 (nit) — the cap is one-dimensional

`MAX_LIVE_REPLAY_LEDGERS` bounds the number of **sessions** in `_liveDuringReplay`; the per-session
`Set<string>` of ids has no bound. Given §1 found no unrecovered stuck-open window, and a session
blocks on one `AskUserQuestion` at a time, the realistic per-window content is 0–1 ids — so the
exposure is negligible. Only flagging it because the comment's framing ("defensively and loudly"
bounded) reads as complete and covers one axis.

## N1 (nit) — stray double blank line

`packages/dashboard/src/store/connection.ts` — both new blocks leave two consecutive blank lines
after `resetReplayReconcile();`. The app file doesn't.

## N2 (nit) — the eviction *consequence* is pinned at the predicate, not at the sweep

The cap test asserts `wasPromptLiveDuringReplay('s0', 'q-0') === false`, which implies the harm the
warn text describes but doesn't demonstrate it. One more line —
`sweepUnansweredPromptsAtReplayEnd('s0', [prompt('q-0')])` returning a `'(resolved)'`-stamped array —
would make the warn's claim and the test's claim the same claim.

## #7477 boundary — confirmed no worse, and the pointers are where the PR says

The refcount never touches `_rebuildBaseline`, and `reconcileReplayEnd`'s baseline branch is
byte-identical to `main` (`closeReplayWindow(sessionId)` replaces `_replayWindowOpen.delete(sessionId)`
in the same position, ahead of the same `_rebuildBaseline.has` check). So end#1 mis-slicing at
start#2's baseline behaves exactly as it did before this PR — including the delta-then-full overlap
(connect replay of the active session still streaming when `switch_session` forces a full rebuild of
the same session), which is the *reachable* form of #7477 and is captured by its acceptance criteria.
The module comment (`replay-reconcile.ts`) and **both** client tests name #7477 with the reason, as
claimed. #7477 is open, correctly scoped, and carries a standalone reproduction.

## Independent verification

**Suites** — my own runs, bare exit codes, log-redirected:

| Suite | Command | Exit | Result |
|---|---|---|---|
| store-core | `npx vitest run` | `0` | 60 files, 2203 tests |
| app store | `npx jest src/__tests__/store --coverage=false` | `0` | 35 suites, 737 tests |
| dashboard store | `npx vitest run src/store/` | `0` | 84 files, 1198 tests |
| store-core typecheck | `npm run typecheck` | `0` | clean |
| app typecheck | `npx tsc --noEmit` | `0` | clean |
| dashboard typecheck | `npx tsc --noEmit` | `0` | clean |

(The app run's "worker process has failed to exit gracefully" is the pre-existing condition on
`main`, not this PR.)

**The client suites really do exercise this PR's store-core**, which is worth stating explicitly in
a worktree with no root `node_modules`: `packages/store-core/package.json` sets `main`/`exports .`
to `./src/index.ts` (only `./crypto` is built to `dist`, and only `dist/crypto.{js,d.ts}` are
tracked), and both client shadows symlink `node_modules/@chroxy/store-core -> ../../../store-core`
**inside the worktree**. So neither a stale `dist` nor the main checkout's copy is in play — the
`dropReplaySessionState` / `getReplayWindowDepth` imports would not have resolved otherwise.

---

## CI at `0b2ddb750` — the two red rows are cancellations, not failures

`gh pr checks` shows `Server Lint` and `Dashboard Tests` as `fail`, which is worth not taking at
face value. Both jobs report `"conclusion": "cancelled"` with an **empty failing-step list**
(jobs `99047306149`, `99047306153`), and the run itself
([33232374786](https://github.com/blamechris/chroxy/actions/runs/33232374786), head
`0b2ddb750`) is still `queued` with 11 contexts pending. `gh pr checks` renders `cancelled` as
`fail`, so there is no verdict on this head yet — 9 pass / 2 cancelled / 11 pending. Nothing to
diagnose; it needs a re-read once the run settles (and `mergeStateStatus: BLOCKED` with
`mergeable: MERGEABLE` is consistent with an incomplete run plus the Copilot ruleset, not with a
defect).

blamechris:
Thanks — all four folded in at `ffcbde293`. Mapping, finding by finding.

## S1 — client-level M2 test (taken verbatim)

Added to the dashboard harness exactly as written. Reproduced both directions on my side:

- unmutated → `Tests 1 passed`, exit `0`
- under M2 (`if (depth === 1)` → `if (depth >= 1)`, grep-verified as landed, restored from a byte backup) → `AssertionError: expected '(resolved)' to be undefined`, exit `1`

Same string, same exit codes. The test carries a comment recording why the delta shape is the one that works — with two full starts, end#1 slices the pre-`start#2` racer out of `messages` entirely, so the assertion could never be reached, and that is #7477 rather than the ledger — plus your note that overlapping pure-delta replays guard the logic rather than reproduce a production sequence.

**PR body corrected too.** The M2 row now reads `store-core ×1, dashboard ×1`, and the "clients survive by design" framing is replaced with the accurate reason. You were right that it wasn't quite the reason: the tests as written couldn't catch it, which is a different claim from the mutation being unreachable client-side.

## S2 — the cap is one-dimensional

Reworded. It now says plainly that it bounds the number of session **keys** and puts no bound on the `Set<string>` inside any one of them, and gives the asymmetry that makes that the right axis to cap: a session blocks on one `AskUserQuestion` at a time so a realistic window holds 0–1 ids, while `subscribe_sessions` can open many windows at once. The "defensively and loudly" phrasing that read as a complete bound is gone.

## N1 — stray blank lines

Both collapsed; dashboard `connection.ts` now matches the app file.

## N2 — took it

It was one statement on the existing cap test, and your framing settled it: the warn text claims the evicted session's question becomes stampable, and the test only asserted the predicate that implies it. It now closes the window for the evicted session and asserts `sweepUnansweredPromptsAtReplayEnd` returns it stamped `'(resolved)'` — so the warn's claim and the test's claim are the same claim.

## Two notes on your findings, for the record

**§2, the `reconcileReplayEnd`-returns-the-ledger hardening.** Agreed that returning is not consuming and so dodges the `updateSession` early-return objection that ruled out the consuming read. I didn't take it here because it changes the return shape at four client call sites for an invariant that M7 already pins from the client side (app ×4), and the failure direction is fail-closed. Recording it as the known cheap hardening if the coupling is ever touched again.

**§3, the `end#1`→`start#2` gap.** Confirmed #7457 independently — same conclusion, including that `resendPendingPermissions` walks only `_pendingPermissions` while `user_question` does reach the ring buffer, so there is no rescue path. Neither widened nor narrowed here; the pre-#7455 `Set` had the identical gap.

## Suites after the folds (bare exit codes, log-redirected)

| Suite | Command | Exit | Result |
|---|---|---|---|
| store-core | `npx vitest run` | `0` | 60 files, 2203 tests |
| store-core typecheck | `npm run typecheck` | `0` | clean |
| app store | `npx jest src/__tests__/store --coverage=false` | `0` | 35 suites, 737 tests |
| app typecheck | `npx tsc --noEmit` | `0` | clean |
| dashboard store | `npx vitest run src/store/` | `0` | 84 files, **1199** tests (+1, S1) |
| dashboard typecheck | `npx tsc --noEmit` | `0` | clean |

`git status --short` empty in the worktree.

